PR #1, opened 2015-10-05
Add first unit tests

asaunier:
Still very basic :P 

I have copied/pasted tests from the API and removed all DB-related code since the UI has no DB connection.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+sudo: false
+language: python
+
+python:
+- 2.7
+
+install:
+- make -f config/travis .build/dev-requirements.timestamp
+- make -f config/travis install
+
+script:
+- make -f config/travis check
+
+notifications:
+  email: false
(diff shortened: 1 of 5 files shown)


asaunier:
Looks good?

asaunier:
What other tests could be added?

tsauerwein:
Yes, looks good!

Are you already making calls to the REST API from the Travis tests?

asaunier:
> Are you already making calls to the REST API from the Travis tests?

No, but I will add such tests. Thanks for the suggestion.

asaunier:
> Are you already making calls to the REST API from the Travis tests?

The problem is that we cannot use a travis-created API to make test calls since the API is a separated app. Does this mean that we have to maintain a demo API app on some server and make travis request it through the UI tests?

I think such tests may be added in a new PR. Merging the current one as is.

PR #4, opened 2015-10-14
Editing

asaunier:
Still todo:
 * [x] route forms
 * [x] feed forms with AJAX
 * [x] client side validating
 * [x] direct request to API using Angular controllers
 * [x] add geom support (at least coordinates inputs for WP)
 * [x] for routes as well
 * [ ] ~~server errors feedback => show messages in form when recording failed?~~ => see #9 and #8
 * [x] redirect to created/updated doc page if success
 * [ ] ~~use templates for errors + fix has-errors~~ => see #9 
 * [ ] ~~when translating a document, the doc exists but not the culture => we must retrieve the figures/geometry without getting a 404 error because the culture does not exist yet~~ => see #10


asaunier:
Interesting resources:
https://scotch.io/tutorials/angularjs-form-validation
https://scotch.io/tutorials/angularjs-form-validation-with-ngmessages
https://scotch.io/tutorials/submitting-ajax-forms-the-angularjs-way
https://docs.angularjs.org/api/ngMessages

asaunier:
The PR is not ready yet but it's close. Feel free to start reviewing.
Actually there will be many things to do to improve the editing but perhaps it's better to make additional PRs and merge this one when it is quite satisfaying.

The approach has somehow changed since the first commits:
* editing forms are no longer filled using the server side but using AJAX requests to the API
* I have created a "documentediting" directive (and its controller) that is added as attributes to the editing form and that takes care of feeding the form and then pushing the form values to the API.
* at this point document types specific controllers are not needed => I have finally dropped them and restored the initial MainController.

As for now the supported features are:
* editing form is fed by AJAX for both waypoints and routes
* basic validating and error displaying (with hard-coded config in the mako templates) are available
* creating a doc and being redirected works

Still todo:
* geometries are not supported yet for WP (nor in routes but they are mandatory for WPs)
* update requests to the API fail (missing attribute?)
* server errors handling is too basic

Something that should be solved (in an other PR?): when translating a document, the doc exists but not the culture => we must retrieve the figures/geometry without getting a 404 error because the culture does not exist yet.

asaunier:
There are still problems with the editing but I have created dedicated issues.

Among them I get validation errors from the API because of some attributes that are marked required (honestly I think some are actually not required even though the tables provided by the client say they are => I will review more precisely the "common" part and make change proposals).

Since this PR is growing to big, I think it is time to review it and possibly merge it before going further.

Please review :)

c2corg_ui/attributes.py
```diff
@@ -43,3 +43,21 @@
     'paragliding_takeoff',
     'paragliding_landing'
 ]
+
+# FIXME: incorrect values
```
tsauerwein: Do you want to use the file from the common project?

These are the change I made to the api project:
https://github.com/c2corg/v6_api/commit/ff5b61cfb9b8c55c109fdd294418afe8f85b887b

asaunier: Yes, it's in the todo list. But I thought it could be done in a separated PR. But no problem doing it in this branch if you think it makes sense.

tsauerwein: No, it's ok. I missed that there is already a ticket for it.

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+ * @param {angular.Scope} $scope Scope.
+ * @param {angular.JQLite} $element Element.
+ * @param {angular.Attributes} $attrs Attributes.
+ * @param {angular.$http} $http
+ * @param {string} apiUrl Base URL of the API.
+ * @constructor
+ * @ngInject
+ * @export
+ */
+app.DocumentEditingController = function($scope, $element, $attrs, $http,
```
tsauerwein: The editing controller is used to edit a document but also to create a new document? Maybe you can add a short note in the comment block (e.g. "Controller to create a new document and to edit an existing document."). Just to make it clear.

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+    );
+  }
+};
+
+
+/**
+ * @const
+ * @type {string}
+ */
+app.DocumentEditingController.VIEW_PROJ = 'EPSG:4326';
```
tsauerwein: `VIEW_PROJ` might be confusing, because for the map you'll also use "epsg:3857".

c2corg_ui/static/js/documentediting.js
asaunier: Yes, the same form is used both for creating a new doc or editing an existing one: the "only" main difference is that the controller makes a request to the API to get the values of the existing doc, which is not needed for a new doc.
The mako template is the same in both cases as well.

I will add such a comment. If relevent I can rename the controller and the form template, but I thought that "editing" made sense, even for creating a doc.

c2corg_ui/static/js/documentediting.js
asaunier: Indeed. What about FORM_PROJ?

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+      response.status + ' ' + response.statusText);
+};
+
+
+/**
+ * @param {boolean} isValid True if form is valid.
+ * @export
+ */
+app.DocumentEditingController.prototype.submitForm = function(isValid) {
+  if (!isValid) {
```
tsauerwein: Maybe it's too early, but using the classes `ng-dirty` and `ng-valid`/`ng-invalid` you can easily indicate if a field is valid/invalid. Something like:

    .ng-dirty.ng-valid {
      border-color: green;
    }
    .ng-dirty.ng-invalid {
      border-color: red;
    }

https://docs.angularjs.org/api/ng/directive/form

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  // push to API
+  var data = this.scope_[this.modelName_];
+  if (!goog.isArray(data['locales'])) {
+    // FIXME
+    // With ng-model="route.locales[0].description" route.locales is taken
+    // as an object instead of an array.
+    var locale = data['locales']['0'];
+    delete data['locales'];
```
tsauerwein: This delete is not needed.

c2corg_ui/static/js/documentediting.js
asaunier: Actually I have used the approach described in this article: https://scotch.io/tutorials/angularjs-form-validation-with-ngmessages#adding-bootstrap-error-classes
See https://github.com/c2corg/v6_ui/blob/editing/c2corg_ui/templates/waypoint/edit.html#L29
You may test it in my instance at http://c2corgv6-demo.gis.internal:6555/waypoints/add

asaunier: Actually the submit button is clickable only if the form is valid:
https://github.com/c2corg/v6_ui/blob/editing/c2corg_ui/templates/waypoint/edit.html#L83

So maybe this "isValid" test in the submitForm() function is not needed or not that important (just to make sure the form is not submitted if not valid).

c2corg_ui/static/js/documentediting.js
asaunier: Right!

c2corg_ui/static/js/documentediting.js
tsauerwein: Nice!

Could you mark the required fields? So that you see right from the start which fields are required?

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+    // TODO: better handling?
+    alert('Form is not valid');
+  }
+
+  // push to API
+  var data = this.scope_[this.modelName_];
+  if (!goog.isArray(data['locales'])) {
+    // FIXME
+    // With ng-model="route.locales[0].description" route.locales is taken
+    // as an object instead of an array.
```
tsauerwein: I think it's because you are not using `ng-repeat`. But `ng-repeat` wouldn't make sense in this case. So this solution seem to be ok for me.

c2corg_ui/static/js/documentediting.js
asaunier: That would make sense indeed. The required fields have a "required" attribute. So I guess it is possible to apply a special style on this attribute. I don't know if we should do this now or wait for the UX/layout instructions...

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+    var locale = data['locales']['0'];
+    delete data['locales'];
+    data['locales'] = [locale];
+  }
+
+  // Convert geometry data
+  if ('geometry' in data && data['geometry']) {
```
tsauerwein: lon/lat are required fields, right? So then why these checks?

c2corg_ui/static/js/documentediting.js
```diff
@@ -0,0 +1,269 @@
...
+ * @return {string} URL.
+ * @private
+ */
+app.DocumentEditingController.prototype.buildUrl_ = function(type) {
+  var url = this.apiUrl_;
+  url += url.substr(-1) != '/' ? '/' : '';
+  url += this.module_;
+  switch (type) {
+    case 'read':
+      url += '/' + this.id_ + '?l=' + this.culture_;
```
tsauerwein: A nice way to construct an URL is also do something like the following:

      this.url_ =
          'http://provelobern-geomapfish.prod.sig.cloud.camptocamp.net/' +
          '{profile}/viaroute?loc={from}&loc={to}&instructions=false&alt=false' +
          '&z={zoom}&output=json';

     var url = this.osrmUrl_
        .replace('{profile}', this.getProfile_())
        .replace('{from}', from[1] + ',' + from[0])
        .replace('{to}', to[1] + ',' + to[0])
        .replace('{zoom}', '20');

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -1,23 +1,94 @@
...
+    <label>Elevation</label>
+    <input type="number" name="elevation" ng-model="waypoint.elevation" class="form-control" required ng-maxlength="4" /> m
+    <div class="help-block" ng-messages="editForm.elevation.$error" ng-if="editForm.elevation.$touched">
+      <p ng-message="required">Elevation is required.</p> 
+      <p ng-message="maxlength">Elevation must be smaller than 9999 m.</p>
+    </div>
+  </div>
+  <div id="longitude-group" class="form-group" ng-class="{ 'has-error': editForm.longitude.$touched && editForm.longitude.$invalid }">
+    <label>Longitude</label>
+    <input type="number" name="longitude" ng-model="waypoint.geometry.geom.longitude" class="form-control" required /> °E
```
tsauerwein: What about setting the `min`/`max` attributes for lon/lat? Also for elevation.

    <input type="number" name="input" ng-model="example.value"
           min="0" max="99" required>

See: https://docs.angularjs.org/api/ng/input/input%5Bnumber%5D

c2corg_ui/static/js/documentediting.js
asaunier: Actually they are required only for WP. Not for books documents for instance, by the way books won't have any geometry.
And for routes, we might have geometries but they are not mandatory and won't use lon/lat fields.

c2corg_ui/static/js/documentediting.js
asaunier: OK, thanks, I felt indeed a bit unhappy with those concatenations and looked for something like the python %. Your way looks fine.

c2corg_ui/templates/waypoint/edit.html
asaunier: Makes sense, thanks for the tips.

c2corg_ui/views/route.py
```diff
@@ -10,28 +11,39 @@ class Route(Document):
...
+
+    @view_config(route_name='routes_add',
+                 renderer='c2corg_ui:templates/route/edit.html')
+    @view_config(route_name='routes_edit',
+                 renderer='c2corg_ui:templates/route/edit.html')
+    def edit(self):
+        try:
+            id, culture = self._validate_id_culture()
+        except Exception:
+            id = None
```
tsauerwein: When will you get in here?
- When you try to edit a document with an invalid id and culture? I think this case should be handled with a generic error handler?
- When you try to add a new document? There are cleaner ways to check if you want to add a new document.

c2corg_ui/static/js/documentediting.js
tsauerwein: Yes maybe.

tsauerwein:
Thanks, that looks good!

c2corg_ui/views/route.py
asaunier: _validate_id_culture() is used to figure out if id/culture parameters have been provided and to check they are of the right form (integer id and culture among the available cultures).
There is no API side check that is done to verify that a document actually exists for the given id/culture (the form directive makes a client side request to the API).
https://github.com/c2corg/v6_ui/blob/editing/c2corg_ui/views/document.py#L32

> When you try to add a new document? There are cleaner ways to check if you want to add a new document.

In that case id/culture are actually None and the Exception is catch to set id/culture to None. That could be done directly in _validate_id_culture(). What cleaner ways to figure out we're creating a new document are you thinking of?

asaunier:
Thanks for reviewing @tsauerwein 
I'm currently addressing your suggestions and will update the PR.

c2corg_ui/views/route.py
tsauerwein: I would write something like:

        if 'id' in self.request.matchdict:
            # creating a new document
            id = None
            culture = None
        else:
            id, culture = self._validate_id_culture()

asaunier:
I have just pushed changes following @tsauerwein suggestions.
I have also made sure that only the current culture is retrieved when loading a document since now the API provides an "available_cultures" attribute in the document (useful to build links towards other available locales of the document). See e5e0488

PR #12, opened 2015-10-29
Use c2corg_common module

### LINKED ISSUE #7: Shared config/attributes between the UI and the API
Some config/attributes are shared between the UI and the API. For instance:
* cultures/langs
* activities
* waypoint_types
* ???

Perhaps also the attributes of the documents, their type and if they are mandatory (for validating in the UI).

As for now we have:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/attributes.py
https://github.com/c2corg/v6_api/blob/master/c2corg_api/attributes.py

@tsauerwein suggested that
> We could have a project "c2corg_common" or "c2corg_constants" in which all constants, attributes, etc. are defined. This project would be a dependency of the api and ui project.

asaunier:
Fix #7


tsauerwein:
Looks good!

PR #14, opened 2015-10-29
Convert WP coordinates before passing them to templates

asaunier:
The conversion is done in the view. I wonder if it should be done in the mako templates instead (since we  will also need the geometry for the map).

### DIFF
--- a/c2corg_ui/templates/waypoint/view.html
+++ b/c2corg_ui/templates/waypoint/view.html
@@ -3,6 +3,7 @@
 <%
 waypoint_id = waypoint['document_id']
 other_cultures = filter(lambda l: l != culture, waypoint['available_cultures'])
+geometry4326 = transform(geometry, 'epsg:3857', 'epsg:4326')
 %>\
 
 <%block name="pagetitle">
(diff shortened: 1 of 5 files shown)


tsauerwein:
> I wonder if it should be done in the mako templates instead (since we will also need the geometry for the map).

Especially since we will have to reproject again to get the coordinates in 'epsg:3857' for the map.

asaunier:
New version: a "transform" function is passed to the template and may be applied on the geometry directly in the template.

Next step is to add tests about geometry reprojecting.

asaunier:
> Next step is to add tests about geometry reprojecting.

Done with 6465fd9

Please review.

tsauerwein:
Sorry, I think I misunderstood you. I thought you would want to do the reprojection in Javascript (which doesn't sound like a good idea in this case in hindsight ;).

I would have passed the geometry in the two different projections to the template. But feel free to leave it like this.

            'geometry': self._get_geometry(waypoint['geometry']['geom']),
            'geometry4326': self.transform_(...)

asaunier:
> I would have passed the geometry in the two different projections to the template.

That's an option indeed. Do you think your approach makes much sense?
I thought that was the responsability of the template to adapt the data according to what it needs to do. We might have to reproject to additional coordinates system as well. For instance EPSG:21781 in Switzerland or even lon/lat in deg/min/sec.
See for instance http://www.camptocamp.org/summits/40507/fr/li-blanche

tsauerwein:
If you see reprojection like for example formatting a number, I think it is ok to call the transformation from the template.

asaunier:
I have pushed a new version of the tests, using assertAlmostEqual and testing the EPSG:21781 projection as well.

Since you seem OK with this PR, I will merge it soon if no last objection is raised.

tsauerwein:
:+1:

PR #15, opened 2015-10-29
Add pages and API calls tests for routes and waypoints

asaunier:
This PR adds a few tests for the UI app.
It's not easy to create relevent tests because:
* we can't assume that the API actually has documents and if yes, what documents.
* creating/editing a doc is done using client-side request to the API, not server-side.

Any ideas of what tests could be added?

As for now I have created similar tests for both routes and waypoints: do you think I should factorize them in the BaseTestUi class and call the factorized methods in WP/Route tests? https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/tests/views/__init__.py#L4

I have also added tests of "private" methods like _call_api or _get_documents. Is it done the right way?
Since both tests pretty much test the same thing (that calling /waypoints from the API returns a list), I wonder if ``test_get_documents`` is really needed.

### DIFF
--- a/c2corg_ui/tests/views/__init__.py
+++ b/c2corg_ui/tests/views/__init__.py
@@ -1,4 +1,5 @@
-from c2corg_ui.tests import BaseTestCase
+from c2corg_ui.tests import BaseTestCase, settings
+from pyramid import testing
 
 
 class BaseTestUi(BaseTestCase):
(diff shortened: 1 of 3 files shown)


tsauerwein:
> we can't assume that the API actually has documents and if yes, what documents.

You can add documents in the tests (using the API).

At some point I think it will also make sense to add tests for the Javascript code.

tsauerwein:
> As for now I have created similar tests for both routes and waypoints: do you think I should factorize them in the BaseTestUi class and call the factorized methods in WP/Route tests?

That would make sense.

asaunier:
I have pushed a new version that factorizes the code of the tests.

> You can add documents in the tests (using the API).

I will think about it. Isn't it a problem if the tests add new documents in the API? I mean it will not add them in the API tests DB but in the standard DB. Not really a big deal for developers instances but what if the tests are run on the prod for instance?

> At some point I think it will also make sense to add tests for the Javascript code

Indeed. Most likely not in this PR :P

asaunier:
What about merging the PR as is? Document creation tests using the API could be added later and the current PR would be useful to add tests in PR #14 .

tsauerwein:
Sure, go ahead.

PR #19, opened 2015-11-05
Editing form fixes

asaunier:
Commit e2f0914 makes sure that the changes done on the data (for instance converting the coordinates to some GeoJSON) do not impact the model. Which may result in changing the form inputs content if the form submit fails.

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -222,11 +222,6 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
     }
   }
 
-  // FIXME: workaround for WP
-  if (!('activities' in data && data['activities'])) {
-    data['activities'] = [];
-  }
-
   var config = {
     'headers': { 'Content-Type': 'application/json' }
   };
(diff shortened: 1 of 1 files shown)


c2corg_ui/static/js/documentediting.js
```diff
@@ -196,7 +199,7 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
   }
 
   // push to API
-  var data = this.scope_[this.modelName_];
+  var data = angular.copy(this.scope_[this.modelName_]);
```
asaunier: Or perhaps ``goog.object.clone`` instead of ``angular.copy``?

asaunier: It seems it is not equivalent.
``goog.object.clone`` does not prevent the geometry field of the model from being updated when changing it in the copy. Maybe because it is only a flat clone:
https://closure-library.googlecode.com/git-history/docs/namespace_goog_object.html
whereas ``angular.copy`` is a deep copy:
https://docs.angularjs.org/api/ng/function/angular.copy

asaunier:
Ready for reviewing.

asaunier:
Any comment for this PR? If not I would like to merge it as soon as possible.

tsauerwein:
:+1:

PR #22, opened 2015-11-05
Map directive

asaunier:
This PR creates a dedicated directive/controller to manage maps.

As for now it is possible to provide the coordinates of a waypoint and have the map recentered on it.

It would be interesting to be able to link a model in order to update show the the WP in the editing page as well, update its position when coords are changed (and of course update the form when a point is clicked on the map).

The current feature feeding is a little too simple as for now (especially when showing a route). We might need to be able to feed the features using an API call that would return a FeatureCollection of the object and its associated object.
Or to get the geometries of the features in a list (index) page.
Most likely done in another PR.

### DIFF
--- a/c2corg_ui/static/js/map.js
+++ b/c2corg_ui/static/js/map.js
@@ -3,15 +3,18 @@ goog.provide('app.mapDirective');
 
 goog.require('app');
 goog.require('ngeo.mapDirective');
+goog.require('ol.Feature');
 goog.require('ol.Map');
 goog.require('ol.View');
+goog.require('ol.geom.Point');
 goog.require('ol.layer.Tile');
+goog.require('ol.layer.Vector');
 goog.require('ol.source.OSM');
+goog.require('ol.source.Vector');
 
 
 /**
- * This directive gets a reference to the map instance through the "app-map"
- * attribute.
+ * This directive is used to display a pre-configured map in v6_ui pages.
  *
  * @return {angular.Directive} The directive specs.
  * @ngInject
@@ -19,9 +22,9 @@ goog.require('ol.source.OSM');
 app.mapDirective = function() {
   return {
     restrict: 'E',
-    /*scope: {
-      'map': '=appMap'
-    },*/
+    scope: {
+      'center': '=appMapCenter'
+    },
     controller: 'AppMapController',
     controllerAs: 'mapCtrl',
     bindToController: true,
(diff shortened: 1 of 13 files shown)


c2corg_ui/static/js/map.js
```diff
@@ -0,0 +1,106 @@
...
+        features: [feature]
+      })
+    });
+
+    // Use vectorLayer.setMap(map) rather than map.addLayer(vectorLayer). This
+    // makes the vector layer "unmanaged", meaning that it is always on top.
+    vectorLayer.setMap(this.map);
+
+    var view = this.map.getView();
+    view.setCenter(center);
```
tsauerwein: Why are you setting center/zoom again here?

tsauerwein:
Thanks, I think that's the right direction.

c2corg_ui/static/js/map.js
asaunier: You're right, it's useless. Thanks for catching this!

asaunier:
Thanks for reviewing. I have pushed a fix for your comment.

I would like also to see if I can link the directive to the editing form model.

asaunier:
> I would like also to see if I can link the directive to the editing form model.

Done with 5067b59
At least it display the waypoint on the map when the editing form has finished loading the waypoint model from the API.
WP position on the map is not updated (yet) when the lon/lat values are changed in the form. And vice versa.

I think the PR is ready for merging.

Just one thing I'm wondering: the WP geometry is loaded from the API in EPSG:3857, then converted to 4326 for displaying in the form. To avoid having tons of decimals, those lon/lat are rounded at 6 digits. Will it cause slight differences when reprojecting to 3857 and, as a result, a new version in the database even though the user has not changed the coordinates?
See https://github.com/c2corg/v6_ui/blob/map-directive/c2corg_ui/static/js/documentediting.js#L167-L172

tsauerwein:
> Will it cause slight differences when reprojecting to 3857 and, as a result, a new version in the database even though the user has not changed the coordinates?

When the geometry is included in an update request, a new geometry version is always created. If the geometry is not included in the request, it will not change the geometry. So, it's in the responsibility of the client to only send the geometry if changes have been made.

asaunier:
Mmmm OK.
At the moment the geometry is always sent to the API even though no change has been made (at least for WP since the coords are shown in the form, which will not be the case for lines/polygons). Which means a new geom version is created every time the form is posted.
I will create an issue about this.

PR #28, opened 2015-11-10
Show list items on the map

asaunier:
This PR adds a FeatureCollection object in the waypoint list page, it is then passed to the map controller in order to show the features on the map.

As for now the collection is passed using a global JS variable and the Angular $window object. Perhaps using a module value (or a service?) would be cleaner but I have not succeeded yet: I got an error saying that the ``app`` module does not (yet?) exist when writing
```
var module = angular.module('app');
module.value('mapFeatureCollection', ...);
```
in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/index.html
Perhaps because this code should be rather added at the end of the HTML page?

Geometries must be provided in the list response from the API => relies on https://github.com/c2corg/v6_common/pull/5

### DIFF
--- a/c2corg_ui/templates/waypoint/index.html
+++ b/c2corg_ui/templates/waypoint/index.html
@@ -1,30 +1,32 @@
-<%inherit file="../base.html"/>
+<%inherit file="../base.html"/>\
 
 <%block name="pagetitle">
 Waypoints 
-</%block>
+</%block>\
+
+<%block name="moduleConstantsValues">
+% if len(waypoints):
+module.value('mapFeatureCollection', {
+  "type": "FeatureCollection",
+  "features": [
+    <%
+    from json import dumps, loads
+    features = [];
+    for waypoint in waypoints:
+        features.append({
+            "type": "Feature",
+            "geometry": loads(waypoint['geometry']['geom'])
+        })
+    features = map(lambda f: dumps(f), features)
+    %>
+    ${', '.join(features) | n}
+  ]
+});
+% endif
+</%block>\
 
 <h1>${self.pagetitle()}</h1>
 % if len(waypoints):
-  <script>
-    // FIXME: use service or module.value
-    var mapFeatureCollection = {
-      "type": "FeatureCollection",
-      "features": [
-        <%
-        from json import dumps, loads
-        features = [];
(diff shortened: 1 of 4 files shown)


asaunier:
The PR is (almost?) ready to be reviewed.

With no surprise Travis fails as reported in an other PR (https://github.com/c2corg/v6_ui/pull/29).

Perhaps the code could be improved concerning the ways to provide the features to display on the map. As for now there are 3 ways:
* setting a FeatureCollection in a ``mapFeatureCollection`` variable that is then retrieved using ``$window['mapFeatureCollection']`` => mmm perhaps we could use a cleaner solution than a JS global variable.
* setting coordinates in an ``app-map-center`` attribute => that's easy and fast (used for showing the geolocalisation of a WP in its detail page) but I wonder if using the FeatureCollection above - though more verbose - would not be better, especially if we would like to pass properties such as the waypoint_type and the title. To be removed?
* providing a ``DocumentEditingController`` instance: its model is used to display the edited waypoint on the map, including when the coords are updated.

asaunier:
> Perhaps the code could be improved concerning the ways to provide the features to display on the map.

I have made changes so that one may provide a "mapFeatureCollection" module.value instead of the JS global var, and use it for the WP view map instead of the "center" attribute (removed).

asaunier:
Travis passes! :)
OK if I merge the PR?

asaunier:
If no one objects by then, I will merge the PR on friday morning when I'm back from Basel.

c2corg_ui/templates/waypoint/index.html
```diff
@@ -1,11 +1,34 @@
...
+<%inherit file="../base.html"/>\
 
 <%block name="pagetitle">
 Waypoints 
-</%block>
+</%block>\
+
+<%block name="moduleConstantsValues">
+% if len(waypoints):
+module.value('mapFeatureCollection', {
```
tsauerwein: I guess this can be refactored in a separate method once we need it for other pages.

asaunier: Right. I haven't done it yet because I thought we should wait for the integration of the other doc types.

PR #30, opened 2015-11-12
Fix response status check

asaunier:


tsauerwein:
Looks good.

asaunier:
Travis still fails...
Where do those ``Bad response: 500 Internal Server Error (not 404)`` come from in Travis reports?
I have no such errors when running the tests in my instance.

asaunier:
@tsauerwein Do you think we still need this PR after the changes you have made to fix Travis tests?

tsauerwein:
It is still needed, because otherwise it will fail when the API server does not respond or responds with an error.

gberaudo:
+1 for merging.

Issue #6, opened 2015-10-15
Localisation of the interface


asaunier:
As for now, no translation of the interface strings are done (except a small example in the home page, based upon the usual angular-gettext tool provided by pyramid_closure).

That's a point we will have to deal with. 2 approachs are possible:
- making UI **server side translations using gettext** in mako templates
  - pros: simple and reliable, translations are immediatly available at page loading since in the HTML code
  - cons: HTML code must be generated for each language (remember that the pages URLs do not include parameter, only the document cultural version) 
- making **client-side translations (in javascript) using angular-gettext** as provided by pyramid_closure (and with ngeo for the maps)
  - pros: only one HTML code generated for all languages, no need to reload the page if the user switch the interface language, base code provided by pyramid_closure, needed for ngeo (maps)
  - cons: translations might be done with a little delay since the lang dictionary must be dowloaded (size of the files?? => how many strings must be translated?) in addition to the current page (only for the first page?) and then applied (default strings are in english)

asaunier:
Interesting resource: http://nimbupani.com/declaring-languages-in-html-5.html

In linked PR, the client-side approach is used => interface strings are translated on the fly using angular-gettext (in javascript).

asaunier:
TODO:
* [ ] add ``lang="en"`` attribute to interface strings elements since they are in english even though the main content might be in an other language (``<html lang="...">``uses the document lang).

asaunier:
The interface localisation uses angular-gettext. This issue could be closed but I just wonder if we should do something about the ``lang`` attribute as described in the comment above.

PR #31, opened 2015-11-12
Use angular-gettext to translate interface strings

asaunier:
Use built-in angular-gettext [1] to translate the interface strings in french.

Some points still need to be fixed:
* [x] form input placeholders or logo title
* [ ] ng-messages (form validation errors)
* [x] values of ng-repeat combos (for ex in editing form)
* [x] strings containing Mako variables (for ex ``${total} waypoints``) 
* [x] use doc culture as ``<html lang="{lang}">`` instead of interface culture
* [x] add interface culture selector

Related to https://github.com/c2corg/v6_ui/issues/6

[1] https://angular-gettext.rocketeer.be/dev-guide/getting-started/

### DIFF
--- a/c2corg_ui/static/js/maincontroller.js
+++ b/c2corg_ui/static/js/maincontroller.js
@@ -7,32 +7,14 @@ goog.require('app');
 /**
  * @param {angular.Scope} $rootScope The rootScope provider.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
- * @param {string} langUrlTemplate Language URL template.
  * @constructor
  * @export
  * @ngInject
  */
-app.MainController = function($rootScope, gettextCatalog, langUrlTemplate) {
+app.MainController = function($rootScope, gettextCatalog) {
 
   /**
-   * @type {angularGettext.Catalog}
-   * @private
-   */
-  this.gettextCatalog_ = gettextCatalog;
-
-  /**
-   * @type {string}
-   * @private
-   */
-  this.langUrlTemplate_ = langUrlTemplate;
-
-  /**
-   * @type {string}
-   * @export
-   */
-  this.lang;
-
-  /**
+   * TODO: use langController instead?
    * Put angular-gettext's translate function on the scope to translate
    * ng-options based dropdown lists. See
    * https://github.com/rubenv/angular-gettext/issues/58
@@ -42,20 +24,6 @@ app.MainController = function($rootScope, gettextCatalog, langUrlTemplate) {
   $rootScope['translate'] = function(str) {
     return gettextCatalog.getString(str);
   };
-
-  this.switchLanguage('fr');
-};
-
-
-/**
- * @param {string} lang Language code.
- * @export
- */
-app.MainController.prototype.switchLanguage = function(lang) {
-  this.gettextCatalog_.setCurrentLanguage(lang);
-  this.gettextCatalog_.loadRemote(
-      this.langUrlTemplate_.replace('__lang__', lang));
-  this.lang = lang;
 };
 
 
--- a/c2corg_ui/static/js/maincontroller.js
+++ b/c2corg_ui/static/js/maincontroller.js
@@ -5,13 +5,14 @@ goog.require('app');
 
 
 /**
+ * @param {angular.Scope} $rootScope The rootScope provider.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @param {string} langUrlTemplate Language URL template.
  * @constructor
  * @export
  * @ngInject
  */
-app.MainController = function(gettextCatalog, langUrlTemplate) {
+app.MainController = function($rootScope, gettextCatalog, langUrlTemplate) {
 
   /**
    * @type {angularGettext.Catalog}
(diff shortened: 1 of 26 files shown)


asaunier:

It uses the angular-gettext based translation system that comes with pyramid_closure. Pages are translated on the fly using javascript replacements and lang detection. The server side is not aware of the user culture at all. Which is consistent with the choice of having a stateless server side UI, knowing nothing about the user (including his ID). The pages will be generated only once for all languages (and most likely cached using Varnish).

Drawbacks are that the client browser has more work to do to render the page (really a problem?) and that JSON translation files (downloaded by angular-gettext when the page lang is changed) might grow a little big depending on the numbers of strings to translate.

See my instance for a demo: http://c2corgv6-demo.gis.internal:6555/

I have addressed most of the TODO items even though I have not succeeded in having the form validation error messages translated yet (?) (pb with ng-messages?).

The "lang" attribute in ``<html lang="...">`` is now set to the document culture currently shown, for "view" pages. For other pages (list, editing => not reachable by search engine anyway), "lang" is set to "mainCtrl.lang". I just realize that I have dropped this property from the main controller but anyway I am not sure it is really relevent to have such a value for "lang" since Angular properties are not supposed to be interpreted by search engines => should I remove "lang" for those pages?

In the server generated HTML, interface strings are then in english. Is it a problem for SEO if the interface uses a different language then the content? We could minimize the problem (and reduce a bit the pages sizes and i18n dictionaries sizes) by using minimalistic msgids instead of real sentences/expressions (EN actually has a translation file).

I have moved most of the lang-related code from App.MainController to App.LangController. A "translate" function remains for the "ng-options" translations (or maybe I can use the "translate" function from langCtrl?). App.MainController is now a little empty. I don't know yet what we will use it for, maybe for linking the different subcontrollers together?

Points that still need to be covered:
* currently the user browser languages prefs are used to figure the best culture, for every single page => we should do it only once and then save the browsing language to some web storage about the user (probably something to link with the user auth).
* the culture dropdown value is not correctly initialized even though the user prefs are actually detected. Don't know why the dropdown is not updated using its ng-model?!

c2corg_ui/static/js/maincontroller.js
```diff
@@ -5,45 +5,25 @@ goog.require('app');
...
-/**
- * @param {string} lang Language code.
- * @export
- */
-app.MainController.prototype.switchLanguage = function(lang) {
-  this.gettextCatalog_.setCurrentLanguage(lang);
-  this.gettextCatalog_.loadRemote(
-      this.langUrlTemplate_.replace('__lang__', lang));
-  this.lang = lang;
+  $rootScope['translate'] = function(str) {
```
gberaudo: In the `langController` you are prefixing the translate function by the controller name.
Would it work to put it on the main controller and prefix by the main controller name instead?

c2corg_ui/templates/base.html
```diff
@@ -25,10 +25,10 @@
   <body>
 
     <div id="page-header">
-    <a href="${request.route_url('index')}" title="Back to homepage"><img src="${request.static_url('c2corg_ui:static/img/logo.png')}" alt="Camptocamp.org" /></a>
+    <a href="${request.route_url('index')}" title="{{'Back to homepage'|translate}}"><img src="${request.static_url('c2corg_ui:static/img/logo.png')}" alt="Camptocamp.org" /></a>
     <ul>
-      <li><a href="${request.route_url('waypoints_index')}">Waypoints</a></li>
-      <li><a href="${request.route_url('routes_index')}">Routes</a></li>
+      <li><a href="${request.route_url('waypoints_index')}" translate>Waypoints</a></li>
```
gberaudo: It is better to use `ng-href`.

asaunier: Why? This is a standard link, not an expression. And it's better to use HTML links so that search engines may access those pages.

gberaudo: Right, thought it was an angular expression.

c2corg_ui/static/js/maincontroller.js
asaunier: The function is indeed duplicated. That would indeed make sense to have it only once.
I have tried to use ``$rootScope.translate()`` in the langDirective but with no luck. Your solution might work, actually I would like to find a way to completely remove the culture-related code from the main controller and have every thing in the langController. Maybe I could simply move ``$rootScope['translate'] = function(str) {`` in the langController instead of the main controller?

tsauerwein:
This is looking good, thanks!

asaunier:
In the last commits I have removed ``lang="{mainCtrl.lang}"`` from the html opening tag (still set to the doc culture for the main doc pages) since mainCtrl has no longer a "lang" property and that this value does not seem that relevent anyway.

I have removed the "translate()" function from the MainController, only using the one available in LangController. MainController is now an empty shell that does nothing... I guess we might need it later so I don't remove it yet.

Merging when Travis approves.

asaunier:
> I have removed the "translate()" function from the MainController, only using the one available in LangController.

Oops, actually it doesn't work. I had to add back a translate() function to the MainController:
https://github.com/c2corg/v6_ui/commit/92a985cdaa28eda3ab0f99e8d637ec7adf18213e

I still can't figure out why the lang dropdown is not updated when the user's browser lang prefs are detected. The ``<select>``tag actually has a ``ng-model="langCtrl.culture"``attribute and ``this.culture is updated when the user's lang is detected: 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/lang.js#L64

Is there something missing somewhere? A $scope.$apply() for instance?

PR #34, opened 2015-11-25
Pagination

asaunier:
As for now no pagination link is available in the list pages. But it is possible to add pagination filters to the list URL (for instance ``/waypoints/limit/10/offset/500``) which will be passed as query string params in the API URL (for instance ``/waypoints?limit=10&offset=500``).

With commit 04cd9a2ac8118678f27530ea7f78d447757fad98 I have set more specific routes adding validation on the matchdict params (id = integer, culture in the available cultures list). It's nice but it makes the tests fail because incorrect document URL such as ``waypoints/1/foo`` or ``waypoints/foo/fr``) are not caught by the "view" route but by the "index" route, taking ``1/foo`` or ``foo/fr`` as filter params instead of document ids => I think I'll have to revert the change and stick to the old route and id/culture validation in the view itself.

Next step is to generate the "next/privious" links in the UI list pages + adding some tests about the filters detection.


tsauerwein:
> It's nice but it makes the tests fail because incorrect document URL such as waypoints/1/foo or waypoints/foo/fr) are not caught by the "view" route but by the "index" route, taking 1/foo or foo/fr as filter params instead of document ids

Would it change something if you first define the route for `view` and then the route for `index`? Because the routes are matched in the order in which you define them.

asaunier:
> Would it change something if you first define the route for view and then the route for index? Because the routes are matched in the order in which you define them.

The "index" route is already at the end, because ``*filters`` in ``/waypoints*filters`` catches every route starting by ``waypoints`` (including the "view" URL).

asaunier:
I'm a bit in trouble about those routing issues.

With commit 04cd9a2 and following we have
```
    id = '{id:\d+}'
    culture = '{culture:(%s)}' % '|'.join(default_cultures)

    config.add_route('waypoints_view', '/waypoints/%s/%s' % (id, culture))
    config.add_route('waypoints_index', '/waypoints*filters')
```
Which means that only URL like ``waypoints/1234/fr`` are detected as a "view" page. If the id is not an integer or the culture is not among the defined list of cultures, errors are not raised (as it was the case as for now) and the URL is taken as an "index" page (list), with the last parts of the URL understood as index "filters" (which they are not).

On the other hand, if I rather use:
```
config.add_route('waypoints_view', '/waypoints/{id}/{culture}')
config.add_route('waypoints_index', '/waypoints*filters')
```
id and culture are correctly detected and errors are raised (by checks in the view code) as expected in case they are wrong. But if an "index" URL only has 2 parts (eg. ``waypoints/limit/30``), it will actually match the "view" route with id=limit and culture=30...

I think the id/culture type checks in the route definitions are required (first case above) even though no error is raised when those id/culture are incorrect (in v5 there a 404 error is returned).
In addition we will have to create additional "view" routes since URL such as ``waypoints/1234`` or ``waypoints/1234/fr/some-slug-string`` should be accepted too. If I had those routes without type checks, we'll get even more routing conflicts.

Another solution would be to use a "singular" form for "view" pages. For instance ``waypoint/1234/fr`` instead of ``waypoints/1234/fr`` (no "s" at "waypoint"). Could it create some other problem?
One problem would be that it would break the backward (ie. v5>v6) compatibility for view URLs. Well since "summits", "huts", ... URLs are anyway replaced by "waypoints" URLs we will have to add some rewrite rules conversions...

@tsauerwein @gberaudo What do you think?

gberaudo:
The solution distinguishing singulars and plurals looks reasonable to me.

asaunier:
I have pushed additional commits and dropped the intial routing commit 04cd9a2 => routing now distinguishes "singular" routes (view/edit/add) and "plural" ones (index).

Review welcome!

c2corg_ui/views/document.py
```diff
@@ -68,13 +74,34 @@ def _get_document(self, id, culture):
...
-            return [], 0
+            documents = []
+            total = 0
+        return documents, total, filters
+
+    def _get_filter_params(self):
+        params = []
+        if 'filters' in self.request.matchdict:
+            filters = self.request.matchdict['filters']
+            # If number of filters is odd, add an empty string at the end:
```
gberaudo: Not sure to understand. Which case is it?

asaunier: I though we might have the case of filter params without value (a bit like ``?debug``). I don't remember if this is already possible in v5.
Maybe you're right and this is not useful.

asaunier:
I realize that my changes about the filters params are a bit confusing. With e7eaa89 I have added some comments to make the filters processing a bit more explicit.

@gberaudo I prefer to keep the oddity check (even though the case is quite unlikely) because the filters key-value grouping creates an error if an odd number of params is provided.

gberaudo:
@asaunier, I prefer code that fails early but feel free to keep it your way.

asaunier:
> routing now distinguishes "singular" routes (view/edit/add) and "plural" ones (index).

@gberaudo made me realize that v5 uses an explicit "list" keyword in list pages URLs (eg. ``routes/list/area/1234-5678``) even though the base URL (``routes`` - with no filter) works as well.
I have then added this "list" keyword and reverted the "singular"/"plural" changes in the URL.

asaunier:
I think the PR is ready now. Any objection against merging it?

gberaudo:
LGTM

Issue #8, opened 2015-10-27
Add a generic messaging system for the actions feedback


asaunier:
A generic tool that would display various messages coming from the UI application (error messages, confirmations of a successful action, welcome message after logging in, etc.) would be useful.

asaunier:
Not sure it should be a separated tool or not, but we also need a "waiting" message (with a spinner?) displayed when an AJAX request or a redirection is done.

tsauerwein:
For the loading indicator, this might be useful:
http://stackoverflow.com/a/15976991/119937
https://github.com/cgross/angular-busy

asaunier:
For actions feedback (AJAX requests etc.) that require displaying some message (error, confirmation, etc.), we might use Bootstrap's alert, showing a small (closable/vanishable) rectangle in some page corner:
http://angular-ui.github.io/bootstrap/#/alert

What do you think?
Do you have other tools to suggest?

gberaudo:
There is also http://bootboxjs.com/ we used for schweizmobil.

asaunier:
Thanks for the tip @gberaudo 
We need non modal messages (that can be closed by the user or autofade after some timeout), else they would be too obstrusive. Not sure those bootboxes can do (I have tested quite fast)?

gberaudo:
There is a cross to close the alert boxes, would it do?
It is possible to set a timeout and call `hide` on the box.
Don't know about fading.

asaunier:
> There is also http://bootboxjs.com/ we used for schweizmobil.

OK bootbox might be OK for showing feedback messages or modals in some cases.

I have "reviewed" a bit the AngularUI Boostrap tools as well as other [AngularUI modules](http://angular-ui.github.io/) and they might be interesting for other features we will need. For instance:
* Carousel http://angular-ui.github.io/bootstrap/#/carousel for showing images in the documents
* Datepicker http://angular-ui.github.io/bootstrap/#/datepicker
* Modal http://angular-ui.github.io/bootstrap/#/modal
* Popover http://angular-ui.github.io/bootstrap/#/popover for inline help messages
* Progressbar http://angular-ui.github.io/bootstrap/#/progressbar for image uploading or document creation process status?
* Tabs http://angular-ui.github.io/bootstrap/#/tabs for document content?
* Typeahead http://angular-ui.github.io/bootstrap/#/typeahead for autocompletion (is it what we use for the simple search widget?)
[...]

PR #36, opened 2015-11-25
Add sign in/up/out actions

asaunier:
* [x] Create minimal user directive to handle user-related tools in page header
* [x] Create sign in button redirecting to auth interface
* [x] Create base auth page
* [x] Add sign in form in auth page (with AJAX request to API)
* [x] Add sign up form in auth page (with AJAX request to API)
* [x] Store user info (username, role, lang) in local storage
* [x] Show username when authenticated
* [x] Add sign out button when authenticated
* [x] Use consistent routes/URLs name with API: register/login/logout
* [ ] Show errors if login or register failed
* [ ] "Remind me" checkbox 
* [x] Whitelist login/register requests
* [x] Redirect to original page after login
* [x] Show register form only on request


gberaudo:
@asaunier, you can rebase onto master to get ride of the ngeo errors.

asaunier:
Yep, done

gberaudo:
You are using `signin/signup`, whereas in the API it is `login/logout`. Do we want to use different names?

I started implementing storage and injection in [this branch](https://github.com/c2corg/v6_ui/tree/auth_token). It already works, with an hardcoded token that you can generate manually with:
- register a user on API: curl --data '{"username": "guillaume", "password":"toto", "email":"s@h.co"}' 'http://127.0.0.1:6546/users/register' -H 'Accept: application/json' -H 'Content-Type: application/json; charset=UTF-8'
- get a token from API: curl --data '{"username": "guillaume", "password":"toto"}' 'http://127.0.0.1:6546/users/login' -H 'Accept: application/json' -H 'Content-Type: application/json; charset=UTF-8'

It is designed so that we can easily change the auth mechanism later if/when we want to switch to
cookie based authentication.

On the API, I did:
```diff
diff --git a/config/dev b/config/dev
index 322cd26..f0be914 100644
--- a/config/dev
+++ b/config/dev
@@ -4,5 +4,5 @@ export version = 0
 
 host = c2corgv6-demo.gis.internal
 
-export cors_origins = $(url):*
-export noauthorization = True
+export cors_origins = *
+export noauthorization = False
```

asaunier:
I have also noticed that we used different names. That would be indeed better to user consistent names. No problem to use yours: register/login/logout
I'll make a change in my branch.

c2corg_ui/static/js/auth.js
```diff
@@ -0,0 +1,114 @@
...
+
+
+/**
+ * @export
+ */
+app.AuthController.prototype.signIn = function() {
+  this.http_.post(this.buildUrl_('login'), this.scope_['signIn'], {
+    'headers': {
+      'Content-Type': 'application/json; charset=UTF-8',
+      'Accept': 'application/json, text/plain, /'
```
asaunier: @gberaudo Is the ``Accept`` value correct? (what about the trailing ``/``?). I have copied it from one of your jabber messages.

gberaudo: @asaunier, have you tried with only 'application/json'?

asaunier: OK thanks, that works with this value.

asaunier:
> No problem to use yours: register/login/logout

Done

Login and register actions are now functional, it's possible to get a token from the API using the UI. We know have to store it as you suggested.

Among things to do: handle errors returned by the API (eg. auth failed). We need the generic "display message" tool mentioned in issue https://github.com/c2corg/v6_ui/issues/8

gberaudo:
@asaunier, OK I will integrate my commits in your branch.

gberaudo:
@asaunier, I created https://github.com/c2corg/v6_api/pull/88 to fix expiration handling and moved logout.

asaunier:
@gberaudo I have noticed that when POSTing to ``users/login`` or ``users/register`` in the API, the Authentication service complains because the user tries to request those URL in POST whereas users are obviously not authenticated when trying to login :p 
Should we consider this case or do you think it's not a big deal?

gberaudo:
@asaunier, yes we can whitelist these two paths.

asaunier:
> yes we can whitelist these two paths.

Could you take care of this please?

asaunier:

3 points are still opened:
* Show errors if login or register failed => I think it should be postponed when we the message system is ready, see issue #8
* "Remind me" checkbox => as for now, the API always reminds token. see issue #42
* Whitelist login/register requests

Concerning the last point it seems that other URL are intercepted as well. When loading the UI I see in my dev console:
```
FIXME auth false /static/partials/user.html
FIXME auth false http://c2corgv6-demo.gis.internal:6555/static/build/locale/fr/c2corg_ui.json
```
Perhaps that's OK? In that case we could remove the ``console.log('FIXME auth', needed, url);`` line in https://github.com/c2corg/v6_ui/blob/auth/c2corg_ui/static/js/httpinterceptor.js#L30 ?
By the way I feel a bit uncomfortable with those ``console.log()`` in the source code. Will we keep them or should we remove them in this PR?

@gberaudo Can we drop branch "auth_token" https://github.com/c2corg/v6_ui/tree/auth_token ?
I think it is no longer need since we have cherry-picked your changes in the current branch.

c2corg_ui/static/js/auth.js
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * @param {Object} response Response from the API server.
+ * @private
+ */
+app.AuthController.prototype.successLogin_ = function(response) {
+  this.appAuthentication_.setUserData(response['data']);
+  // redirect to previous page
+  var url_from = this.ngeoLocation_.hasParam('from') ?
+      goog.string.urlDecode(this.ngeoLocation_.getParam('from')) : '/';
```
gberaudo: Have you tried using `decodeURIComponent`?

asaunier: What's the problem with ``goog.string.urlDecode``?

c2corg_ui/static/js/auth.js
```diff
@@ -0,0 +1,176 @@
...
+};
+
+
+/**
+ * @param {string} action Action.
+ * @return {string} URL.
+ * @private
+ */
+app.AuthController.prototype.buildUrl_ = function(action) {
+  return '{base}/users/{action}'
```
gberaudo: `base` is constant. Is the function really needed?

asaunier: It's a way to avoid having a bunch of concat signs which are quite unreadable.

asaunier:
@gberaudo Thanks for your changes and OK for the ``goog.DEBUG`` solution with ``console.log()``.

With ea73a9e I have moved the ``setUserData`` to group methods together. I have also added some missing ``@export`` keywords, make the ``needAuthorization`` code a bit clearer (not easy not to get confused when too many conditions are tested at once) and remove the initial test in ``addAuthenticationHeaders`` because it is already done in ``needAuthorization`` (or else we should call ``needAuthorization`` directly inside ``addAuthenticationHeaders``. OK for you?

By the way, I was wondering if it would not make sense to rename ``addAuthenticationHeaders`` to ``addAuthorizationHeaders`` (or even "...Header" with no "s") since the added header is ``Authorization``? So would also be more consistent with ``needAuthorization``. What do you think?

c2corg_ui/static/js/auth.js
gberaudo: It relies on the goog library instead of standard javascript.
The less we depend on goog now, the easier it will be to get ride of it in the future.

c2corg_ui/static/js/authservice.js
```diff
@@ -0,0 +1,181 @@
...
+   */
+  this.userData_ = null;
+};
+
+
+/**
+ * @return {boolean}
+ * @export
+ */
+app.Authentication.prototype.isAuth = function() {
```
gberaudo: Have you considered using the full name `isAuthenticated`?

c2corg_ui/static/js/auth.js
asaunier: OK, I will test if it works as you suggest.

c2corg_ui/static/js/authservice.js
asaunier: Well, since exported functions are not shortnamed by closure, I wanted to use a short name. We could of course rename it as you suggest it but is it really that confusing?

c2corg_ui/static/js/auth.js
asaunier: ``goog.string.urlDecode`` is based upon ``decodeURIComponent`` but adds an additional ``+`` replacement:
https://closure-library.googlecode.com/git-history/docs/local_closure_goog_string_string.js.source.html#line437
I think it is safe to use the standard function instead.
Ditto for the encoding function, defined a few lines above.

Done at f84e3e1

c2corg_ui/static/js/authservice.js
asaunier: > Have you considered using the full name isAuthenticated?

Done at 7691dd3

tsauerwein: > Well, since exported functions are not shortnamed by closure, I wanted to use a short name.

Don't worry about this. Internally a "shortnamed" name will be used and only at the end there will be something like "app.Authentication.prototype.isAuth = Xy.fo".

Additionally the html and js files will be gzipped, so there is no need to use shorter variable or function names! :)

c2corg_ui/static/js/auth.js
gberaudo: OK, as you wish.

gberaudo:
@asaunier, I added the type to auth data and made it immutable.
Now it is returned directly to the calling code.

I changed the way the user data is loaded from the storage and added multi-tabs
support: if you have several tabs opened and you login/logout  in one of them, then the other tabs
are notified and all UIs are updated.

I ended up using `app.Authentication.addAuthorizationToHeaders` and clarified even more the
documentation. The method is public, that is why it needs the url check; I restored it. I think it is critical to prevent authorization data to be sent to external urls.

asaunier:
> I ended up using app.Authentication.addAuthorizationToHeaders and clarified even more the
documentation. The method is public, that is why it needs the url check; I restored it. I think it is critical to prevent authorization data to be sent to external urls.

You're right but then it means that the test is made twice since you already call "needAuthorization" before calling "addAuthorizationToHeaders". By the way your embedded test is uncomplete since you only test the passed URL is an API URL.
Why not embedding "needAuthorization" in "addAuthorizationToHeaders" instead of calling it in the httpinterceptor?

asaunier:
> Why not embedding "needAuthorization" in "addAuthorizationToHeaders" instead of calling it in the httpinterceptor?

Discussed verbaly with @gberaudo => his proposition makes sense.

gberaudo:
Added token renewal.

asaunier:
> Added token renewal.

I might have found an error (?), see previous comment. Else looks good to me.

asaunier:
:+1:

gberaudo:
Remember me added in #46.

Issue #37, opened 2015-11-26
Upgrade AngularJS and extensions


asaunier:
We use Angular 1.3.5 (default version shipped with pyramid_closure) whereas latest angular version is 1.4.8 (1.4.7 in ngeo).

We also use angular-gettext 1.1.3 whereas latest version is 2.1.2 according to https://github.com/rubenv/angular-gettext/blob/master/bower.json

gberaudo:
I have just done this on my branch.

PR #39, opened 2015-11-26
Update angular and angular-gettext

gberaudo:


gberaudo:
Related to #37.

gberaudo:
The update of angular creates a regression with selects translation not updated anymore.
With @asaunier, language select has been `fixed` by removing one time binding.

asaunier:
If no other features is broken by this update (still possible to create a WP?), I think it's OK to merge the PR.

I have just created an issue about the i18n problem:
https://github.com/c2corg/v6_ui/issues/40

asaunier:
I have just rebased the PR and fixed a conflict in the Makefile.
Tests pass (still waiting for Travis) and the UI still works as expected except that the dropdown lists in the editing forms are no longer translated (already reported in issue #40 ).

I suggest that we merge the PR since we won't stick to Angular 1.3 and need a recent version.
@gberaudo @tsauerwein OK for merging?

gberaudo:
+1

PR #45, opened 2015-12-04
Add config/gberaudo_remote

gberaudo:


asaunier:
I don't think you need a review for that one :P

gberaudo:
;)

Issue #42, opened 2015-12-01
Manage "remind me" case when a user authenticates


asaunier:
Actually a user is always reminded until his token expires. We should provide a way to end the authentication when the browser is closed.

tsauerwein:
Would it help to set a short expiration time when "remind me" is not checked when logging in?

PR #46, opened 2015-12-04
Handle remember me using session storage

gberaudo:


gberaudo:
Closes: #42.

c2corg_ui/static/js/user.js
```diff
@@ -113,6 +113,7 @@ app.UserController.prototype.errorLogout_ = function(response) {
   alert('logout error');
   console.log('logout error');
   console.log(response);
+  this.auth.removeUserData();
```
asaunier: So you remove the client-side user data even though the logout failed API side?
Does this mean that the logout will then be effective (since token is lost client side even though it is still available in the API DB)?

asaunier:
Just a question else looks good to me.

c2corg_ui/static/js/user.js
gberaudo: Yes, I think it is important for the user experience and the security that the token be always removed from the browser when the user asks for a logout.

asaunier: OK, makes sense

gberaudo:

Issue #41, opened 2015-11-26
Error 400, activities not iterable when creating a Route


gberaudo:
The UI displays and sends a string instead of a collecition: `{activities: "rock_climbing",…}`

tsauerwein:
Could this be a bug in the API?

asaunier:
I guess that the API expects a list of activities but gets a string.

ginold:
is this issue still valid? it's been a while.

asaunier:
A list of activities is actually passed.

PR #47, opened 2015-12-07
Use multiple select for route activities

gberaudo:


gberaudo:
Closes: https://github.com/c2corg/v6_ui/issues/41

asaunier:
Thanks, that should actually solve the linked issue.

We can merge as is but I would prefer that we use checkboxes instead of "multiple" selects because it's a known UX problem in v5 that many people don't know that it's possible to select several options by pressing the Ctrl key and that it's also possible to remove a selected option this way. In v5, we have added a "blank" option at the beginning of such selects so that users may also set no option at all. But that sucks. Checkboxes are a better way to do this (but that's most likely a bit more complicated to handle in an angular model).

asaunier:
In the meantime :+1: for merging this PR.

gberaudo:
For now I will simply merge this fix and we can improve later as you suggested.
See issue https://github.com/c2corg/v6_ui/issues/49.

PR #52, opened 2015-12-09
Add search control

tsauerwein:
Not ready, but reviews already welcome.

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -2,6 +2,8 @@ goog.provide('app.DocumentEditingController');
 goog.provide('app.documentEditingDirective');
 
 goog.require('app');
+goog.require('app.utils');
+goog.require('goog.asserts');
 goog.require('ol.format.GeoJSON');
 goog.require('ol.geom.Point');
 
@@ -265,10 +267,9 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
  */
 app.DocumentEditingController.prototype.successSave_ = function(response) {
   // redirects to the document view page
-  var url = '/{module}/{id}/{culture}'
-      .replace('{module}', this.module_)
-      .replace('{id}', response['data']['document_id'])
-      .replace('{culture}', this.culture_);
+  goog.asserts.assert(this.culture_ !== null);
+  var url = app.utils.buildDocumentUrl(
+      this.module_, response['data']['document_id'], this.culture_);
   window.location.href = url;
   // FIXME: use $window.location.href instead?
   // TODO: add a loading message
--- a/c2corg_ui/static/js/searchdirective.js
+++ b/c2corg_ui/static/js/searchdirective.js
@@ -1,0 +1,184 @@
+goog.provide('app.SearchController');
+goog.provide('app.searchDirective');
+
+goog.require('app');
+goog.require('app.utils');
+goog.require('ngeo.searchDirective');
+
+
+/**
+ * The directive for the auto-complete search field shown in the header
+ * of every page.
+ *
+ * @return {angular.Directive} Directive Definition Object.
+ */
+app.searchDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppSearchController',
+    bindToController: true,
+    controllerAs: 'searchCtrl',
+    templateUrl: '/static/partials/search.html',
+    link:
+        /**
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Atttributes.
+         */
+        function(scope, element, attrs) {
+          // Empty the search field on focus and blur.
+          element.find('input').on('focus blur', function() {
+            $(this).typeahead('val', '');
+          });
+        }
+  };
+};
+app.module.directive('appSearch', app.searchDirective);
+
+
+
+/**
+ * @constructor
+ * @param {angular.Scope} $rootScope Angular root scope.
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {string} apiUrl Base URL of the API.
+ * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
+ * @param {Array.<string>} cultures All cultures.
+ * @ngInject
+ */
+app.SearchController = function(
+    $rootScope, $compile, apiUrl, gettextCatalog, cultures) {
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.apiUrl_ = apiUrl;
+
+  /**
+   * @type {Array.<string>}
+   * @private
+   */
+  this.cultures_ = cultures;
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.gettextCatalog_ = gettextCatalog;
+
+  /**
+   * @type {TypeaheadOptions}
+   * @export
+   */
+  this.options = /** @type {TypeaheadOptions} */ ({
+    highlight: true,
+    hint: true,
+    minLength: 3
+  });
+
+  /**
+   * @type {Array.<TypeaheadDataset>}
+   * @export
+   */
+  this.datasets = [
+    this.createDataset_('waypoints', 'Waypoints'),
+    this.createDataset_('routes', 'Routes')
+  ];
+
+  /**
+   * @type {ngeox.SearchDirectiveListeners}
+   * @export
+   */
+  this.listeners = /** @type {ngeox.SearchDirectiveListeners} */ ({
+    select: app.SearchController.select_.bind(this)
+  });
+};
+
+
+/**
+ * @param {string} type The document type.
+ * @param {string} title The category title.
+ * @return {TypeaheadDataset} A data set.
+ * @private
+ */
+app.SearchController.prototype.createDataset_ = function(type, title) {
+  var bloodhoundEngine = this.createAndInitBloodhound_(type);
+  return /** @type {TypeaheadDataset} */({
+    source: bloodhoundEngine.ttAdapter(),
+    display: function(doc) {
+      return doc.label;
+    },
+    limit: 20,
+    templates: {
+      header: (function() {
+        return '<div class="header">' +
+            this.gettextCatalog_.getString(title) + '</div>';
+      }).bind(this),
+      suggestion: function(doc) {
+        return '<p>' + doc.label + '</p>';
+      }
+      // Note: The templates for `notFound` and `pending` are not set, because
+      // they would be shown for each dataset.
+    }
+  });
+};
+
+
+/**
+ * @param {string} type The document type.
+ * @return {Bloodhound} The bloodhound engine.
+ * @private
+ */
+app.SearchController.prototype.createAndInitBloodhound_ = function(type) {
+  var url = this.apiUrl_ + '/search?q=%QUERY';
+  var bloodhound = new Bloodhound(/** @type {BloodhoundOptions} */({
+    limit: 10,
+    queryTokenizer: Bloodhound.tokenizers.whitespace,
+    datumTokenizer: Bloodhound.tokenizers.obj.whitespace('label'),
+    remote: {
+      url: url,
+      wildcard: '%QUERY',
+      rateLimitWait: 50,
+      filter: (function(/** appx.SearchResponse */ resp) {
+        var documentResponse =
+            /** @type {appx.SearchDocumentResponse} */ (resp[type]);
+        var documents = documentResponse.documents;
+        var currentCulture = this.gettextCatalog_.currentLanguage;
+        var cultures = this.cultures_;
+
+        return documents.map(function(/** appx.SearchDocument */ doc) {
+          var locale = app.utils.getBestLocale(doc, currentCulture, cultures);
+          doc.label = locale.title;
+
+          if (currentCulture !== locale.culture) {
+            doc.label += ' (' + locale.culture + ')';
+          }
+
+          doc.bestCulture = locale.culture;
+          doc.documentType = type;
+          return doc;
+        });
+      }).bind(this)
+    }
+  }));
+  bloodhound.initialize();
+  return bloodhound;
+};
+
+
+/**
+ * @param {jQuery.Event} event Event.
+ * @param {appx.SearchDocument} doc Suggested document.
+ * @param {TypeaheadDataset} dataset Dataset.
+ * @this {app.SearchController}
+ * @private
+ */
+app.SearchController.select_ = function(event, doc, dataset) {
+  var culture = doc.bestCulture;
+  var url = app.utils.buildDocumentUrl(
+      doc.documentType, doc.document_id, culture);
+  window.location.href = url;
+};
+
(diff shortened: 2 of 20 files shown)


tsauerwein:
Ready from my side.

.gitignore
```diff
@@ -11,3 +11,4 @@
 *.pyc
 *.egg-info
 *.po~
+/.idea
```
gberaudo: Have you considered adding this to your global gitignore instead?

c2corg_ui/static/js/documentediting.js
```diff
@@ -265,10 +267,9 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
  */
 app.DocumentEditingController.prototype.successSave_ = function(response) {
   // redirects to the document view page
-  var url = '/{module}/{id}/{culture}'
-      .replace('{module}', this.module_)
-      .replace('{id}', response['data']['document_id'])
-      .replace('{culture}', this.culture_);
+  goog.asserts.assert(this.culture_ !== null);
+  var url = app.utils.buildDocumentUrl(
```
gberaudo: There are many possible URLs for a document. What about renaming to `buildGetDocumentUrl`?

c2corg_ui/static/js/utils.js
```diff
@@ -0,0 +1,58 @@
...
+goog.require('goog.asserts');
+
+
+/**
+ * @param {string} module The document type.
+ * @param {string} documentId The document id.
+ * @param {string} culture Culture.
+ * @return {string} Url.
+ */
+app.utils.buildDocumentUrl = function(module, documentId, culture) {
```
gberaudo: I think it would be more clear to name the argument `type` instead of `module`.

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+ * @return {angular.Directive} Directive Definition Object.
+ */
+app.searchDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppSearchController',
+    bindToController: true,
+    controllerAs: 'ctrl',
+    template:
+        '<input type="text" placeholder="search…" ' +
```
gberaudo: Do we need a `translate` here?

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+ * of every page.
+ *
+ * @return {angular.Directive} Directive Definition Object.
+ */
+app.searchDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppSearchController',
+    bindToController: true,
+    controllerAs: 'ctrl',
```
gberaudo: Have you considered using a more specific name like `searchCtrl`?

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+  this.gettextCatalog_ = gettextCatalog;
+
+  /**
+   * @type {TypeaheadOptions}
+   * @export
+   */
+  this.options = /** @type {TypeaheadOptions} */ ({
+    highlight: true,
+    hint: true,
+    minLength: 2
```
gberaudo: Do we get useful results with `2`?
Have you considered a bigger number to save resources?

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+    this.createDataset_('waypoints', 'Waypoints'),
+    this.createDataset_('routes', 'Routes')
+  ];
+
+  /**
+   * @type {ngeox.SearchDirectiveListeners}
+   * @export
+   */
+  this.listeners = /** @type {ngeox.SearchDirectiveListeners} */ ({
+    select: angular.bind(this, app.SearchController.select_)
```
gberaudo: app.SearchController.select_.bind(this)

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+app.SearchController.prototype.createDataset_ = function(type, title) {
+  var bloodhoundEngine = this.createAndInitBloodhound_(type);
+  return /** @type {TypeaheadDataset} */({
+    source: bloodhoundEngine.ttAdapter(),
+    display: function(suggestion) {
+      return suggestion['label'];
+    },
+    limit: 20,
+    templates: {
+      header: angular.bind(this, function() {
```
gberaudo: Idem, you can use ES5 `bind` instead.

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+ * @param {string} title The category title.
+ * @return {TypeaheadDataset} A data set.
+ * @private
+ */
+app.SearchController.prototype.createDataset_ = function(type, title) {
+  var bloodhoundEngine = this.createAndInitBloodhound_(type);
+  return /** @type {TypeaheadDataset} */({
+    source: bloodhoundEngine.ttAdapter(),
+    display: function(suggestion) {
+      return suggestion['label'];
```
gberaudo: Would be great to have a typedef for `suggestion`.

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+    },
+    limit: 20,
+    templates: {
+      header: angular.bind(this, function() {
+        return '<div class="header">' +
+            this.gettextCatalog_.getString(title) + '</div>';
+      }),
+      suggestion: function(suggestion) {
+        return '<p>' + suggestion['label'] + '</p>';
```
gberaudo: Idem https://github.com/c2corg/v6_ui/pull/52/files#r47204626

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+  var url = this.apiUrl_ + '/search?q=%QUERY';
+  var bloodhound = new Bloodhound(/** @type {BloodhoundOptions} */({
+    limit: 10,
+    queryTokenizer: Bloodhound.tokenizers.whitespace,
+    datumTokenizer: Bloodhound.tokenizers.obj.whitespace('label'),
+    remote: {
+      url: url,
+      wildcard: '%QUERY',
+      rateLimitWait: 50,
+      filter: angular.bind(this, function(resp) {
```
gberaudo: ES5 bind

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+        var documents = resp[type]['documents'];
+        var currentCulture = this.gettextCatalog_.currentLanguage;
+        var cultures = this.cultures_;
+        return documents.map(function(document) {
+          var locale = app.utils.getBestLocale(
```
gberaudo: We should have a typedef for this kind of document response and not use brackets.
Typedefs can be added to c2corg_ui/externs/appx.js

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+    }
+  }));
+  bloodhound.initialize();
+  return bloodhound;
+};
+
+
+/**
+ * @param {jQuery.Event} event Event.
+ * @param {Object} document Suggested document.
```
gberaudo: A typedef should be passed for the document.

c2corg_ui/static/js/searchdirective.js
```diff
@@ -0,0 +1,184 @@
...
+ * @param {TypeaheadDataset} dataset Dataset.
+ * @this {app.SearchController}
+ * @private
+ */
+app.SearchController.select_ = function(event, document, dataset) {
+  var culture = document['bestCulture'];
```
gberaudo: Minor suggestion: maybe use `doc` to avoid shadowing the global `document` object.

c2corg_ui/static/js/utils.js
```diff
@@ -0,0 +1,58 @@
...
+    .replace('{culture}', culture);
+};
+
+
+/**
+ * Cultures ordered by their relevance.
+ * @const
+ * @type {Array.<string>}
+ */
+app.utils.LANG_PRIORITY = ['fr', 'en', 'it', 'de', 'es', 'ca', 'eu'];
```
gberaudo: Why do we need a lang priority?
Shouldn't this come from the `common` project?

c2corg_ui/static/js/utils.js
```diff
@@ -0,0 +1,58 @@
...
+ * `preferedCulture` (usually the interface language), if not the locale in
+ * the "most relevant" culture.
+ *
+ * @param {Object} document The document.
+ * @param {string} preferedCulture The prefered culture
+ * @param {Array.<string>} cultures All cultures.
+ * @return {Object} The "best" culture.
+ */
+app.utils.getBestLocale = function(document, preferedCulture, cultures) {
+  var locales = document['locales'];
```
gberaudo: Should use a typedef and maybe avoid shadowing `document`.

c2corg_ui/static/js/utils.js
```diff
@@ -0,0 +1,58 @@
...
+  var availableLocales = {};
+  $.each(locales, function(i, locale) {
+    availableLocales[locale['culture']] = locale;
+  });
+
+  if (availableLocales[preferedCulture] !== undefined) {
+    return availableLocales[preferedCulture];
+  } else {
+    var localeToUse = null;
+    $.each(app.utils.LANG_PRIORITY, function(i, culture) {
```
gberaudo: You can use ES5 forEach to iterate the array.

gberaudo: Not seen the return. You can use standard `some()` in this case, and return true.

gberaudo:
I added some comments. Otherwise, great work @tsauerwein.

tsauerwein:
"Some" comments... ;) Thanks, I will work through them.

.gitignore
tsauerwein: Good idea!

c2corg_ui/static/js/documentediting.js
asaunier: On the other hand the document URL is the main one (view page). For instance https://www.camptocamp.org/waypoints/12345/fr (by the way we will have to add the slug at the end some day).

Other URL are annex URLs: editing, history, archived versions, etc.

c2corg_ui/static/js/searchdirective.js
asaunier: Yep. I think that can be done using ``placeholder="{{'Search...'|translate}}"``.
See for instance: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/edit.html#L75

We also have to find a way the string detected by the angular-gettext extractor. I think that if you had a partial in https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials it will be automatically detected.

tsauerwein: Yes, exactly!

c2corg_ui/static/js/searchdirective.js
asaunier: We rather used ``goog.bind`` in other controllers. See 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/documentediting.js#L104
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/auth/auth.js#L87

c2corg_ui/static/js/searchdirective.js
asaunier: goog.bind! :P

c2corg_ui/static/js/utils.js
asaunier: Yes and no. "module" is a bit more generic and MVC-related I think.
"type" could also be confusing since we have for instance waypoints types.

c2corg_ui/static/js/utils.js
asaunier: Because we have to figure out what culture of the document should be shown when several cultures are available.
See comments in https://github.com/c2corg/v6_api/issues/83

c2corg_ui/static/js/searchdirective.js
tsauerwein: I set the `minLength` to 3, let's see how this works.

c2corg_ui/static/js/searchdirective.js
tsauerwein: `angular.bind` and `goog.bind` is both used in ngeo. I don't have a preference. What do you think?

c2corg_ui/static/js/searchdirective.js
tsauerwein: Used on an anonymous function it doesn't even look too bad:

    (function() {
    
    }).bind(this)

c2corg_ui/static/js/utils.js
gberaudo: Do you mean the `model` of MVC?
We also have `_API_ROUTE` in `document.py`.
I see `type` is reserved in python, so maybe better to use something else.

tsauerwein: I used `document_type`.

c2corg_ui/static/js/searchdirective.js
asaunier: I prefer that we are consistent with other controllers and use ``goog.bind``.
I don't really mind that we use another syntax if it works but then please change in the other controllers as well :p

c2corg_ui/static/js/utils.js
asaunier: > Do you mean the model of MVC?

No, I mean "module" (like "waypoints", "routes"). But that's maybe only a name used in Symfony (v5 framework).

c2corg_ui/static/js/searchdirective.js
gberaudo: Well, wouldn't it be better to consistently use the javascript standard?

tsauerwein: Would be ok for me.

c2corg_ui/static/js/utils.js
gberaudo: I see, thanks.

c2corg_ui/static/js/searchdirective.js
tsauerwein: :+1:

c2corg_ui/static/js/utils.js
tsauerwein: So far this is the only place where the list is needed. Later, when needed, we can move it to common.

c2corg_ui/static/js/searchdirective.js
gberaudo: :wink:

tsauerwein:
Ok, I should have addressed all comments.

asaunier:
OK for me

c2corg_ui/static/js/utils.js
gberaudo: Fine with me.

gberaudo:
+1

tsauerwein:
Thanks for your reviews.

PR #51, opened 2015-12-09
Templatecache

asaunier:
This PR introduces the "templatecache" system used in ngeo [1] and Schweizmobil [2] for caching the directives partials.

[1] https://github.com/camptocamp/ngeo/blob/master/buildtools/templatecache.mako.js
[2] https://github.com/camptocamp/schweizmobil_re3/blob/master/templatecache.mako.js

It should hopefully solve by the way the path problem observed on the demo server (that has a "/ui/" path in the URL, breaking the partials loading).

There's probably something still missing because partials are still loaded the standard way instead of using the cache. Working on it...

PR #54, opened 2015-12-09
History

gberaudo:


gberaudo:
See: https://github.com/c2corg/v6_ui/issues/48

gberaudo:
Ready for review.

c2corg_ui/views/route.py
```diff
@@ -46,3 +46,8 @@ def edit(self):
             'route_id': id
         })
         return self.template_input
+
+    @view_config(route_name=_API_ROUTE + '_history',
```
asaunier: Why using _API_ROUTE in the route_name whereas it has not been used for other route_names?
I think it is unnecessarly confusing, I would prefer we use the explicit route_name here.

c2corg_ui/views/waypoint.py
```diff
@@ -47,3 +47,8 @@ def edit(self):
             'waypoint_id': id
         })
         return self.template_input
+
+    @view_config(route_name=_API_ROUTE + '_history',
```
asaunier: Ditto

c2corg_ui/views/document.py
```diff
@@ -115,6 +115,26 @@ def _get_filter_params(self):
                     params.append(filters[i:i+2])
         return params
 
+    def _get_history(self):
+        id, culture = self._validate_id_culture()
+        url = '%s/%s/%d/%s/%s' % (
+            self.settings['api_url'], 'document', id, 'history', culture)
```
asaunier: Why not simply using
```
url = '%s/document/%d/history/%s' % (self.settings['api_url'], id, culture)
```
?

c2corg_ui/templates/document/history.html
```diff
@@ -0,0 +1,27 @@
...
+      <td>Radios ${v['version_id']}</td>
+      <td><a href="${request.route_url(module+'_view', id=document_id, culture=culture)}#TODOversion">{{'${1000 * v['written_at']}' | date:'medium'}}</a></td>
+      <td><a href="#TODO profile for ${v['user_id']}">${v['username']}</a></td>
+      <td>${v['comment']}</td>
+    </tr>
+  % endfor
+</table>
+
+<ul>
+  <li><a href="${request.route_url(module+'_view', id=document_id, culture=culture)}" translate>View</a></li>
```
asaunier: Wouldn't flake8 require blank space around "+"?

gberaudo:
PR updated.

asaunier:
You also have a conflict with the PO files I guess. I'm trying to fix it in my PR too: #51 
I don't know if we should try to fix it in parallel or if it is better to wait that it's OK for one of us.

gberaudo:
OK, waiting for your merge.

asaunier:
My PR is merged.

gberaudo:
Rebased, thanks.

PR #55, opened 2015-12-10
Make interface language selection persistent

asaunier:


c2corg_ui/static/js/lang.js
```diff
@@ -84,6 +84,7 @@ app.LangController.prototype.updateCulture = function() {
   this.gettextCatalog_.setCurrentLanguage(this.culture);
   this.gettextCatalog_.loadRemote(
       this.langUrlTemplate_.replace('__lang__', this.culture));
+  window.localStorage.setItem('interface_lang', this.culture);
```
gberaudo: Should by in a try/catch.

gberaudo:
Added one comment, otherwise +1.

c2corg_ui/static/js/lang.js
asaunier: You think that window.localStorage might not be available and would cause an error?
By the way, I considered using sessionStorage as you did for the auth but thought  that it was a bit overkill since it's not  a sensitive info.

asaunier:
I have added the try/catch as you suggested. Merging...

c2corg_ui/static/js/lang.js
gberaudo: I agree non sensitive info should simply be stored in local storage.
The try catch is for iOS in incognito mode.

PR #56, opened 2015-12-11
Make sure that user data are removed from all possible storages when …

asaunier:
…logging out

In my instance I observed that I was always automatically reauthenticated even after pressing the "log out" button. I finally found out that for some reason "userData" was set both in localStorage and in sessionStorage while I had not checked the "remember me" box => using sessionStorage. Perhaps the localStorage was set from an old attempt?

Anyway to avoid that kind of problem I think we should make sure that "userData" is removed from both localStorage and sessionStorage (even though only one storage is actually used) when logging out.


gberaudo:
LGTM

PR #60, opened 2015-12-14
Add make target to rebuild CSS files

tsauerwein:
Run `make -f config/... less` to rebuild the CSS files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ help:
 	@echo "- clean			Remove generated files"
 	@echo "- cleanall		Remove all the build artefacts"
 	@echo "- compile-catalog	Compile the translation catalog"
-	@echo "- flake8			Run flake8 checker on the Python code"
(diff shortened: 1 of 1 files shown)


asaunier:
That's indeed useful, thanks!

Makefile
```diff
@@ -29,7 +29,7 @@ help:
 	@echo "- clean			Remove generated files"
 	@echo "- cleanall		Remove all the build artefacts"
 	@echo "- compile-catalog	Compile the translation catalog"
-	@echo "- flake8			Run flake8 checker on the Python code"
+	@echo "- flake8		Run flake8 checker on the Python code"
```
gberaudo: indentation

tsauerwein: When I run `make help` it is perfectly aligned. Not for you?

gberaudo: I see, the indentation appears unregular in this github page (and in vim), but you are right it is aligned when run in the shell.

PR #61, opened 2015-12-14
Diff view

tsauerwein:
TODO 

- [x] Translations for fields
- [x] Map to show difference between geometries (or later?)
- [ ] ?

### DIFF
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -23,6 +23,8 @@ def main(global_config, **settings):
     config.add_route('waypoints_history', '/waypoints/history/{id}/{culture}')
     config.add_route('waypoints_archive',
                      '/waypoints/{id}/{culture}/{version:\d+}')
+    config.add_route(
+        'waypoints_diff', '/waypoints/diff/{id}/{culture}/{v1}/{v2}')
 
     config.add_route('routes_view', '/routes/{id}/{culture}')
     config.add_route('routes_index_default', '/routes')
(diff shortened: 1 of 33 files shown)


tsauerwein:
Ready to review.

It looks like this:
c2corg-diff

asaunier:
I see that you have added a dedicated less file for the diff. It's fine for me. I'm just wondering if we could have a rule for this? For instance say that styles only needed for a given page should be added in a separated less file.

What about the history page styles? As for now they are in the main less file since there are not plenty of them:
https://github.com/c2corg/v6_ui/blob/master/less/c2corg_ui.less#L39-L54
Ditto for "archived pages":
https://github.com/c2corg/v6_ui/blob/master/less/c2corg_ui.less#L33-L38

Should we move those styles in dedicated less files?

tsauerwein:
I think if there are only 1-2 styles, we can keep them in the main style file. Otherwise it would be better to create a new file, so that the main file does not grow too big.

asaunier:
OK, then we can keep "history" and "achive" styles as is unless they grow much bigger in the future. Thanks.

asaunier:
The PR looks good to me.

Just one remark: I notice that you have created a dedicated ``diffMap`` directive. When creating the standard ``map`` directive, I thought we could use it for any map in the UI (with extensions?). This way when changing for instance the common tools or the background layers list, we would not have to port the changes in several map directives. We have discussed it a bit with @gberaudo (not sure he was fully of my opinion :P). Perhaps we should discuss this point (in a dedicated issue?).

tsauerwein:
>  This way when changing for instance the common tools or the background layers list, we would not have to port the changes in several map directives.

In general I think that is a good idea. The diff map is a bit special because you have a custom styling and don't need other controls (like the search, layer manager, ...). But once we have different background layers, it would make sense to share common code between the two directives.

tsauerwein:
Thanks for the review.

PR #64, opened 2015-12-16
Add Angular-UI Boostrap + alert directive/controller

asaunier:
Here is a first version of an action feedback/alert system based upon [Angular-UI Bootstrap's alert](http://angular-ui.github.io/bootstrap/#/alert) itself based upon [Bootstrap alert](http://getbootstrap.com/components/#alerts).

Angular-UI Bootstrap (and linked libs) could be used for other widgets as well. See details at https://github.com/c2corg/v6_ui/issues/8#issuecomment-163985899

Related to #8

### DIFF
--- a/c2corg_ui/static/js/alert.js
+++ b/c2corg_ui/static/js/alert.js
@@ -2,6 +2,7 @@ goog.provide('app.AlertController');
 goog.provide('app.alertDirective');
 
 goog.require('app');
+goog.require('app.ControllerHub');
 
 
 /**
@@ -27,16 +28,19 @@ app.module.directive('appAlert', app.alertDirective);
 
 /**
  * @constructor
+ * @param {app.ControllerHub} appControllerHub Controller hub service
  * @export
  * @ngInject
  */
-app.AlertController = function() {
+app.AlertController = function(appControllerHub) {
 
   /**
    * @type {Array.<appx.AlertMessage>}
    * @export
    */
   this.alerts = [];
+
+  appControllerHub.register('alert', this);
 };
 
 
@@ -50,12 +54,11 @@ app.AlertController.prototype.close = function(index) {
 
 
 /**
- * @param {string|Object} msg Alert data.
+ * @param {string|appx.AlertMessage} msg Alert data.
  * @export
  */
 app.AlertController.prototype.addAlert = function(msg) {
-  msg = (typeof msg == 'string') ? {'msg': msg} : msg;
-  this.alerts.push(msg);
+  this.alerts.push(typeof msg == 'string' ? {'msg': msg} : msg);
 };
 
 
--- a/c2corg_ui/static/js/alert.js
+++ b/c2corg_ui/static/js/alert.js
@@ -1,0 +1,62 @@
+goog.provide('app.AlertController');
+goog.provide('app.alertDirective');
+
+goog.require('app');
+
+
+/**
+ * This directive is used to display feedbacks to user actions.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.alertDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppAlertController',
+    controllerAs: 'alertCtrl',
+    bindToController: true,
+    templateUrl: '/static/partials/alert.html'
+  };
+};
+
+
+app.module.directive('appAlert', app.alertDirective);
+
+
+
+/**
+ * @constructor
+ * @export
+ * @ngInject
+ */
+app.AlertController = function() {
+
+  /**
+   * @type {Array.<appx.AlertMessage>}
+   * @export
+   */
+  this.alerts = [];
+};
+
+
+/**
+ * @param {number} index Index of alert to close.
+ * @export
+ */
+app.AlertController.prototype.close = function(index) {
+  this.alerts.splice(index, 1);
+};
+
+
+/**
+ * @param {string|Object} msg Alert data.
+ * @export
+ */
+app.AlertController.prototype.addAlert = function(msg) {
+  msg = (typeof msg == 'string') ? {'msg': msg} : msg;
+  this.alerts.push(msg);
+};
+
+
+app.module.controller('AppAlertController', app.AlertController);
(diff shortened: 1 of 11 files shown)


c2corg_ui/static/js/alert.js
```diff
@@ -0,0 +1,61 @@
...
+ * @constructor
+ * @export
+ * @ngInject
+ */
+app.AlertController = function() {
+
+  /**
+   * @type {Array.<Object>}
+   */
+  this['alerts'] = [];
```
gberaudo: Better to use the dotted notation with an `@export`.

c2corg_ui/static/js/alert.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @constructor
+ * @export
+ * @ngInject
+ */
+app.AlertController = function() {
+
+  /**
+   * @type {Array.<Object>}
```
gberaudo: Would be great to have a typedef instead of `Object`.

asaunier: I was waiting for it! :p 
But is it reaaaaally needed?

gberaudo:
Looks like a good start. I added some comments. +1 for merging when they are addressed.

asaunier:
I have just updated the PR to add a typedef for alert messages and use the dotted notation.

c2corg_ui/static/js/alert.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {number} index Index of alert to close.
+ * @export
+ */
+app.AlertController.prototype.close = function(index) {
+  this.alerts.splice(index, 1);
+};
+
+
+/**
+ * @param {string|Object} msg Alert data.
```
gberaudo: s/Object/appx.AlertMessage

gberaudo:
I added one more comment, otherwise I think it is ready for merging.

c2corg_ui/static/js/alert.js
asaunier: Good catch, thanks.

asaunier:
I'm working on a ``app.ControllerHub`` service so that it is possible to call this alert tool from other controllers. Since it is closely related to the current branch, it's a bit simpler to add those changes to the same PR.

asaunier:
I have added a ``ControllerHub`` service in which it is possible to store references to controllers. Which is needed to call the ``alertController`` from the ``authController`` for instance and display error messages related to login/register actions.

There is still a problem with the alert directive: HTML code in message is not interpreted and tags are rendered as is.

Perhaps a solution found here: http://stackoverflow.com/a/27047501

c2corg_ui/static/js/auth/auth.js
```diff
@@ -111,10 +119,11 @@ app.AuthController.prototype.successLogin_ = function(remember, response) {
...
  */
 app.AuthController.prototype.errorLogin_ = function(response) {
-  // TODO
-  alert('login error');
-  console.log('login error');
-  console.log(response);
+  // TODO: i18n
+  this.ctrlHub_.alert.addAlert({
+    type: 'danger',
+    msg: 'Login failed'
```
gberaudo: How is translation handled?

gberaudo: Hmm, I see there is a TODO ;)

asaunier: Yes, we will have to use angular-gettext

c2corg_ui/static/js/auth/auth.js
```diff
@@ -151,10 +161,19 @@ app.AuthController.prototype.successRegister_ = function(response) {
...
  */
 app.AuthController.prototype.errorRegister_ = function(response) {
-  // TODO
-  alert('register error');
-  console.log('register error');
-  console.log(response);
+  // TODO: i18n
+  // FIXME: HTML is not interpreted in alert messages
+  var msg = 'Registration failed because of:';
+  var errors = response['data']['errors'],
```
gberaudo: It would be great to have a typedef for error responses.

c2corg_ui/static/js/auth/auth.js
```diff
@@ -151,10 +161,19 @@ app.AuthController.prototype.successRegister_ = function(response) {
...
+      len = errors.length;
+  if (len > 0) {
+    msg += '<ul>';
```
gberaudo: Have you considered doing the formatting in `c2corg_ui/static/partials/alert.html` instead?
Passing a list of submessages should avoid the sanitization issue.

gberaudo:
Controler hub code looks good to me.

c2corg_ui/static/js/auth/auth.js
asaunier: Maybe but that's not easy anyway since the alert system does not know what HTML code will be passed. In some cases we might want to display a link etc.

asaunier: As for now I have been able to render the HTML code in the messages using a ``trustAsHtml`` filter as in http://stackoverflow.com/a/30620547
But then the "close" button is lost :(

asaunier: > But then the "close" button is lost :(

The problem is the combination of the ``ng-include`` in AngularUI Bootstrap alert (used to pass the content to display to the alert template) and of ``trustAsHtml`` filter since the latter requires to change
```
<uib-alert ng-repeat="alert in alertCtrl.alerts" template-url="/static/partials/alert-tpl.html">{{alert.msg}}</uib-alert>
```
to
```
<uib-alert ng-repeat="alert in alertCtrl.alerts" template-url="/static/partials/alert-tpl.html" ng-bind-html="{{alert.msg | appTrustAsHtml}}"></uib-alert>
```
"ng-bind-html" then replaces most of the content of ``/static/partials/alert-tpl.html`` including the "close" button code.

We could also skip using AngularUI Bootstrap alert, using instead a single template with no "ng-include" but with the "appTrustAsHtml" filter. It works: the formatted message as well as the close button are displayed. But then we lose the timeout closing feature, which requires a controller instance for each alert message (and its own $timeout)...
See https://github.com/angular-ui/bootstrap/blob/master/src/alert/alert.js

asaunier: > Have you considered doing the formatting in c2corg_ui/static/partials/alert.html instead?
Passing a list of submessages should avoid the sanitization issue.

I have tried to add HTML formating directly in the "alert" directive partial: it indeed works. The close button and the timeout closing too.
But as I said, it's not possible to know in advance what will be the content of the message (lists, links, bold/italic words, etc.).

asaunier:
I have trouble making this code actually support all 3 needs:
* autoclose after some timeout
* provide a close button
* support of any HTML content

PR #68 might replace the current one.

PR #65, opened 2015-12-16
Remove OL3 objects from document editing model

gberaudo:
OL3 objects are not copyable / serializable.
This change fixes the edition of waypoints.

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -97,6 +97,12 @@ app.DocumentEditingController = function($scope, $element, $attrs, $http,
    */
   this.culture_ = null;
 
+  /**
+   * @type {ol.format.GeoJSON}
+   * @private
+   */
+  this.geojsonFormat_ = new ol.format.GeoJSON();
+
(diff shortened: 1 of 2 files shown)


gberaudo:
The geometry is kept unparsed in the model.

c2corg_ui/static/js/documentediting.js
```diff
@@ -97,6 +97,11 @@ app.DocumentEditingController = function($scope, $element, $attrs, $http,
    */
   this.culture_ = null;
 
+  /**
+   * @private
```
asaunier: Don't we need a type here?

asaunier:
Is the geometry string actually updated (and pushed to the API) when one changes the coordinates in the form?
Else looks good to me.

c2corg_ui/static/js/documentediting.js
gberaudo: done

gberaudo:
Yes @asaunier, it works.

asaunier:
Thanks, please merge then.

gberaudo:
Thank you for the review.

PR #66, opened 2015-12-17
Styling and waypoint interaction

gberaudo:
Add hover and click interactions to the waypoint list maps.
Features are styled with an icon according to their type.

On mouseover, the feature is magnified and its title is displayed.
On click, the location is changed to display the target feature.

The styling and the mouseover interaction are also used when viewing a given waypoint.


asaunier:
What about moving the PO files changes (line numbers...) in a dedicated commit? They kind of pollute the real changes I think.

gberaudo:
OK, done.

asaunier:
Does the interaction work in view/edit pages as well?

asaunier:
Please merge if view/edit pages are OK as well.

gberaudo:
Thanks for the review, I added the icon cache, renamed the directory and implemented a best lang function taking an array. If there is no other comments I will merge this PR.

gberaudo:
Updated with your comments.

asaunier:
:+1:

gberaudo:
I deployed on the demo server but the UI shows 0 waypoints, which is strange since the database is far from empty. Do someone has some idea about what is happening?

asaunier:
That's because I have set the API down.
The sysadmins have created the new vhosts and I'm trying to install the API in a separated vhosts. As for know there is stil a problem. I will put the standard API instance back but it might change soon.

asaunier:
Actually I prefer leaving the API down because its config cause a "duplicate" error with the new one. So skip the main instance until the new API is on.

gberaudo:
OK

gberaudo:
Thanks again for the review.

PR #68, opened 2015-12-18
Add an action feedback system

asaunier:
I was not happy with PR #64 since I had troubles having alert messages that
* autoclosed after some timeout
* had a close button
* supported any HTML content

In addition I am not that sure we needed a ``ControllerHub`` service: we could simply have an alert service to which controllers could pass their alerts.

This new PR no longer uses Angular-UI Boostrap alert directive. It rather uses app-specific directives that somehow mimic the ones from Angular-UI Boostrap but with changes that support the 3 points above.

Related to #8

### DIFF
--- a/c2corg_ui/static/js/alerts.js
+++ b/c2corg_ui/static/js/alerts.js
@@ -1,0 +1,107 @@
+goog.provide('app.AlertController');
+goog.provide('app.AlertsController');
+goog.provide('app.alertDirective');
+goog.provide('app.alertsDirective');
+
+goog.require('app');
+goog.require('app.Alerts');
+goog.require('app.trustAsHtmlFilter');
+
+
+/**
+ * This directive is used to display feedbacks to user actions.
+ * Inspired by angular-ui boostrap alert:
+ * https://github.com/angular-ui/bootstrap/blob/master/src/alert/alert.js
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.alertsDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppAlertsController',
+    controllerAs: 'alertsCtrl',
+    bindToController: true,
+    templateUrl: '/static/partials/alerts.html'
+  };
+};
+
+
+app.module.directive('appAlerts', app.alertsDirective);
+
+
+
+/**
+ * @constructor
+ * @param {app.Alerts} appAlerts Alert service.
+ * @export
+ * @ngInject
+ */
+app.AlertsController = function(appAlerts) {
+
+  /**
+   * @type {Array.<appx.AlertMessage>}
+   * @export
+   */
+  this.alerts = appAlerts.get();
+};
+
+
+/**
+ * @param {number} index Index of alert to close.
+ * @export
+ */
+app.AlertsController.prototype.close = function(index) {
+  this.alerts.splice(index, 1);
+};
+
+
+app.module.controller('AppAlertsController', app.AlertsController);
+
+
+/**
+ * This directive is used to display one alert message.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.alertDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppAlertController',
+    controllerAs: 'alertCtrl',
+    bindToController: true,
+    templateUrl: '/static/partials/alert.html',
+    scope: {
+      'type': '@',
+      'close': '&',
+      'timeout': '@',
+      'msg': '@'
+    }
+  };
+};
+
+
+app.module.directive('appAlert', app.alertDirective);
+
+
+
+/**
+ * @param {angular.$timeout} $timeout Angular timeout service.
+ * @constructor
+ * @ngInject
+ * @export
+ */
+app.AlertController = function($timeout) {
+  if (this['timeout'] && this['close']) {
+    var timeout = parseInt(this['timeout'], 10);
+    if (timeout) {
+      $timeout(goog.bind(function() {
+        this['close']();
+      }, this), timeout);
+    }
+  }
+};
+
+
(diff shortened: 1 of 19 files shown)


c2corg_ui/static/js/auth/auth.js
```diff
@@ -151,10 +162,35 @@ app.AuthController.prototype.successRegister_ = function(response) {
...
+ * @private
+ */
+app.AuthController.prototype.formatErrorMsg_ = function(response) {
+  // TODO: i18n
+  // TODO: filter strings from response using $sanitize or another tool
+  var errors = response['data']['errors'],
+      len = errors.length,
+      msg = '';
+  if (len == 1) {
+    msg = errors[0]['description'];
```
asaunier: As suggested by @gberaudo it would make sense to escape this ``description`` field since the standard escaping AngularJS normally performs on expressions is disabled in the alert partial using the ``appTrustAsHtml`` filter.

We could for instance use ``$sanitize`` but it requires installing a separated package.
https://docs.angularjs.org/api/ngSanitize/service/$sanitize

c2corg_ui/static/js/alerts.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param {angular.$timeout} $timeout Angular timeout service.
+ * @constructor
+ * @ngInject
+ * @export
+ */
+app.AlertController = function($timeout) {
+  if (this['timeout'] && this['close']) {
+    var timeout = parseInt(this['timeout'], 10);
+    if (timeout) {
+      $timeout(goog.bind(function() {
```
tsauerwein: Didn't we want to use the native `bind`?

asaunier: I have not understood yet the difference with ``goog.bind``. We use ``goog.*`` functions in many parts of the code, what's the problem with using ``goog.bind``?

Can we replace all ``goog.bind`` by native bind in all our code? If yes, we should create an issue because they are used at some other places, for instance in the ``documentediting`` or ``auth`` directives.

c2corg_ui/static/js/alertservice.js
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * @type {Array.<appx.AlertMessage>}
+   * @private
+   */
+  this.alerts_ = [];
+};
+
+
+/**
+ * @param {string|appx.AlertMessage} msg Alert data.
```
tsauerwein: Instead of mixing the types, I would make `type` and `timeout` optional arguments. So that you could call the method with:

    this.alerts_.add(msg);

or with:


    this.alerts_.add(msg, 'danger', 5000);

asaunier: Yes, or I could force using an ``appx.AlertMessage`` object since its ``type``and ``timeout`` attributes are optional anyway:
```
this.alerts_.add({'msg': msg});
```

tsauerwein: Or that, yes.

c2corg_ui/static/js/auth/auth.js
```diff
@@ -151,10 +162,35 @@ app.AuthController.prototype.successRegister_ = function(response) {
...
+  } else if (len > 0) {
+    msg += '<ul>';
+    for (var i = 0; i < len; i++) {
+      msg += '<li>' + errors[i]['description'] + '</li>';
```
tsauerwein: i18n would also be needed here?

asaunier: Yep, that's the reason of the ``TODO: i18n`` at the beginning of the function.

tsauerwein: Sorry, I missed that.

c2corg_ui/static/js/alerts.js
gberaudo: `goog.bind` is basically a wrapper around the standard `bind`.
There is no reason to use it. I will create an issue about the use of goog in general.

asaunier: OK, I'm updating the PR for adressing your comments, will use native bind as well.

gberaudo: See https://github.com/c2corg/v6_ui/issues/70

asaunier:
The branch is up-to-date compared to master and I have commited changes about the native bind and the msg format.

In the todo remain:
* translating the feedback messages
* escaping the feedback messages

Concerning escaping: last week I have tried adding the $sanitize service. Not that a big deal except that's another package to install. Do you think it is worth installing? In addition it means another file to load at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L68-L72
Do you think we should build all those library files in a single one? directly in ``build.js``?

gberaudo:
There is also`goog.string.htmlEscape` that you may be interested to try.

asaunier:
> There is alsogoog.string.htmlEscape that you may be interested to try.

Thanks for the tip. Done at 4084c4b

asaunier:
I have just spent a bunch of time fixing conflicts in the PO files. Please DON'T commit changes to PO files from your local machines, only from the dev server, hopefully this will reduce the amount of changes in the PO files and possibly the conflicts as well!

The PR is ready for reviewing!

asaunier:
Would be cool if we could merge the PR soon, to avoid other PO conflicts and because we need it for other error handling.

tsauerwein:
This looks fine to me. Thanks!

gberaudo:
+1

asaunier:
Thanks all for reviewing :)

gberaudo:
Great work on this feature @asaunier.

asaunier:
Thanks @gberaudo I hope that Santa will notice it :P

gberaudo:
Sure, I hope Santa will bring lots of presents for the good boys (and girls) having
commited lots of great code this year.

PR #69, opened 2015-12-21
Follow-up on #61

tsauerwein:
#61


gberaudo:
+1 when Travis is happy.

PR #71, opened 2015-12-21
Use more consistent name for search directive file

asaunier:
Minor change but since as for now no directive filename contains "directive", I thought it would be more consistent to remove it from the search directive filename as well.


tsauerwein:
Ok for me.

PR #75, opened 2016-01-05
List requirements

tsauerwein:


README.md
```diff
@@ -1,6 +1,15 @@
 UI Application for camptocamp.org v6
 ====================================
 
+Requirements
+------------
+
+ * Python 2.7
+ * Node (4.0 or higher)
```
gberaudo: Why do we need a node version so new?
I have been using v0.10.25 (from Ubuntu 15.10) without any issue.

tsauerwein: Version v0.10.29 is the one we are having problems with. I know that it works with Node 4.0. But we can take a lower version if you tell me which one is ok. :)

mfournier: I confirm node v4 isn't required, from what I've seen. I only needed v4 to get a slightly more descriptive error message which allowed me to undertand node was attempting to run java behind the scene.

tsauerwein: Ok, I changed it to:

> Node (0.10.x or higher, there are known issues with 0.10.29)

asaunier:
:+1:

PR #77, opened 2016-01-07
Do not use http://pypi.camptocamp.net/pypi

gberaudo:
Using it was not necessary and produced security warnings.


tsauerwein:
Thanks! Could you do the same for the API project?
https://github.com/c2corg/v6_api/blob/master/requirements.txt

PR #81, opened 2016-01-08
[WIP] List card layout

ginold:
* [x] changed the simple waypoint list into a card layout, and it's responsive
* [x] added compatibility with IE9 for media queries and columns - loads a different CSS file 
* [x] Helpers buttons (pagination) are now stylized and have icons

TODO
* [x] same layout for routes
* [x] click on whole card in order to see details, and not only on ```<a>``` tag
* [ ] change icons depending on waypoint type

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 /apache/app-c2corg_ui.wsgi
 /apache/wsgi.conf
 /MANIFEST
+/nbproject
 *.pyc
 *.egg-info
 *.po~
--- a/c2corg_ui/locale/c2corg_ui-client.pot
+++ b/c2corg_ui/locale/c2corg_ui-client.pot
@@ -4,11 +4,16 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
+#: c2corg_ui/templates/waypoint/index.html:75
+msgid "${waypoint['waypoint_type']}"
+msgstr ""
+
 #: c2corg_ui/templates/route/index.html:11
 msgid "Add a route"
 msgstr ""
 
-#: c2corg_ui/templates/waypoint/index.html:48
+#: c2corg_ui/templates/waypoint/index.html:54
+#: c2corg_ui/templates/waypoint/index.html:87
 msgid "Add a waypoint"
 msgstr ""
 
@@ -16,11 +21,11 @@ msgstr ""
 msgid "Author"
 msgstr ""
 
-#: c2corg_ui/templates/base.html:27
+#: c2corg_ui/templates/base.html:31
 msgid "Back to homepage"
 msgstr ""
 
-#: c2corg_ui/templates/base.html:45
+#: c2corg_ui/templates/base.html:49
 msgid "Browse site in"
 msgstr ""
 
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -8,6 +8,7 @@
 /apache/app-c2corg_ui.wsgi
 /apache/wsgi.conf
 /MANIFEST
+/nbproject
```
gberaudo: What is `nbproject`?

ginold: it's a folder created by Netbeans project, I'm using it locally only.

gberaudo: I see. In this case you should instead configure a global ignore like explained here: https://help.github.com/articles/ignoring-files/.

asaunier: Thanks for the tip @gberaudo 
This means that this commit should be dropped (using ``git rebase -i`` as we did it for other tree changes).

asaunier:
Nice job @ginold !
I'll make a code review later at the office (we'll look at the conflict as well) but I have already tested your branch in my instance.

Here are some screenshots showing how it looks like:

screencapture-c2corgv6-demo-gis-internal-6555-waypoints-1452362473382

screencapture-c2corgv6-demo-gis-internal-6555-waypoints-1452362507620

Disclaimer: this is not the final layout (still being created) but @ginold is already implementing a first draft of the cards concept and adding some responsivity. It will be useful to show those concepts to the association without waiting for the final layout.

My comments about the look-and-feel:
* Background seems no longer white. Is it on purpose? At least it makes some white background appears behind the logo and search field.
* All list items are impacted by the changes: see the "menu" at the top of the page and the content of a waypoint page.
* I'm not found of the gray block with its inner scrollbar for the cards area, it uselessly takes extra space (margins) and crops the content. I think it's better to make the block expands a bit more to the sides (less margins) and to the bottom (no cropping => show all cards) and to get rid of the inner scrollbar. When the user will scroll down, I think that some parts will  shrink and stick to the top of the window (as in https://twitter.com/camptocamporg). The map will remain as well. Don't spend too much time implementing this part, the web design company will give instructions for this I think.
* Could you move the map to the right side of the page as in https://www.airbnb.fr/s/La-Ciotat?checkin=15%2F07%2F2016&checkout=23%2F07%2F2016&room_types%5B%5D=Entire+home%2Fapt&ss_id=qcmjsy0g&page=2 ? Airbnb search is a good example of what we want. Please note the fixed map when scrolling down as well as the shrinking "search criterias" section (we'll add it later).
* The "Add waypoint" might not be placed in this page, so it's a bit useless to duplicate it. Could you move it above the pagination buttons (or as the first card)?
* The pagination buttons + "add WP" at the end of the cards list are a bit useless. Could you remove them? If needed we'll add them back later.

asaunier:
@ginold As detailed in my comments, could you review all your changes and make sure the coding standard is OK, especially indentations, blank spaces and lines?

asaunier:
I have rebased the branch to fix the conflicts. By the way I have dropped the commits about IE9 and NetBeans.

Since force-pushing the rebased branch would make many of the comments above lost, I have prefered creating a new pull request #98.
All changes must be done in the latter pull request.

PR #84, opened 2016-01-10
Switch to Python 3.4

tsauerwein:
See https://github.com/c2corg/v6_api/pull/124

The only Python package that does not support Python 3 is closure-linter. Because the package is only needed to run the linter, I simply create a second virtualenv with Python 2.7 and install closure-linter in there. There are [no plans](https://github.com/google/closure-linter/issues/81#issuecomment-170260192) to add support for Python 3, so in the long run we have to find a replacement for closure-linter.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,16 @@
+{
+  "extends": "openlayers",
+  "globals": {
+    "app": false,
+    "goog": false,
+    "ol": false,
+    "ngeoModule": false,
+    "angular": false,
+    "$": false,
+    "Bloodhound": false
+  },
+  "rules": {
+    "no-console": 1,
+    "valid-jsdoc": 0
+  }
+}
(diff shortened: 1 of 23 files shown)


tsauerwein:
> The only Python package that does not support Python 3 is closure-linter. Because the package is only needed to run the linter, I simply create a second virtualenv with Python 2.7 and install closure-linter in there. There are no plans to add support for Python 3, so in the long run we have to find a replacement for closure-linter.

Problem solved by replacing closure-linter (and JsHint) with ESLint. The same rules as for ol3 are used with some exceptions.

tsauerwein:
Ok for merging this PR as well? Any opinions on ESLint?

asaunier:
OK for me

gberaudo:
+1 for switching to ESLint and basing our config on the one from OpenLayers.
The only drawback seems to be the spacing change before constructors.

tsauerwein:
So Python 3 it is. :)

PR #86, opened 2016-01-11
Make map and editing form interact

asaunier:
* [x] Update WP feature on the map when changing coordinates in the editing form
 * [x] bug: a new feature is created every time coords change
* [x] Click on the map to place a feature
* [x] Modify a WP feature on a the map
* [x] Styling of features on the map => currently not visible...
* [x] Update the coordinates in the editing form when using the map to place a feature
* [x] Keep other feature types (eg. lines) in mind
* [ ] Upgrade ngeo/OL3?


asaunier:
@gberaudo In commit 0084bb9 I have removed ``data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(point);`` because it seems to be done again (or cancelled) a few lines later. Looks good to you?

In the latter commit I also wondered if we should not move the ``data['geometry']`` creation/update and rather trigger it when the coordinates fields are changed. But I thought it was safer to keep it when submitting the form to make sure that it's actually up-to-date. Some lines are quite the same in both parts but I don't know if it is worth factorizing since the common part is only about a few lines.

At ff5eff9 I have somehow disabled your feature styling function when using the map in the editing page to make sure at least a default style is applied (for instance when no WP type is selected).

Commit a63e346 is still WIP: it is actually possible to draw points on the map and to move them but:
- they are not rendered on the map after being drawn (styling issue?)
- the initial feature (when editing an existing doc) is not editable (maybe not added in the right feature collection?)
- I need to find a way to avoid that only one point is drawn at a time on the map.

gberaudo:
> @gberaudo In commit 0084bb9 I have removed data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(point); because it seems to be done again (or cancelled) a few lines later. Looks good to you?

Yes, looks good.

asaunier:
The PR is ready for review!

Upgrading OL3/ngeo is not required to make it works but that could be the opportunity to have more recent versions. What do you think? If yes is it reasonable to upgrade to the latest commits?

Another point that could be discussed (in the future?): the PR adds feature editing interactions in the map but those tools are probably not needed in other maps. I wonder if we should have a set of map directives/controllers that would all extend a base couple of directive/controller (common features) but would contain the special tools.

asaunier:
Thanks @gberaudo for the review and the tips. I have made changes accordingly with e7727e7

gberaudo:
LGTM

asaunier:
Thanks, I have made the last change you suggested about ``drawType`` as well.
I'll merge when Travis gives a thumb up if no one else objects.

PR #88, opened 2016-01-11
Internal api url

### LINKED ISSUE #18: Public/internal URL for the API?
Since the API is requested by the UI client-side, we obviously need a public API URL.
But the UI also makes server-side calls to the API. Would it be then relevent to use an internal URL (or even localhost)?

asaunier:
Fixes #18

The default behaviour is unchanged => should work as is on dev instances.
If a ``api_url_internal`` is set in the config, it is used instead of ``api_url`` when the UI server requests the API. In that case, one may specify a Host header by setting ``api_url_host``.

### DIFF
--- a/c2corg_ui/tests/views/__init__.py
+++ b/c2corg_ui/tests/views/__init__.py
@@ -14,28 +14,28 @@ class BaseTestUi(BaseTestCase):
         self.settings = settings
 
     def _test_pages(self):
-        response = self.app.get(self._prefix, status=200)
+        route = '/%s' % self._prefix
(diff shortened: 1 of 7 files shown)


asaunier:
Fixes #18

asaunier:
Tested successfully on the main instance (apache mode) with config https://github.com/c2corg/v6_ui/blob/internal-api-url/config/main#L6-L8 => API requests from the UI server actually use localhost.

Please review.

tsauerwein:
That looks good to me, thanks.

gberaudo:
+1

PR #97, opened 2016-01-13
Make sure empty other_cultures block is not displayed in view pages

asaunier:
Using ``filter()``, ``if other_cultures:`` is always evaluated to ``True`` even though the doc is available in no other culture. In addition the list comprehension is a bit more readable. :P


gberaudo:
+1

PR #100, opened 2016-01-15
Add summary field to waypoints and routes

### LINKED ISSUE #91: Add "summary" on waypoint page
Interesting to have because ElasticSearch indexes title and summary

asaunier:
Fixes #91

6156e9a is only line number changes in PO files caused by the new HTML added in the templates. Would be cool if we could omit those line numbers => we wouldn't have those quite useless commits about the PO files.


tsauerwein:
LGTM

asaunier:
Thanks @tsauerwein !

PR #106, opened 2016-01-17
Disable MouseWheelZoom in maps

### LINKED ISSUE #105: Disable scrolling control by default on maps
Else scrolling the page also impacts the map when the cursor is above the map.

asaunier:
Fixes #105

### DIFF
--- a/c2corg_ui/static/js/map/diffmap.js
+++ b/c2corg_ui/static/js/map/diffmap.js
@@ -67,6 +67,7 @@ app.DiffMapController = function(mapFeatureCollection) {
    * @export
    */
   this.map = new ol.Map({
+    interactions: ol.interaction.defaults({mouseWheelZoom: false}),
     layers: [
       new ol.layer.Tile({
         source: new ol.source.OSM()
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -34,7 +34,8 @@ app.mapDirective = function() {
     restrict: 'E',
     scope: {
       'editCtrl': '=appMapEditCtrl',
-      'drawType': '@appMapDrawType'
+      'drawType': '@appMapDrawType',
+      'mouseWheel': '@appMapMouseWheel'
     },
     controller: 'AppMapController',
     controllerAs: 'mapCtrl',
@@ -97,7 +98,9 @@ app.MapController = function($scope, mapFeatureCollection) {
    * @export
    */
   this.map = new ol.Map({
-    interactions: ol.interaction.defaults({mouseWheelZoom:false}),
+    interactions: ol.interaction.defaults({
+      mouseWheelZoom: this['mouseWheel'] || false
+    }),
     layers: [
       new ol.layer.Tile({
         source: new ol.source.OSM()
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -97,6 +97,7 @@ app.MapController = function($scope, mapFeatureCollection) {
    * @export
    */
   this.map = new ol.Map({
+    interactions: ol.interaction.defaults({mouseWheelZoom:false}),
     layers: [
       new ol.layer.Tile({
         source: new ol.source.OSM()
--- a/c2corg_ui/templates/index.html
+++ b/c2corg_ui/templates/index.html
@@ -1,4 +1,4 @@
 <%inherit file="base.html"/>
 
 <h1 translate>Welcome to Camptocamp.org</h1>
-<app-map></app-map>
+<app-map app-map-mouse-wheel="true"></app-map>
--- a/c2corg_ui/templates/route/index.html
+++ b/c2corg_ui/templates/route/index.html
@@ -41,6 +41,6 @@
 </section>
 % if len(routes):
 <div class="map-right">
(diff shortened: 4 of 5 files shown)


tsauerwein:
Me personally I would not disable the mouse-wheel zoom. I think users can understand that the map is zoomed when you scroll over the map and if not the page is scrolled. Even Google does it that way (e.g [here](https://www.google.fr/maps/search/dentist/@45.644461,5.8797226,13.75z?hl=en)).

gberaudo:
I think we should try with the final design to see how it feels.
In general, I find it natural and convenient to be able to zoom with the scroll.

asaunier:
> I think users can understand that the map is zoomed when you scroll over the map and if not the page is scrolled.

You may be right concerning users understanding the difference etc. but the fact is that it's really a pain in the neck when you scroll the page and eventually get your cursor above the map: the map is then scrolled whereas you wanted to scroll the page. In addition is not user friendly to have to move the cursor outside of the map to be able to scroll the page.
I have experienced it myself many times and that's the reason I have opened this issue.
In v5 map scroll is disabled for embedded maps, enable for the main map page (with only the map in it). I really do think that map scrolling is not adapted to embedded maps, only to whole-page maps. No matter the design.

asaunier:
Have you tested the new design created by @ginold ?
I think we still have the problem with the mousewheel when trying to scroll down the page. I think the mousewheel support is not adapted to embedded maps...

tsauerwein:
> Have you tested the new design created by @ginold ?
> I think we still have the problem with the mousewheel when trying to scroll down the page. I think the mousewheel support is not adapted to embedded maps...

I just tested. On which page do you think it is a problem?

asaunier:
All pages with a map :)
Whenever you want to scroll the page down, if your cursor gets over the map, it's unexpectedly hugely zoomed in/out instead of moving down the page.
In addition in the list page, if the map extent is used to filter the results list, this will trigger a request to the API.

gberaudo:
+1 to disable for all full width maps.

asaunier:
Some feedbacks from the client:

> Sinon, il y a un truc pète c.... avec la carto : comme elle est en haut, je veux tout de suite voir le texte et donc je donne un coup de roulette... Dès que la carto passe sous la souris (ou si elle y est dès le départ), ça dézoome la carto et ne fait pas défiler le texte.
Idem quand je veux revenir en haut : un coup de roulette jusqu'en haut, mais dès que la carte passe sous la souris, le défilement s'arrête, et à la place la carte est zoomée.
Il faudrait que la carto soit insensible à la souris tant qu'on n'a pas cliqué dedans (un clic non relaché suffirait).
Et si on clique en dehors (ou même dès que la souris sort de la carto), on perd le focus et elle redevient insensible à la souris.

and also

>  j'avais pas osé le signaler, c'est tellement mal foutu que je pensais que ça faisait partie des dysfonctionnements normaux pour le moment

Someone suggested that

> Sur movescount y'a un truc pas mal (je sais pas si c'est dur à reproduire), mais si on scrolle depuis l'extérieur de la carte, ça ne zoom pas dedans, par contre si on arrête la souris dans la carte, le zoom est effectif. C'est assez pratique comme effet

Example: http://www.movescount.com/fr/moves/move89249382
Don't know if this is possible (and not to difficult).

asaunier:
@tsauerwein what do you think? Testers feedbacks say that the current mouse wheel interaction conflicts badly with the page scrolling.
In addition I feel concerned that accidentally wheel-zooming the map in the list/filter page will trigger unexpected requests to the API, with then an unwanted refresh of the results list.

In my opinion it is better to disable the mouse wheel map interaction, enabling it back later if we can find a satisfying solution like in movescount.

tsauerwein:
@asaunier Ok, this seems to be a bigger problem than I thought. But do you really want do disable mouse-wheel scrolling on all maps? For example on the filter page, where the map takes the full height, I would keep it.

The effect on Movescount looks interesting, and I think it is not that difficult to implement. I will give it a try.

asaunier:
I suggest that we merge the PR as is for now (for all maps) and that we open an issue about the movescount approach to implement it later (for all maps) when possible.

tsauerwein:
Feel free to merge, but I don't see why the scrolling should be disabled on the filter page, where the map takes the full height.

asaunier:
I can try adding an attribute to the ``app-map`` directive to switch on the mouse wheel? Would it make sense?

tsauerwein:
Yes, and the default can be `false` if you want.

asaunier:
OK, I have updated the branch and added the possibility to enable the mouse wheel interaction for some maps.

tsauerwein:
Thanks, that is looking good!

asaunier:
Thanks, still an indentation problem (I had not run ``make check`` because I have other files in the way). I will fix it and merge when Travis greenlights.

asaunier:
What about the diff map?
Since it's not really a navigation map (could be static actually), I could simply disable the mouse wheel interaction anyway. OK for you?

tsauerwein:
Yes, that is fine for me.

tsauerwein:
Yay, check this example: http://bl.ocks.org/tsauerwein/raw/38dc9e71a6781da42017/

The code is here: https://gist.github.com/tsauerwein/38dc9e71a6781da42017

PR #107, opened 2016-01-18
Automatic login to discourse using iframe

gberaudo:
When logging is initiated by the UI, an iframe is used to silently
log the user to discourse.

See https://github.com/c2corg/v6_forum/issues/5


asaunier:
No objection from my perspective.

gberaudo:
Rebased on master.

asaunier:
Looks good to me

PR #111, opened 2016-01-19
Make sure the model is created when creating a new document

### LINKED ISSUE #104: Georef on map does not work when creating a new document

asaunier:
As for now, when creating a new document, the model is only created when feeding the editing form. We then get an error when clicking on the map to georef a WP (undefined model).
This PR makes sure at least a default model is created and then updated when georeferencing a WP on the map.

Fixes #104


asaunier:
Any objection?

tsauerwein:
No, feel free to merge.

PR #112, opened 2016-01-19
Move feedback formatting to alert service + make documentediting use it

asaunier:
Please note that errors triggered when trying to save a document without being authenticated cannot be managed the right way because the API does not return a correct error message, see https://github.com/c2corg/v6_api/issues/85

### DIFF
--- a/c2corg_ui/externs/appx.js
+++ b/c2corg_ui/externs/appx.js
@@ -58,7 +58,7 @@ appx.SearchDocumentLocale;
 
 /**
  * @typedef {{
- *  msg: string,
+ *  msg: (string|Object),
  *  type: ?string,
  *  timeout: ?number
  * }}
--- a/c2corg_ui/static/js/alertservice.js
+++ b/c2corg_ui/static/js/alertservice.js
@@ -5,28 +5,38 @@ goog.require('app');
 
 
 /**
+ * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @constructor
  * @export
  */
-app.Alerts = function() {
+app.Alerts = function(gettextCatalog) {
 
   /**
    * @type {Array.<appx.AlertMessage>}
    * @private
    */
   this.alerts_ = [];
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.gettextCatalog_ = gettextCatalog;
 };
 
 
 /**
- * @param {appx.AlertMessage} msg Alert data.
+ * @param {appx.AlertMessage} data Alert data.
  * @export
  */
-app.Alerts.prototype.add = function(msg) {
+app.Alerts.prototype.add = function(data) {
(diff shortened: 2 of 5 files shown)


asaunier:
The main changes of this PR are:
* ``formatErrorMsg_`` and ``filterStr_`` functions are moved to the alerts service since usefull for most alerts (not only auth alerts).
* the documentediting tool also uses the alerts service.

asaunier:
Any objection?

asaunier:
Could anyone review this PR please?
I need it for PR #116

c2corg_ui/externs/appx.js
```diff
@@ -58,7 +58,7 @@ appx.SearchDocumentLocale;
 
 /**
  * @typedef {{
- *  msg: string,
+ *  msg: (string|Object),
```
tsauerwein: @gberaudo would want a type for this. ;)

c2corg_ui/static/js/documentediting.js
```diff
@@ -221,7 +231,7 @@ app.DocumentEditingController.prototype.errorRead_ = function(response) {
  */
 app.DocumentEditingController.prototype.submitForm = function(isValid) {
   if (!isValid) {
-    // TODO: use a common messaging tool to display the error
+    // Should not happen: form validity is tested before being able to submit.
```
tsauerwein: If this never happens, could you remove the `alert` below?

tsauerwein:
Two small comments, otherwise it looks ok.

c2corg_ui/static/js/documentediting.js
asaunier: Indeed. But we never know: perhaps the validity check has been omitted in the editing form.
I think I will use the standard alerts service instead of a raw alert(), using the regular way to give feedbacks to the user.

c2corg_ui/externs/appx.js
asaunier: Yes, I also expected @gberaudo would say something here :p 
That would make sense indeed, especially since the alerts service expects this is a ``response`` object with a ``response['data']['errors']`` property. But are the API responses always formatted the same way? I'm also afraid that the type is a bit complicated.

c2corg_ui/static/js/alertservice.js
```diff
@@ -41,10 +51,45 @@ app.Alerts.prototype.get = function() {
 
 
 /**
+ * @param {Object} response Response from the API server.
+ * @return {string}
+ * @private
+ */
+app.Alerts.prototype.formatErrorMsg_ = function(response) {
+  var errors = response['data']['errors'],
```
asaunier: Perhaps I should add an ``assert`` here to make sure that ``response['data']['errors']`` is available.

asaunier:
> Two small comments, otherwise it looks ok.
I will make the few changes I have just mentioned and will merge if there is no other objections.

PR #113, opened 2016-01-19
Fix list + editing page titles

asaunier:
I am not fully satisfied of this PR:
* with the bracket notation (``{{'Some title' | translate}}``), brackets appear until the translation is done (not very discrete for the page title!). The standard ``<span translate>Some title</span>`` avoids this but cannot be used inside a ``<title>`` element (the contained tags appear) => perhaps we should use the ``translate`` attribute directly in ``<title>``.
* the ``pagetitle`` block is defined in the parent ``base.html`` template. It seems then mako variables such as ``updating_doc`` (a boolean set in the child template) are not available in the block. As a result https://github.com/c2corg/v6_ui/blob/b7f9956dfce9f7dceb77193abc7130f2cdc5302e/c2corg_ui/templates/waypoint/edit.html#L8 always produces "Creating a waypoint" even though an existing document is edited.

### DIFF
--- a/c2corg_ui/locale/c2corg_ui-client.pot
+++ b/c2corg_ui/locale/c2corg_ui-client.pot
@@ -12,6 +12,10 @@ msgstr ""
 msgid "Add a waypoint"
 msgstr ""
 
+#: c2corg_ui/templates/auth.html
+msgid "Authentication"
+msgstr ""
+
 #: c2corg_ui/templates/document/history.html
 msgid "Author"
 msgstr ""
@@ -47,6 +51,14 @@ msgstr ""
 msgid "Created on"
 msgstr ""
 
+#: c2corg_ui/templates/route/edit.html
+msgid "Creating a route"
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "Creating a waypoint"
+msgstr ""
+
 #: c2corg_ui/templates/waypoint/edit.html
 msgid "Culture"
 msgstr ""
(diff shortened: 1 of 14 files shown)


tsauerwein:
> ... , brackets appear until the translation is done (not very discrete for the page title!). 

I guess [ng-cloak](https://docs.angularjs.org/api/ng/directive/ngCloak) does not work for `<title>`? Could you do it like this?

    <title translate="Default title">Default tile</title>

asaunier:
Thanks for the tip about ng-cloak!

Concerning the ``<title>`` element, as for now we have no access to this tag, since set in the main template https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L12
=> perhaps I'll have to change this.

asaunier:
> with the bracket notation ({{'Some title' | translate}}), brackets appear until the translation is done (not very discrete for the page title!). 

``ng-cloak`` does not work for ``<title>``. By the way a ``translate`` attribute does not work for ``<title>`` either. It seems we would need to use ``ng-bind``, maybe the same for the page title displayed inside the page itself (``ng-cloak`` uses some CSS trick to show the element only when ready, not sure it is what we want for a page title). But I don't want to spend too much time on this issue now.

Ditto for 
> the pagetitle block is defined in the parent base.html template. It seems then mako variables such as updating_doc (a boolean set in the child template) are not available in the block.

=> I have used the passed-to-template variables since the computed ones are not available in the parent template where the ``pagetitle`` block is initially created.
Could be improved in the future.

Please review.

gberaudo:
+1

PR #114, opened 2016-01-20
Disable form submit button when form is still untouched

asaunier:
The goal is to prevent that the form is submitted when no change has been applied yet.


tsauerwein:
:+1:

PR #98, opened 2016-01-13
List card layout

asaunier:
Replaces #81 
NEW
* [x] Lists are on the left and map on the right (ass seen on AirBnB.fr)
* [x] layout changes described at https://github.com/c2corg/v6_ui/pull/81#issuecomment-170268261
* [x] code changes listed in the original PR #81 (indentation, blank spaces etc.)
* [x] change icons depending on waypoint type

TODO 
* [ ] search filters
* [ ] list & map should take whole screen's height (it's too short on large screens)

DONE
* [x] changed the simple waypoint list into a card layout, and it's responsive
* [x] Helpers buttons (pagination) are now stylized and have icons
* [x] same layout for routes
* [x] click on whole card in order to see details, and not only on ```<a>``` tag

PR #116, opened 2016-01-20
Only authenticated users should have access to editing pages

### LINKED ISSUE #103: Make sure unauthenticated users have no access to the editing pages

asaunier:
Fixes #103 

* [x] Redirect unauthenticated users to auth page before giving them access to the editing pages
 * [x] View page > edit current doc
 * [x] View page > translate current doc (x ? other cultures)
 * [x] List page > add a new doc
* [x] When loading the editing page (form) => check if user is auth. If yes => nothing to do. If no:
 * [x] disable the AJAX request feeding the form
 * [x] redirect to auth page
 * [x] show alert message when user tries to save editing form while no longer authenticated
* [ ] Editing page: what if user logs out while on the editing page => form should become inactive?!

### DIFF
--- a/c2corg_ui/static/js/alertservice.js
+++ b/c2corg_ui/static/js/alertservice.js
@@ -42,6 +42,32 @@ app.Alerts.prototype.add = function(data) {
 
 
 /**
+ * @param {(string|Object)} msg
+ * @export
+ */
+app.Alerts.prototype.addSuccess = function(msg) {
+  this.add({
+    'type': 'success',
+    'msg': msg,
+    'timeout': 5000
+  });
+};
+
+
+/**
+ * @param {(string|Object)} msg
+ * @export
+ */
+app.Alerts.prototype.addError = function(msg) {
+  this.add({
+    'type': 'danger',
+    'msg': msg,
+    'timeout': 5000
+  });
+};
+
+
+/**
  * @return {Array.<appx.AlertMessage>}
  * @export
  */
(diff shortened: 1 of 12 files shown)


asaunier:
This PR is ready for review.

Point ``Editing page: what if user logs out while on the editing page => form should become inactive?!`` is still TODO but could be treated in a separated issue/PR.

tsauerwein:
> Point Editing page: what if user logs out while on the editing page => form should become inactive?! is still TODO but could be treated in a separated issue/PR.

When saving the form, the API will return an unauthenticated error. We could then show a message and tell the user to log in.

asaunier:
> When saving the form, the API will return an unauthenticated error. We could then show a message and tell the user to log in.

That's already done by https://github.com/c2corg/v6_ui/commit/5bb21a857e2695dc03f22c2c12f8619db163623c
=> an auth check is done before triggering the request to the API => if false a message is shown
=> if the API is requested anyway and returns a 403, a message is shown as well

But I meant we could make it more obvious that the user is no longer logget out and then cannot push the changes unless logging in first. I admit it's an "optional" improvement, that's why I think it could be done in a separated PR.

tsauerwein:
> But I meant we could make it more obvious that the user is no longer logget out and then cannot push the changes unless logging in first. I admit it's an "optional" improvement, that's why I think it could be done in a separated PR.

I would consider that a *nice-to-have*. I don't think this case occurs that often.

tsauerwein:
That looks great to me! :+1:

asaunier:
Thanks @tsauerwein for the review. I have fixed the typo.

I will merge when PR #98 is merged (I'll probably have to fix conflicts first).

Issue #117, opened 2016-01-21
Line numbers in the PO files creating merge conflicts


asaunier:
The translation strings extractor add comments with the filename and line numbers where the strings are located. This often causes many changes in the PO files when templates are a bit changed (thus shifting the line numbers). This is a problem because:
* it uselessly add not-that-useful changes in the PR/commits history
* it often causes conflicts between branches

It would be cool if we could find a way getting rid of those line numbers, don't you think?

asaunier:
The extraction is done using Angular-Gettext: https://angular-gettext.rocketeer.be/
Perhaps there's an option/way to omit line numbers.

Another option would be to always omit changes (ie. not commit them) concerning line numbers. Only changes about new strings would be commited.

PR #118, opened 2016-01-21
Remove line numbers from pot

gberaudo:
Fix https://github.com/c2corg/v6_ui/issues/117


tsauerwein:
Cool! Thanks for the fix.

asaunier:
Don't you think the extractor will set them back?

asaunier:
Oops, I missed the change in the extractor: https://github.com/c2corg/v6_ui/blob/remove_pot_line_numbers/tools/extract-messages.js#L9

Looks really great, thanks a lot!

PR #120, opened 2016-01-26
Translation issues

ginold:
- drop lists and errors are translated correctly
- added some feedback to editing forms, indicating when a field is ok or not.


asaunier:
I think it is no use having to distinct commits here. Could you merge them into a single one?
I have also added several questions/comments embedded in the commits.

asaunier:
Please merge when Travis passes.

asaunier:
Travis fails because:

> c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po:548: duplicate message definition...
c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po:463: ...this is the location of the first definition
msgmerge: found 1 fatal error

PR #124, opened 2016-01-27
Page header navigation 

ginold:
- page-header is responsive and stylized.
- you can hover with the mouse on the search bar -> it will expand. On small devices it will roll over the other buttons and hide them (triggered by hover or click).
- there is one dropdown-button with sub-menus (login/logout/my account), it changes depending if the user is logged in or not.
- the search list is better lookin'. Lines too long are split in two on small devices.

### DIFF
--- a/c2corg_ui/static/js/lang.js
+++ b/c2corg_ui/static/js/lang.js
@@ -21,6 +21,7 @@ app.langDirective = function() {
     controllerAs: 'langCtrl',
     bindToController: true,
     template: '<select ' +
+        'class="form-control lang-selector" ' +
         'ng-model="langCtrl.culture" ' +
         'ng-options="culture as langCtrl.translate(culture) ' +
         'for culture in langCtrl.cultures" ' +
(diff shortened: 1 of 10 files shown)


asaunier:
Looks like a great job :)

I have made a few comments. In addition I see that some changes revert some others, you also have "fix" commits (eg.  b5aaba3). I wonder if it would not make sense to merge all commits (except the first one) in a single one.

asaunier:
Both in Chrome and Chromium, as well as in FF I see that some glyphicons fonts files are trying to be downloaded but fail with 404 errors:
http://c2corgv6-demo.gis.internal:6555/static/fonts/glyphicons-halflings-regular.woff2

Do we really need them?

If yes I see those files are available in ``/v6_ui/node_modules/bootstrap/fonts`` => the URL is not correct and should be http://c2corgv6-demo.gis.internal:6555/node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff2

asaunier:
Looks good to me, please merge.

PR #127, opened 2016-01-28
Parse documents base markup syntaxes

### LINKED ISSUE #121: Parse BBcode etc. when displaying the content of a document
Documents contain BBcode as well as additional formating syntaxes.

See http://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte for details about the formatting syntaxes available in v5.

* [x] BBcode (see #127)
* [x] Markdown (see #127)
* [ ] Custom tags (see #131)

asaunier:
Fixes #121 

Support the various markups (including custom ones)
* [x] BBcode
* [x] Markdown
* [x] Prevent user submitted HTML/JS from being interpreted

### DIFF
--- a/c2corg_ui/locale/c2corg_ui-client.pot
+++ b/c2corg_ui/locale/c2corg_ui-client.pot
@@ -71,9 +71,45 @@ msgstr ""
 msgid "Describe here the waypoint"
 msgstr ""
 
+#: c2corg_ui/templates/i18n.html
+msgid "Document has been deprotected"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Document has been protected"
+msgstr ""
+
 #: c2corg_ui/templates/route/view.html
 #: c2corg_ui/templates/waypoint/view.html
 msgid "Edit"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in ca"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in de"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in en"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in es"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in eu"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in fr"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Edit in it"
 msgstr ""
 
 #: c2corg_ui/templates/route/edit.html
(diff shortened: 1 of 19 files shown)


asaunier:
Combining (!) python modules ``bbcode`` and ``markdown`` we can quite easily support much (?) of v5 document formatting.

I have also tried ``mistune`` as markdown parser as it cleams to be 4x faster and better designed than ``markdown``. Maybe true but not that obvious at first human sight. In addition the rendering seems the same in both cases but I would need to test on many different documents.

On the other hand ``markdown`` seems to be a more active project with more contributors...

I think the choice will be made on the possibilities to extend the supported syntax in order to convert c2corg custom tags. Ditto for BBcode tags since some tags are quite specific to c2corg such as ``[toc 2 right]``, ``[img=188119 right]Lever de soleil sur le Mont Blanc[/img]`` or ``[[routes/53781|Arête des Bosses]]``. Not to mention the ``L#`` tags...

asaunier:
I would like to merge this PR as is and will open a new PR for custom tags. This way we will already support at least the parsing of standard tags without waiting for custom tags handling which might be a quite huge work.

For the records if we want to use mistune instead of python_markdown:

* requirements.txt:
```
cython==0.23.4
mistune==0.7.1
```
* c2corg_ui/templates/utils/format.py
```
import mistune

text = mistune.markdown(text)
```

tsauerwein:
Could we maybe only use one of the markup languages? Either BBCode or Markdown?

gberaudo:
Same remark as @tsauerwein.

tsauerwein:
> Could we maybe only use one of the markup languages? Either BBCode or Markdown?

On the other hand, Discourse also supports BBCode AND Markdown: https://meta.discourse.org/t/post-format-reference-documentation/19197

asaunier:
> Could we maybe only use one of the markup languages? Either BBCode or Markdown?

That would probably make sense I guess, since . But in v5 document both systems are actually used. Which means that: 
* either we both support them (+ the custom tags)
* or we drop one system. But we'll have to convert dropped tags to the remaining syntax while migrating the DB from v5 to v6.

It also depends if we have to make adaptations to the custom tags (and make conversions while migrating).

Honestly I would prefer to stick to option 1. At least first. And then option 2 if we realize that we have too much trouble supporting the full v5 formatting. As you main notice, supporting both standard BBcode and Markdown is not that a big deal and it's not a huge change if we eventually decide to remove the support of one of them.

asaunier:

I have made changes so that:
- user-submitted tags are escaped at output time
- standard bbcode and markdown are supported. One syntaxe may be disabled if needed. For instance in a version "comment" field, bbcode is useful for instance to add a link towards a contributor or a document, on the other hand we don't want to have titles or paragraphs (markdown at least adds ``<p>`` tags around a simple string).
- text attributes formatting is done for list/view/diff/history pages (not edit page since not using mako).

tsauerwein:
For me it's ok, feel free to merge.

Issue #130, opened 2016-01-29
Improve the view page tools positioning


asaunier:
As for now, in a document view page, buttons and links such as "edit", "translate in ...", history/versions are displayed a bit randomly in the view page.

We need to improve their placement to make them quite obvious (at least the main tools) without disturbing too much the layout of the document. => see the wireframes and layout document.
For instance some tools will be available as floating buttons on the right hand side of the page.

Issue #129, opened 2016-01-29
Organise content/attributes in a document view page


asaunier:
(to be confirmed when wireframes and layout are ready)

As for now a few attributes key/value pairs are shown in WP and routes pages using ``<li>`` tags.
Using ``<li>`` tags (at least not for all attributes) is not good because some fields such as "description" might contain advanced formating (with lists, images, etc.) => remove ``<li>`` tags, at least for text fields.

Move map up, maybe make it not span on the full page width depending on the available page width ? See layout and wireframes.

Some attributes could be grouped together, for instance:
* the gradings (for routes)
* the height up/down (using icons to distinguish up and down etc.)
* for a WP, the elevation could be shown on the same line than the WP name (ie title)

PR #132, opened 2016-02-02
Routes & Waypoints detail view style

ginold:
screenshot from 2016-02-04 12 12 12
screenshot from 2016-02-04 12 12 29
screenshot from 2016-02-04 12 13 02

- data displayed in a list with icons (we will need to change icons and use the correct ones)
- there are floating buttons in the header of the page that adapt their size depending on the title height
- there is sometimes a problem with ng-init= something -> from time to time angular won't display the contents, saying "Lexer Error: Unexpected next at column{1} in expression [{2}]."

Related to #130 and #129


asaunier:
Please check the fuzzy tags in the "locale" files.

asaunier:
I have fixed some points and rebased the commits. Merging.

PR #133, opened 2016-02-02
Add title_prefix for routes titles

### LINKED ISSUE #108: Show title_prefix of routes
`title_prefix` contains the name of the main waypoint of a route and can be shown in:

 * listings
 * detail view
 * simple search results

asaunier:
Fixes #108

### DIFF
--- a/c2corg_ui/externs/appx.js
+++ b/c2corg_ui/externs/appx.js
@@ -50,6 +50,7 @@ appx.SearchDocument;
 /**
  * @typedef {{
  *   title: string,
+ *   title_prefix: ?string,
  *   culture: string
  * }}
  */
--- a/c2corg_ui/static/js/search.js
+++ b/c2corg_ui/static/js/search.js
@@ -180,7 +180,9 @@ app.SearchController.prototype.createAndInitBloodhound_ = function(type) {
 
         return documents.map(function(/** appx.SearchDocument */ doc) {
           var locale = doc.locales[0];
-          doc.label = locale.title;
+          doc.label = type === 'routes' && locale.title_prefix ?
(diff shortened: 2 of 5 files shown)


asaunier:
Please note when using a french interface, the WP name and the route name are separated using `` : `` instead of ``: ``. The interface language is available for the list page and the simple search results, not really for a route view page => I use the culture of the currently shown document, even if it's not fully relevant.

I wonder if we should not simplify by using some other separator such as `` - `` that would be the same whatever the interface language.

tsauerwein:
Thanks! That looks good.

tsauerwein:
> I wonder if we should not simplify by using some other separator such as - that would be the same whatever the interface language.

Good idea!

asaunier:
Thanks @tsauerwein 
I have made the suggested simplification. Hope it will be OK with the client.

tsauerwein:
Could you also use "-" instead of ":" in the API? https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/utils.py#L34

Thanks!

asaunier:
OK, I will make the change as well.
I will confirm with the client that's OK before doing further change though.

asaunier:
As discussed with the client, using `` : `` between ``title_prefix`` and ``title`` for routes, whatever the language, is an acceptable compromise. Using ``:`` is prefered to any other character since:
* it's the one currently used
* it marks a hierarchy WP > route
* other characters such as - or > are already used in some other cases in route names (for instance for route going from one point to another, eg. ``traversée E > W``).

I have made the same change in the API as requested by @tsauerwein See https://github.com/c2corg/v6_api/pull/139

PR #138, opened 2016-02-04
title with brackets resolved (kinda)

ginold:
fixes #113 
screenshot from 2016-02-04 13 53 14
Page titles that are normally translated, for instance 'Waypoints' or 'Routes' have a default title in english that is translated on the go into other languages but will stay in english in the browser history.

Other pages that use waypoint_title or route_title as page title don't have this issue of course.

### DIFF
--- a/c2corg_ui/locale/c2corg_ui-client.pot
+++ b/c2corg_ui/locale/c2corg_ui-client.pot
@@ -21,10 +21,6 @@ msgstr ""
 msgid "Around"
 msgstr ""
 
-#: c2corg_ui/templates/auth.html
-msgid "Authentication"
-msgstr ""
-
 #: c2corg_ui/templates/document/history.html
 msgid "Author"
 msgstr ""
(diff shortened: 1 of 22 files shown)


asaunier:
Instead of having ``- camptocamp.org`` added to the ``<title>`` tag for every single template, I think you should define a helper once in ``templates/helpers.html`` like:
```
<%def name="show_title(title)">\
${title + ' - ' if title else ''}Camptocamp.org\
<%/def>
```

and call it in your pagetitle block:
```
<%block name="pagetitle"><title ng-init="id='${route_id}'" ng-bind="id != 'None' ? ('Editing a route' | translate) : ('Creating a route' | translate)">${show_title('Create/edit route')}</title></%block>\
```
The helper must be included in the template, for instance:
```
<%namespace file="../helpers.html" import="add_pagination_links, show_title"/>
```

asaunier:
I think that the commits could be merged in a single one (except maybe the PO files changes?).

asaunier:
I have fixed some issues and rebased the commits. Merging.

PR #140, opened 2016-02-05
Disable fuzzy matching in PO files

asaunier:
Gettext often tries to guess translations, adding "fuzzy" tags, with more or less luck.
It's not really useful (the client will do the translations) and the suggested translations are often incorrect.
The PR disables this behaviour.
See doc here https://www.gnu.org/software/gettext/manual/html_node/msgmerge-Invocation.html


gberaudo:
+1

PR #146, opened 2016-02-08
Rename API's 'l' parameter to 'lang'

asaunier:
See https://github.com/c2corg/v6_api/issues/101#issuecomment-166672389
Requires https://github.com/c2corg/v6_api/pull/148

Should we rename the "pl" parameter as well?


tsauerwein:
To be honest, contrary to my old comment, I would now prefer the short version `?l=...`. But as you already did the work, I let you decide. :)

asaunier:
OK, no problem to cancel the PRs, I have not spent that much time on them.

PR #147, opened 2016-02-09
waypoint main title + h1 view titles

ginold:


c2corg_ui/templates/waypoint/view.html
```diff
@@ -42,7 +42,7 @@
 </%block>\
 
 <header class="view-details-title">
-  <span class="heading">${locale['title']} <label class="badge" translate>waypoint</label></span>
+  <span class="heading"><h1>${locale['title']} <label class="badge" translate>waypoint</label></h1></span>
```
asaunier: Do you really need to have both ``<span class="heading">`` and ``<h1>``?
If the "heading" class is required, what about simply writing
```
<h1 class="heading">
```
?

c2corg_ui/templates/route/view.html
```diff
@@ -18,7 +18,7 @@
 </%block>\
 
 <header class="view-details-title">
-  <span class="heading">${show_route_title(locale)} <label class="badge" translate>route</label></span>
+  <span class="heading"><h1>${show_route_title(locale)}<label class="badge" translate>route</label></h1></span>
```
asaunier: Same comment about ``<span class="heading"><h1>`` than for waypoints

asaunier:
Else looks good to me (and tests pass in my instance).

PR #148, opened 2016-02-09
filters aligment problem

ginold:


asaunier:
Looks good to me.
Travis failed about some closure JS check (whereas there is no change in JS in your PR!?) => I have made it run once more. Please merge when travis passes.

PR #149, opened 2016-02-09
Side menu

ginold:
responsive side menu

### DIFF
--- a/c2corg_ui/templates/base.html
+++ b/c2corg_ui/templates/base.html
@@ -28,18 +28,11 @@ closure_library_path = settings.get('closure_library_path')
   </head>
   <body>
 
-    <div class="page-menu">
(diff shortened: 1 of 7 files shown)


asaunier:
I feel a little uncomfortable with all that JS code snippets directly embedded in the mako templates. Are you really sure they cannot be added to the angular files even though they use no angular?

@tsauerwein @gberaudo what do you think we could do?

tsauerwein:
I am surprised to find that much CSS inside JavaScript. 95% of the functionality could be done in pure CSS (for example checks like `if ( (window.innerWidth > tabletScreen && ...`). Here is a good [example](https://codyhouse.co/gem/responsive-sidebar-navigation/) with a similar menu which uses a more CSS-based approach.

asaunier:
I guess this PR can be closed and the branch deleted?

PR #154, opened 2016-02-11
Check for unused `goog.require`s

### LINKED ISSUE #76: Check for unused 'goog.require's
See https://github.com/camptocamp/ngeo/pull/469

tsauerwein:
Closes #76 

Directives that are used in partials can be flagged with `@suppress {extraRequire}`, e.g.:

    /** @suppress {extraRequire} */
    goog.require('ngeo.searchDirective');

### DIFF
--- a/build.json
+++ b/build.json
@@ -41,41 +41,10 @@
       "node_modules/ngeo/externs/ngeox.js"
     ],
     "jscomp_error": [
-      "accessControls",
-      "ambiguousFunctionDecl",
-      "checkEventfulObjectDisposal",
-      "checkRegExp",
-      "checkTypes",
-      "checkVars",
-      "const",
-      "constantProperty",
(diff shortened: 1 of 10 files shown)


asaunier:
Not sure to understand what this "suppress" flags does and why/when it is needed?

tsauerwein:
For example when you are using a directive in a partial, you have to require e.g. `ngeo.searchDirective`. Because you are not using `ngeo.searchDirective` in the JavaScript code, the compiler will complain. But you actually need the require, so you can tell the compiler to ignore the warning.

asaunier:
OK thanks for the explanations. Looks good to me.

gberaudo:
I think each suppress warning deserves a detailed comment (in the code) explaining why it is needed.
Otherwise it looks like black magics.

PR #156, opened 2016-02-11
Parse c2corg custom wikilinks

asaunier:
This PR adds a first extension to the standard markdown/bbcode syntaxes: wikilinks.

Links such as ``[[routes/12345/etc|Some label]]`` are converted to ``<a href="/routes/12345/etc">Some label</a>``.

One problem left: flake8 complains because some of the methods used in the markdown extension class are not lowercased:
> c2corg_ui/format/wikilinks.py:19:9: N802 function name should be lowercase
c2corg_ui/format/wikilinks.py:31:9: N802 function name should be lowercase
c2corg_ui/format/wikilinks.py:44:5: N802 function name should be lowercase

but I can't change the case of this functions since they are overloading methods defined in the Markdown classes. Is there a way to tell flake8 to ignore those errors here?

### DIFF
--- a/c2corg_ui/format/README.md
+++ b/c2corg_ui/format/README.md
@@ -1,0 +1,18 @@
+Parsing the custom formating syntax of camptocamp.org
+=====================================================
+
+Base syntaxes
+-------------
+
+Camptocamp.org uses several base syntaxes to format the documents text attributes.
+External python libraries are used to parse those base syntaxes:
+* Markdown: [Python-Markdown](https://github.com/waylan/Python-Markdown)
+* BBcode: [bbcode](https://github.com/dcwatson/bbcode)
+
+Custom tags
+-----------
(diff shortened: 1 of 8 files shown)


tsauerwein:
> Is there a way to tell flake8 to ignore those errors here?

Yes with `# noqa`:

    def setUpClass(cls): # noqa

https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/__init__.py#L115

tsauerwein:
That looks good. What about a test?

asaunier:
> Yes with # noqa:

Thanks for the tip.

I have added some tests as well.

tsauerwein:
Thanks! Feel free to merge.

asaunier:
Thanks @tsauerwein for reviewing!

Issue #157, opened 2016-02-11
Redirection to document view page after editing is broken


asaunier:
Maybe related to the recent changes in the API about document editing but the id of the document is no longer correctly detected in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/documentediting.js#L341 (set to "undefined"). As a result the redirected URL is wrong.

It seems that ``response.data`` has no longer the appropriate attribute (ie. document_id).

OK when creating a document.
KO when updating it.

tsauerwein:
Yes, the document id is no longer returned when doing an update. You could use the following instead:

    var url = app.utils.buildDocumentUrl(this.module_, this.id_, this.lang_);

asaunier:
Thanks for the tip, I have created a PR #159

PR #159, opened 2016-02-12
Fix missing id when redirecting after updating document

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/157


tsauerwein:
:+1:

Issue #160, opened 2016-02-12
Egg generation by Makefile


gberaudo:
Does anyone know what the egg targets in the Makefile are supposed to do?

I would expect the `$(SITE_PACKAGES)/c2corg_ui.egg-link` target to create an
`.build/venv/lib/python3.4/site-packages/c2corg_ui.egg-link` file, but it is currently not the case.

Is there something broken?

asaunier:
@gberaudo is this issue still valid?

PR #161, opened 2016-02-12
Only refresh requirements.txt when needed

gberaudo:
Currently, each time we run the Makefile with the serve target,
the egg target is run and the requirements.txt are refreshed.
This is due to the egg-link never being created.

- Remove broken egg target.
- Depend on pip file instead of a directory.
- Only trigger a requirements.txt refresh when needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,16 +75,13 @@ flake8: .build/venv/bin/flake8
 lint: .build/node_modules.timestamp .build/eslint.timestamp
 
 .PHONY: install
-install: build install-dev-egg template .build/node_modules.timestamp
+install: build template .build/node_modules.timestamp
 
 .PHONY: template
 template: $(TEMPLATE_FILES)
 
 .PHONY: less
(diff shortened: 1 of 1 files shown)


gberaudo:
Related to #160.

gberaudo:
Any review?

asaunier:
For me it's OK but I'm not sure I really understand your changes, thus not feeling allowed to give you the clearance :)

gberaudo:
If there is no objection I am going to merge this PR.
@tsauerwein?

tsauerwein:
The `-e .` installs the project in *[dev mode](https://pip.pypa.io/en/stable/reference/pip_install/#editable-installs)*, so I am not sure why we had the target for the egg-link. If you confirm that everything still works properly, feel free to merge.

gberaudo:
I have been using this branch for a week without noticing any issue, so merging.
Thanks for the review.

PR #143, opened 2016-02-05
Show associations + maps and areas

gberaudo:
Show a simple list of maps, areas and associated documents at the bottom of the waypoint and route views.

PR #162, opened 2016-02-12
Minimal and temporary layout for the associations

asaunier:
This PR moves the associations links recently added by @gberaudo (see https://github.com/c2corg/v6_ui/pull/143) to some temporary (but nicer :P) block below the description block.


gberaudo:
+1

Issue #136, opened 2016-02-03
Document-associating tool


asaunier:
Issue #135 deals with displaying associated documents.

In order to add (or remove!) associations, a directive is needed. Short spec (to be discussed):
* available to auth users only
* based upon Elastic Search => the user types keyword, the tool suggests matching documents
* suggest "in the area" documents (based upon the geometry) (already exists in v5)

Questions:
* do we need to restrict on some document_types?

Based upon API's https://github.com/c2corg/v6_api/pull/125

PR #165, opened 2016-02-12
Add associations

gberaudo:
- [x] add a directive to search for documents;
- [x] add a directive for displaying an associated document card;
- [x] display error message when associating a document fails;
- [x] use directives in waypoint and route views.

Fixes https://github.com/c2corg/v6_ui/issues/136

### DIFF
--- a/c2corg_ui/static/js/alertservice.js
+++ b/c2corg_ui/static/js/alertservice.js
@@ -26,6 +26,20 @@ app.Alerts = function(gettextCatalog) {
 
 
 /**
+ * Use this function to annotate a string for extraction by the
+ * gettext extract tool. The call to this function will be eliminated
+ * by the Closure compiler when minifying, so it should not have any
+ * performance effect.
+ * See https://angular-gettext.rocketeer.be/dev-guide/annotate-js/
+ * @param {string} str String to have extracted by gettext tool
+ * @return {string}
+ */
+app.Alerts.prototype.gettext = function(str) {
+  return str;
+};
+
+
+/**
  * @param {appx.AlertMessage} data Alert data.
  * @export
  */
--- a/c2corg_ui/static/js/associationcarddirective.js
+++ b/c2corg_ui/static/js/associationcarddirective.js
@@ -1,0 +1,17 @@
+goog.provide('app.associationCardDirective');
+
+goog.require('app.utils');
+
+
+/**
+ * @return {angular.Directive} Directive Definition Object.
+ */
+app.associationCardDirective = function() {
+  var url = app.utils.buildDocumentUrl('{{::doc.documentType}}', '{{::doc.document_id}}', '{{::doc.locales[0].lang}}');
+  return {
+    restrict: 'E',
+    template: '<a ng-href="' + url + '">{{::doc.locales[0].title}}</a>'
+  };
+};
+       
+app.module.directive('appAssociationCard', app.associationCardDirective);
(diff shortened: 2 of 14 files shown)


gberaudo:
Depend on https://github.com/c2corg/v6_ui/pull/161.

gberaudo:
Ready for review.

gberaudo:
I think I addressed all (or most?) comments.
Notably I improved the API service by having it handle failures.
The commit about console.log was removed and an issue opened: #169.

c2corg_ui/static/js/alertservice.js
```diff
@@ -26,6 +26,20 @@ app.Alerts = function(gettextCatalog) {
...
 /**
+ * Use this function to annotate a string for extraction by the
+ * gettext extract tool. The call to this function will be eliminated
+ * by the Closure compiler when minifying, so it should not have any
+ * performance effect.
+ * See https://angular-gettext.rocketeer.be/dev-guide/annotate-js/
+ * @param {string} str String to have extracted by gettext tool
+ * @return {string}
+ */
+app.Alerts.prototype.gettext = function(str) {
```
tsauerwein: Is this function supposed to be used only for alerts? If it is a generic function, maybe it would make sense to put it into some utils module and not on `app.Alerts`.

c2corg_ui/static/js/associationcarddirective.js
```diff
@@ -0,0 +1,17 @@
...
+
+
+/**
+ * @return {angular.Directive} Directive Definition Object.
+ */
+app.associationCardDirective = function() {
+  var url = app.utils.buildDocumentUrl('{{doc.documentType}}', '{{doc.document_id}}', '{{doc.locales[0].lang}}');
+  return {
+    restrict: 'E',
+    template: '<a ng-href="' + url + '">{{doc.locales[0].title}}</a>'
```
tsauerwein: `{{doc.locales[0].title}}` is not going to change, right? You could use an one-time-binding: `{{::doc.locales[0].title}}`

tsauerwein:
I did a quick review. This looks very good!

c2corg_ui/static/js/alertservice.js
gberaudo: This function deals with a very specific use case and I think we currently only need it on `app.Alerts`. I would rather keep it there for now and only create the generic version if
we really need it.

tsauerwein: > I would rather keep it there for now and only create the generic version if we really need it.

Ok for me.

asaunier:
I wonder if the ``appSelect`` changes do not impact the normal search tool as well. In my instance http://c2corgv6-demo.gis.internal:6555/waypoints/641547/fr
* the association tool actually works
* but it's no longer possible to select a result in the simple search tool at the top of the page!?

gberaudo:
>  it's no longer possible to select a result in the simple search tool at the top of the page!?

Indeed the binding broke the normal search. Thanks for finding this. I added a commit with a fix.
An alternative would be to create a tiny wrapping directive.

@asaunier, @tsauerwein a preference?

asaunier:
Your solution looks good to me and seems fairly simple.
Using wrapping directives (extending a base search directive if I have correctly understood) would also be a solution but maybe a bit more "verbose" and time-consuming to develop?

asaunier:
Thanks for the changes. After rebasing, if no one objects about your ``apiService`` approach, feel free to merge. Please then open an issue to remind us that we'll have to use the ``apiService`` for all directives.

gberaudo:
Thanks for the review. I squashed the commits and will merge on Travis green light.

Issue #82, opened 2016-01-08
Filepaths vs web caches


asaunier:
To prevent problems with reverse proxies or browser caches, the URL of static files should change when updating the UI app, for instance using uuids.

Not sure something is already available in the project. What about the ``version`` config in https://github.com/c2corg/v6_ui/blob/master/common.ini.in#L10 ?

tsauerwein:
See also: http://docs.pylonsproject.org/projects/pyramid/en/1.6-branch/narr/assets.html#cache-busting

gberaudo:
What about doing like in schweizmobil where we prefix by the git hash?

asaunier:
> What about doing like in schweizmobil where we prefix by the git hash?

Yep. By the way we have a ``version`` param in https://github.com/c2corg/v6_ui/blob/master/common.ini.in#L10 that is set using the git hash in the default config: https://github.com/c2corg/v6_ui/blob/master/config/default#L4 but set back to 0 in config/dev (?) https://github.com/c2corg/v6_ui/blob/master/config/dev#L3

asaunier:
> See also: http://docs.pylonsproject.org/projects/pyramid/en/1.6-branch/narr/assets.html#cache-busting

Requires Pyramid 1.6 which will be provided by https://github.com/c2corg/v6_ui/issues/139

asaunier:
Since we now use Pyramid 1.6, I have taken a look at Pyramid cache busting tools quoted by @tsauerwein : 
http://docs.pylonsproject.org/projects/pyramid/en/1.6-branch/narr/assets.html#cache-busting

The simplest solution based upon ``QueryStringConstantCacheBuster(str(int(time.time())))``works out of the box by adding the example code to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/__init__.py#L10 => all ``static/`` URLs added by ``request.static_url`` in the mako templates get a ``x=123456`` parameter added to their querystring.

But as the doc says:
> This is almost never what you want in production as it does not allow fine-grained busting of individual assets.

Not to mention that

> Many caching HTTP proxies will fail to cache a resource if the URL contains a query string.

I have also tried the ``GitCacheBuster`` class that uses the file git hash but:
[...]

PR #174, opened 2016-02-17
Add cache busting

asaunier:
GMF 2 also uses Pyramid 1.6 and its cache busting system.

I have then ported the GMF code
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/__init__.py#L344-L352
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/lib/cacheversion.py
to our project with a few adaptations (python3, no server caching yet).

It works pretty nice and almost out of the box.

Instead of a random uuid string, I would like to use the short last commit hash provided by https://github.com/c2corg/v6_ui/blob/cache_buster2/config/default#L4
but I don't know how to get a config parameter from within a standard function (say ``get_cache_version``). I have tried to use some ``CACHE_VERSION`` variable set in the ``main()`` function but it's then not available in ``get_cache_version``. flake8 by the way says that:
> c2corg_ui/__init__.py:3:1: F401 'CACHE_VERSION' imported but unused
c2corg_ui/__init__.py:14:5: F811 redefinition of unused 'CACHE_VERSION' from line 3
c2corg_ui/__init__.py:14:5: F841 local variable 'CACHE_VERSION' is assigned to but never used

Anyone knows what's the best way to do this? Thanks!

Fixes https://github.com/c2corg/v6_ui/issues/82

### DIFF
--- a/c2corg_ui/lib/cacheversion.py
+++ b/c2corg_ui/lib/cacheversion.py
@@ -1,0 +1,35 @@
+# Inspired from GeoMapFish
+# noqa https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/lib/icacheversion.py
+
+from urllib.parse import urljoin
+
+
+CACHE_PATH = []
+
+
+def version_cache_buster(request, subpath, kw):
+    # TODO get cache version from server caching?
+    cache_version = request.registry.settings.get('cache_version')
+    return urljoin(cache_version + '/', subpath), kw
+
+
+class CachebusterTween:
+    ''' Get back the cachebuster URL. '''
+    def __init__(self, handler, registry):
+        self.handler = handler
+
+    def __call__(self, request):
+        path = request.path_info.split('/')
+        if path[1] in CACHE_PATH:
+            # remove the cache buster
+            path.pop(2)
+            request.path_info = '/' .join(path)
+
+        response = self.handler(request)
+
+        if path[1] in CACHE_PATH:
+            response.headers['Access-Control-Allow-Origin'] = '*'
+            response.headers['Access-Control-Allow-Headers'] = \
+                'X-Requested-With, Content-Type'
+
+        return response
--- a/c2corg_ui/lib/cacheversion.py
+++ b/c2corg_ui/lib/cacheversion.py
@@ -17,19 +17,50 @@ class CachebusterTween:
     ''' Get back the cachebuster URL. '''
     def __init__(self, handler, registry):
         self.handler = handler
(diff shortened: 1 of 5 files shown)


c2corg_ui/lib/cacheversion.py
```diff
@@ -0,0 +1,41 @@
...
+
+# TODO cache
+def get_cache_version():
+    """Return a cache version that is regenerate after each cache invalidation"""
+    # return CACHE_VERSION
+    return uuid.uuid4().hex
+
+
+def version_cache_buster(request, subpath, kw):  # pragma: nocover
+    return urljoin(get_cache_version() + "/", subpath), kw
```
tsauerwein: Here you can get the Git revision with:

    request.registry.settings.get('cache_version')

tsauerwein:
This looks good. How does it work in development mode with `make serve`? The cache is not used in that case?

gberaudo:
We need to do something to prevent cache poisoning, see https://github.com/camptocamp/schweizmobil_re3/issues/516. I can have a look at it before merging.

tsauerwein:
By cache poisoning you mean that old versions of static files are not removed from Varnish?

gberaudo:
No, I mean that an old version of the file is used instead of the correct one.

asaunier:
I have updated the PR to take @tsauerwein 's suggestions.

Please note that the tool does not work for the ``node_modules`` and ``closure`` static views. I see that in GMF some special code is used: https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/__init__.py#L301-L318
Do we really need this? For instance if finally those ``node_modules`` files are also built together or downloaded from a CDN?

> How does it work in development mode with make serve? The cache is not used in that case?

Mmm you probably right, I guess the cache is used until you make a new commit. Perhaps that was the goal of the ``version`` parameter set either to 0 or nothing in ``config/dev`` or ``development.ini.in``... (not sure it would work out of the box though).
Anyway is it really a problem? When developing, don't you simply disable your browser cache?

tsauerwein:
> No, I mean that an old version of the file is used instead of the correct one.

I thought Varnish by default takes query parameters into account?

gberaudo:
Rebased and added support for debug mode and preventing cache poisoning.

asaunier:
Looks good. Don't forget my question about the long URL here: https://github.com/c2corg/v6_ui/blob/cache_buster2/c2corg_ui/lib/cacheversion.py#L2-L3

PR #178, opened 2016-02-18
Use apiservice

gberaudo:


c2corg_ui/static/js/documentediting.js
```diff
@@ -314,56 +266,25 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
...
-        this.successSave_.bind(this),
-        this.errorSave_.bind(this)
-    );
+    this.api_.createDocument(this.module_, data).then(function(response) {
+      this.id_ = response['data']['document_id'];
+      window.location.href = app.utils.buildDocumentUrl(
+        this.module_, this.id_, this.lang_);
+    }.bind(this));
   }
   // TODO: show a loading message until doc saving is done
```
asaunier: This line (TODO) can be removed since a loading indicator has been added last week.

asaunier:
If the authentication fails (when typing a random user or password), the JS console reports an error because:
> TypeError: Cannot read property 'data' of undefined
    at app.AuthController.successLogin_ (auth.js:133)

When trying to register, if I provide a wrong email address (for instance ``sdfsdfs@sdfsdf`` => validates in the form but rejected by the API) I get at the same time:
* a success message ("you will receive a confirmation email")
* an error message ("invalid email address")
=> both the success and error callbacks have been called.

Ditto if I open a WP editing page, then switch my API off, when I press the "save" button I am redirected to the document view page (as if the editing succeeded) instead of being warned that the editing failed.

It seems that the success/error callbacks are not called as expected! Don't you think that the ``then()`` function expects 2 callbacks (1 success and 1 error)? In your changes, ``then()`` sometimes have only one callback instead of 2 previously.

gberaudo:
Thanks @asaunier, I fixed this. That is fine to pass only one function to `then`.
The handling of the catch promise was wrong.
I think I addressed all issues ; I'll merge with the Travis green light.

asaunier:
Wait, there is still a problem!

If the API is down when you open the editing page, you get an error "Failed saving the changes", which is obviously not the correct error message.

gberaudo:
Thanks, fixed it.

gberaudo:
Thanks for the review.

PR #179, opened 2016-02-18
Unassociate document

gberaudo:
Handle both Mako and Angular cases:
- Mako: remove card DOM element;
- Angular: splice added documents array.


asaunier:
This PR includes commits from PR #178. I guess you'll rebase when the latter PR is merged?

gberaudo:
Indeed, rebased before merging. Thanks.

Issue #185, opened 2016-02-19
Missing i18n


asaunier:
When running the make command, changes to the PO/POT files are reported because the extracted strings are not yet available in the PO/POT files.
Probably something to do.

Issue #176, opened 2016-02-18
Position of the "archived version" warning


asaunier:
As for now, the block saying that "you are viewing an archived version of the document" is displayed in the "more" tool, which is way to hidden.
It's important this warning is displayed more obviously, at the beginning of the document.

PR #187, opened 2016-02-23
History versions

ginold:
- "you are watching an old version of the document" is on the top of the document
- minor layout changes in differences and history pages
- there were problems with search associations bar -> not anymore
- New icon

Fixes https://github.com/c2corg/v6_ui/issues/176
Fixes https://github.com/c2corg/v6_ui/issues/185


asaunier:
Looks good to me except that the radio buttons in the history page are not rendered as expected (?):
screencapture-c2corgv6-demo-gis-internal-6555-routes-history-512952-fr-1456234900091
Having 2 cols of radios is a bit more intuitive, isn't it? See wikipedia for instance: https://fr.wikipedia.org/w/index.php?title=Abbaye_de_Beaugerais&action=history

I would also keep the "compare" button at the top of the list as well because if we have many versions, the bottom button might not be visible at first sight.

asaunier:
Looks good to me. I have made a few fixes in the locale files (needed for travis to be happy). Please merge when Travis is OK.

PR #194, opened 2016-02-23
[WIP] Associated document cards and all route activity-specific fields

ginold:
some changes in the detailed view of the route.
- if available, all possible fields for any given route should appear and not give any error
- associated routes and waypoints are displayed as cards, similar to those found in the list view
- most parts of the page are collapsible


c2corg_ui/templates/helpers/view.html
```diff
@@ -81,45 +81,91 @@
...
 <%def name="show_areas(document)">\
-  <ul>
-  % for a in document['areas']:
-    <li>${a['area_type']} ${a['locales'][0]['title']}</li>
-  % endfor
-  </ul>
+  % if len(document['areas']) > 0:
+    <ul>
+    % for a in document['areas']:
+    <li><span translate>${a['area_type']}</span>: <b>${a['locales'][0]['title']}</b></li>
```
asaunier: You cannot use the ``translate`` directive here, else ``${a['area_type']}`` is added to the PO files.

c2corg_ui/templates/route/detailedattributes.html
```diff
@@ -0,0 +1,250 @@
+<%namespace file="../helpers/view.html" import="show_areas, show_attr"
+ />
```
asaunier: Should be on the same line then the previous one

c2corg_ui/templates/route/detailedattributes.html
```diff
@@ -0,0 +1,250 @@
...
+      </div>
+      <span class="detail-text accordion">
+        ${show_areas(route)}
+        
+        % if route.get('orientations'):
+          <p><span translate>orientations</span>: <b>${route['orientations']}</b></p>
+        % endif
+        
+        % if not route.get('orientations') and len(show_areas(route)) == 0:
+        <p><span translate>No location info</span></p>
```
asaunier: Some indentation here would be cool and more consistent with the lines above.

c2corg_ui/templates/helpers/view.html
```diff
@@ -1,7 +1,7 @@
 <%!
 from c2corg_ui.templates.utils import get_attr
-%>
```
asaunier: Unneeded blank line here

c2corg_ui/templates/helpers/view.html
```diff
@@ -81,45 +81,91 @@
...
+    <div class="float-button float-comments" tooltip-placement="left" uib-tooltip="{{'Comments' | translate}}">
+      <span class="glyphicon glyphicon-comment"></span>
+      <p class="float-button-text" translate>Comments</p>
+    </div>
+    <div class="btn-group float-more float-button " uib-dropdown tooltip-placement="left" uib-tooltip="{{'More' | translate}}">
+      <span class="glyphicon glyphicon-th-list dropdown-toggle" uib-dropdown-toggle></span>
+      <p class="float-button-text" translate>More</p>
+      <ul class="dropdown-menu"  role="menu" aria-labelledby="dLabel">
+        % if version:
+        <li>${show_archive_data('routes', route, locale, version)}</li>
```
asaunier: We don't need to display the "you are viewing an archived version" warning here since it is already shown at the top of the page.

c2corg_ui/templates/route/detailedattributes.html
```diff
@@ -0,0 +1,250 @@
+<%namespace file="../helpers/view.html" import="show_areas, show_attr"
+ />
+ 
+## LOCATION
+<%def name="get_route_location(route)">\
+
```
asaunier: Perhaps you could remove the unneeded blank lines at the beginning and end of the function?

c2corg_ui/templates/route/view.html
```diff
@@ -4,8 +4,13 @@
 <%inherit file="../base.html"/>
 <%namespace file="../helpers/view.html" import="show_attr, show_short_attr, show_missing_langs_links,
     show_other_langs_links, show_archive_data, show_route_title, show_associated_waypoints,
-    show_associated_routes, show_areas, show_maps"/>
+    show_associated_routes, show_areas, show_float_buttons, show_maps"
+ />
+<%namespace file="detailedattributes.html" import="get_route_gear, get_route_location, get_route_rating, get_route_general, 
+  get_route_heights,  get_route_access"
```
asaunier: extra space :P

c2corg_ui/templates/route/view.html
```diff
@@ -4,8 +4,13 @@
...
 <%namespace file="../helpers/view.html" import="show_attr, show_short_attr, show_missing_langs_links,
     show_other_langs_links, show_archive_data, show_route_title, show_associated_waypoints,
-    show_associated_routes, show_areas, show_maps"/>
+    show_associated_routes, show_areas, show_float_buttons, show_maps"
+ />
+<%namespace file="detailedattributes.html" import="get_route_gear, get_route_location, get_route_rating, get_route_general, 
+  get_route_heights,  get_route_access" 
+ />
 <%
+
```
asaunier: extra blank line

c2corg_ui/templates/route/view.html
```diff
@@ -4,8 +4,13 @@
 <%inherit file="../base.html"/>
 <%namespace file="../helpers/view.html" import="show_attr, show_short_attr, show_missing_langs_links,
     show_other_langs_links, show_archive_data, show_route_title, show_associated_waypoints,
-    show_associated_routes, show_areas, show_maps"/>
+    show_associated_routes, show_areas, show_float_buttons, show_maps"
+ />
+<%namespace file="detailedattributes.html" import="get_route_gear, get_route_location, get_route_rating, get_route_general,
```
asaunier: Since the file only contains functions (helpers), what about making the filename more explicit, including "helpers" somewhere or even by creating an ``helpers/`` directory?

c2corg_ui/templates/helpers/view.html
ginold: Luckily it did't add anything. I just had to add the translations by hand.

asaunier: Are you sure?
I can see it here: https://github.com/c2corg/v6_ui/blob/associated-document-cards/c2corg_ui/locale/c2corg_ui-client.pot#L8 :P

c2corg_ui/templates/route/detailedattributes.html
ginold: I found it easier to read.

c2corg_ui/templates/helpers/view.html
tsauerwein: How do we want to handle this? We want translations for all the enum types, so we would have to mark it with `translate`.

Just an idea, we could create a special directive `translate-enum` which is an alias for the existing `translate` directive (untested):

    app.module.directive('translate-enum', angular.module('gettext').directive('translate'));

This will prevent that these strings land in the PO files and we should have the same behaviour as with `translate` (I hope).

asaunier: Come on! Is this that easy to create an alias for a directive? That the solution I've been looking for (not that actively :P) for weeks!

That would be much more convenient then the current solution (``{{mainCtrl.translate('${a['area_type']}')}}``) which is verbose and adds ``{{}}`` until the translation is done (which is ugly).

I don't think ``translate-enum`` is a relevant name since we would need it for other "variable" content. What about ``translate-var``?

tsauerwein: > Come on! Is this that easy to create an alias for a directive? That the solution I've been looking for (not that actively :P) for weeks!

I haven't tested it! I hope I didn't make you any false hopes... ;)

> I don't think translate-enum is a relevant name since we would need it for other "variable" content. What about translate-var?

Actually it's for constants? `translate-const`?

asaunier:
Thanks @ginold for the changes. As far as I can see it looks OK for merging.

PR #196, opened 2016-02-25
Waypoints + routes specific fields and associated-documents cards

ginold:
Same layout as already done for routes, but this time for waypoints with a few upgrades like a super-duper sticky-note with location info and an animated shiny bling bling menu button, all in CSS ! 
screenshot from 2016-02-25 15 46 17

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -126,22 +126,22 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   // When creating a new document, the model is not created until
   // the form is touched. At least create an empty object.
   this.scope_[this.modelName_] = {};
-
-  if (this.id_ && this.lang_) {
-    if (this.auth_.isAuthenticated()) {
-      // Get document attributes from the API to feed the model:
+  if (this.auth_.isAuthenticated()) {
+    if (this.id_ && this.lang_) {
+     // Get document attributes from the API to feed the model:
       goog.asserts.assert(!goog.isNull(this.id_));
       goog.asserts.assert(!goog.isNull(this.lang_));
       this.api_.readDocument(this.module_, this.id_, this.lang_).then(
           this.successRead_.bind(this)
       );
-    } else {
-      // Redirect to the auth page
-      var current_url = window.location.href;
-      window.location.href = '{authUrl}?from={redirect}'
-          .replace('{authUrl}', this.authUrl_)
-          .replace('{redirect}', encodeURIComponent(current_url));
     }
+  } else {
+    // Redirect to the auth page
+    var current_url = window.location.href;
+    window.location.href = '{authUrl}?from={redirect}'
+        .replace('{authUrl}', this.authUrl_)
+        .replace('{redirect}', encodeURIComponent(current_url));
+    return;
   }
 
   this.scope_.$root.$on('mapFeatureChange', function(event, feature) {
--- a/c2corg_ui/static/js/search.js
+++ b/c2corg_ui/static/js/search.js
@@ -6,7 +6,6 @@ goog.require('app.utils');
 /** @suppress {extraRequire} */
 goog.require('ngeo.searchDirective');
 
-
 /**
  * The directive for the auto-complete search field shown in the header
  * of every page.
@@ -40,32 +39,20 @@ app.searchDirective = function() {
               $('.page-header').removeClass('show-search');
             }
           });
-          // Add class only on hover && when screen width < @phone (defined in LESS)
-          $('.page-header').on('mouseenter', function() {
-            if (window.innerWidth < phoneScreen) {
-              $('.page-header').find('input').addClass('show-search');
-            }
-          });
+
           // Trigger focus on search-icon click for .search
-          $('.page-header').on('click', function(event) {
+          $('.page-header .search-icon').on('click', function(event) {
             if (window.innerWidth < phoneScreen) {
+              $('.page-header').find('.search').toggleClass('show-search');
+              $('.page-header').find('.search').focus();
               event.stopPropagation();
-              $('.page-header').find('input').addClass('show-search');
-              $('.page-header').find('.search').triggerHandler('focus');
             }
           });
-          // If the input is focused, don't remove the class
-          $('.page-header').on('mouseleave', function() {
-            if (window.innerWidth < phoneScreen && !$('.search').is(':focus')) {
-              $('.page-header').find('.show-search').removeClass('show-search');
-            }
+
+          $('body:not(app-search)').click(function() {
+            $('.tt-dataset.empty').remove();
           });
-          // If you click outside the search input, it has to be closed on @phone
-          if (window.innerWidth < phoneScreen) {
-            $('html').not('.search').not('.glyphicon-search').click(function() {
-              $('.page-header').find('.show-search').removeClass('show-search');
-            });
-          }
+
           //show spinning gif while waiting for the results
           element.on('typeahead:asyncrequest', function() {
             element.find('input').addClass('loading-gif-typehead');
@@ -82,14 +69,14 @@ app.module.directive('appSearch', app.searchDirective);
 
 /**
  * @constructor
- * @param {angular.Scope} $rootScope Angular root scope.
+ * @param {angular.Scope} $scope Angular scope.
  * @param {angular.$compile} $compile Angular compile service.
  * @param {angular.Attributes} $attrs Angular attributes.
  * @param {string} apiUrl Base URL of the API.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @ngInject
  */
-app.SearchController = function($rootScope, $compile, $attrs, apiUrl, gettextCatalog) {
+app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog) {
 
   /**
    * Bound from directive.
@@ -97,6 +84,12 @@ app.SearchController = function($rootScope, $compile, $attrs, apiUrl, gettextCat
    * @export
    */
   this.selectHandler;
+
+  /**
+   * @type {angular.$compile}
+   * @private
+   */
+  this.compile_ = $compile;
 
   if (!$attrs['appSelect']) {
     // Angular puts a noop function when mapping an attribute to a
@@ -110,6 +103,13 @@ app.SearchController = function($rootScope, $compile, $attrs, apiUrl, gettextCat
    * @private
    */
   this.apiUrl_ = apiUrl;
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
+
 
   /**
    * @type {angularGettext.Catalog}
@@ -162,7 +162,7 @@ app.SearchController.prototype.createDataset_ = function(type, title) {
     limit: 20,
     templates: {
       header: (function() {
-        return '<div class="header">' +
+        return '<div class="header ' + title + '">' +
             this.gettextCatalog_.getString(title) + '</div>';
       }).bind(this),
       suggestion: function(doc) {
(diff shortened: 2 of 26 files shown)


tsauerwein:
> ... like a super-duper sticky-note with location info and an animated shiny bling bling menu button, all in CSS ! 

Do you have some screenshots of these goodies? ;)

ginold:
Menu button as in
http://callmenick.com/_development/css-hamburger-menu-icons/
Screenshot added :)

On Thu, Feb 25, 2016 at 3:43 PM, Tobias Sauerwein <notifications@github.com>
wrote:

> ... like a super-duper sticky-note with location info and an animated
> shiny bling bling menu button, all in CSS !
>
> Do you have some screenshots of these goodies? ;)
>
> —
> <https://github.com/c2corg/v6_ui/pull/196#issuecomment-188815455>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

tsauerwein:
Nice, that looks good!

The sticky note is not really a map marker? When you pan the map it does not move?

ginold:
sticky note is sticky, so it will stay in the same place even when you pan around

asaunier:
Really cool @ginold !

I think the "maps" section at the end takes too much place. It's not that important. You could add the maps list into the "information" block. By the way there is a "map infos" attribute that is used to specified maps not automatically detected, for instance in countries where we have no map footprints. Both block should be combined. See maps list for instance here: http://www.camptocamp.org/summits/37527/fr/grand-paradis

Concerning the associated documents, I'm not sure it is relevant to have blocks saying for instance "no associated route". I think it is wasted place. What about simply showing nothing in that case?

Cool to have embedded the "remarks" and "summary" fields in the description but I notice there formatting is no longer interpreted (bbcode and markdown).

I find the sticky note on the map really cool. We'll see what the client thinks of it.

asaunier:
The scroll bar in the associated docs cards sections is a good idea, it avoids the page grows too big when there are many associated docs. Perhaps you could make the minimal height a little bigger though.

Please note that we need to consider the case of routes with many associated outings. See for instance
http://www.camptocamp.org/routes/154701/fr/cogne-valeille-e-tutto-relativo#outings

Perhaps the scroll bar solution is enough. On the other hand it makes the page heavier if we have dozens/hundreds of outings (should we consider loading only the first 10? outings in mako and get the extra outings by AJAX?).

tsauerwein:
> Concerning the associated documents, I'm not sure it is relevant to have blocks saying for instance "no associated route". I think it is wasted place. What about simply showing nothing in that case?

What about something like "no associated route. add a new one" to encourage the user.

About the sticky note: The pin gives the impression that it marks a location on the map, which is not the case. Also I am worried that it takes too much place in the map area (without providing much information), especially once we have the map controls. But it looks good... :)

tsauerwein:
> Perhaps the scroll bar solution is enough. On the other hand it makes the page heavier if we have dozens/hundreds of outings (should we consider loading only the first 10? outings in mako and get the extra outings by AJAX?).

I thought we only show the last 10 (?) and then give a link to a page like this:
http://www.camptocamp.org/outings/list/routes/154701/orderby/date/order/desc

asaunier:
> What about something like "no associated route. add a new one" to encourage the user.

That makes sense for the "associated routes" block in WP pages.
Or for outings in route pages.

A little less for the "associated routes" of route pages (that's not that frequent to have routes associated together).

> About the sticky note: The pin gives the impression that it marks a location on the map, which is not the case. Also I am worried that it takes too much place in the map area (without providing much information), especially once we have the map controls. But it looks good... :)

Good point. Perhaps the pin could be omitted? And if the map is small (mobile screen?), the note is transfered out of the map?

> I thought we only show the last 10 (?) and then give a link to a page like this:

Right. That's what the wireframes say I think (I didn't remember this change in v6).

c2corg_ui/templates/route/helpers/detailed_route_attributes.html
```diff
@@ -51,10 +51,6 @@
       <p><span translate>configuration</span>: <b>${','.join(route['configuration'])}</b></p>
```
asaunier: ``${','.join(route['configuration'])}`` will probably not be translated. Shouldn't you use something like https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/view.html#L15 ?

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -0,0 +1,365 @@
...
+    <span class="glyphicon glyphicon-asterisk"></span>
+    <p translate>General</p>
+  </div>
+  <span class="detail-text accordion">
+    % if waypoint.get('waypoint_type'):
+    <p><span translate>waypoint_type</span>: <b x-translate>${waypoint['waypoint_type']}</b></p>
+    % endif
+
+    % if waypoint.get('orientation'):
+      <p><span translate>orientation</span>: <b>${waypoint['orientation']}</b></p>
```
asaunier: I think we need an ``x-translate`` for ``${waypoint['orientation']}`` since the shorts for cardinal points are different from a language to another, especially E and W.

By the way (but it's a nice to have), it would be cool to show orientations graphically, using a compass.

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -0,0 +1,365 @@
...
+    % endif
+
+    % if waypoint.get('climbing_indoor_types'):
+    <article>
+      <b translate>climbing_indoor_types</b><br>
+      <ul>
+        % for style in waypoint['climbing_indoor_types'] :
+        <li>${style}</li>
```
asaunier: x-translate

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -0,0 +1,365 @@
...
+      </ul>
+    </article>
+    % endif
+
+    % if waypoint.get('rock_types'):
+    <article>
+      <b translate>rock_types</b><br>
+      <ul>
+        % for type in waypoint['rock_types'] :
+        <li>${type}</li>
```
asaunier: x-translate

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
asaunier: Please note that "orientation" is an array and could have several values at a time.
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L78

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -0,0 +1,365 @@
...
+      </ul>
+    </article>
+    % endif
+
+    % if waypoint.get('climbing_outdoor_types') and waypoint['climbing_outdoor_types'][0] != '':
+    <article>
+      <b translate>climbing_outdoor_types</b><br>
+      <ul>
+        % for type in waypoint['climbing_outdoor_types'] :
+        <li>${type}</li>
```
asaunier: x-translate

Ditto for other "types" below.

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -0,0 +1,365 @@
...
+      <ul>
+        % for type in waypoint['weather_station_types'] :
+        <li>${type}</li>
+        % endfor
+      </ul>
+    </article>
+    % endif
+
+    % if waypoint.get('rain_proof'):
+    <p><span translate>rain_proof</span>: <b>${waypoint['rain_proof']}</b></p>
```
asaunier: x-translate?

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -0,0 +1,365 @@
...
+        <span class="glyphicon glyphicon-sunglasses"></span>
+        <p translate>Style</p>
+      </div>
+      <span class="detail-text accordion">
+
+        % if waypoint.get('climbing_styles'):
+        <b translate>climbing_styles</b><br>
+        <ul>
+          % for style in waypoint['climbing_styles'] :
+          <li>${style}</li>
```
asaunier: x-translate

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
tsauerwein: > By the way (but it's a nice to have), it would be cool to show orientations graphically, using a compass.

:+1:

ginold: you say it - you got it (next week) :dancers:

asaunier:
There's a mako error when displaying a archived version:
> File "/home/alex/Bureau/v6_ui/c2corg_ui/templates/waypoint/view.html", line 79, in render_body
    % if 'routes' in waypoint['associations'] or 'waypoint' in waypoint['associations']  :
KeyError: 'associations'

c2corg_ui/static/js/documentediting.js
```diff
@@ -126,22 +126,21 @@ app.DocumentEditingController = function($scope, $element, $attrs,
...
-      window.location.href = '{authUrl}?from={redirect}'
-          .replace('{authUrl}', this.authUrl_)
-          .replace('{redirect}', encodeURIComponent(current_url));
     }
+  } else {
+    // Redirect to the auth page
+    var current_url = window.location.href;
+    window.location.href = '{authUrl}?from={redirect}'
+        .replace('{authUrl}', this.authUrl_)
+        .replace('{redirect}', encodeURIComponent(current_url));
```
asaunier: Good catch!
Would it be useful to add a ``return`` here, so that the rest of the function is not executed if the redirection is not instantaneous? Or is the function already exited when setting ``window.location.href``?

asaunier:
Minor issue: the "create new ..." buttons are smaller in the association search tool than in the general search tool.

c2corg_ui/static/js/search.js
```diff
@@ -98,6 +85,12 @@ app.SearchController = function($rootScope, $compile, $attrs, apiUrl, gettextCat
    */
   this.selectHandler;
 
+/**
```
asaunier: The indentation of the comment block is wrong, should be shifted 2 spaces to the right.

c2corg_ui/static/js/documentediting.js
ginold: "The Javascript session will struggle to continue it's mission in the face
of impending doom. Coding in this manner is considered unpredictable.
...and cruel."
So yeah, let's add a return statement.

On Fri, Feb 26, 2016 at 3:35 PM, Alex Saunier <notifications@github.com>
wrote:

> In c2corg_ui/static/js/documentediting.js
> <https://github.com/c2corg/v6_ui/pull/196#discussion_r54249971>:
>
> >      }
> > +  } else {
> > +    // Redirect to the auth page
> > +    var current_url = window.location.href;
> > +    window.location.href = '{authUrl}?from={redirect}'
> > +        .replace('{authUrl}', this.authUrl_)
> > +        .replace('{redirect}', encodeURIComponent(current_url));
>
> Good catch!
> Would it be useful to add a return here, so that the rest of the function
> is not executed if the redirection is not instantaneous? Or is the function
> already exited when setting window.location.href?
>
> —
> <https://github.com/c2corg/v6_ui/pull/196/files#r54249971>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

asaunier:
When the remaining issues are addressed, I think it is OK to merge the PR.
@tsauerwein OK for you? If yes, we would like to merge and update the demo instance before the WE.

asaunier:
I have noticed that the "change comment" is not translated in the archived version view. Could you add a ``<span x-translate>`` around ``${show_version_comment(version)}`` at https://github.com/c2corg/v6_ui/blob/waypoints-associated-cards/c2corg_ui/templates/helpers/view.html#L29
?

asaunier:
Please merge.

Issue #171, opened 2016-02-17
Quotes in version comment make the history page crash


asaunier:
In http://c2corgv6-demo.gis.internal:6555/waypoints/history/37355/fr (Mont Blanc), there's a comment in french that say
> Ajustement d'un titre d'itinéraire

=> that creates a syntax error and most of the page does not behave as expected

tsauerwein:
Wait, why is it `{{mainCtrl.translate('Ajout d'un itinéraire : Traversée Arête des bosses - Aiguille du Midi')}} `? This part should not be translated.

asaunier:
> This part should not be translated.

Indeed. On the other hand we have default comments added by the system such as "edit in fr", "imported from v4", etc. that would make sense to translate!?

ginold:
is it better to create new strings like 'edit in fr' => 'traduire en francais' or use existing ones, like for example translate(edit in) + translate(fr) ?

asaunier:
The less translations we have, the lighter the i18n files to load, the better.
But we have to find a compromise with complexity and flexibility.

> translate(edit in) + translate(fr)

This won't work in language like german since you can have words around the language name ("ins Deutsche übersetzen"?). But it seems you can use expressions in the translation strings, for instance ``translate in {{lang}}``.
Please note that in the history page we anyway have comments coming from the DB such as ``edit in fr``, that we cannot split as you suggest.

tsauerwein:
The system message are not only in English? E.g. if a user with the interface language set to French makes a modification, the system message will also be in French?

If that is the case, I would not try to translate. It is as if the user has entered the message them self, and we don't translate custom messages.

asaunier:
> The system message are not only in English?

The default history/version comments are only in english.

Issue #195, opened 2016-02-24
Alias for the "translate" directive that would not be extracted


asaunier:
When translating data that are for instance provided by the API (using a mako variable or some angular expression), one cannot use the usual ``translate`` directive because then the variable is added to PO files by the translations extracting tool.

A workaround is to use ``{{mainCtrl.translate('${some_variable}')}}`` since it is not recognized by the extractor. The downside is that we have ``{{ }}`` that appear in the page until the translation is done, wich is ugly.

A solution would be to have an alias for the usual ``translate`` directive that could be use the same way but won't be "extracted" to the PO files.

@tsauerwein has suggested to write something like
```
app.module.directive('translate-enum', angular.module('gettext').directive('translate'));
```

but I get an Closure compiler error:
> ERROR - actual parameter 2 of angular.Module.prototype.directive does not match formal parameter
ERR! compile found   : angular.Module
ERR! compile required: (Array<(Function|null|string)>|Function|null|undefined)
ERR! compile app.module.directive('translate-enum', angular.module('gettext').directive('translate'));

tsauerwein:
As Angular allows to prefix attributes with `x-` or `data-` (see [Directives - normalization](https://docs.angularjs.org/guide/directive#normalization)), we can use `x-translate` which is not extracted:

    <span x-translate>${a["area_type"]}</span>

asaunier:
Really cool! Thanks!

Then we'll have to replace most of the ``{{mainCtrl.translate()}}`` by this ``x-translate``.

asaunier:
I have tested the ``x-translate`` directive and it works as expected:
* translations are done, as with ``translate``
* embedded strings are not extracted to the PO files.

@ginold since you make a lot of changes in the template, you're probably the best one to integrate this ``x-translate`` directive instead of the ``{{mainCtrl.translate()}}`` workarouds, this would avoid conflicts among PR.

PR #197, opened 2016-02-25
Simplify some translate attr + use x-translate for variables

asaunier:
Related to #195 

I have not made changes to */view.html nor view helpers to avoid conflicts with current changes of @ginold in #196


asaunier:
A "side effect" of the use of "x-translate" is that we no longer have the "quote escaping" problem in history pages, which means the PR fixes https://github.com/c2corg/v6_ui/issues/171

OK for merging?

tsauerwein:
:+1:

Issue #27, opened 2015-11-10
Error handling on the server-side


tsauerwein:
See https://github.com/c2corg/v6_ui/issues/25#issuecomment-155056958

stef74:
@tsauerwein this issue is manage #25 ?

Issue #25, opened 2015-11-06
Fallback when API is down


asaunier:
As for now when the API is down (for instance pserve off), the UI WP URL shows a pyramid error page with message

>   File "/home/alex/v6_ui/c2corg_ui/views/document.py", line 72, in _get_documents
>     return content if resp.status == 200 else []
> AttributeError: 'dict' object has no attribute 'status'

tsauerwein:
It's because you are returning a dictionary here:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L35

Maybe you want to return the original response instead?

asaunier:
Are you sure that's the problem?
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L35 actually returns a "resp" output with a "status" attribute, doesn't it?

By the way I have noticed that sometimes the API fails but returns an HTML page as the "content" (instead of a JSON output - or maybe as the value of a JSON attribute?). I don't remember in what cases.

tsauerwein:
> Are you sure that's the problem?
> https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L35 actually returns a "resp" output with a "status" attribute, doesn't it?

It returns a dictionary, so you would have to use `resp['status']`.

> By the way I have noticed that sometimes the API fails but returns an HTML page as the "content" (instead of a JSON output - or maybe as the value of a JSON attribute?). I don't remember in what cases.

Could you please log the request when this happens again?

asaunier:
> It returns a dictionary, so you would have to use resp['status'].

Do you mean we should use
```
return content if resp['status'] == 200 else []
```
instead of
```
return content if resp.status == 200 else []
```
at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L72 ?

> Could you please log the request when this happens again?

OK

tsauerwein:
Yes. Or return the original response (which you get from the `http.request()` call).

asaunier:
Perhaps we don't need that try/except? I don't remember why I have used it? Maybe to handle contents (errors) that do not fit in ``json.loads`` but is it really a problem?

Maybe
```
    def _call_api(self, url, method='GET', body=None, headers=None):
        http = httplib2.Http()
        resp, content = http.request(
            url, method=method, body=body, headers=headers
        )
        return resp, json.loads(content)
```
would be enough?

tsauerwein:
[...]

PR #200, opened 2016-02-26
ApiError directive

asaunier:
The directive is used to display an alert (using ``AlertService``) when the API fails (for instance down) while loading the WP or routes list pages (as for now we only show pages with 0 item).

It does not support errors when loading a document view page (yet?), because the UI already raises a 500 or 404 error:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L85-L89
In that case it's probably better to customize the 500/404 pages.

Related to https://github.com/c2corg/v6_ui/issues/25 and https://github.com/c2corg/v6_ui/issues/27 and https://github.com/c2corg/v6_ui/issues/201

### DIFF
--- a/c2corg_ui/static/js/alertservice.js
+++ b/c2corg_ui/static/js/alertservice.js
@@ -44,14 +44,15 @@ app.Alerts.prototype.gettext = function(str) {
  * @export
  */
 app.Alerts.prototype.add = function(data) {
-  this.addLoading_();
+  var timeout = data['timeout'] || 0;
+  this.addLoading_(timeout);
   var msg = data['msg'];
   msg = msg instanceof Object ? this.formatErrorMsg_(msg) :
       this.filterStr_(msg);
   this.alerts_.push({
     type: data['type'] || 'warning',
     msg: msg,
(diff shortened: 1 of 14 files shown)


asaunier:
Review welcome :)

c2corg_ui/static/js/alertservice.js
```diff
@@ -44,14 +44,15 @@ app.Alerts.prototype.gettext = function(str) {
  * @export
  */
 app.Alerts.prototype.add = function(data) {
-  this.addLoading_();
+  var timeout = data['timeout'] || 0;
```
gberaudo: Is `timeout` in the extern?

gberaudo:
Shouldn't we instead consistently throw a 500 error when the API is down?

c2corg_ui/static/js/alertservice.js
asaunier: No clue :)

asaunier:
> Shouldn't we instead consistently throw a 500 error when the API is down?

That would make sense indeed.

PR #218, opened 2016-03-02
Automatically logs out based on an URL parameter

gberaudo:
Used together with the discourse `logout_redirect = http://c2corgv6.demo-camptocamp.com/auth?logout` setting.


asaunier:
The commit message lacks an "out", else looks good to me, thanks.

gberaudo:
See https://github.com/c2corg/v6_forum/issues/5.

gberaudo:
Thanks for the review.

PR #220, opened 2016-03-03
Edit/create waypoint attributes

ginold:
screenshot from 2016-03-03 17 06 44
screenshot from 2016-03-03 17 09 29
screenshot from 2016-03-03 17 09 44

that was HELLOTA work !! and hopefully it works as it should. Gimmie a break

- step by step waypoint creation/edition
- animated progress bar that adapts to the number of steps (4-5)
- the progress bar titles change depending on the waypoint type and its contents
- nice SVG orientation picker
- form inputs change depending on the waypoint type
- the *required* values change depending on the waypoint type
- lists checkboxes should have their values prechecked if the waypoint contains them

- maybe to do -> validation : user should not be allowed to step further if he hasn't filled all the required inputs.

### DIFF
--- a/c2corg_ui/locale/c2corg_ui-client.pot
+++ b/c2corg_ui/locale/c2corg_ui-client.pot
@@ -4,6 +4,10 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "&"
+msgstr ""
+
 #: c2corg_ui/templates/route/helpers/detailed_route_attributes.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "Access"
@@ -101,6 +105,7 @@ msgstr ""
 msgid "Comparing documents"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "Contact"
 msgstr ""
@@ -127,6 +132,14 @@ msgstr ""
 
 #: c2corg_ui/templates/helpers/view.html
 msgid "Date"
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "Describe access"
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "Describe access_period"
 msgstr ""
 
 #: c2corg_ui/templates/route/edit.html
@@ -189,6 +202,7 @@ msgstr ""
 msgid "Email"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "Equipment"
 msgstr ""
@@ -287,8 +301,8 @@ msgstr ""
 msgid "Logout"
 msgstr ""
 
-#: c2corg_ui/templates/waypoint/edit.html
-msgid "Longitude"
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+msgid "Longitude and latitude are required."
 msgstr ""
 
 #: c2corg_ui/templates/route/helpers/detailed_route_attributes.html
@@ -317,6 +331,7 @@ msgid "Next"
 msgstr ""
 
 #: c2corg_ui/templates/route/filters.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/filters.html
 msgid "No"
 msgstr ""
@@ -335,6 +350,10 @@ msgstr ""
 
 #: c2corg_ui/templates/helpers/view.html
 msgid "No maps associated"
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "Numbers"
 msgstr ""
 
 #: c2corg_ui/templates/auth.html
@@ -407,8 +426,13 @@ msgstr ""
 
 #: c2corg_ui/templates/auth.html
 #: c2corg_ui/templates/route/edit.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
 #: c2corg_ui/templates/waypoint/edit.html
 msgid "This field is required."
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "This field must be  than 9999 m."
 msgstr ""
 
 #: c2corg_ui/templates/route/edit.html
@@ -422,10 +446,6 @@ msgstr ""
 
 #: c2corg_ui/templates/waypoint/edit.html
 msgid "This field must be between -90° and 90°."
-msgstr ""
-
-#: c2corg_ui/templates/waypoint/edit.html
-msgid "This field must be smaller than 9999 m."
 msgstr ""
 
 #: c2corg_ui/templates/route/edit.html
@@ -444,6 +464,10 @@ msgstr ""
 
 #: c2corg_ui/templates/helpers/view.html
 msgid "Translate into an other lang"
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "Types & styles"
 msgstr ""
 
 #: c2corg_ui/templates/auth.html
@@ -473,6 +497,7 @@ msgid "Welcome to Camptocamp.org"
 msgstr ""
 
 #: c2corg_ui/templates/route/filters.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/filters.html
 msgid "Yes"
 msgstr ""
@@ -482,14 +507,20 @@ msgid "You are viewing an archived version of this document."
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/view.html
 msgid "access"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "access_period"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "access_time"
 msgstr ""
@@ -530,6 +561,8 @@ msgid "bergschrund"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "best_periods"
 msgstr ""
@@ -542,8 +575,18 @@ msgstr ""
 msgid "bivouac"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "blanket_unstaffed"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "boat"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "bus"
 msgstr ""
 
 #: c2corg_ui/templates/document/diff.html
@@ -556,13 +599,21 @@ msgid "ca"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
+msgid "cable_car"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
 msgid "camp_site"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "capacity"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "capacity_staffed"
 msgstr ""
@@ -571,6 +622,8 @@ msgstr ""
 msgid "cave"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "children_proof"
 msgstr ""
@@ -583,6 +636,8 @@ msgstr ""
 msgid "climbing_indoor"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "climbing_indoor_types"
 msgstr ""
@@ -592,21 +647,34 @@ msgid "climbing_outdoor"
 msgstr ""
 
 #: c2corg_ui/templates/route/helpers/detailed_route_attributes.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "climbing_outdoor_types"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "climbing_rating_max"
+msgstr ""
+
 #: c2corg_ui/templates/i18n.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "climbing_rating_median"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "climbing_rating_min"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "climbing_styles"
 msgstr ""
@@ -619,10 +687,17 @@ msgstr ""
 msgid "confluence"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "contact"
+msgstr ""
+
 #: c2corg_ui/templates/i18n.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
 msgid "country"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "custodianship"
 msgstr ""
@@ -636,11 +711,13 @@ msgid "dec"
 msgstr ""
 
 #: c2corg_ui/templates/route/edit.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
 #: c2corg_ui/templates/waypoint/edit.html
 msgid "description"
 msgstr ""
 
 #: c2corg_ui/templates/route/filters.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
 #: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/filters.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
@@ -652,6 +729,8 @@ msgstr ""
 msgid "elevation min/max"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "elevation_min"
 msgstr ""
@@ -664,10 +743,16 @@ msgstr ""
 msgid "engagement_rating"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "equipment"
+msgstr ""
+
 #: c2corg_ui/templates/route/helpers/detailed_route_attributes.html
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
 msgid "equipment_rating"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "equipment_ratings"
 msgstr ""
@@ -688,6 +773,8 @@ msgstr ""
 msgid "exposed"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "exposition_rating"
 msgstr ""
@@ -704,6 +791,8 @@ msgstr ""
 msgid "fr"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "gas_unstaffed"
 msgstr ""
@@ -712,6 +801,10 @@ msgstr ""
 #: c2corg_ui/templates/route/helpers/detailed_route_attributes.html
 #: c2corg_ui/templates/route/view.html
 msgid "gear"
+msgstr ""
+
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "geolocation"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
@@ -742,6 +835,8 @@ msgstr ""
 msgid "ground_types"
 msgstr ""
 
+#: c2corg_ui/templates/waypoint/create_edit_summary.html
+#: c2corg_ui/templates/waypoint/edit.html
 #: c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
 msgid "heating_unstaffed"
 msgstr ""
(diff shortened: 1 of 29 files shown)


tsauerwein:
The screenshots look promising! :+1:

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+          <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.orientation.$valid"></span>
+        </div>
+
+        ## WEATHER STATION TYPES
+        <div class="form-group data half" ng-if="waypoint.waypoint_type == 'weather_station'"
+             ng-class="{ 'has-error': editForm.weather_station_types.$touched && editForm.weather_station_types.$invalid, 'has-success': editForm.weather_station_types.$valid }">
+          <label><span translate>weather_station_types</span></label>
+          <ul class="types">
+            <li ng-repeat="type in ${weather_station_types}"  ng-click="editCtrl.pushToArray(waypoint, 'weather_station_types', type, $event)">
+              <div class="checkbox"><input type="checkbox" ng-checked="waypoint.weather_station_types.indexOf(type) > - 1"><span translate>{{type}}</span></div>
```
asaunier: Please use ``x-translate`` here instead of ``translate``.
Ditto for translating the other ``{{type}}`` occurences.

c2corg_ui/locale/c2corg_ui-client.pot
```diff
@@ -4,6 +4,10 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
+#: c2corg_ui/templates/waypoint/edit.html
+msgid "&"
```
asaunier: Do you really need to translate this?

c2corg_ui/static/js/maincontroller.js
```diff
@@ -69,3 +69,17 @@ app.MainController.prototype.isPath = function(path) {
   }
 };
 
+/**
+ *
+ * @export
+ */
+app.MainController.prototype.animateHeaderIcon = function(e) {
```
asaunier: This function should rather be moved to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/utils.js

asaunier:
Maybe not that a big deal but it's better to avoid using "/" in a branch name...

c2corg_ui/static/js/documentediting.js
```diff
@@ -117,12 +124,33 @@ app.DocumentEditingController = function($scope, $element, $attrs,
    */
   this.alerts_ = appAlerts;
 
+
```
asaunier: Duplicated blank line (ditto at lines 145 and 153)

c2corg_ui/static/js/documentediting.js
```diff
@@ -33,7 +33,14 @@ app.documentEditingDirective = function() {
     scope: true,
     controller: 'appDocumentEditingController',
     controllerAs: 'editCtrl',
-    bindToController: true
+    bindToController: true,
+    link: function(scope, el, attrs, ctrl) {
+      // The directive executes after the dom is loaded but we have to wait until the expressions are evaluated
```
asaunier: Is it bullet-proof?

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+            ## MAPS REF
+            <div  id="maps_info-group"  class="input-group">
+              <label translate>maps_references</label>
+              <div ng-repeat="map in waypoint.maps" class="map-info">
+                <input  value="{{map.code}} - {{map.editor}} -  {{map.locales[0].title}}" class="form-control">
+                <span class="glyphicon glyphicon-trash"></span>
+              </div>
+              <div>
+                <input type="text" name="maps_info" ng-model="waypoint.maps_info" class="form-control" placeholder="{{maps_info}}"/>
+                <button id="document-add-maps" type="button" class="btn btn-default" ng-click="editCtrl.addAnotherMapsInfo()"><span class="glyphicon glyphicon-plus"></span></button>
```
asaunier: Why do you need this? "Maps info" is a simple text attribute. I don't think it's worth complicating the HTML code and editing controller by adding line for every maps.

c2corg_ui/static/js/documentediting.js
```diff
@@ -354,5 +382,204 @@ app.DocumentEditingController.prototype.getCoordinatesFromPoint_ = function(
...
+
+
+/**
+ * Navigate through creation/editing steps
+ * @param {number} step
+ * @param {string} direction
+ * @export
+ */
+
+app.DocumentEditingController.prototype.step = function(step, direction) {
```
asaunier: Does it work for other document types that would not have the same number of steps?

c2corg_ui/static/js/documentediting.js
```diff
@@ -354,5 +382,204 @@ app.DocumentEditingController.prototype.getCoordinatesFromPoint_ = function(
...
+
+app.DocumentEditingController.prototype.setOrientation = function(orientation, waypoint) {
+
+  if (!waypoint['orientation']) {
+    waypoint.orientation = '';
+  }
+
+  waypoint.orientation = orientation;
+
+  $('.circle-selected').removeClass('circle-selected');
```
asaunier: Won't such code be needed in other directives? For instance for filtering lists of documents?
If yes, shouldn't you move the code to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/utils.js or some other dedicated files?

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+
+            ## SLOPE
+            <div id="slope-group" class="form-group data half" ng-if="waypoint.waypoint_type == 'paragliding_takeoff' || waypoint.waypoint_type == 'paragliding_landing'"
+                 ng-class="{ 'has-error': editForm.slope.$touched && editForm.slope.$invalid, 'has-success': editForm.slope.$valid && editForm.slope.$touched }">
+              <label><span translate>slope</span> *</label>
+              <div class="input-group">
+                <input type="number" name="slope" min="0" ng-model="waypoint.slope" class="form-control" max="9999"/>
+                <span class="input-group-addon">m</span>
+              </div>
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.slope.$valid && editForm.slope.$touched""></span>
```
asaunier: There's an extra double quote " at the end of the line.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+                                      'has-success': editForm.height_max.$valid && editForm.height_max.$touched}">
+              <label><span translate>height_max</span></label>
+              <div class="input-group">
+                <input type="number" name="height_max" min="0" ng-model="waypoint.height_max" class="form-control" max="9999"/>
+                <span class="input-group-addon">m</span>
+              </div>
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.height_max.$valid && editForm.height_max.$touched"></span>
+              <span class="glyphicon glyphicon-warning-sign form-control-feedback" ng-show="editForm.height_max.$invalid && editForm.height_max.$touched"></span>
+              <div class="help-block" ng-messages="editForm.height_max.$error">
+                <p ng-message="number"  style="{{ (editForm.height_max.$invalid && editForm.height_max.$touched) ? 'display:inline' : ''}}" translate>This field must be a number.</p>
```
asaunier: Faster then hawkeye @gberaudo : there's an extra blank space after ``ng-message="number"``.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
-      <label><span translate>Longitude</span> *</label> 
-     <div class="input-group">
-        <input type="number" name="longitude" ng-model="waypoint.lonlat.longitude" ng-change="editCtrl.updateMap()" class="form-control" required min="-180" max="180"/>
-        <span class="input-group-addon">°E</span>
+
+    </section>
+  </div>
+
+
+      <div class="step two">
```
asaunier: The indentation of the whole following block looks weird. You should probably shift it leftward a bit.

c2corg_ui/templates/waypoint/helpers/create_edit_summary.html
```diff
@@ -0,0 +1,288 @@
+
```
asaunier: This file contains no helper functions. It should be moved to the root of ``c2corg_ui/templates/waypoint/``.

c2corg_ui/templates/waypoint/helpers/create_edit_summary.html
```diff
@@ -0,0 +1,288 @@
...
+
+
+<figure>
+
+  ## TITLE & TYPES
+  <h3>ttle & types</h3>
```
asaunier: ``ttle``?
Isn't a ``translate`` missing? Or rather:
```
<h3><span translate>title</span> & <span translate>types</span></h3>
```

c2corg_ui/templates/waypoint/helpers/create_edit_summary.html
```diff
@@ -0,0 +1,288 @@
...
+    <input disabled class="form-control" value="{{waypoint.locales[0].title}}">
+  </p>
+  <p><label translate>waypoint_type</label> 
+    <input disabled class="form-control" value="{{mainCtrl.translate(waypoint.waypoint_type)}}">
+  </p>
+  <p><label translate>lang</label>
+    <input disabled class="form-control" value="{{mainCtrl.translate(waypoint.locales[0].lang)}}">
+  </p>
+  <p ng-if="waypoint.waypoint_type == 'paragliding_takeoff' || waypoint.waypoint_type == 'paragliding_landing' || waypoint.waypoint_type == 'climbing_outdoor'">
+    <label translate>orientation</label>  {{waypoint.orientation}}</p>
```
asaunier: extra blank space before ``{{waypoint.orientation}}``?

c2corg_ui/templates/waypoint/helpers/create_edit_summary.html
```diff
@@ -0,0 +1,288 @@
...
+      <li ng-repeat="type in waypoint['rock_types']">{{type| translate}}</li>
+    </ul>
+  </article>
+
+</figure>
+
+## DESCRIPTION
+<figure>
+  <h3>text</h3>
+  <p><label translate>description</label>
```
asaunier: Dont you need closing ``</p>`` here and below?

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+                <p ng-message="max"  style="{{ (editForm.elevation_min.$invalid && editForm.elevation_min.$touched) ? 'display:inline' : ''}}" translate>This field must be  than 9999 m.</p>
+                <p ng-message="number"  style="{{ (editForm.elevation_min.$invalid && editForm.elevation_min.$touched) ? 'display:inline' : ''}}" translate>This field must be a number.</p>
+              </div>
+            </div>
+
+            ##  PROMINENCE
+            <div id="prominence-group" ng-if="waypoint.waypoint_type == 'summit'" class="form-group data half" ng-class="{ 'has-error': editForm.prominence.$touched && editForm.prominence.$invalid, 'has-success': editForm.prominence.$valid }">
+              <label><span translate>prominence</span> *</label>
+              <div class="input-group">
+                <input type="number" name="prominence" min="0" ng-model="waypoint.prominence" class="form-control" required max="9999"/>
```
asaunier: ``prominence`` is not a required attribute.
See https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_waypoint.py#L29-L33

asaunier: I think you can remove the ``max`` validation, it's a bit ridiculous anyway to say that a prominence should be < 9999m.
By the way there is a typo in the ``ng-message``:
> This field must be **smaller** than 9999 m.

asaunier:
The orientation picker is REALLY COOL! Thanks :)

Please note ``orientation`` is a multiselect attribute (array):
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L78
Then it should be possible to select several orientations.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+              <div class="input-group">
+                <input type="tel" name="message" ng-model="waypoint.phone_custodian" class="form-control " placeholder="phone_custodian" />
+                <span class="input-group-addon"><span class="glyphicon glyphicon-earphone"></span></span>
+              </div>
+            </div>
+
+            ## CUSTODIANSHIP
+            <div class="form-group data half" ng-if="waypoint.waypoint_type == 'gite' || waypoint.waypoint_type == 'hut' || waypoint.waypoint_type == 'camp_site'">
+              <label><span translate>custodianship</span> <span ng-if="waypoint.waypoint_type == 'gite' || waypoint.waypoint_type == 'camp_site' || waypoint.waypoint_type == 'hut'">*</span></label>
+              <input type="tel" name="message" ng-model="waypoint.custodianship" class="form-control " placeholder="custodianship" ng-required="waypoint.waypoint_type == 'gite' || waypoint.waypoint_type == 'camp_site' || waypoint.waypoint_type == 'hut'"/>
```
asaunier:
```
ng-required="waypoint.waypoint_type == 'gite' || waypoint.waypoint_type == 'camp_site' || waypoint.waypoint_type == 'hut'"
```
could be replaced by
```
ng-required="['gite', 'camp_site', 'hut'].indexOf(waypoint.waypoint_type) > -1"
```

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+                <span class="input-group-addon"><span class="glyphicon glyphicon-resize-vertical"></span></span>
+              </div>
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.lift_access"></span>
+            </div>
+
+            ## PARKING FEE
+            <div class="form-group data half">
+              <label translate>parking_fee</label>
+              <div class="input-group">
+                <input type="number" name="parking_fee" ng-model="waypoint.parking_fee" class="form-control" min="1"/>
```
asaunier: This field is not a numeric field (with CHF or EUR). It's a dropdown
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L135
whose values are:
https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L195-L199

asaunier:
As mentioned above, ``prominence`` is not a required attribute.

To take benefit of the "common" attributes configuration:
https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_waypoint.py
I wonder if we could make a mako helper that would read the file above and print a ``required`` or ``ng-required=...`` attribute?

asaunier:
I don't think we need "glyphicon-ok" validation (green check mark appearing when a value is selected) for non required dropdown fields.

asaunier:
Please try to use a few new translations as possible, adding translation as an impact on the weight of the translations file to download and on the workload of translators. Prefer recycling existing translations or combining them.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+                <p ng-message="number"  style="{{ (editForm.capacity_staffed.$invalid && editForm.capacity_staffed.$touched) ? 'display:inline' : ''}}" translate>This field must be a number.</p>
+              </div>
+            </div>
+
+          </div>
+        </section>
+
+        ## MAPS
+        <section class="section maps-info">
+          <h2 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#edit-maps"><span translate>maps_info</span><span class="glyphicon glyphicon-menu-down"></span></h2>
```
asaunier: "Maps info" is not a relevant name for this section. Prefer something like "Geolocation".

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+              <span class="glyphicon glyphicon-warning-sign form-control-feedback" ng-show="editForm.latitude.$invalid && editForm.latitude.$touched"></span>
+              <div class="help-block" ng-messages="editForm.latitude.$error">
+                <p ng-message="required"  style="{{ (editForm.latitude.$invalid && editForm.latitude.$touched) ? 'display:inline' : ''}}" translate>This field is required.</p>
+                <p ng-message="number"  style="{{ (editForm.latitude.$invalid && editForm.latitude.$touched) ? 'display:inline' : ''}}" translate>This field must be a number.</p>
+                <p ng-message="min"  style="{{ (editForm.latitude.$invalid && editForm.latitude.$touched) ? 'display:inline' : ''}}" translate>This field must be between -90° and 90°.</p>
+                <p ng-message="max"  style="{{ (editForm.latitude.$invalid && editForm.latitude.$touched) ? 'display:inline' : ''}}" translate>This field must be between -90° and 90°.</p>
+              </div>
+            </div>
+
+            <app-map app-map-edit-ctrl="editCtrl" app-map-draw-type="Point" class="map-edit"></app-map>
```
asaunier: The map is a bit too small in height. We need a bigger map since we will probably add tools to switch the background, show other objects (waypoints for instance), etc.

In addition for routes or outings, the line might take room vertically. As a consequence users will have to zoom out to see the whole feature.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+              <select ng-model="waypoint.equipment_rating" ng-options="rat as rat for rat in ${equipment_ratings}" class="form-control"></select>
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.equipment_ratings.$valid"></span>
+              <span class="glyphicon glyphicon-warning-sign form-control-feedback" ng-show="editForm.equipment_ratings.$invalid && editForm.equipment_ratings.$touched"></span>
+            </div>
+
+            ## CLIMBING RATING MIN
+            <div class="form-group data half rating" ng-if="waypoint.waypoint_type == 'climbing_outdoor' || waypoint.waypoint_type == 'climbing_indoor'"
+                 ng-class="{ 'has-error': editForm.climbing_rating_min.$touched && editForm.climbing_rating_min.$invalid, 'has-success': editForm.climbing_rating_min.$valid }">
+              <label><span translate>climbing_rating_min</span></label><br>
+              <select ng-model="waypoint.climbing_rating_min" ng-options="rat as rat for rat in ${equipment_ratings}" class="form-control"></select>
```
asaunier: ``equipment_ratings`` is not the correct source for this attribute.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.climbing_rating_min.$valid"></span>
+              <span class="glyphicon glyphicon-warning-sign form-control-feedback" ng-show="editForm.climbing_rating_min.$invalid && editForm.climbing_rating_min.$touched"></span>
+            </div>
+
+            ## CLIMBING RATING MAX
+            <div class="form-group data half rating" ng-if="waypoint.waypoint_type == 'climbing_outdoor' || waypoint.waypoint_type == 'climbing_indoor'"
+                 ng-class="{ 'has-error': editForm.climbing_rating_max.$touched && editForm.climbing_rating_max.$invalid, 'has-success': editForm.climbing_rating_max.$valid }">
+              <label><span translate>climbing_rating_max</span></label><br>
+              <select ng-model="waypoint.climbing_rating_max" ng-options="rat as rat for rat in ${equipment_ratings}" class="form-control"></select>
```
asaunier: ``equipment_ratings`` is not the correct source for this attribute.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -7,139 +9,774 @@
...
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.climbing_rating_max.$valid"></span>
+              <span class="glyphicon glyphicon-warning-sign form-control-feedback" ng-show="editForm.climbing_rating_max.$invalid && editForm.climbing_rating_max.$touched"></span>
+            </div>
+
+            ## CLIMBING RATING MEDIAN
+            <div class="form-group data half rating" ng-if="waypoint.waypoint_type == 'climbing_outdoor' || waypoint.waypoint_type == 'climbing_indoor'"
+                 ng-class="{ 'has-error': editForm.climbing_rating_median.$touched && editForm.climbing_rating_median.$invalid,
+                                      'has-success': editForm.climbing_rating_median.$valid }">
+              <label><span translate>climbing_rating_median</span></label><br>
+              <select ng-model="waypoint.climbing_rating_median" ng-options="rat as rat for rat in ${equipment_ratings}" class="form-control"></select>
```
asaunier: ``equipment_ratings`` is not the correct source for this attribute.

asaunier:
Lots of comments but what a huge and great work @ginold !

c2corg_ui/static/js/documentediting.js
```diff
@@ -354,5 +382,204 @@ app.DocumentEditingController.prototype.getCoordinatesFromPoint_ = function(
...
+ */
+
+app.DocumentEditingController.prototype.step = function(step, direction) {
+  if (step === 1) {
+    $('.editing').animate({left: '0'});
+    this.animateBar_(step, direction);
+    this.previous_step = 0;
+    this.next_step = 2;
+  }
+  if (step === 2) {
```
tsauerwein: This and the following `if`s could be `else if`s. Or maybe use `switch`.

c2corg_ui/static/js/documentediting.js
tsauerwein: Which expressions have to be evaluated? Is there an event that you could listen to? Waiting for 500ms looks a bit fragile.

c2corg_ui/templates/waypoint/edit.html
ginold: I doesn't work, it's just for layout purpose. I thought that it might be useful as we are creating or editing a waypoint?

c2corg_ui/templates/waypoint/edit.html
ginold: well spotted !

c2corg_ui/templates/waypoint/edit.html
ginold: after a 100000th copy & paste that could have passed unseen.. :dart:

c2corg_ui/templates/waypoint/edit.html
asaunier: Actually we well need something like that in the outings (for providing snow conditions at various places such as in http://www.camptocamp.org/outings/699400/fr/tricotage-a-la-croix-de-l-ecuelle-et-a-la-pointe-des-follys-depuis-la-lanche#description). Then you may save the code somewhere to reuse it later.

But we don't need it here and should stay simple (ie. use a small text field). So I'd prefer you remove it: only do things that are actually needed :)

c2corg_ui/static/js/documentediting.js
ginold: it works for 5 or less steps. You're right saying that it should be the most generic and work for event 100 steps, but I've created max 5 divs with 18% of width of the parent div that only slides to the left or right. My argument is that 5 steps is already a lot and we should not have more than that :)

asaunier: Right. The wireframes only suggest 4 steps for the outings. So I guess 5 steps is enough anyway.

c2corg_ui/static/js/documentediting.js
ginold: max_steps ( = $(.step).length ) has to be evaluated and it's the number of .step classes that depend on ng-if="waypoint.waypoint_type", some have more or less steps and I wait until for this eval (not sure how to do better). If I don't do that, max_steps will first be initialized as 4 while it's real value will have changed to 5 in between. I've had this problem when changing waypoint type and max_steps was always equal to it's previous value, not the selected one!

c2corg_ui/static/js/documentediting.js
```diff
@@ -354,5 +382,204 @@ app.DocumentEditingController.prototype.getCoordinatesFromPoint_ = function(
...
+/**
+ * Animate the progress bar
+ * @param {number} step
+ * @param {string} direction
+ * @private
+ */
+
+app.DocumentEditingController.prototype.animateBar_ = function(step, direction) {
+  var percent = 100 / this.max_steps;
+  var that = this;
```
tsauerwein: You can avoid the `that` by using `bind`:

     var timeout = setInterval(function() {...}.bind(this));

c2corg_ui/static/js/documentediting.js
```diff
@@ -117,12 +124,33 @@ app.DocumentEditingController = function($scope, $element, $attrs,
    */
   this.alerts_ = appAlerts;
 
+
+  /**
+   * first next step would be 2 by default
+   * @export
+   */
+  this.next_step = 2;
```
tsauerwein: We usually use camel-case `nextStep` instead of `next_step` in Javascript.

c2corg_ui/static/js/documentediting.js
```diff
@@ -354,5 +382,204 @@ app.DocumentEditingController.prototype.getCoordinatesFromPoint_ = function(
...
+  }, 10);
+};
+
+/**
+ * Update steps
+ * @export
+ */
+
+app.DocumentEditingController.prototype.updateMaxSteps = function() {
+  var that = this;
```
tsauerwein: Same here: Use `bind` instead of `that`.

c2corg_ui/static/js/documentediting.js
tsauerwein: These 500ms are also to wait for the request which fetches the document data? Can't you set `maxSteps` once you have the data?
https://github.com/c2corg/v6_ui/blob/edit/create-route/waypoint-attributes/c2corg_ui/static/js/documentediting.js#L229

And then do the "calculation" how many steps there are in the controller.

c2corg_ui/static/js/documentediting.js
ginold: thanks! i didn't know this little thing.

c2corg_ui/static/js/documentediting.js
ginold: not the document data, but it gives it a little bit of time for the ng-ifs. How do i know when the ng ifs are finished evaluating? I changed to
 `    
link: function(scope, el, attrs, ctrl) {
     ctrl.updateMaxSteps();
    }`

and it works ok, but when I set it in the controller like `this.max_steps = this.updateMaxSteps()` it never gets the correct value (at time).

tsauerwein: > not the document data, but it gives it a little bit of time for the ng-ifs. How do i know when the ng ifs are finished evaluating?

When do the `ng-if`s start evaluating? When `this.scope_['waypoint']` is set. And that is when the request has finished loading. I wouldn't try to count the elements with a class `.steps`. When you have the waypoint (when the data is loaded), I would determine how many steps you need depending on the type and then directly set `this.maxSteps_`.

ginold: I thought about it but it implies to specify a number of steps for every waypoint type and it's some lines of code more...or add this propriety to the DB? i don't know.

asaunier: The number of steps has no connection with the API/DB. It is only related with the UI and the way it works. So it's not a  good idea to add it to the DB.
But it could be set in some config in the UI, for instance in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/constants.js ?

ginold: Good idea, I can try that.
But I'm not really sure to understand what is the problem with waiting 500ms to evaluate this max_steps (which isn't immediately needed neither) ? The DOM has loaded and the user can start editing/creating a document, this variable will just appear a 0.5s later.

tsauerwein: The problem is not that it will appear 0.5 seconds later. The problem is that you have no guarantee that 500ms are are always long enough. Are the elements that you want to count always ready at that moment?

It would be good if you could avoid this timeout.

ginold: in my case yes, but maybe it's thanks to a fast computer and internet
connection. :D
Trying to do that with constants.

On Fri, Mar 4, 2016 at 1:33 PM, Tobias Sauerwein <notifications@github.com>
wrote:

> In c2corg_ui/static/js/documentediting.js
> <https://github.com/c2corg/v6_ui/pull/220#discussion_r55024386>:
>
> > @@ -33,7 +33,14 @@ app.documentEditingDirective = function() {
> >      scope: true,
> >      controller: 'appDocumentEditingController',
> >      controllerAs: 'editCtrl',
> > -    bindToController: true
> > +    bindToController: true,
> > +    link: function(scope, el, attrs, ctrl) {
> > +      // The directive executes after the dom is loaded but we have to wait until the expressions are evaluated
>
> The problem is not that it will appear 0.5 seconds later. The problem is
> that you have no guarantee that 500ms are are always long enough. Are the
> elements that you want to count always ready at that moment?
>
> It would be good if you could avoid this timeout.
>
> —
> <https://github.com/c2corg/v6_ui/pull/220/files#r55024386>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

tsauerwein: > in my case yes, but maybe it's thanks to a fast computer and internet connection. :D

Exactly that is the point. :)

asaunier:
elevation is not a required attribute for "climbing_indoor":
https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_waypoint.py#L95-L99

c2corg_ui/static/js/maincontroller.js
```diff
@@ -69,3 +69,10 @@ app.MainController.prototype.isPath = function(path) {
   }
 };
 
+
+/**
+ * @export
+ */
+app.MainController.prototype.animateHeaderIcon = function(e) {
+  app.utils.animateHeaderIcon(e);
```
asaunier: Can't you directly use ``app.utils.animateHeaderIcon(e);`` in the HTML instead of this alias in the main controller?

asaunier:
Thanks for the changes. Since the remaining desired changes seem done in https://github.com/c2corg/v6_ui/pull/224 and if no one objects, I'm OK for merging the PR.

c2corg_ui/static/js/maincontroller.js
ginold: I tried but it doesn't seem to work, as if it was awaiting a controller function. But I find that 'main controller' is supposed to represent a controller that can be used application-wide, and in this case we may want to use this particular function in multiple parts of the app.

PR #221, opened 2016-03-04
Returns a 500 error when API fails responding to a list request

asaunier:
Replaces #200 

I have cherry-picked some commits from #200 that were relevant anyway.


asaunier:
Any objection?
If none I will merge this PR and close #200 .

PR #222, opened 2016-03-07
Always regenerate template files

gberaudo:
This PR makes sure the files are generated and up-to-date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 SITE_PACKAGES = $(shell .build/venv/bin/python -c "import distutils; print(distutils.sysconfig.get_python_lib())" 2> /dev/null)
-TEMPLATE_FILES_IN = $(filter-out ./.build/% ./node_modules/%, $(shell find . -type f -name '*.in'))
+TEMPLATE_FILES_IN = $(filter-out ./.build/% ./node_modules/% ./v6_api/%, $(shell find . -type f -name '*.in'))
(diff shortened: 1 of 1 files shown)


asaunier:
Could you give more details? Why do we need this? What was the problem?

gberaudo:
- when you did make check from a fresh clone it would error since no test.ini was generated;
- when you did make -f config/alex install; then make -f config/main install it would use generated files for alex.

gberaudo:
Added a fix a strange bug: when building on Travis, v6_api path was included in the template path.
Don't know why it worked previously.

tsauerwein:
> Added a fix a strange bug: when building on Travis, v6_api path was included in the template path.
Don't know why it worked previously.

Before, the ui project was built, after that the api project was checked out and built. So, the moment the templates were generated for the ui project, the api project was not there yet. But now, when you start the tests for the ui, the template files are re-generated.

gberaudo:
Thanks for the explanations @tsauerwein. It makes sense.

> The template files are re-generated when one of the config/* files has changed?
Do you mean the dependency is now useless and I should remove it?

gberaudo:
Removed useless dependency on config files.

tsauerwein:
I think that looks good!

gberaudo:
Thanks for the review @tsauerwein and @asaunier. I will create a similar PR for the API.

PR #224, opened 2016-03-08
Edit create route attributes

ginold:
screenshot from 2016-03-08 16 38 53
screenshot from 2016-03-08 16 39 26

same idea here as for waypoints :
- step by step creation/edition flow
- fields change depending on route activities


asaunier:
If I get it well, the 5 first commits of this PR come from https://github.com/c2corg/v6_ui/pull/220
Then the relevant commits are c83a2b0 and de7662d , correct?

asaunier:
Making it possible to associate WP directly from the route editing form is indeed a good idea and will most likely encourage people to create those associations (in v5 most people don't do it, perhaps because they have not noticed it was possible).

But I guess it implies changes in the API to actually create the associations...

tsauerwein:
> But I guess it implies changes in the API to actually create the associations... 

You mean in one request?

asaunier:
Maybe not. It would be cool to be able to reuse existing tools/services, if possible/not too complicated.

ginold:
Isn't it one of the main points we discussed - to be able, and forced, to
associate a waypoint while creating a new route?
why does it imply changes to the api? it is already available in the
detailed-view page, isn't it?

On Tue, Mar 8, 2016 at 5:02 PM, Alex Saunier <notifications@github.com>
wrote:

> Making it possible to associate WP directly from the route editing form is
> indeed a good idea and will most likely encourage people to create those
> associations (in v5 most people don't do it, perhaps because they have not
> noticed it was possible).
>
> But I guess it implies changes in the API to actually create the
> associations...
>
> —
> <https://github.com/c2corg/v6_ui/pull/224#issuecomment-193839860>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

tsauerwein:
When you create/modify a route, you can specify a main waypoint. I think Alex is talking about adding/updating arbitrary associations when updating a route (in one request)?

asaunier:
> why does it imply changes to the api? it is already available in the
detailed-view page, isn't it?

Yes but it means that the associations will be done even though the document editing is finally not confirmed (maybe not a problem).

Another problem is that when creating a new document, the associations won't work since we the document id is not available until the document is actually created => we need to postpone the associations the creation is done, thus we cannot use the association tool as is.

asaunier:
> When you create/modify a route, you can specify a main waypoint

Can we already set the main WP when editing a route?
If yes, how does it work? One should provide the main WP id in the POST/PUT JSON request?

tsauerwein:
Like so: https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_route.py#L391-L411

asaunier:
OK, then we need in the editing form to be able to list the associated WPs (with the current main WP being marked as "main"), pick one and then update the ``main_waypoint_id`` attribute.

tsauerwein:
Note that currently no association is created when setting the main waypoint. But it would make sense, see https://github.com/c2corg/v6_api/issues/177.

asaunier:
Associations and main waypoint could be managed in separated PR. I have created issues about those topics: https://github.com/c2corg/v6_ui/issues/227 https://github.com/c2corg/v6_ui/issues/228

asaunier:
No objections against merging this PR when #220 is merged and conflicts are managed (most likely you will have to rebase and drop some commits).

asaunier:
Do you need to have both ``activities.svg`` and other SVG files for every single activity?

asaunier:
The organisation of the attributes among the steps and sections could probably be discussed. I see things that are not very consistent. For instance the "exposition" attributes in the "gear" section.
It would make sense to show the summary field before the description field. Maybe also to show the title field in the same step.
 
I think the map diserve its own step, should be bigger and with a tool to upload a GPS track.

But well, you don't have to guess how to organise the attributes. I suggest we keep them like this and I will ask the client to give us instructions.

c2corg_ui/static/img/fonts/icomoon.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" >
+<svg xmlns="http://www.w3.org/2000/svg">
+<metadata>Generated by IcoMoon</metadata>
```
asaunier: Is it a good idea to name a file according to the tool used to generate it?

c2corg_ui/static/img/news.svg
```diff
@@ -0,0 +1,44 @@
...
+	<path style="fill:#030104;" d="M81.114,30.146C73.196,11.725,60.456-2.306,54.009,0.316C43.062,4.765,60.53,26.133,6.774,47.984
+		c-4.643,1.887-5.82,9.441-3.877,13.957c1.939,4.515,8.291,9.01,12.936,7.123c0.803-0.326,3.754-1.275,3.754-1.275
+		c3.316,4.45,6.787,1.811,8.018,4.639c1.479,3.398,4.697,10.785,5.789,13.298c1.096,2.511,3.576,4.837,5.377,4.153
+		c1.795-0.684,7.91-3.011,10.25-3.901c2.338-0.89,2.896-2.974,2.182-4.619c-0.771-1.772-3.932-2.292-4.832-4.36
+		c-0.902-2.068-3.848-8.696-4.695-10.785c-1.152-2.841,1.295-5.152,4.85-5.522c24.467-2.55,29.041,12.561,37.371,9.173
+		C90.343,67.243,89.03,48.568,81.114,30.146z M78.358,60.03c-1.432,0.582-11.061-7.013-17.215-21.334
+		c-6.152-14.318-5.379-27.408-3.949-27.989c1.43-0.582,10.822,8.58,16.975,22.898C80.321,47.923,79.79,59.448,78.358,60.03z"/>
+</g>
+<g>
+</g>
```
asaunier: ??
Do we need all those empty tags?

Shouldn't we put this file in an SVG font containing all the icons?

ginold: actually all the svgs are already added to the icomoon fonts. Those svgs are here just in case i want to make changes to them and reupload a font (so theoretically you don't need them and i could have them on my pc only?)

c2corg_ui/static/img/fonts/icomoon.svg
ginold: don't know, it's been autogenerated

c2corg_ui/templates/helpers/view.html
```diff
@@ -97,15 +97,22 @@
...
+      % if 'title_prefix' in route['locales'][0] and route.locales[0]['title_prefix']:
+        <a href="${request.route_url('routes_view', id=route['document_id'], lang=route['locales'][0]['lang'])}">${route['locales'][0]['title']}</a>
       % else :
-        <a href="${request.route_url('routes_view', id=r['document_id'], lang=r['locales'][0]['lang'])}">${r['locales'][0]['title']}</a>
+        <a class="item-title" href="${request.route_url('routes_view', id=route['document_id'], lang=route['locales'][0]['lang'])}">${route['locales'][0]['title']}
+          <div class="route-activities">
+            ${get_route_activities(route, 'top')}
+          </div>
+          <div>
+            <p><b translate>elevation</b>  min ${route['elevation_min']} / max ${route['elevation_max']}</p>
```
tsauerwein: Note that `elevation_min/max` are not required fields, so you might end up with: "elevation min / max".

c2corg_ui/static/img/news.svg
asaunier: OK. Having the source SVG files in the repo is probably a good idea (unless we can easily extract them from the fonts?). To make it clearer that they are not used as is, perhaps we could store them in a ``src/`` directory?

asaunier:
When you click on checkboxes (for instance rock types), it adds some space at the bottom of the page (the scroll bar handle shrinks) whereas additional page is not needed!?

ginold:
thats because your summary pages gets bigger and new values are appearing.

asaunier:
> thats because your summary pages gets bigger and new values are appearing.

When you check more rock types?

ginold:
when you check or type or select anything.

On Fri, Mar 11, 2016 at 2:25 PM, Alex Saunier <notifications@github.com>
wrote:

> thats because your summary pages gets bigger and new values are appearing.
>
> When you check more rock types?
>
> —
> <https://github.com/c2corg/v6_ui/pull/224#issuecomment-195361420>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

c2corg_ui/static/js/lang.js
```diff
@@ -70,7 +74,7 @@ app.LangController = function(gettextCatalog, langUrlTemplate,
    */
   this.lang = this.cookies_.get('interface_lang') ||
       ngeoGetBrowserLanguage(this['langs']) || 'fr';
-  this.updateLang();
+  this.updateLang(null);
```
asaunier: I think you can avoid passing a ``null`` argument by adding a default value in the ``updateLang`` definition.
For instance
```
app.LangController.prototype.updateLang = function(lang) {
  lang = lang || null;
```

asaunier: Or maybe simply:
```
app.LangController.prototype.updateLang = function(lang) {
  if (lang || false) {
```

tsauerwein: Maybe this should rather be?
```js
var lang = this.cookies_.get('interface_lang') ||
       ngeoGetBrowserLanguage(this['langs']) || 'fr';
this.updateLang(lang);
```

asaunier:
I was just wondering why checking additional rock types (when some were already checked for instance) change the page height.
But that's a really minor issue.

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -35,6 +35,7 @@ <h1 ng-init="id = ${waypoint_id if waypoint_id else 0}" class="text-center">
     <span class="ng-hide">
       {{editCtrl.updateMaxSteps(waypoint.waypoint_type)}} 
       {{editCtrl.waypoint_type = waypoint.waypoint_type}}
+      {{document = waypoint}}
```
asaunier: Does this mean this hidden span is used to set vars?
Isn't it a bit of a hack?
Can't you an ``ng-init`` for instance?

ginold: it is a hack but tried only after seeing that it doesn't work with ng init.
Probably ng init comes to early and has false values, whereas this set can
be modified and updated on the fly.

On Fri, Mar 11, 2016 at 2:45 PM, Alex Saunier <notifications@github.com>
wrote:

> In c2corg_ui/templates/waypoint/edit.html
> <https://github.com/c2corg/v6_ui/pull/224#discussion_r55830010>:
>
> > @@ -35,6 +35,7 @@ <h1 ng-init="id = ${waypoint_id if waypoint_id else 0}" class="text-center">
> >      <span class="ng-hide">
> >        {{editCtrl.updateMaxSteps(waypoint.waypoint_type)}}
> >        {{editCtrl.waypoint_type = waypoint.waypoint_type}}
> > +      {{document = waypoint}}
>
> Does this mean this hidden span is used to set vars?
> Isn't it a bit of a hack?
> Can't you an ng-init for instance?
>
> —
> <https://github.com/c2corg/v6_ui/pull/224/files#r55830010>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

ginold:
not anymore #updated

On Fri, Mar 11, 2016 at 1:27 PM, Alex Saunier <notifications@github.com>
wrote:

> Do you need to have both activities.svg and other SVG files for every
> single activity?
>
> —
> <https://github.com/c2corg/v6_ui/pull/224#issuecomment-195348048>.
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

c2corg_ui/static/js/lang.js
ginold: changed to this.updateLang(this.lang);
and lang = lang || null;  

and it works well :)) thx

asaunier:
Thanks for the changes. Looks good to me.

PR #236, opened 2016-03-15
Mockup style

ginold:
mostly css modifications in order to get closer to the look of mock-ups.

screenshot from 2016-03-15 13 14 36
screenshot from 2016-03-15 13 14 45
screenshot from 2016-03-15 13 15 43


asaunier:
A few minor comments else that looks good to me thanks.
I'm not very found of the green you used though => too fluo!
Please stick to the color proposed by Altimax.

asaunier:
The menu could probably be a bit improved. On my laptop the ad banner overflow above the navigation buttons. If the screen heigh shrinks those buttons can no longer be used.

Could you make sure that:
* the navigation buttons (main content of the left bar) are always in front of the banner and the other contents.
* remove the "ad" word (is it needed anyway?)
* is the 140x140 banner the size suggested by Altimax? Could you use the precise format provided in their layout?
* remove the "portals" button in the menu (does not exist)

I think we should remove the logo from the left bar and always put it (smaller) to the left of the main column. As when the screen is narrower (tablet or phone). By the way in tablet size, the logo has no margin and touch the border of the page, which is not nice.

What could we do when the screen is not big enough to show all the content? Have a scroll bar?

ginold:
- nav buttons done
- ad word removed 
- banner is 16/9 now
- portals removed

but i'm not sure how you want the logo to be and if I do understand - i'm completely not with you! 
could you draw how you see it? 
Neither I see any problems with logo margins or its touching the page borders...

if you have a screen soooooooooooooo small in height then there is a scroll bar, but you should use a nokia 3310 to have a screen this small :P

asaunier:
> nav buttons done
ad word removed
banner is 16/9 now
portals removed

OK thanks!

The buttons still go under the banner:
screencapture-c2corgv6-demo-gis-internal-6555-routes-1458154504473

> Neither I see any problems with logo margins or its touching the page borders...

You're right, it's OK with this branch (but not with the master branch, see the demo).

> if you have a screen soooooooooooooo small in height then there is a scroll bar, but you should use a nokia 3310 to have a screen this small :P

Well, see my screenshot above. OK I have reduced **a bit** the window height (I have a macbook pro retina screen) but that's not that marginal. What about mobile screens? Especially if you use your phone in landscape mode.

> but i'm not sure how you want the logo to be and if I do understand

I meant for instance:
* remove the logo from the left menu bar
* always show it as in tablet/mobile mode like in:
![screencapture-c2corgv6-demo-gis-internal-6555-routes-1458155637139]
(https://cloud.githubusercontent.com/assets/1192331/13825642/d291001a-ebb3-11e5-8ea3-ee48c5741954.png)

By the way in the screenshot above, do we really need the gray borders around the hamburger button or the one above the "Home" button?

By the way it seems there are some problems with the filters section which overflows in the search page (maybe not related to this PR):
screencapture-c2corgv6-demo-gis-internal-6555-routes-1458155135044
=> when "more filters" is clicked the filter section should take the whole height.

gberaudo:
@ginold if you think you addressed @asaunier comments I am fine if you merge the PR.

c2corg_ui/templates/sidemenu.html
```diff
@@ -47,39 +41,45 @@
...
-      <span class="icon-google-plus"></span>
-      <span class="icon-facebook"></span>
-      <a href="http://www.freepik.com/" class="copyright">some icons designed by freepik</a>
+      <div class="icons">
+        <a href="https://twitter.com/camptocamp"><pan class="icon-twitter"></span></a>
+        <a href="https://plus.google.com/104270548458536561874/posts"><span class="icon-google-plus"></span></a>
+        <a href="https://www.facebook.com/Camptocamp-207324185960679/"><span class="icon-facebook"></span></a>
+      </div>
+      <div class="copyright-text">
+        <a href="http://www.freepik.com/" class="copyright">some icons designed by freepik</a>
```
tsauerwein: Do we want to keep the copyright for the icons in such a prominent place? Can't we only show it in the imprint?

> How to attribute it?
> For websites:
> Insert the attribution on the page of the icon (for example in the page footer) **or on the imprint page**.

http://support.flaticon.com/hc/en-us/articles/207248209-How-I-must-insert-the-attribution-

asaunier:
I have only one comment: GREAT JOB! Please merge.

Or maybe another one:
I agree with @tsauerwein about the icons copyright: if it's allowed to give credits in a separated impressum/credits page, I would do so. Anyway we will have many credits to give to people/sponsors/libs.

Perhaps you could simply remove the mention in this PR and create an issue about the credits (I can do it and ask the client what should appear in it).
By the way, since c2corg uses CC licenses, "© 1997-2016 all rights reserved" could be replaced by "© 1997-2016 some rights reserved" :P

PR #237, opened 2016-03-17
Handle validation of registration email

gberaudo:
At the end of the process, the registration is completed and
the user logged in.

Useful with https://github.com/c2corg/v6_api/pull/190.

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -241,6 +241,19 @@ app.Api.prototype.register = function(data) {
 
 
 /**
+ * @param {string} nonce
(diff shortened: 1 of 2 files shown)


asaunier:
As far as I can see it works the way you described me last week: when a new user clicks on the link in the validation email message it is then directly logged in without having to enter his username/password.
Correct?

asaunier:
And the "password forgotten" feature would work the same: one would receive an email with a link to click that would log him in instead of receiving a new email. Still correct?

gberaudo:
Yes in both cases the user is directly logged in.

asaunier:
Looks good to me

Issue #85, opened 2016-01-11
Merge JS lib and CSS files


asaunier:
As for now, we have several node_modules files that are loaded from the HTML including in standard mode: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L74-L80

Merging those lib files in a single one would be useful.

tsauerwein:
For libraries like jQuery, Angular and TypeAhead we could also consider using CDN URLs so that these libraries might already be in the browser cache from visiting other sites that use the same libraries.

tsauerwein:
See also: http://stackoverflow.com/a/21099939/119937

asaunier:
OK, that indeed makes sense!

gberaudo:
By using a CDN you expose user's privacy to the CDN (referrer, cookie).
There is also the risk of the CDN being hacked and serving malwares.
It also adds a failure point: if the CDN is down c2c becomes unusable.

In modern browsers it is possible to protect from CDN malwares by using subresource integrity: https://hacks.mozilla.org/2015/09/subresource-integrity-in-firefox-43/

For the first point, it is a choice to make between user's privacy and speed.

tsauerwein:
Or we concatenate all dependencies into a single file. We would have a `deps.js` and a `build.js` file. The `build.js` is going to change more often, so we wouldn't have to invalidate the cache for the `deps.js` file every time.

gberaudo:
Concatenating all dependencies into a signle `deps.js` file looks good to me.

asaunier:
Sounds good to me to have a deps.js file. As mentioned by @tsauerwein upgrading the libs will occur not that often.

tsauerwein:
Same for CSS files (I updated the title).

PR #240, opened 2016-03-18
Outings

ginold:
continuing with another document type - outings. Looks good to me! 

I updated views in other documents and helpers in order to support outings.

I also added moment.js and angular-moment to translate dates and have some easy date manipulation, so we will easily be able to add stuff like '8 minutes ago'  or 'outing of 16th may 2016' in all languages.

outing/route/waypoint cards are generated everywhere by the same helper.

Edit form supports localstorage - > if user exists the window and comes back, the unfinished form will be filled with data previously saved (each 5 secs). It is removed and replaced when user changes of document.

screenshot from 2016-03-18 17 39 21
screenshot from 2016-03-18 17 38 24
screenshot from 2016-03-18 17 37 48
screenshot from 2016-03-18 17 38 13

screenshot from 2016-03-18 17 38 56

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -47,6 +47,7 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @param {angular.Attributes} $attrs Attributes.
  * @param {app.Authentication} appAuthentication
  * @param {app.Alerts} appAlerts
+ * @param {amMoment} amMoment angular moment directive.
  * @param {app.Api} appApi Api service.
  * @param {string} authUrl Base URL of the authentication page.
  * @constructor
@@ -54,7 +55,7 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @export
  */
 app.DocumentEditingController = function($scope, $element, $attrs,
-    appAuthentication, appAlerts, appApi, authUrl) {
+    appAuthentication, appAlerts, appApi, authUrl, amMoment) {
 
   /**
    * @type {angular.Scope}
@@ -118,6 +119,12 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   this.alerts_ = appAlerts;
 
   /**
+   * @type {amMoment}
+   * @private
+   */
+  this.amMoment_ = amMoment;
+
+  /**
    * first next step would be 2 by default
    * @export
    */
@@ -154,6 +161,10 @@ app.DocumentEditingController = function($scope, $element, $attrs,
    */
   this.scope_.waypoint = {};
 
+  /**
+   * @export
+   */
+  this.differentDates;
 
   /**
    * @type {app.Api}
@@ -186,26 +197,15 @@ app.DocumentEditingController = function($scope, $element, $attrs,
     this.handleMapFeatureChange_(feature);
     this.scope_.$apply();
   }.bind(this));
+
+  this.putLocalstorage_();
 };
-
-
-/**
- * @const
- * @type {string}
- */
-app.DocumentEditingController.FORM_PROJ = 'EPSG:4326';
-
-
-/**
- * @const
- * @type {string}
- */
-app.DocumentEditingController.DATA_PROJ = 'EPSG:3857';
 
 
 /**
  * @param {Object} response Response from the API server.
  * @private
+ * @suppress {checkTypes}
  */
 app.DocumentEditingController.prototype.successRead_ = function(response) {
   var data = response['data'];
@@ -236,7 +236,24 @@ app.DocumentEditingController.prototype.successRead_ = function(response) {
     this.isNewLang_ = true;
   }
 
-  this.scope_[this.modelName_] = data;
+  var document = JSON.parse(window.localStorage.getItem(this.modelName_));
+
+  // if document does'nt exist in localstorage or it's another document, set the data got from API
+  if (document === null || document['document_id'] !== data['document_id']) {
+    this.scope_[this.modelName_] = data;
+    // if it exists and it's the same document (for ex reloaded the page), replace whit what's in localstorage
+  } else if (document['document_id'] === data['document_id']) {
+    this.scope_[this.modelName_] = document;
+  }
+
+  if (this.modelName_ === 'outing') {
+    var outing = this.scope_['outing'];
+    outing['date_start'] = app.utils.formatDate(outing['date_start']);
+    outing['date_end'] = app.utils.formatDate(outing['date_end']);
+    // to show or hide date_end in the form
+    this.differentDates = app.utils.areDifferentDates(outing['date_start'], outing['date_end']);
+  }
+
   this.scope_.$root.$emit('documentDataChange', data);
 };
 
@@ -270,8 +287,8 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
     var lonlat = data['lonlat'];
     if ('longitude' in lonlat && 'latitude' in lonlat) {
       var point = new ol.geom.Point([lonlat['longitude'], lonlat['latitude']]);
-      point.transform(app.DocumentEditingController.FORM_PROJ,
-                      app.DocumentEditingController.DATA_PROJ);
+      point.transform(app.constants.documentEditing.FORM_PROJ,
+                      app.constants.documentEditing.DATA_PROJ);
       // If creating a new document, the model has no geometry attribute yet:
       data['geometry'] = data['geometry'] || {};
 
@@ -318,6 +335,8 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
         this.module_, this.id_, this.lang_);
     }.bind(this));
   }
+  // remove the edited item from localstorage
+  window.localStorage.removeItem(this.modelName_);
 };
 
 
@@ -343,8 +362,7 @@ app.DocumentEditingController.prototype.updateMap = function() {
     var lonlat = data['lonlat'];
     if ('longitude' in lonlat && 'latitude' in lonlat) {
       var point = new ol.geom.Point([lonlat['longitude'], lonlat['latitude']]);
-      point.transform(app.DocumentEditingController.FORM_PROJ,
-                      app.DocumentEditingController.DATA_PROJ);
+      point.transform(app.constants.documentEditing.FORM_PROJ, app.constants.documentEditing.DATA_PROJ);
       // If creating a new document, the model has no geometry attribute yet:
       data['geometry'] = data['geometry'] || {};
       data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(point);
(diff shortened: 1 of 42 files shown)


asaunier:
There's probably something missing in the PR: opening an outing detail view makes a crash:

>   File "/home/alex/Bureau/v6_ui/c2corg_ui/templates/outing/view.html", line 11, in render_body
    <%
  File "/usr/lib/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'NoneType'

c2corg_ui/templates/base.html
```diff
@@ -109,6 +118,15 @@
...
+    <script src="${request.static_url('%s/moment/min/moment.min.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/angular-moment/angular-moment.min.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/ca.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/fr.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/en-gb.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/de.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/es.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/eu.js' % node_modules_path)}"></script>
+    <script src="${request.static_url('%s/moment/locale/it.js' % node_modules_path)}"></script>
     <script src="${request.static_url('c2corg_ui:static/build/build.js')}"></script>
```
asaunier: If the files are the same in both standard and debug modes, you could include them only once, outside of the ``if debug:`` blocks.

Isn't it possible to load those locale files only when the interface lang changes? (as done for angular-gettext).
If not we will have to make sure that those locale files are included in a global libaries build. See https://github.com/c2corg/v6_ui/issues/85

ginold: does it happen to every outing or just to a particular one ? i don't have this issue

asaunier:
Whatever the list page (outings, routes, WP), the creation button is "Add a WP".

asaunier:
Is it normal that in the routes list page, activities icons are no longer displayed?

asaunier:
> does it happen to every outing or just to a particular one ? i don't have this issue

Yes, all the outings I have tried. It happens in both my local and dev server instances: 
http://c2corgv6-demo.gis.internal:6555/outings

ginold:
oO everything is ok in my instance. Maybe run upgrade-dev or something. How can I debug this? maybe your instance has a problem.

c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
```diff
@@ -0,0 +1,199 @@
...
+<%def name="get_outing_location(waypoint, geometry4326)">\
+<div class="name-icon-value " ng-click="detailsCtrl.openTab($event)">
+  <h4><span translate>Location</span><span class="glyphicon glyphicon-menu-right"></span></h4>
+  <div class="name-icon">
+    <span class="glyphicon glyphicon-map-marker"></span>
+    <p translate>Location</p>
+  </div>
+  <span class="detail-text accordion" ng-init="point = ${geometry4326['geom']}">
+  x: {{point.coordinates[0] | number: 0}}
+  y:  {{point.coordinates[1] | number: 0}}
```
asaunier: We don't need this for outings: the default point geometry is only used to recenter the map or make geographic filtering. But having the coordinates of a point is not relevant for outings: they are not points.

c2corg_ui/templates/outing/view.html
```diff
@@ -0,0 +1,161 @@
...
+<%inherit file="../base.html"/>
+<%namespace file="../helpers/common.html" import="show_title"/>
+<%namespace file="helpers/detailed_outings_attributes.html" import="get_outing_associated_maps, get_outing_snow, get_outing_access, get_outing_participants, get_outing_location, get_outing_general, get_outing_heights"/>
+<%namespace file="../helpers/view.html" import="get_document_min_max, get_document_locale_text, show_attr, show_missing_langs_links,
+    show_other_langs_links, show_archive_data, show_route_title, get_route_activities, 
+    show_areas, show_maps, show_float_buttons, get_associated_documents"/>
+<%
+outing_id = outing['document_id']
+other_langs, missing_langs = get_lang_lists(outing, lang)
+geometry4326 = outing['geometry']
```
asaunier: We don't need ``geometry4326`` for outings. As explained for function ``get_outing_location``, no coordinates are shown.

c2corg_ui/templates/outing/view.html
```diff
@@ -0,0 +1,161 @@
...
+point = json.loads(outing['geometry']['geom'])
```
asaunier: This line is responsible of the crashes I have observed when loading an outing detail view => if the outing has no geometry set (``None``), python crashes saying that ``TypeError: the JSON object must be str, not 'NoneType'``.

My API instance seems outdated since I get no error when using the API demo instance.
In the new DB structure ``outing['geometry']['geom']`` is the default point geomtry and is always available for all outings. @tsauerwein is it correct?

Anyway I don't see where the ``point`` variable is used in the template!?

tsauerwein: > In the new DB structure outing['geometry']['geom'] is the default point geomtry and is always available for all outings. @tsauerwein is it correct?

Yes. Except that there is no guarantee that there is always a default geometry. You might end up with an outing that has no default geometry.

asaunier: OK thanks. This means we need to check the geom is not null before using ``json.loads``. For instance:
```
point = json.loads(outing['geometry']['geom']) if outing['geometry']['geom'] not None else None
```

asaunier:
Button "Add my outing" in the top banner is not functional.
By the way rather name it "Add an outing" ("my" is so old schooool :P )

Concerning the outing editing/creating form process:
* have you really used the steps described in the wireframes for creating an outing? It seems not. For instance the GPS track should be provided at step 1 (to suggest associated routes in the neighborhood), ditto for routes association tool, useful to suggest an outing title. See https://basecamp.com/1760892/projects/9067037/todos/234805483
* what's the difference between the "Save" and "Post" buttons?
* why do we have those 2 (actually 3) buttons only in the first step?

ginold:
- save/post ? i don't know, i just copied the mockup
- it's not exactely like seen on wireframe, because they (of course) did't anticipate so much data and most likely didn't event know what are the resting inputs so I made a few choices for not squishing it all together. They forgot 'date end' for example, or doesn't it exist anymore? I saw 'parcours'  and 'itineraire' section, I thought it was clearly the better place for 'itineraire suivi'. What is the difference between 'itineraire suivi' & 'itineraire emprunte' ?
- but "ajouter ma sortie" is not oldschool ?:P

asaunier:
OK keep the steps as is, but as for the WP and routes, I think it will have to be reviewed by the client.

asaunier:
I have updated (checked) some of the issues reported previously.

Other problems I see:
* [ ] The category dropdown is set to "routes" when showing the outings list page (should be "outings").
* [ ] Could you change the order of the categories? I recommend routes > waypoints > outings.
* [ ] In outings list view maps are not needed, only for WP and routes detail views (should probaby be removed on API as well)
* [ ] I know the layout overviews minimize the date but it's an essential info that should be more obvious. I think it should appear in the title or as subtitle, above the map (as in the wireframes).
* [ ] Do we really need to have the activities displayed twice (icon in the title + name in the figures)?
* [ ] The associated routes should also be more obvious, and be shown higher in the page.
* [ ] "Drescription" is not a relevant title (in the editing form as well). Should be "comments" (or "personal comments").

asaunier:
I had some precision from the client for outings dates (can be managed in a separated PR if you prefer):
* [ ] date to show in outings lists is ``date_end``
* [ ] showing the date as subtitle of outings detail view page would be nice to make the date more obvious
* [ ] if date_end != date_start, if possible it would look like:
```
12 - 16 mars 2016
26 mars - 3 avril 2016
27 décembre 2015 - 4 janvier 2016
```

The client asked if adding the weekday ("dimanche") would be (reasonably!) possible? Perhaps with momentjs?

c2corg_ui/static/js/documentediting.js
```diff
@@ -565,6 +573,43 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
...
+app.DocumentEditingController.prototype.openDateEnd = function() {
+  this.dateEndOpened = true;
+};
+
+
+/**
+ * @export
+ * @suppress {missingProperties}
+ */
+app.DocumentEditingController.prototype.formatDate = function(date) {
```
asaunier: I'm a bit reluctant with having document_type specific functions and properties in this controller, that should be kept as generic as possible. Should we consider using inheriting controllers when the generic editing controller is not enough?

Perhaps at least this ``formatDate`` function could be moved to ``app.utils``?

c2corg_ui/static/js/lang.js
```diff
@@ -102,6 +111,14 @@ app.LangController.prototype.updateLang = function(lang) {
     'path': '/',
     'expires': this.todayInOneYear_()
   });
+
+  if (lang === 'en') {
+    lang = 'en-gb';
+  }
+  $.get('/node_modules/moment/locale/' + lang + '.js', function() {
```
asaunier: :+1:

c2corg_ui/templates/helpers/view.html
```diff
@@ -146,6 +93,205 @@ <h4 translate>No maps associated</h4>
...
+     ## all activities are stacked one upon another and only the #target is displayed
+       <span class="route-activity icon-${activity}" uib-tooltip="{{ mainCtrl.translate('${activity}') }}" tooltip-placement="${tooltipPosition}"></span>
+    % endfor 
+  % endif
+</%def>
+
+
+
+<%def name="get_associated_documents(document, associatedDocuments)">\
+
```
asaunier: The indentation of the if/elif blocks looks a bit weird in this function.

c2corg_ui/static/js/documentediting.js
asaunier: > I'm a bit reluctant with having document_type specific functions and properties in this controller, that should be kept as generic as possible. Should we consider using inheriting controllers when the generic editing controller is not enough?

I have created an issue for this: https://github.com/c2corg/v6_ui/issues/246

Perhaps you could make the suggested change for ``formatDate`` anyway?

asaunier:
I have moved other change requests to dedicated issues https://github.com/c2corg/v6_ui/issues/245 https://github.com/c2corg/v6_ui/issues/246 https://github.com/c2corg/v6_ui/issues/247

Still 2 minors comments. Else :+1: for merging, thanks!

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/244

asaunier:
Personal comment field at step1 of the editing form:
* typo in field title
* the placeholder "write your comment" should disappear when the focus is on the field
* What is difference between this field and "description" at step 4?

asaunier:
@ginold I have (temporarly) neutralised your changes about saving the editing form content in the localstorage. See 58be436

I don't have time to fix the errors reported by the Closure compiler today and I think there are many things that should be improved before it is ready:
* depending if the user has checked "remember me" when authenticating, we should use ``sessionStorage`` instead of ``localStorage``. We should do something like in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/auth/authservice.js#L79 By the way should we introduce a common storage service that could be use for auth and documentediting?
* I think that when the localstorage is used to fill the editing form, it's no use making an API request to fill the form. Couldn't you check for the value of ``this.id_`` to figure out if the ids match?

I think it is better to make those changes in a new PR, taking the comments above into account. I will create an issue for this.

Issue #210, opened 2016-03-01
Menu layout and content


asaunier:
Layout: see Altimax

Content to add:
* banner block
* footer/disclaimer/partners block
* moving the interface lang selector to the menu (as in the wireframes) would make sense (the string "browse site in" is needed to help clarifying it's not the document lang version selector).
* what about moving the social networks links here as well?

asaunier:
The banner block should use a standard size. See http://blog.kinoa.com/wp-content/uploads/2013/06/format_publicite_internet_netoides1.jpg

Most likely 160x320.

What's the actual banner size in Altimax' layout?

ginold:
I use a standard size (16/9) already, it's just not 160x320px because our menu isnt' that wide max 250px).

asaunier:
We need to use an official banner size. See http://blog.kinoa.com/wp-content/uploads/2013/06/format_publicite_internet_netoides1.jpg for examples.

> 160x320px because our menu isnt' that wide (max 250px).

160px is the width here.

ginold:
*320x160 :) our is 240px on 135px 16/9

asaunier:
* [ ] The client told me that the logo of Fondation Petzl is not required. It will appear in some of the banners.
Than it can be removed from the menu.

asaunier:
* [ ] The URLs of Facebook and Twitter buttons are the ones of Camptocamp SA (G+ is OK). The correct URLs are https://twitter.com/camptocamporg and https://www.facebook.com/camptocamp.org/

asaunier:
* [ ] Having a clearer background-color (white?) for the simple search input would make it more visible and nicer.

Issue #234, opened 2016-03-13
Feedbacks on route editing


asaunier:
(more to come)

* [x] Icons for activities "escalade" and "rocher haute montagne" are switched
* [x] Icon for activity "mountain bike" is missing (at least in the demo)

ginold:
probably because of the corrected typo in 'moutain biking'

asaunier:
Same problem than for waypoint editing (see https://github.com/c2corg/v6_ui/issues/242#issuecomment-199018593) => fields groups are sometimes not relevant + too many steps.

Here is a recommendation (similar to WP editing):

* step 1: text attributes and map
 * title
 * lang
 * map (should be big enough to draw a route) + a file upload field to upload a GPS track
 * summary
 * description (big enough to type a lot of text)
 * remarks
 * gear
* step 2: figures
 * activities
 * "figures" attributes depending on the activities
* step 3: associations...
* step 4: preview

ginold:
i still don't know where to put : 

glacier_gear 
locales.slope
ski_exposition
hiking_mtb_exposition
lift_access

asaunier:
* [ ] attributes "route_configuration_types", "duration_types" and "glacier_gear" are not required attributes, whatever the activity, see https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py

* [ ] "Elévation min" => this attribute name or translation looks weird.

* [ ] Not sure what causes the error, but when editing an existing route, one gets the following error:
> "None" is not a mapping type: Does not implement dict-like functionality.

* [ ] "duration_types": this attribute (available in the editing form) does not exist in the model. The real name is "durations" and it's a multi-select field: https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/route.py#L64

asaunier:
Discussed with the client, here is a proposal for the route editing form structure:

### step 1

title + lang
activities

### step 2

map + upload GPX
associations of WP and routes

### step 3

(configuration => or maybe in step1 if that's too many items in step3)

orientations
route_types
route_configuration_types
(climbing_outdoor_types)
(rock_types)

("numbers" => we must find another name, "numbers" does not mean much)
altitudes
deniv
longueur

(ratings)
cotations ...
exposition ...
slope

### step 4

summary
description
remarks
gear + glacier_gear
route_history
external_resources (biblio) + associations livres

quality

### step 5

preview

ginold:
[...]

Issue #233, opened 2016-03-11
Use ?a=0 option to request data for edition


tsauerwein:
With https://github.com/c2corg/v6_api/pull/149 it is now possible to request data for the edition page without associations and area. E.g. `/waypoints/38575?a=0`.

fjacon:
Are we going to use it ?

asaunier:
Is this issue still valid?
Hasn't ``?a=0`` been replaced by ``?e=1``? I think I have used it in the editing directive.
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/apiservice.js#L207-L211

tsauerwein:
Yes right, it is `e=1` [now](https://github.com/c2corg/v6_api/blob/65241c6/c2corg_api/views/waypoint.py#L148-L150).

PR #261, opened 2016-03-30
lots of small fixes

ginold:
Lots of small fixes in this PR, but most noticeably :
- rearranged waypoint and route editing forms
- removed editButton directive -> moved the functionality to the main controller
- associated users/contributors are shown as participants (outing view)
- waypoint type icons (everywhere)
- HTML/angular card templates for outing, waypoint and routes
- main waypoint has a different border color (route details and edit view) and you can change the main waypoint (edit view)
- conditions_levels is displayed properly (outing details view)
- associated outings moved to the top and with a different border colour (to be disscussed?)
- "show more" button appears when there are 10 or more documents, but it's not functional yet (details view -> associations)  
- "add an outing" buttons (details view)
- you can set different dates (outing edit)
- badges are clickable and redirect to list-view (details view)


c2corg_ui/static/js/documentediting.js
```diff
@@ -559,6 +542,52 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
   }
 }
 
+app.DocumentEditingController.prototype.pushToArray = function(object, property, value, event) {
+  var pushed = app.utils.pushToArray(object, property, value);
+  var checkbox = $(event.currentTarget).find('input');
+
+  if (pushed) {
```
tsauerwein: This could be:
```
checkbox.prop('checked', pushed);
```

c2corg_ui/static/js/documentediting.js
```diff
@@ -559,6 +542,52 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
...
+    checkbox.prop('checked', true);
+  } else {
+    checkbox.prop('checked', false);
+  }
+}
+
+app.DocumentEditingController.prototype.pushToArray = function(object, property, value, event) {
```
tsauerwein: Am I missing something or is this function defined twice?

ginold: even 3 times! surely something went wrong when resolving conflicts. Thank's for telling.

config/agina~
```diff
@@ -0,0 +1,6 @@
+include config/dev
```
tsauerwein: Did you mean to commit this file?

asaunier:
I have not looked at the code yet.

Here are my comments about the UI:

> rearranged waypoint and route editing forms

Thanks.
It seems OK for the WP (but I have not tested all waypoint types).
For the routes it's not consistent with https://github.com/c2corg/v6_ui/issues/234#issuecomment-203324058 (updated this morning).

> associated users/contributors are shown as participants (outing view)

OK, 2 comments though:
* [ ] show contributors first and then the text attribute
* [ ] show the topo_name instead of the username => should we create a helper to centralize the way contributors links are rendered in any pages?

> waypoint type icons (everywhere)

Nice!
It seems "virtuel" has no icon. Should we have a default WP icon that could be use when no icon is available?

For instance http://c2corgv6-demo.gis.internal:6555/waypoints/734138/fr
It's not that a big deal in the detail view. But in the list page, it impacts the card height, shifting the cards vertically.

> HTML/angular card templates for outing, waypoint and routes

Cool

For WP cards, what about showing the WP_type icon floating to the right or to the left? The waypoint_type would appear only has tooltip when hovering the icon.

Thanks for changing the underlining when hovering the cards.
I think changing only the style of the title is a bit to restrictive. It would be cool for instance that the border color changes.

> main waypoint has a different border color (route details and edit view) and you can change the main waypoint (edit view)

OK cool. Not sure it will be obvious that this card is the main WP. Perhaps we should make it appear at the beginning of the list?

> conditions_levels is displayed properly (outing details view)

Nice!

> associated outings moved to the top and with a different border colour (to be disscussed?)

In a WP page, outings are not the most interesting associated documents => routes are. I think it is better to display the associated outings at the end of the page, in a separated block, more distinct.
The dark gray color is a bit sad.

> "show more" button appears when there are 10 or more documents, but it's not functional yet (details view -> associations)

Currently it is not obvious that the "show more" button at the bottom of an outings list is related to the list and to the associations below (except the color) => it is better to make the blocks more distinct.
outings_associations

> "add an outing" buttons (details view)

OK! Should it be displayed near the "show more" button at the bottom of the list? In addition having the button before the outings block make a white space appear at the beginning of the section.

I wonder if the "add an outing" and "show more" buttons should not be included in the flow of the cards? For instance "add an outing" at the beginning (a bit like a "here could be your outing" card) and "show more" at the end (a card that could represent the other cards available).

> you can set different dates (outing edit)

Is it already possible to save outings?

> badges are clickable and redirect to list-view (details view)

Nice!

Additional comments:

Route editing form:
* [ ] don't show the associated outings, there may be too many of them, making the form unsuable. Perhaps a change is needed in the API @tsauerwein ?

tsauerwein:
> don't show the associated outings, there may be too many of them, making the form unsuable. Perhaps a change is needed in the API @tsauerwein ?

This https://github.com/c2corg/v6_ui/issues/233?

asaunier:
> This #233?

Yes and no.
We would need the associations but not the outings associations :P

tsauerwein:
Ok, I see. I opened a ticket for this: https://github.com/c2corg/v6_api/issues/212

ginold:
- 'we should use topo_name' -> i don't see this property, it doesn't exist or it is just not returned?

tsauerwein:
The old field `topo_name` is now simply `name`.

c2corg_ui/static/js/documentediting.js
```diff
@@ -552,25 +535,40 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
...
 
-  if (pushed) {
-    checkbox.prop('checked', true);
+  checkbox.prop('checked', pushed);
+}
+
+/**
+ * @export
+ */
+app.DocumentEditingController.prototype.formatDate = function(date) {
```
asaunier: To move in ``app.utils``?
Date manipulating might be useful in the outings advanced search...

asaunier:
> HTML/angular card templates for outing, waypoint and routes

I see you have created ``c2corg_ui/templates/document/outing_card.html``, ``c2corg_ui/templates/document/route_card.html`` and ``c2corg_ui/templates/document/waypoint_card.html``.

It would be more consistent to have ``card.html`` templates in ``c2corg_ui/templates/waypoint``, ``c2corg_ui/templates/route`` and ``c2corg_ui/templates/outing``.

asaunier:
Commit b43b6851c397b826d91cdc18bc455b185823bcc8 "Temporarly neutralise localstorage from document editing" seems not related to myself and the message does not match what the commit actually does.

c2corg_ui/templates/document/route_card.html
```diff
@@ -1,6 +1,6 @@
 <a>
   <span class="list-item-title">{{route['locales'][0]['title']}}</span>
-  <span class="list-item-lang"> {{route['locales'][0]['lang']}}</span><br>
+  <span class="list-item-lang"> ({{route['locales'][0]['lang']}})</span><br>
```
asaunier: The lang info is actually needed only when the document's lang is not the current interface language.
Could we show it only in that case? For instance with
```ng-show="{{route['locales'][0]['lang']}} != <interface lang>"```
?

File: c2corg_ui/templates/outing/edit.html [hunk 6]
```diff
@@ -23,10 +23,11 @@ <h1 ng-cloak ng-init="id=${outing_id if outing_id else 0}" ng-bind="id ? ('Editi
   % endif 
     name="editForm" novalidate ng-submit="editCtrl.submitForm(editForm.$valid)" class="document-editing">
     
-    ## Init max steps. It won't work with ng-init
+    ## It won't work with ng-init
     <span class="ng-hide">
       {{editCtrl.max_steps = 5}} 
       {{document = outing}}
+      {{quality_types = [ 'stub', 'medium', 'correct', 'good', 'excellent']}}
```
asaunier: Hard coding values is evil!
It would be better to retrieve those values from ``c2corg_common/attributes.py`` even though it is not available in this file yet. I have open an issue in the API: https://github.com/c2corg/v6_api/issues/213

asaunier: Anyway I don't really understand why you need to set an angular expression for this?
It's a standard document attribute...

c2corg_ui/templates/helpers/view.html
```diff
@@ -108,26 +109,33 @@ <h4 translate>No maps associated</h4>
...
-        % endfor
-        <article class="list-item" ng-repeat="doc in added | filter:{'documentType': '${associatedDocuments}'}">
-          <app-association-card parent-id="${document['document_id']}" doc="doc" added-documents="added"></app-association-card>
-        </article>
-      </div>
-    </article>
+      <article>
+        <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#associated-${associatedDocuments}">
+          <span class="glyphicon glyphicon-map-marker"></span>
+          <span translate>Associated ${associatedDocuments}</span><span class="glyphicon glyphicon-menu-down"></span>
```
asaunier: You cannot use ``translate`` here since the string to translate contains a Mako variable: ``Associated ${associatedDocuments}``.

c2corg_ui/templates/helpers/view.html
```diff
@@ -108,26 +109,33 @@ <h4 translate>No maps associated</h4>
...
+          <div class="list-item">
+            ${route_card(route)}
+            <app-delete-association parent-id="${document['document_id']}" child-id="${route['document_id']}"></app-delete-association>
+          </div>
+          % endfor
+          <article class="list-item" ng-repeat="doc in added | filter:{'documentType': '${associatedDocuments}'}">
+            <app-association-card parent-id="${document['document_id']}" doc="doc" added-documents="added"></app-association-card>
+          </article>
+        </div>
+        % if len(document['associations']['routes']) == 10:
```
asaunier: I think we want to show all the associated routes (on the contrary to associated outings).

c2corg_ui/templates/helpers/view.html
```diff
@@ -173,10 +197,10 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
   % if text == 'weather':
```
asaunier: Using ``text`` as variable name is confusing since it's more a type than a description of what the variable is. ``attr`` or ``attr_name`` would have been clearer I think.

c2corg_ui/templates/helpers/view.html
```diff
@@ -221,72 +245,61 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
...
-    <a href="${request.route_url('outings_view', id=outing['document_id'], lang=outing['locales'][0]['lang'])}">
-      <div class="outing-author text-center">
-        <div class="outing-date">{{ '${outing['date_start']}' | date: 'MM/dd/yyyy'}}</div>
-        <span class="icon-user"></span><br>
-        <b class="author-name">${outing['author']['username']}</b>
+<%def name="outing_card(outing)">\
+  <a href="${request.route_url('outings_view', id=outing['document_id'], lang=outing['locales'][0]['lang'])}">
+
+    <div class="outing-author text-center">
+      <div class="outing-date">{{ '${outing['date_start']}' | date: 'MM/dd/yyyy'}}</div>
```
asaunier: Can't we use ``dd/MM/yyyy`` instead?

c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
```diff
@@ -122,10 +122,20 @@
...
         % endif
-
-        % if outing['locales'][0]['participants']:
-        <p><b>${outing['locales'][0]['participants']}</b></p>
-        % endif
+        
+        <p>
+          <b>
+            % if outing['locales'][0]['participants'] and outing['associations']['users']:
+              ${outing['locales'][0]['participants']},
```
asaunier: Displaying contributors before non-registered participants would make more sense.

c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
```diff
@@ -111,8 +111,8 @@
 
 ## PARTICIPANTS
 <%def name="get_outing_participants(outing)">\
-  % if outing.get('participant_count')  or outing['locales'][0]['participants']:
-    <div class="name-icon-value" ng-click="detailsCtrl.openTab($event)">
+  % if outing.get('participant_count')  or outing['locales'][0]['participants'] or outing['associations']['users']:
+    <div class="name-icon-value" ng-click="detailsCtrl.openTab($event)" ng-init='users = ${outing['associations']['users']}'>
```
asaunier: Do we really need to use angular expressions to display the users?

c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
```diff
@@ -160,3 +170,36 @@
...
+      <div class="view-details-info col-xs-12" ng-init="conditions = ${locale['conditions_levels']}">
+        <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#conditions_levels">
+          <span translate>conditions_levels</span><span class="glyphicon glyphicon-menu-right"></span>
+        </h3>
+        <section class="collapse" id="conditions_levels">
+          <div class="lead table-responsive">
+            <table class="table table-striped conditions-levels">
+              <thead>
+                <tr>
+                  <th class="location" translate>location / altitude / orientation / time</th>
```
asaunier: I think using separated translations here would help reducing the number of strings to translate:
```
<th class="location"><span translate>location</span> / <span translate>elevation</span> / <span translate>orientation</span> / <span translate>time</span></th>
```

c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
```diff
@@ -160,3 +170,36 @@
...
+                  <th class="soft-snow" translate>soft snow</th>
+                  <th class="total-snow" translate>total snow</th>
+                  <th class="comment" translate>comment</th>
+                </tr>
+              </thead>
+              <tbody>
+                <tr ng-repeat="condition in conditions">
```
asaunier: I think using Mako ``for`` loop is more appropriated than using Angular here. Python can also read JSON easily.
Golden rule: use Mako/python whenever it is possible, Angular only if you need advanced interactivity.

c2corg_ui/static/js/maincontroller.js
```diff
@@ -76,3 +91,17 @@ app.MainController.prototype.isPath = function(path) {
 app.MainController.prototype.animateHeaderIcon = function(e) {
   app.utils.animateHeaderIcon(e);
 }
+
+
+/**
+ * @export
+ */
+app.MainController.prototype.redirect = function(url) {
```
asaunier: The function name is not explicit enough if outside of the ``editButton`` controller.

asaunier:
> removed editButton directive -> moved the functionality to the main controller

Are you sure there are no place where this directive is used?
For instance https://github.com/c2corg/v6_ui/blob/2326403977d59da3da3d124ea828f2d981a9b978/c2corg_ui/templates/helpers/view.html#L48

You should also merge the 2 related commits: https://github.com/c2corg/v6_ui/pull/261/commits/d290989de588b725786abe8c699310ddcfe7cd3c and https://github.com/c2corg/v6_ui/pull/261/commits/1da2ae1ff448fa078bc1b44d5dcf8c04e8802314

c2corg_ui/templates/outing/edit.html
ginold: i'll leave it as is until we create this array in attributes.py

c2corg_ui/templates/helpers/view.html
ginold: so should i use x-translate?

asaunier: At least yes.

c2corg_ui/templates/helpers/view.html
ginold: 'unlike' or 'as opposed to' instead of 'on the contrary' :p

asaunier: Are you reviewing my review? :P

c2corg_ui/static/js/maincontroller.js
ginold: I'm not sure how you want to be more explicit about a 'redirect' function than calling it 'redirect' ? :kissing:

asaunier: It's not obvious that the function is used for protected URLs which require to be authenticated. You don't want to use it for public URLs.
Maybe ``redirectToUrlWhichRequiresToBeAuthenticated`` :P  or ``checkAuthAndRedirect`` or ```redirectToProtectedUrl``?

c2corg_ui/templates/helpers/view.html
ginold: sometimes i'm a grammar nazi

asaunier:
I have pushed a syntax error fix with 8411eff
@ginold could you please make sure this is the correct fix? If yes rebase this commit with some other.

asaunier:
This PR (including the last commits I have just pushed) fixes https://github.com/c2corg/v6_ui/issues/210

PR #262, opened 2016-03-30
Add MyAccount page

gberaudo:
When authenticated, allow changing:
- password;
- email (through validation);
- toponame / forum username.

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -310,4 +310,29 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
 };
 
 
+/**
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.readAccount = function() {
+  var promise = this.getJson_('/users/account');
+  promise.catch(function(response) {
+    this.alerts_.addError(response);
+  }.bind(this));
+  return promise;
+};
+
+
+/**
+ * @param {Object} data
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.updateAccount = function(data) {
+  var promise = this.postJson_('/users/account', data);
+  promise.catch(function(response) {
+    this.alerts_.addError(response);
+  }.bind(this));
+  return promise;
+};
+
+
 app.module.service('appApi', app.Api);
--- a/c2corg_ui/templates/account.html
+++ b/c2corg_ui/templates/account.html
@@ -27,9 +27,9 @@
           <input type="email" name="email" ng-model="account.email" class="form-control" />
         </div>
 
-        <div id="account-toponame-group" class="form-group" ng-class="{ 'has-error': accountForm.toponame.$touched && accountForm.toponame.$invalid }">
-          <label translate>toponame</label>
-          <input type="text" name="toponame" ng-model="account.toponame" class="form-control" />
+        <div id="account-name-group" class="form-group" ng-class="{ 'has-error': accountForm.name.$touched && accountForm.name.$invalid }">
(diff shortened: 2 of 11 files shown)


gberaudo:
Depend on https://github.com/c2corg/v6_api/pull/211.

c2corg_ui/static/js/apiservice.js
```diff
@@ -310,4 +310,29 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
 };
 
 
+/**
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
```
asaunier: Just curious: don't we need ``@export`` for this function and the others?

c2corg_ui/static/js/apiservice.js
```diff
@@ -310,4 +310,29 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
+ */
+app.Api.prototype.readAccount = function() {
+  var promise = this.getJson_('/users/account');
+  promise.catch(function(response) {
+    this.alerts_.addError(response);
+  }.bind(this));
+  return promise;
```
asaunier: Finally all the functions in the API service use the following code:
```
   promise.catch(function(response) {
     this.alerts_.addError(response);
   }.bind(this));
   return promise;
```
Shouldn't it be factorized somewhere?

c2corg_ui/templates/account.html
```diff
@@ -0,0 +1,40 @@
+<%inherit file="base.html"/>
+<%namespace file="helpers/common.html" import="show_title"/>
+<%block name="pagetitle"><title ng-bind="mainCtrl.page_title('Account')">${show_title('Account')}</title></%block>
+<br /><br /><br /><br />
```
asaunier: Is this the right way to add blank lines?

tsauerwein: Or do you maybe set a margin/padding in the CSS? ;)

c2corg_ui/static/js/apiservice.js
gberaudo: No since they should only be called from the controllers.

c2corg_ui/static/js/apiservice.js
gberaudo: I considered it but decided against it since it would not gain much and is more readable this way.

c2corg_ui/templates/account.html
gberaudo: Right, the appearance would need some love.
The final design of the page is not yet decided so I just did the minimum to avoid the floating bar at the top.
I will add some temporary CSS then.

c2corg_ui/static/js/apiservice.js
asaunier: But then they are public methods anyway since called outside of the class?

c2corg_ui/static/js/apiservice.js
asaunier: Well, I can count this exact piece of code 12 times in the file :P

c2corg_ui/static/js/apiservice.js
gberaudo: Yes.

asaunier: Then why not tagging them as ``@export``?

gberaudo: Being public does not mean they are not or should not be minified. It is an independant concept.

asaunier: OK

asaunier:
Is this PR ready for merging?

gberaudo:
Finally got ride of the brs!
I also added to commits fixing bugs related to user management.

PR #265, opened 2016-04-05
make remember me the default option

ginold:
set's "remember me" checkbox by default + some styling.

Fixes https://github.com/c2corg/v6_ui/issues/259

### DIFF
--- a/c2corg_ui/templates/auth.html
+++ b/c2corg_ui/templates/auth.html
@@ -4,105 +4,141 @@
 
 <div app-auth class="auth-form">
 
-  <div ng-show="authCtrl.uiStates.showLoginForm">
-    <h1 translate>Login</h1>
+  <section>
+    <div ng-show="authCtrl.uiStates.showLoginForm">
+      <h1 translate>Login</h1>
 
-    <form name="loginForm" novalidate ng-submit="authCtrl.login()">
-      <div id="login-username-group" class="form-group" ng-class="{ 'has-error': loginForm.username.$touched && loginForm.username.$invalid }">
-        <label translate>Username</label>
-        <input type="text" name="username" ng-model="login.username" class="form-control" required />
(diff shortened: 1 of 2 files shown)


gberaudo:
LGTM

ginold:
LGBT

asaunier:
It would have been easier to read/review to have splited the changes in 2 commits, 1 for the "remember me" default status, 1 for the styling.
Currently it is very hard to figure out where this change about "remember me" is done...

ginold:
to be honest, the only line that would make "remember me" default is ` <input type="checkbox" name="remember" ng-model="login.remember" ng-checked="true">`, the rest is styling.

asaunier:
> to be honest, the only line that would make "remember me" default is <input type="checkbox" name="remember" ng-model="login.remember" ng-checked="true">, the rest is styling.

Yep, that's why I think it should have been done a commit, the rest in a second "styling" one.
Looking a your PR, it's not obvious where the change "make remember me the default option" is done.

gberaudo:
> LGBT

@ginold what do you mean?

asaunier:
OK for the "remember me" change.

Concerning the styling changes, they look fine except that:
* it would make more sense to place the "no account yet?" button right next buttons "Login", "Forgot password?" (and not in some deported place to the right) => all options are at the same place, easy to see and use.
* why using a different text styling for "no account yet?" and "Forgot password?" compared to "Login"?
* the showing of the various forms ("login", "register", "reset password") is not consistent => depending on what buttons you clicked and in what order you see different combinations of forms. For instance if you click on "Forgot" then on "No account?" you see the register + reset forms => what's the point? I think the "reset" form should only be displayed when the user clicks on "forgot" and be hidden when clicking on "no account" or "have an account".

asaunier:
Thanks for the changes!
Looks good to me.

asaunier:
Feel free to merge @ginold thanks!

Issue #245, opened 2016-03-23
Date formating for outings


asaunier:
I had some precision from the client for outings dates (can be managed in a separated PR if you prefer):
* [x] date to show in outings lists is ``date_end``
* [x] showing the date as subtitle of outings detail view page would be nice to make the date more obvious
* [x] if date_end != date_start, if possible it would look like:
```
12 - 16 mars 2016
26 mars - 3 avril 2016
27 décembre 2015 - 4 janvier 2016
```

The client asked if adding the weekday ("dimanche") would be (reasonably!) possible? Perhaps with momentjs?

ginold:
it is possible but i don't see having such a long date. lundi 27 décembre 2015 - 4 vendredi janvier 2016, longer than the title :P

tsauerwein:
Maybe show the short format and set the long format in the title attribute, so that if you hover the data you can see the weekdays.

asaunier:
Fixed by #266

Issue #243, opened 2016-03-21
How to display ratings in route pages


asaunier:
Here is a proposal from the client. There are still blur points. I have asked the client to make changes and will update the scheme below as soon as possible:

```
The ratings are organized by block. Each block depends on one or several activities. A block is showed if at least one of its activities is selected in the outing.
A space to showed is coded as #. The blocks are concatenated with 2 space as separator.

Block A : snow_ice_mixed, mountain_climbing, rock_climbing, ice_climbing, via_ferrata
global_rating ## rock_free_rating>rock_required_rating[A0] # aid_rating # ice_rating # mixed_rating # via_ferrata_rating ## engagement_rating # risk_rating ## equipment_rating # exposition_rock_rating

- ">rock_required_rating" is showed only if : rock_free_rating is filled, and rock_free_rating is > rock_required_rating.
- [A0] (without bracket) is showed only if equipment_rating = P1 or P1+, and if aid_rating is empty.

Block B : skitouring
ski_rating/ski_exposition # labande_global_rating/labande_ski_rating

Block C : hiking
hiking_rating

Block D : snowshoeing
snowshoe_rating

Block E : mountain_biking
mtb_up_rating # mtb_down_rating

Block F : hiking, mountain_biking
hiking_mtb_exposition
```

asaunier:
Examples:

Cascade et NGM :             D  4+ M1                   II X1    P4 
Rocher  et escalade :        D  6a>5cA0 A1           II X1    P4 E1       (le A1 correspond à l'artif obligatoire) 
Rocher + Cascade  :         D  6a>5cA0 A1 4+ M2  II X1    P4 E1

ginold:
updated, should be ok now.

ginold:
can be closed?

asaunier:
Of course since the PR is merged.

asaunier:
Reopening the issue since it seems there are still problems according to https://github.com/c2corg/v6_ui/issues/433#issuecomment-226810272

stef74:
@shortoland please test

shortoland:
@stef74 

I can still see some problems:

- I don't think blocks are separated by 2 spaces.
- For a ski and hiking route, hiking rating is placed in between ski ratings: pb ?
- Block A: order is not respected
- Rock_climbing: rock_free_rating is displayed twice

ginold:
could you give an example where `rock_free_rating` is displayed twice in a `rock_climbing` route?

[...]

Issue #242, opened 2016-03-20
Feedbacks on waypoint editing


asaunier:
* [x] Yes/no attributes (eg. "gas unstaffed") should also be set to none (if the info is not available) => the radio buttons should be uncheckable. Perhaps use a "exclusive" checkboxes that supports 0 or 1 value.
* [x] paragliding_ratings: this is not a quality rating based on stars :P Please use a standard dropdown with values 1 to 5.

asaunier:
Concerning the organisation of the attributes (steps), it is currently a bit complicated and not intuitive. Probably too many steps.

Here is a recommendation:
* step 1: text fields and map
 * title
 * lang
 * map
 * long/lat/elevation
 * summary
 * description
 * map info (with auto-associated maps list)
* step 2: figures
 * waypoint_type
 * other "figures" attributes (including orientation etc.)
* step 3: preview before saving

ginold:
I did it, but putting everything in figures makes the form very long. You forgot about contact, access, transports, ratings...
But if it's better that way :no_mouth:

ginold:
isn't a 'Rating' most commonly represented with stars ? If it's not a special value like those letters in other ratings why not use stars ? 5 = 5 stars, it's very speaking :star: :star: :star: :star: :star: 
Or tell me that users will not get what the stars are and are so used to a scale 1-5, then ok :P

tsauerwein:
It's more about the difficulty of a route. For the real *rating* there is a field `awesomeness` for outings. That's where we can use stars.

ginold:
that's exactly what I did, this time i would not let it go if it weren't stars :gun:  
got it!

asaunier:
> isn't a 'Rating' most commonly represented with stars ? 

That's not a quality rating. It's a difficulty indicator. A paragliding takeoff with difficulty 5 is very difficult, not very cool :P

ginold:
whoa so difficult it must be awesome hardcore!

asaunier:
> I did it, but putting everything in figures makes the form very long. You forgot about contact, access, transports, ratings...
But if it's better that way

Don't make too much changes about this. It's still discussed by the client. I'm urging them to make a choice.

asaunier:
Errors reported about the WP editing form:

* [ ] Type "canyon" is missing in https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L17-L55

Climbing indoor:
[...]

PR #266, opened 2016-04-05
All view fixes 2

ginold:
The second version of all-view-fixes, the previous pr. I took under account the most recent clients' and your feedback.  Now everything should be fine. Should be...but will surely be not, so i'm awaiting your lovely comments so I can make it the perfect version!

Fixes https://github.com/c2corg/v6_ui/issues/210
Replaces https://github.com/c2corg/v6_ui/pull/261


asaunier:
If this PR replaces https://github.com/c2corg/v6_ui/pull/261 can we close the latter PR?

asaunier:
Does this PR addresses all wishes from https://github.com/c2corg/v6_ui/issues/245 ?

asaunier:
What about the ratings shown in the route detail view (or in the cards):
https://github.com/c2corg/v6_ui/issues/243

ginold:
I didn't address ALL wishes as they were still not confirmed or 'blurry'. for example do we want to use dates as `12 - 16 mars 2016
26 mars - 3 avril 2016
27 décembre 2015 - 4 janvier 2016` or numeric format only?
as for #243, you say 'There are still blur points. I have asked the client to make changes and will update the scheme below as soon as possible.' so I didn't wan't to do something, that is not sure.

asaunier:
> as 12 - 16 mars 2016 26 mars - 3 avril 2016 27 décembre 2015 - 4 janvier 2016 or numeric format only?

We have to be consistent with the one-day outings that show "4 janvier 2016". If we have 2 dates, we should at least show them the same way.

asaunier:
This comment from the other PR about the removed editButton directive seems still valid:
https://github.com/c2corg/v6_ui/pull/261#issuecomment-203848026

asaunier:
Thanks for making the hovering of cards much more obvious :)
The way to mark and change the main waypoint for routes seem nice. Perhaps the "main waypoint" mention could be a little less raw :P 

* [ ] Icons for activities rock_climbing and mountain_climbing are once again switched.

Editing forms in general:
* [ ] it should be possible to provide no value for a non required single-choice dropdown field. As for now, if the user selects a value, he cannot undo it and choose to leave the field blank. See for instance "rain_proof" or "children_proof" for waypoint>climbing_outdoor.

Waypoint editing (see also https://github.com/c2corg/v6_ui/issues/242):
* [ ] When selecting "summit" as waypoint_type, the steps bar shows a "figures" that is not actually displayed (neither needed)
* [ ] step "comments" => this title is a bit confusing since it is also possible to make comments about a document. "description" as the section title is probably clearer.
* [ ] climbing_indoor: numbers and ratings sections could be merged
* [ ] Contact>URL: is @ the right symbol for a website URL? I think many people will think it's the email address that is expected!
* climbing_outdoor:
 * [ ] "best periods" should be displayed near "orientations" (those are info helping people to figure out if it is possible to climbing here at a given date and weather).
 * [ ] the section title "transport & Access" is not relevant since it contains access time (by foot!), rain proof and children proof! It's better to simplify and to move all those fields to the first section ("Types and styles").
 * [ ] as for climbing_indoor, numbers and ratings sections could be merged
 * [ ] text fields "access" and "access_periods" are missing in the "associations & comments" section.
 * [ ] Again the latter step could be simply renamed "descriptions"
* hut:
 * [ ] "custodianship" is not a text field, it's a yes/no/null attribute (like "matress_unstaffed" for instance): https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L142
 * [ ] "custodianship" should be moved in the "numbers" section, before "capacity"
* access:
 * [ ] "Access" should be in the last steps with the other text fields
* paragliding takeoff:
 * [ ] slope => in ° not in m
 * [ ] title "Types & styles" does not mean anything in step 2 since it only contains an orientation field! All fields of this step could be gathered in the same section ("figures")

Routes editing (see also https://github.com/c2corg/v6_ui/issues/234):
* [ ] "lift_access" should rather be in the "numbers" section (with the elevations and height diffs) than in "ratings"

Outings editing (see also https://github.com/c2corg/v6_ui/issues/263):
* step1 => the order is quite different from https://github.com/c2corg/v6_ui/issues/263 ...
 * [ ] "general informations" and "route" sections should be merged and we don't need an "associations" section since it is already managed by the "route_taken" field (could be more than 1 route!)
 * [ ] "figures" such as elevation min/max, length etc. should come at the end of the section (might depend on the associated routes)
 * [ ] "partial trip": yes/no is enough (people do know if they have followed the whole route or not => no need for "no info")
 * [ ] move the map near the top of the section, say below the activities.
 * [ ] move the upload GPS track button above the map
* step2
 * [ ] "End date?" is not explicit enough for the date button => "Several days?" would be clearer
 * [ ] "duration_difficulties" => you cannot reuse the "duration_types" of the routes. Anyway this field will probably be dropped soon (being discussed) => simply remove it.
 * [ ] "condition_levels" is not text field => it should be a table with several columns and possibly several lines. See v5
 * [ ] "snow_quantity"/"snow_quality" => those fields are not integer fields! See https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/outing.py#L76-L78
 * [ ] Fields in "snow" should go to "conditions".
 * [ ] "avalanches_info" should only appear if "avalanche_sign" is set and not "no"
 * [ ] "Access" section => the title is not relevant since it also contains hut info => "Access & hut"?
* step3
 * [ ] remove field "summary"
 * [ ] "outing name" => "outing title"

ginold:
> climbing_indoor: numbers and ratings sections could be merged
this cannot be done or implies code duplication. I can't merge sections depending on waypoint type.

asaunier:
Outing detail view:
* [ ] no need to bold the participants text field, it appears too loud now
* [ ] title: the badge should be after the date (the date is somehow part of the title).

Route detail view:
* [ ] Isn't the "more outings" button missing when there are more than 10 outings?

c2corg_ui/templates/outing/edit.html
```diff
@@ -23,19 +23,19 @@ <h1 ng-cloak ng-init="id=${outing_id if outing_id else 0}" ng-bind="id ? ('Editi
   % endif 
     name="editForm" novalidate ng-submit="editCtrl.submitForm(editForm.$valid)" class="document-editing">
     
-    ## Init max steps. It won't work with ng-init
+    ## It won't work with ng-init
     <span class="ng-hide">
-      {{editCtrl.max_steps = 5}} 
+      {{editCtrl.max_steps = 4}} 
       {{document = outing}}
+      {{quality_types = [ 'stub', 'medium', 'correct', 'good', 'excellent']}}
```
asaunier: I think it's not possible to keep this as is. The list value cannot be hard-coded.
At least we will have to create an issue and fix this when https://github.com/c2corg/v6_common/issues/16 is done.

c2corg_ui/templates/outing/edit.html
```diff
@@ -81,22 +59,151 @@ <h2 class="heading show-phone">
...
+              </div>
+
+          </div>
+        </section>
+        
+        <section class="section associations">
+          <h2 class="heading show-phone"><span translate>associations </span></h2>
+          <div class="content route-associations" id="title-edit">
+            <app-add-association parent-id="document.document_id" added-documents="added"></app-add-association>
+            <%include file="../document/associated_routes.html"/>
```
asaunier: Move to ``templates/routes/associated.html`` instead? Ditto for outings and waypoints.

asaunier:
Else the code looks good to me, thanks for having addressed my comments in https://github.com/c2corg/v6_ui/pull/261

asaunier:
> Else the code looks good to me

Well it seems there are problems with the ``translate`` directives since there are many "illegal" msgids in the PO files:

```
#: c2corg_ui/templates/helpers/view.html
msgid "${attr}"
msgstr ""

#: c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
msgid "${outing['condition_rating']}"
msgstr ""

#: c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
msgid "${outing['duration_difficulties']}"
msgstr ""

#: c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
msgid "${outing['frequentation']}"
msgstr ""

#: c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
msgid "${outing['partial_trip']}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_down'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_max'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_min'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_up'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop}"
msgstr ""
```

asaunier:
* [ ] Could it be possible to change the mouse cursor when hovering the activities icons in the editing forms?

* [ ] The outing add page makes a Mako crash:
> TemplateLookupException: Can not locate template for uri 'c2corg_ui:templates/outing/../document/associated_waypoints.html'

* [ ] In waypoint > paragliding, there is a "configuration" title that appears to the right of the section header  and "Styles & Types" to the left...

ginold:
updated

asaunier:
Thanks for updating!

Outings:
* [ ] "routeS taken": should be restricted to routes (no use to return WP)
* [ ] "condition_levels" should only appear for activities "ski", "snow, ice, mixed", "ice_climbing" and "snowshoeing"
* [ ] "condition_levels" a first line with input should appear by default
* [ ] "condition_levels" button "Add Row" has no effect$
* [ ] "associated_participants" => we need to have a user association tool (as for routes) but restricted to users

asaunier:
Concerning the quality field, you may ignore my comments. I will create an issue about this (we most likely have several things to do about it).

asaunier:
* [ ] activity "paragliding" is missing in the outing editing form (should not appear in route editing form)
* [ ] icon "paragliding" is not displayed in the detail view of outings such as http://c2corgv6-demo.gis.internal:6555/outings/736815/fr (which mentions "paragliding" in the activities text field)

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -68,6 +90,47 @@ app.ViewDetailsController.prototype.openModal = function(selector) {
   this.modal_.open({animation: true, size: 'lg', template: template});
 };
 
+
+/**
+ * @param {Date} dateStart
+ * @param {Date} dateEnd
+ * @suppress {checkTypes} it told (found scope | null, required -> scope) at compile()(this.scope)...
```
asaunier: Is it the right way to get rid of closure complaints? :P

ginold: to just 'get rid of', probably yes :P apparently it works, but if someone has a better solution, it will be welcome. 
I've also got a problem with `app.ViewDetailsController.prototype.displayDate` at `if (this.firstTime_)` - read the comment.

asaunier:
* Make sure the association searches only query ES for the needed types:
 * [ ] Outings: routes for the routes association tool, users for the participants association tool
 * [ ] Routes: WP and routes
 * [ ] WP: WP and routes

c2corg_ui/static/js/maincontroller.js
```diff
@@ -76,3 +91,16 @@ app.MainController.prototype.isPath = function(path) {
 app.MainController.prototype.animateHeaderIcon = function(e) {
   app.utils.animateHeaderIcon(e);
 }
+
+/**
+ * @export
+ */
+app.MainController.prototype.redirectToProtectedUrl = function(url) {
```
asaunier: Thanks :)

c2corg_ui/static/js/search.js
```diff
@@ -259,4 +256,18 @@ app.SearchController.select_ = function(event, doc, dataset) {
...
+
+/**
+ * @param {string} dataset
+ * @private
+ */
+app.SearchController.prototype.createDatasets_ = function(dataset) {
+  if (!dataset) { // default
+    return  [this.createDataset_('waypoints', 'Waypoints'), this.createDataset_('routes', 'Routes')];
+  } else {
+    return [this.createDataset_(dataset, dataset.charAt(0).toUpperCase() + dataset.slice(1))];
```
asaunier: Shouldn't we use the same case on both side of the search tool (client vs API)?

c2corg_ui/templates/helpers/view.html
```diff
@@ -15,9 +15,9 @@
       <p><b translate>Changes</b>: <span x-translate>${show_version_comment(version)}</span></p>
     </div>
     <p class="text-center">
-        <a href="${request.route_url(module + '_view', id=document['document_id'], lang=locale['lang'])}">
-          <button class="btn btn-default" translate>See the latest version</button>
-        </a>
+      <a href="${request.route_url(module + '_view', id=document['document_id'], lang=locale['lang'])}">
+        <button class="btn btn-defaul\t" translate>See the latest version</button>
```
asaunier: Are you sure of ``btn btn-defaul\t``?

c2corg_ui/templates/helpers/view.html
```diff
@@ -85,9 +85,10 @@ <h4 translate>No maps associated</h4>
 
 <%def name="show_areas(document)">\
   % if document.get('areas'):
+  <% areas = document.get('areas') %>\
```
asaunier: It is clearer to indent all the content of the ``ìf`` block, for reading reasons.

asaunier:
When compiling I get the following error:

> ERR! compile /home/alex/Bureau/v6_ui/c2corg_ui/static/js/documentdetailsview.js:57: WARNING - parameter amMoment does not appear in app.ViewDetailsController's parameter list
ERR! compile app.ViewDetailsController = function($uibModal, $compile, $scope) {

ginold:
retry now

asaunier:
* Outing editing form:
 * [ ] Why not using cards (or...) to list the associated participants in the outing editing form? As far as I can see, you list them in a text field that can be edited. What if we made a mistake and want to remove an associated participant?
 * [ ] The cross near the "end_start" field does not react.
 * [ ] I think "avalanches_infos" should only appear when "avalanche_sign" is not empty or "no"

* Outing detail view:
 * [ ] is it really relevant to show the condition_levels fields in a dedicated block? It should rather be moved to "weather and conditions". (the rendering is cool anyway!) 
 * [ ] could you rename the "description" block into "personal comments" (as in the editing form)

* Route detail view:
 * [ ] Inserting the "add outing" and "show more outings" buttons in the outing cards flow is cool. Just find it not consistent that the card reacts when hovering but that it does not trigger the action as when clicking on the buttons themselves.
 * [ ] Could you make the "show more outings" button open the outing lists (later it will add a parameter to filter the shown outings)?
 * [ ] Ratings are not shown at all!? Edit: ratings appear with some activities, but not for skitouring for instance. For instance http://localhost:6555/routes/263430/fr vs

asaunier:
Maybe already OK:
* [ ] The quality field should default to "medium" (level 2) for new documents. Documents migrated from v5 will receive a default "correct" quality (level 3). See https://github.com/c2corg/v6_ui/issues/272 and https://github.com/c2corg/v6_api/issues/220

asaunier:
* [ ] Please note that some waypoint_types will be dropped (and the icons as well even though we might recycle some of them for another type) See https://github.com/c2corg/v6_common/issues/15

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -68,6 +89,47 @@ app.ViewDetailsController.prototype.openModal = function(selector) {
...
+ * @export
+ */
+app.ViewDetailsController.prototype.displayDate = function(dateStart, dateEnd) {
+
+  //angular executes this function so many times it tells 'Error: 10 $digest() iterations reached. Aborting' so
+  // with this quick n dirty method there is no error. Any idea how to do it in the pro way ?
```
asaunier: We have to find why this function is called many times!
It seems to be called only once in this expression: 
https://github.com/c2corg/v6_ui/blob/all-view-fixes-2/c2corg_ui/templates/outing/view.html#L31
```
<span class="outing-date" class="ng-cloak">{{detailsCtrl.displayDate('${outing['date_start']}', '${outing['date_end']}') }}</span>
```
Can't we have a one-time binding?
Anyway do we really need angular here? Why not using a Mako def to do this in python?

tsauerwein: > We have to find why this function is called many times!

This is the normal behavior for bound values. In its digest cycle Angular checks if the value has changed. Because this function returns a new object on every call, Angular assumes that something has changed and runs the digest again. You could cache the object that you are returning and only create a new one when dateStart or dateEnd have changed. One-time binding could also work.

Maybe it would be even better to create a separate directive for this part. This would make it easier to reuse it elsewhere.

asaunier: Actually I think this part could be done in Mako and that we don't need angular (except for translating the dates using "amDateFormat" filters).
I'm currently trying to port the function in a Mako def.

c2corg_ui/static/js/documentdetailsview.js
gberaudo: You can not never rely on `@suppress {checkTypes}` to silent the compiler.
The only appropriate way is to fix the underlying issue:
```diff
diff --git a/c2corg_ui/static/js/documentdetailsview.js b/c2corg_ui/static/js/documentdetailsview.js
index 35857ce..2d39c1e 100644
--- a/c2corg_ui/static/js/documentdetailsview.js
+++ b/c2corg_ui/static/js/documentdetailsview.js
@@ -47,7 +47,7 @@ app.module.directive('appViewDetails', app.viewDetailsDirective);
 
 /**
  * @param {Object} $uibModal modal from angular bootstrap
- * @param {angular.Scope} $scope Scope.
+ * @param {!angular.Scope} $scope Scope.
  * @param {angular.$compile} $compile Angular compile service.
  * @constructor
  * @export
@@ -67,7 +67,7 @@ app.ViewDetailsController = function($uibModal, $compile, $scope) {
   this.compile_ = $compile;
 
   /**
-   * @type {angular.Scope}
+   * @type {!angular.Scope}
    * @private
    */
   this.scope_ = $scope;
@@ -93,7 +93,6 @@ app.ViewDetailsController.prototype.openModal = function(selector) {
 /**
  * @param {Date} dateStart
  * @param {Date} dateEnd
- * @suppress {checkTypes} it told (found scope | null, required -> scope) at compile()(this.scope)...
  * @export
  */
 app.ViewDetailsController.prototype.displayDate = function(dateStart, dateEnd) {
@@ -125,7 +124,7 @@ app.ViewDetailsController.prototype.displayDate = function(dateStart, dateEnd) {
                 ' - <span class="date-locale">{{\' ' + dateEnd.toISOString() + ' \'  | amDateFormat:"Do MMMM YYYY" }})</span>')(this.scope_);
       }
     }
-    $('.outing-date').append(dateHTML);
+    $('.outing-date').append(dateHTML.get());
   }
 };
 
diff --git a/c2corg_ui/static/js/documentediting.js b/c2corg_ui/static/js/documentediting.js
index 1ca4076..cbc2f67 100644
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -199,7 +199,6 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 /**
  * @param {Object} response Response from the API server.
  * @private
- * @suppress {checkTypes}
  */
 app.DocumentEditingController.prototype.successRead_ = function(response) {
   var data = response['data'];
```

asaunier: Do your changes fix the "suppress" problem? Including in ``documentediting.js``?
As said in an other comment, we might not need function ``displayDate`` so it's not worth spending time fixing it, but would be cool for the other.

gberaudo: It fixes both issues. Feel free to keep the one you need. ^^

asaunier:
It seems there are incorrect "translate" directives in c2corg_ui/templates/helpers/view.html

```
#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_down'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_max'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_min'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop + '_up'}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${prop}"
msgstr ""

#: c2corg_ui/templates/helpers/view.html
msgid "${text}"
msgstr ""
```

asaunier:
Except the "translate" issue reported above, the code looks good and the pages seem to work fine. So for me it's OK for merging when the "translate" issue is addressed.

Issue #254, opened 2016-03-27
Navigating in the guidebook


asaunier:
I had several feedbacks saying that making the "guidebook" menu pointing to the routes list page and that having then to select an entry in the document types dropdown is not that intuitive.
In addition it makes an extra request to the API to show the list of routes even though the user is trying to reach the list of WP or outings.

We need to find a solution. At least 2 possibilities:
- replace the guidebook "home page" by a somehow static page asking "what are you looking for" and showing buttons towards the routes, WP, outings, images, books, ...
- making the later item directly available in the menu (but not as first level entries - even though having "outings" in the main menu would make sense since it's rather a "live conditions" section whereas the routes/WP/books are a kind of knowledge database). A bit like post.ch ?
- ???

The aim is to avoid having to open several pages to find what you want in the guidebook and to make the guidebook structure (WP, routes, outings, ...) more obvious.

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/244

fjacon:
@ginold @asaunier where do we stand for this issue ?

tsauerwein:
> replace the guidebook "home page" by a somehow static page asking "what are you looking for" and showing buttons towards the routes, WP, outings, images, books, ...

Yes, I agree that it would very useful to have a  "home page" for the guidebook. Maybe something similar to http://www.outdooractive.com/en/guide/ where you have a list of activities, areas/regions, waypoint types, ... which link to a search for routes, waypoints, ...

And a quick search "I'd like to go climbing/hiking/... in ..." like [here](http://www.outdooractive.com/en/) would also be nice.
![outdooractive-search](https://cloud.githubusercontent.com/assets/266474/14554750/21171808-02ef-11e6-8193-dfd0a3f7cef1.png)

asaunier:
Some people at the association could not find the way to open the advanced search pages for WP or outings (default is routes) => the current dropdown to choose the document type seems not intuitive enough.

ginold:
don't forget to have a look at #267

fjacon:
To be discussed after alpha 3

stef74:
@asaunier @fjacon can we close ?

tsauerwein:
I am preparing a proposition on how this "homepage for the guidebook" could look like. I will share something in the next days.

asaunier:
Anyway I think this issue is still relevant and should not be closed.

tsauerwein:
Something like this:

[...]

PR #267, opened 2016-04-05
guidebook nav

ginold:
here is a quick 'n dirty "proof of concept" (thus please don't mind all the displaying bugs) referring to the guidebook navigation issue #254 . 

I've made two options, you will have to un/comment the appropriate parts in sidemenu.css (.submenu) and sidemenu.html (li -> topoguide), they are titled with FIRST and SECOND captions so you should not miss them.

So, you will get an idea of what i thought of, maybe we can implement one of those options, or think of another one.

### DIFF
--- a/c2corg_ui/templates/sidemenu.html
+++ b/c2corg_ui/templates/sidemenu.html
@@ -15,12 +15,62 @@
         <span class="menu-text" translate>Home</span>
       </a>
     </li>
+    ## FIRST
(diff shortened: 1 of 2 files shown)


tsauerwein:
Would you mind sharing a screenshot?

tsauerwein:
Is this PR still relevant?

asaunier:
Yes, the question is still open.

tsauerwein:
Related to https://github.com/c2corg/v6_ui/issues/756, please make sure to use relative urls.

Issue #202, opened 2016-02-26
Support geometry for route/outing documents


asaunier:
- [ ] Show geometry in list pages
- [x] Show line geometry in route view page
- [x] Enable geometry modification in map
- [x] Send updated geometry to API when saving document

PR #273, opened 2016-04-07
Show line geometry in routes and outings detail views

asaunier:
An improvement would be to show the name of the line near the mouse cursor when it hovers the line (as for waypoints).

Related to https://github.com/c2corg/v6_ui/issues/202

### DIFF
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -73,6 +73,16 @@ app.MapController = function($scope, mapFeatureCollection) {
   this.vectorLayer_ = null;
 
   /**
+   * @type {Object.<string, ol.style.Icon>}
+   */
+  this.iconCache = {};
+
+  /**
+   * @type {Object.<string, ol.style.Style|Array.<ol.style.Style>>}
+   */
+  this.styleCache = {};
+
+  /**
    * @type {?app.DocumentEditingController}
    * @private
    */
@@ -247,14 +257,6 @@ app.MapController.prototype.getVectorLayer_ = function() {
  * @private
  */
 app.MapController.prototype.createStyleFunction_ = function(scale) {
-  /**
-   * @type {Object.<string, ol.style.Icon>}
-   */
-  var iconCache = {};
-  /**
-   * @type {Object.<string, ol.style.Style|Array.<ol.style.Style>>}
-   */
-  var cache = {};
   return (
       /**
        * @param {ol.Feature|ol.render.Feature} feature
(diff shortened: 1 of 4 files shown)


c2corg_ui/static/js/map/map.js
```diff
@@ -242,71 +253,114 @@ app.MapController.prototype.getVectorLayer_ = function() {
 
 
 /**
- * @param {number} scale
+ * @param {boolean} hilight
```
tsauerwein: highlight

c2corg_ui/static/js/map/map.js
```diff
@@ -242,71 +253,114 @@ app.MapController.prototype.getVectorLayer_ = function() {
...
-              textBaseline: 'middle'
-            });
-          }
-
-          style = new ol.style.Style({
-            image: icon,
-            text: text
-          });
-          cache[key] = style;
+        // TODO: rename to document_type?
```
gberaudo: Misplaced comment?

asaunier: Why? It's related to the next line, I was wondering if using "document_type" (or "documentType") instead of "module" would not make sense.

c2corg_ui/static/js/map/map.js
asaunier: Thanks, fixed.

asaunier:
By the way, we will have to show on the map as well:
* for routes: the associated WP 
* for WP: the associated WP and routes 

but those features are not directly available. I guess we will have to make an additional API request (with a GeoJSON response?). Could be done in another PR (along with the title stuff mentioned above), in this PR I only do the essential.

tsauerwein:
> I guess we will have to make an additional API request (with a GeoJSON response?).

See https://github.com/c2corg/v6_api/issues/184

c2corg_ui/static/js/map/map.js
gberaudo: OK, thought you might be referring to the top of the removed block:
`var type = /** @type {string} */ (feature.get('type'));`.

`Module` sounds good, it is the name of the feature property.

asaunier:
OK it would ease displaying associated WP on the map!
But it won't work for associated routes (or maybe we don't show the geometries of associated routes?).

c2corg_ui/static/js/map/map.js
asaunier: OK, I have removed the comment then.

asaunier:
OK for merging?

I also plan to do the same for the editing forms (+ add GPX upload and track editing). But most likely in another PR.

gberaudo:
+1

PR #276, opened 2016-04-08
Map svg icons

ginold:


asaunier:
That looks good, thanks.
Perhaps we could remove the old PNG icons?
https://github.com/c2corg/v6_ui/tree/map-svg-icons/c2corg_ui/static/img/icons

Issue #280, opened 2016-04-11
Waypoint detail view


asaunier:
* [x] For "access" waypoints, the ``access period`` attribute is displayed in the "figures" section, which does not make sense since it is a text attribute that can potentially be quite long. See for instance http://c2corgv6-demo.gis.internal:6555/waypoints/735236/en => this attribute (and all other text attribute) should be displayed in the usual text section along with description.

asaunier:
* [ ] Yes/no/no info attributes are not correctly displayed. For instance if a user has checked "no heating" in a hut editing form, the info will not be displayed in the detail view (as if "no info" was checked) => should display "Heating: no".

ginold:
I will check that, can you give me an id of a wapoint with heating: no?

asaunier:
Just create one! :P

ginold:
are there other fields that are required to be showed even if their value = false ? 
for example, we have discussed that lift_access doesn't have to be shown if it's value = false, just show properties that are set to true.

asaunier:
For me if you have yes/no/no info => "yes" and "no" should be displayed, "no info" should not.
Concerning "lift_access", we could drop value "no" and follow the above rule as well.

asaunier:
* [ ] Don't use angular when Mako can be used for displaying data. For instance http://c2corgv6-demo.gis.internal:6555/waypoints/636651/fr ("climbing_outdoor" WP), ``rock_types`` is formatted using Angular like this:
```
      <article ng-init="rock_types = [&#39;gres&#39;]">
        <span translate class="value-title">rock_types</span><br>
        <span x-translate ng-repeat="type in rock_types" class="value">{{type}}{{$last ? '' : ', '}}</span>
      </article>
```
=> this is not the good way. It shows ugly angular brackets whereas Mako provides a boolean ``${loop.last}`` that can be used to know if this the last iteration of the loop. See http://docs.makotemplates.org/en/latest/runtime.html#the-loop-context

```
      <article>
        <span translate class="value-title">rock_types</span><br>
        % for rock_type in rock_types:
          <span x-translate class="value">${rock_type}</span>${'' if loop.last else ', '}
        % endfor
      </article>
```
Ditto for other data if needed.

asaunier:
> Yes/no/no info attributes are not correctly displayed. For instance if a user has checked "no heating" in a hut editing form, the info will not be displayed in the detail view (as if "no info" was checked) => should display "Heating: no".

[...]

Issue #277, opened 2016-04-11
Routes editing form


asaunier:
* [x] It seems that when creating a skitouring (or snowshoeing) route, some mountainbiking fields still appear in the editing form: mtb_length_trail, mtb_height_diff_portages

asaunier:
* [ ] Some list array attributes are not correctly set when no value is selected. For instance "route_types". Or maybe an API issue ? Related to https://github.com/c2corg/v6_ui/issues/110#issuecomment-208433615 and https://github.com/c2corg/v6_api/issues/145
* [ ] When updating a route with ``geometry: null``, the API says:
> {"description": "\"None\" is not a mapping type: Does not implement dict-like functionality.", "name": "document.geometry", "location": "body"}
Also an API issue?

tsauerwein:
> When updating a route with geometry: null, the API says: > {"description": "\"None\" is not a mapping type: Does not implement dict-like functionality.", "name": "document.geometry", "location": "body"} Also an API issue?

Yes.

asaunier:
OK I have created an issue in the API https://github.com/c2corg/v6_api/issues/226 and removed striked it through here.

Issue #110, opened 2016-01-19
Some WP cannot be edited


asaunier:
For instance id 636651 => API returns a 400 error

tsauerwein:
This is the error returned by the API:

    {
      "status": "error",
      "errors": [{
        "location": "body",
        "description": "Required",
        "name": "document.climbing_outdoor_types.0"
      }, {
        "location": "body",
        "description": "\"None\" is not iterable",
        "name": "document.orientation"
      }, {
        "location": "body",
        "description": "\"None\" is not iterable",
        "name": "document.best_periods"
      }, {
        "location": "body",
        "description": "\"None\" is not iterable",
        "name": "document.equipment_ratings"
      }, {
        "location": "body",
        "description": "\"None\" is not iterable",
        "name": "document.climbing_styles"
      }]
    }

asaunier:
Thanks. Seems related to missing attributes not available in the editing form yet. => will most likely be fixed when all the WP attributes are finally added.

tsauerwein:
The fields are actually included in the request, it looks like a problem with the API.

asaunier:
> it looks like a problem with the API.

I think it's a problem with the WP editing form!

I just tried to create a new "climbing_outdoor" waypoint. No pb with the standard attributes. If I add for instance an ``equipment_ratings``(see by the way this issue/PR: #289 ) the API complains that:

> {"status": "error", "errors": [{"description": "\"P1+\" is not iterable", "name": "document.equipment_ratings", "location": "body"}, {"description": "\"None\" is not iterable", "name": "document.climbing_outdoor_types", "location": "body"}, {"description": "\"None\" is not iterable", "name": "document.best_periods", "location": "body"}, {"description": "\"None\" is not iterable", "name": "document.orientation", "location": "body"}, {"description": "\"None\" is not iterable", "name": "document.climbing_styles", "location": "body"}, {"description": "Required", "name": "document.rock_types.0", "location": "body"}]}

=> equipment_ratings, best_periods, orientation, climbing_styles etc. should be arrays in the JSON request. As for now they are scalar:
```
climbing_outdoor_types: null
climbing_styles: null
equipment_ratings: "P1+"
```
They should rather be
```
climbing_outdoor_types: []
climbing_styles: []
equipment_ratings: ["P1+"]
```
@ginold could you have a look please?

tsauerwein:
`null` should also be accepted by the API: https://github.com/c2corg/v6_api/issues/55

[...]

PR #283, opened 2016-04-11
Route-editing + waypoint view fields

ginold:
small update for route-editing form and wp view.

answers #277  and #280

### DIFF
--- a/c2corg_ui/templates/route/edit.html
+++ b/c2corg_ui/templates/route/edit.html
@@ -109,7 +109,7 @@ updating_doc = route_id and route_lang
           <div class="content">
 
             ## ACTIVITIES
-            <div class="form-group data full" id="route-activities" ng-init="activities = ${activities}" ng-class="{ 'has-success': editForm.activities }">
+            <div class="form-group data full" id="route-activities" ng-init="activities = ${activities}">
               <label><span translate>activities</span> *</label>
               <div class="route-activities">
                 <div ng-repeat="activity in activities" class="activity" ng-if="activity !== 'paragliding' ">
@@ -121,17 +121,13 @@ updating_doc = route_id and route_lang
                   </div>
                 </div>
               </div>
-              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="editForm.activities.$valid"></span>
             </div>
 
-            ## CLIMBING OUTDOOR TYPES
-            <div ng-if="route.activities.indexOf('rock_climbing') > -1" class="form-group data half">
-              <label><span translate>climbing_outdoor_types</span> *</label>
-              <ul class="types">
-                <li ng-repeat="type in ${climbing_outdoor_types}" ng-click="editCtrl.pushToArray(route, 'climbing_outdoor_types', type, $event)">
-                  <div class="checkbox"><input type="checkbox" ng-checked="route.rock_types.indexOf(type) > -1"><span>{{type | translate}}</span></div>
-                </li>
-              </ul>
+            ## CLIMBING OUTDOOR TYPE
+            <div ng-if="route.activities.indexOf('rock_climbing') > -1" class="form-group data half" ng-class="{ 'has-success': route.climbing_outdoor_type}">
+              <label translate>climbing_outdoor_type</label>
+              <select class="form-control" ng-model="route.climbing_outdoor_type" ng-options="type as mainCtrl.translate(type) for type in ${climbing_outdoor_types}"><option value=""></option></select>
+              <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="route.climbing_outdoor_type"></span>
             </div>
             
           </div>
@@ -145,13 +141,12 @@ updating_doc = route_id and route_lang
             <div class="form-group data half">
               <label><span translate>orientation</span></label>
               <ul class="types long" id="orientation-list">
-                <li ng-repeat="type in ${orientation_types}" ng-click="editCtrl.pushToArray(route, 'orientation', type, $event)">
-                  <div class="checkbox"><input type="checkbox" ng-checked="route.orientation.indexOf(type) > - 1"><span x-translate>{{type}}</span></div>
+                <li ng-repeat="type in ${orientation_types}" ng-click="editCtrl.pushToArray(route, 'orientations', type, $event)">
+                  <div class="checkbox"><input type="checkbox" ng-checked="route.orientations.indexOf(type) > - 1"><span x-translate>{{type}}</span></div>
                 </li>
               </ul>
               <%include file="c2corg_ui:static/img/orientation.svg" />
             </div>
-            
                
             <div class="form-group data half">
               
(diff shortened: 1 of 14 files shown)


asaunier:
Please use more explicit commit messages like "fix attributes incorrectly shown in route editing form" or "Move access_period to the route text attributes block".
Else we won't have much info about the changes when looking at the commits history.

I have just tested the route editing form: the extra MTB fields no longer show up when selecting skitouring. But neither when selecting MTB?

Thanks for the change about access_period!

asaunier:
Please merge if 
>  the extra MTB fields no longer show up when selecting skitouring. But neither when selecting MTB?

is addressed.

ginold:
i've made a few other small changes -> i used the mako syntax in for loops instead of angular, 
changed url to a <a href> link, deleted a field that i created but didn't even exist, added other one missing.. maybe you want to have a quick look.

But i'm still waiting to clarify the 'orientation(s)' question. I also still see some strange errors while saving some attributes, but i don't know if I should correct it or is it an API problem?

For example : 
`['multi', 'bloc'] is not a string: {'climbing_outdoor_types': ''}` or `"None" is not iterable` even if it can be null and is not required.

asaunier:
>  i'm still waiting to clarify the 'orientation(s)' question

See https://github.com/c2corg/v6_api/pull/227 (and in common)

> "None" is not iterable even if it can be null and is not required.

This seems to be a problem with the API, see https://github.com/c2corg/v6_ui/issues/110#issuecomment-208739547

> ['multi', 'bloc'] is not a string: {'climbing_outdoor_types': ''}

Are you sure that the error?
I guess the API expects to receive:
```
'climbing_outdoor_types': ['multi', 'bloc']
```

c2corg_ui/templates/route/edit.html
```diff
@@ -80,7 +80,7 @@ <h2 class="heading show-phone" translate>general informations</h2>
                 <p ng-message="required" class="label label-danger" style="{{ (editForm.lang.$invalid && editForm.lang.$touched) ? 'display:inline' : '' }}" translate>This field is required.</p>
               </div>
             </div>
-            
+            {{route.orientations}}
```
asaunier: Is this really needed?

c2corg_ui/templates/route/helpers/detailed_route_attributes.html
```diff
@@ -2,22 +2,20 @@
...
-  % if route.get('areas') or route.get('orientation'):
+  % if route.get('areas') or route.get('orientations'):
     <div class="name-icon-value location" ng-click="detailsCtrl.openTab($event)">
       <h4><span translate>Location</span><span class="glyphicon glyphicon-menu-right"></span></h4>
       <div class="name-icon">
         <span class="glyphicon glyphicon-map-marker"></span>
         <p translate>Location</p>
       </div>
-      <span class="detail-text accordion">
+      <span class="detail-text accordion" id="detail-orientation" ng-init="route.orientation = ${route.get('orientations')}">
```
asaunier: Missing "s" in ``ng-init="route.orientation``?

c2corg_ui/templates/route/helpers/detailed_route_attributes.html
```diff
@@ -223,21 +240,23 @@
 
 ## ACCESS
 <%def name="get_route_access(route)">\
-  % if route.get('lift_access') or route.get('height_diff_access'):
+  % if route.get('lift_access') == False or route.get('lift_access') == True or route.get('height_diff_access'):
```
asaunier:
```
% if route.get('lift_access') not None or route.get('height_diff_access'):
```

c2corg_ui/templates/route/helpers/detailed_route_attributes.html
```diff
@@ -223,21 +240,23 @@
...
       <div class="name-icon">
         <span class="glyphicon glyphicon-road"></span>
         <p translate>Access</p>
       </div>
       <span class="detail-text accordion">
-        % if route.get('lift_access') and route['lift_access'] is not 'None':
-          <b translate class="value-title">lift_access</b>
-          <input type="checkbox" checked="checked"><span></span>
+
+        % if route.get('lift_access') == True:
```
asaunier:
```
% if route.get('lift_access'):
```

c2corg_ui/templates/route/helpers/detailed_route_attributes.html
```diff
@@ -223,21 +240,23 @@
...
+        <span x-translate class="value-title">lift_access</span> : <span translate class="value">yes</span><br>
+        % elif route.get('lift_access') == False :
```
asaunier:
```
% elif route.get('lift_access') not None:
```

c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
```diff
@@ -300,29 +316,34 @@
...
+        
       </span>
     </div>
   % endif
 </%def>
 
 ## EQUIPMENT  
 <%def name="get_waypoint_equipment(waypoint)">\
-  % if waypoint.get('matress_unstaffed') or  waypoint.get('gas_unstaffed') or waypoint.get('heating_unstaffed') or waypoint.get('blanket_unstaffed') :
+  % if waypoint.get('matress_unstaffed') == True or waypoint.get('matress_unstaffed') == False or  \
```
asaunier:
```
% if waypoint.get('matress_unstaffed') not None or  \
```
ditto for the following conditions.

ginold:
> I guess the API expects to receive:

`'climbing_outdoor_types': ['multi', 'bloc']` <

that's exactly what i'm sending. I understand it as if API was awaiting a string rather than an array, or maybe JSON.stringified array but that makes little sense.

asaunier:
Indeed, I have the same behaviour.

> I understand it as if API was awaiting a string rather than an array

Actually the API indeed expects a string:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/route.py#L148
not a list of strings. Ditto in the google sheet,

But that consistent: a route cannot be of several types at the same time!

tsauerwein:
If `climbing_outdoor_types` is not an array it should be named `climbing_outdoor_type`.

ginold:
that's exactly what is coming up next in one of Alex'es pr!

requirements.txt
```diff
@@ -14,6 +14,6 @@ Markdown==2.6.5
 # c2corg_common project
 # for development use a local checkout
 # -e ../v6_common
-git+https://github.com/c2corg/v6_common.git@343e1ef
+git+https://github.com/c2corg/v6_common.git@b8e402d
```
asaunier: The hash for c2corg_common with the fixes you suggested about some route/WP attributes is f825f8f

asaunier:
Except for the hash of c2corg_common, your changes look good to me too.

ginold:
I changed the hash. May i merge after travis has passed?

asaunier:
:+1:

PR #284, opened 2016-04-11
fix remember-me checkbox

### LINKED ISSUE #279: Authentication is lost when closing the browser
For some reason, if a user authenticates in the UI and then closes the UI tab in his browser, he won't be recognized next time he opens the UI even though he had checked "remember me".
The issue occurs in the demo instance (updated last friday) and in my own instance (up-to-date on master). This seems a quite recent bug.

ginold:
fixes #279

### DIFF
--- a/c2corg_ui/static/js/auth/auth.js
+++ b/c2corg_ui/static/js/auth/auth.js
@@ -55,6 +55,9 @@ app.AuthController = function($scope, appApi, appAuthentication,
    */
   this.scope_ = $scope;
 
+  // set remember-me to default -> true
+  this.scope_['login'] = {'remember': true};
+
   /**
    * @type {app.Api}
    * @private
--- a/c2corg_ui/templates/auth.html
+++ b/c2corg_ui/templates/auth.html
@@ -25,7 +25,7 @@
         </div>
         <div class="checkbox">
           <label>
-            <input type="checkbox" name="remember" ng-model="login.remember" ng-checked="true"> <span translate>Remember me</span>
+            <input type="checkbox" name="remember" ng-model="login.remember"><span translate>Remember me</span>
           </label>
         </div>
         <p>


gberaudo:
According to https://docs.angularjs.org/api/ng/directive/ngInit the preferred way is to change
the controller.

c2corg_ui/static/js/auth/auth.js
```diff
@@ -55,6 +55,12 @@ app.AuthController = function($scope, appApi, appAuthentication,
    */
   this.scope_ = $scope;
 
+  if (!this.scope_['login']) {
+    this.scope_['login'] = {};
+  }
```
asaunier: I think we rather write
```
this.scope_['login'] = this.scope_['login'] || {};
```

asaunier:
A minor comment. Anyway it looks good to me now.

By the way @ginold add a comment when you make changes to a PR, else no one will notice :P

c2corg_ui/static/js/auth/auth.js
gberaudo: Can it be simplified to `this.scope_['login'] = {'remember': true};`?

ginold: make up your minds and tell me which solution is better :ideograph_advantage:

tsauerwein: I prefer @gberaudo variant, it's less confusing.

tsauerwein: Unless if there is other stuff stored in `this.scope_['login']`?

ginold: yep, there are login.password, .remember and .username

tsauerwein: Ok, then I would prefer your initial proposition. :)

gberaudo: I tested my solution and it works as-is; there is no need to initialize the other fields in the controller.

asaunier: I think @gberaudo solution is OK since we set "login" in the scope with the default values (remember: true) before actually setting the username/password from the form.

asaunier: @ginold Make your choice and merge, thanks :)

PR #287, opened 2016-04-11
Set no mapFeatureCollection for routes with no geometry at all

asaunier:
Else a Mako error is returned when creating a route with no GPS track and no associated WP.


asaunier:
Will merge shortly if no one objects!

Issue #274, opened 2016-04-08
missing langs redirect not working (anymore)


ginold:
in detail view -> more button -> missing langs -> translate into another lang, the redirecting buttons don't work anymore. They had worked when app-edit-button still existed...

It copies a hidden template #missing-langs (= selector) and clones it into the modal :
  `var template = $(selector).clone();`
  `this.modal_.open({animation: true, size: 'lg', template: this.compile_(template)(this.scope_)});`

any ideas?

PR #288, opened 2016-04-11
[WIP] missing-lang redirect fix

ginold:
so i've found a solution, but there is clearly something wrong that i can't achieve.
It seems that redirection works only from a directive (I remember it worked with edit-button, then it stopped). I tried every way around but this is the only one that works.
#274

### DIFF
--- a/c2corg_ui/static/js/protectedurlbtn.js
+++ b/c2corg_ui/static/js/protectedurlbtn.js
@@ -1,0 +1,60 @@
+goog.provide('app.ProtectedUrlBtnController');
+goog.provide('app.protectedUrlBtnDirective');
+
+goog.require('app');
+
+
+/**
+ * @return {angular.Directive} directive for detailed views
+ */
+app.protectedUrlBtnDirective = function() {
+  return {
+    restrict: 'A',
+    controller: 'AppProtectedUrlBtnController',
+    scope: {
+      'lang': '=',
+      'url': '='
+    },
+    link: function(scope, el, attr, ctrl) {
+      el.click(function() {
+        ctrl.redirectToProtectedUrl(scope.url);
+      })
+    }
+  }
+};
+
+app.module.directive('protectedUrlBtn', app.protectedUrlBtnDirective);
+
+
+/**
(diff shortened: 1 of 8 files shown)


asaunier:
Instead of creating a new ``missingLang``directive (that duplicates the ``redirectToProtectedUrl`` method from ``mainController``), what about re-adding the ``editButton`` directive and use it everywhere we need to check the user credentials before granting him access:
https://github.com/c2corg/v6_ui/search?utf8=%E2%9C%93&q=redirectToProtectedUrl

ginold:
i hardly believe that what we want to do is impossible and that we have to create a directive specially for that! can someone have a look and try ?

asaunier:
I would simply get the ``editButton`` directive back:
https://github.com/c2corg/v6_ui/blob/d228cfc4a324873ad5cb88299fe119b44e15e068/c2corg_ui/static/js/editbutton.js

I don't remember exactly but I think I did it this way because it failed the other way. In addition the ``editButton`` directive add a transclusion that allows to easily embed anything inside the button:
https://github.com/c2corg/v6_ui/blob/d228cfc4a324873ad5cb88299fe119b44e15e068/c2corg_ui/static/js/editbutton.js#L25

Renaming the directive is of course an option if we want to use it for anything else than editing actions.

ginold:
maybe simply call it redirect-button ?

asaunier:
As you want provided it is not too long, self-explicity and generic enough.
What about simply ``protected-button``?

ginold:
i chose protected-url-btn.
 Have a look!

c2corg_ui/static/js/protectedurlbtn.js
```diff
@@ -0,0 +1,61 @@
+goog.provide('app.protectedUrlBtnController');
```
asaunier: ``protectedUrlBtnController`` => ``ProtectedUrlBtnController`` (capital initial letter for controllers)

asaunier:
I have not tested but the code looks good.

ginold:
because it is good :+1:  I will update and merge, ok?

PR #289, opened 2016-04-11
Fix typo in equipement_ratings fields in waypoint editing form

asaunier:
While trying to figure out the problem with https://github.com/c2corg/v6_ui/issues/110 I realized that there was a missing trailing "s" in ``equipement_ratings``!

For waypoints (climbing_outdoor), there may be several equipment rating values for the same WP (if it has several routes):
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L67
whereas routes have ``equipment_rating`` (one value per route):
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/route.py#L97

I have used ``equipment_rating`` (no s) for the field title in order not to add another translation.


ginold:
eye of a tiger, Alex! :tiger:

PR #290, opened 2016-04-11
Better Mako syntax for listing participants

asaunier:
@ginold I have finally found out a better way to list participants than the one we have discussed together earlier :)


ginold:
:ok_woman:

Issue #137, opened 2016-02-03
TypeError with some waypoints (with no geometry?)


asaunier:
* type "age" in simple search
* choose "Vallée des perroquets - bloc age"

=> error 500:

```
File "/home/alex/v6_ui/c2corg_ui/views/waypoint.py", line 42, in view
￼￼'geometry': self._get_geometry(waypoint['geometry']['geom']),
File "/home/alex/v6_ui/c2corg_ui/views/document.py", line 174, in _get_geometry
￼￼return asShape(json.loads(data))
File "/usr/lib/python3.4/json/__init__.py", line 312, in loads
￼￼s.__class__.__name__))
TypeError: the JSON object must be str, not 'NoneType'
```

tsauerwein:
To reproduce you can directly open http://c2corgv6-demo.gis.internal/waypoints/103899/fr, right?

asaunier:
Indeed

fbunoz:
It is allowed not to georef a "climbing site" in V5, for sub-site, to have only one WP for the whole site on the map. Then there are 160 WP without georef : http://www.camptocamp.org/sites/list/geom/no
But we'll have to improve the map to allowed that all WP are georef with a good display on the map.

asaunier:
Coordinates are a required field for all waypoints in v6.
Georeferencing sub-sites makes sense, doesn't it? This way you can spot on the map where the subsites are.

fbunoz:
Yes, but in the first version of V5 map, it caused some problem in the display.
There is always a problem : http://www.camptocamp.org/map?map_x=659594.22623803&map_y=5555691.6353353&map_zoom=10&tree_layers=sites
The main WP for a site is masked by the sub WP.
I don't know how to avoid this.

asaunier:
We could use clusters when icons are too close from each other:
http://openlayers.org/en/v3.15.1/examples/earthquake-clusters.html

PR #292, opened 2016-04-12
Support waypoint with no geometry

asaunier:
There are (rare) WP with no geometry imported from v5, see https://github.com/c2corg/v6_ui/issues/137

This PR tests that a geometry is available.

Fixes https://github.com/c2corg/v6_ui/issues/137


asaunier:
Any objections?

PR #293, opened 2016-04-12
don't show associated outings in wp editing form

ginold:
I forgot about this one!


asaunier:
But are the associated outings still listed in the AJAX request?

asaunier:
It seems so...

I know it's possible to make an AJAX request with no association at all using a ``a=0`` parameter: https://github.com/c2corg/v6_ui/issues/233
but then we would not get the other associated.

Let's say your change is OK, I have created an issue in the API: https://github.com/c2corg/v6_api/issues/228

PR #310, opened 2016-04-14
Hide delete button from not authenticated users

### LINKED ISSUE #297: Don't show "bin" icon in the cards when not authenticated
As for now a "bin/trash can" icon appears in the cards of associated documents (for instance in a WP page). It should only appear for users with permission "moderator".

ginold:
fixes #297

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -117,7 +117,7 @@
         % for route in document['associations']['routes']:
           <div class="list-item">
             ${route_card(route)}
-            <app-delete-association parent-id="${document['document_id']}" child-id="${route['document_id']}"></app-delete-association>
+            <app-delete-association parent-id="${document['document_id']}" child-id="${route['document_id']}" ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
           </div>
         % endfor
         <article class="list-item" ng-repeat="doc in added | filter:{'documentType': '${associatedDocuments}'}">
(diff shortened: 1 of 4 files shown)


c2corg_ui/static/js/documentediting.js
```diff
@@ -171,6 +171,8 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   if (this.auth_.isAuthenticated()) {
 
     if (this.id_ && this.lang_) {
+      // check each n seconds if user has logged out in between
+      this.recheckAuth_();
```
asaunier: Why not doing it also when creating a new document?

And honestly I would be rather pissed off if I was redirected  to the auth page while editing a doc (losing my work?).

asaunier:
What the connection of commit "redirect user that has logged out during editing" with the association removal button? Please create a separated PR.

Anyway as commented above, I think this change is not user friendly as is => I would remove the commit from this PR.

ginold:
OK, i'll move it. 
It will not be lost because it will be saved to the localstorage. And what's the point if i disconnect myself and then have to login back in order to save the document, because i have no rights? it would be lost in the same way.

asaunier:
As for now there is no localstorage saving.
It would be better to add your change when the localstorage saving is done (in the same PR?). 
Actually I think it is no use to redirect the user to the auth page while he is editing: we'd better check the user is still authenticated when saving the editing form (we'll have to do it anyway to make sure the changes are not lost), so no need to do it before.

ginold:
I moved the commit to a different branch/pr. 
Besides, ok for merging?

asaunier:
Thanks, please merge.

Issue #309, opened 2016-04-14
Use maintained fork of Typeahead


tsauerwein:
The official version of Typeahead is no longer maintained, we should use a fork that is actively developed: https://github.com/corejavascript/typeahead.js

See also: https://github.com/camptocamp/ngeo/issues/958

asaunier:
Does it simply require to update this line https://github.com/c2corg/v6_ui/blob/master/package.json#L26 ?

tsauerwein:
> Does it simply require to update this line https://github.com/c2corg/v6_ui/blob/master/package.json#L26 ?

Yes, maybe some modifications will be necessary for changes in Typeahead. It would make sense to do this switch as part of https://github.com/c2corg/v6_ui/pull/311.

asaunier:
:+1:

ginold:
Trying to install the maintained forks seems to be 'not-so-well-maintained' :+1: 

`npm install corejs-typeahead
npm ERR! 404 Not Found
npm ERR! 404 
npm ERR! 404 'corejs-typeahead' is not in the npm registry.`

I don't know how I should proceed from here?

tsauerwein:
What irony.. ;)

You can use: `npm install github:corejavascript/typeahead.js#v0.11.1`

ginold:
npm install github:corejavascript/typeahead.js#v0.11.1 `Cloning into bare repository...fatal: remote error: is not a valid repository name`

npm install github:corejavascript/typeahead.js      -||-

npm install corejs-typeahead `Cannot read property 'latest' of undefined` `corejs-typeahead' is not in the npm registry.`

npm install https://github.com/corejavascript/typeahead.js  `Error: ENOENT`

I think we might want to postpone this update maybe :)

tsauerwein:
You seem to be using a different version of npm. Try it like this:

    npm install git://github.com/corejavascript/typeahead.js#v0.11.1

Same syntax as for ol3 or ngeo:
https://github.com/c2corg/v6_ui/blob/master/package.json#L21-L22

ginold:
I updated npm yesterday and it took me some time to realize that it made my builds impossible. When trying to do a make check or make build I got errors like 'command not found, no receipt for build.js' or something like that. Finally @gberaudo helped me out and we had to go back and install an old version of NPM. Maybe we should watch out for this in the future.
Besides, I updated the package.json in and it is good and runnin'.

fjacon:
to be closed ?

[...]

PR #311, opened 2016-04-14
search-tool style update

### LINKED ISSUE #252: Simple search tool
List of reported problems:
* [ ] When there are more than 10 results for a document type, a bouton at the end of the list should allow to open the advanced search/list page with all the results
* [ ] Not sure if it is a UI or an API problem: sometimes the search tool only returns waypoints results, no route (despite they are the routes of the provided WP). For instance with "Mont de grange"

ginold:
- better styling for search results
- you can collapse the results by clicking on the dataset title
- for now, "show more results" button which redirects to list-view of the document type
will fix #252

### DIFF
--- a/c2corg_ui/static/js/search.js
+++ b/c2corg_ui/static/js/search.js
@@ -9,10 +9,10 @@ goog.require('ngeo.searchDirective');
 /**
  * The directive for the auto-complete search field shown in the header
  * of every page.
- *
+ * @param {angular.$compile} $compile Angular compile service.
  * @return {angular.Directive} Directive Definition Object.
  */
-app.searchDirective = function() {
+app.searchDirective = function($compile) {
   return {
     restrict: 'E',
     controller: 'AppSearchController',
@@ -27,12 +27,14 @@ app.searchDirective = function() {
          * @param {angular.JQLite} element Element.
          */
         function($scope, element) {
+
           var phoneScreen = app.constants.SCREEN.SMARTPHONE;
 
           // Empty the search field on focus and blur.
           $('.page-header').find('input').on('focus blur', function() {
             $(this).typeahead('val', '');
           });
+
           //Remove the class 'show-search' when screen width > @phone (defined in LESS)
           $(window).resize(function resize() {
             if ($(window).width() > phoneScreen) {
@@ -41,35 +43,41 @@ app.searchDirective = function() {
             }
           });
 
-          // Trigger focus on search-icon click for .search
-          $('.page-header .search-icon').on('click', function(event) {
+          element.on('click', function(e) {
+
+            // collapse suggestions
+            if ($('app-search .header').is(e.target)) {
+              $(e.target).siblings('.tt-suggestion').slideToggle();
+            }
+
+            // Trigger focus on search-icon click for .search
             if (window.innerWidth < phoneScreen) {
-              $('.page-header').find('.quick-search').toggleClass('show-search');
-              $('.page-header').find('.search').focus();
-              $('.logo.header, .menu-open-close.header').toggleClass('no-opacity');
-              event.stopPropagation();
+              if ($('.page-header .search-icon').is(e.target)) {
+                $('.page-header').find('.quick-search').toggleClass('show-search');
+                $('.page-header').find('.search').focus();
+                $('.logo.header, .menu-open-close.header').toggleClass('no-opacity');
+              }
             }
           });
 
-          $('body:not(.page-header)').click(function(e) {
-            $('.tt-dataset.empty').remove();
-
-            if (!$('.page-header, .show-search, .quick-search, .search').is(e.target)) {
+          // hide the menu when click outside (smartphone)
+          $('main').click(function(e) {
+            if (window.innerWidth < phoneScreen) {
               $('.show-search').removeClass('show-search');
-              if (window.innerWidth < phoneScreen) {
-                $('.logo.header, .menu-open-close.header').removeClass('no-opacity');
-              }
+              $('.logo.header, .menu-open-close.header').removeClass('no-opacity');
             }
           });
 
           //show spinning gif while waiting for the results
           element.on('typeahead:asyncrequest', function() {
             element.find('input').addClass('loading-gif-typehead');
-          })
+          });
           element.on('typeahead:asynccancel typeahead:asyncreceive', function() {
             element.find('input').removeClass('loading-gif-typehead');
           });
+
         }
+
   };
 };
 
@@ -78,14 +86,15 @@ app.module.directive('appSearch', app.searchDirective);
 
 /**
  * @constructor
- * @param {angular.Scope} $scope Angular scope.
+ * @param {!angular.Scope} $scope Angular scope.
  * @param {angular.$compile} $compile Angular compile service.
+ * @param {angular.$templateCache} $templateCache service
  * @param {angular.Attributes} $attrs Angular attributes.
  * @param {string} apiUrl Base URL of the API.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @ngInject
  */
-app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog) {
+app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog, $templateCache) {
 
   /**
    * Bound from directive.
@@ -107,6 +116,13 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
     this.selectHandler = undefined;
   }
 
+
+  /**
+   * @type {angular.$templateCache}
+   * @private
+   */
+  this.templatecache_ = $templateCache;
+
   /**
    * @type {string}
    * @private
@@ -114,7 +130,7 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
   this.apiUrl_ = apiUrl;
 
   /**
-   * @type {angular.Scope}
+   * @type {!angular.Scope}
    * @private
    */
   this.scope_ = $scope;
@@ -141,8 +157,9 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
    * @export
    */
   this.datasets = [
-    this.createDataset_('waypoints', 'Waypoints'),
-    this.createDataset_('routes', 'Routes')
+    this.createDataset_('waypoints'),
+    this.createDataset_('routes'),
+    this.createDataset_('outings')
   ];
 
   /**
@@ -163,13 +180,13 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
 
 /**
  * @param {string} type The document type.
- * @param {string} title The category title.
  * @return {TypeaheadDataset} A data set.
  * @private
  */
-app.SearchController.prototype.createDataset_ = function(type, title) {
+app.SearchController.prototype.createDataset_ = function(type) {
   var bloodhoundEngine = this.createAndInitBloodhound_(type);
   return /** @type {TypeaheadDataset} */({
+    contents : type,
     source: bloodhoundEngine.ttAdapter(),
     display: function(doc) {
       return doc.label;
@@ -177,18 +194,23 @@ app.SearchController.prototype.createDataset_ = function(type, title) {
     limit: 20,
     templates: {
       header: (function() {
-        return '<div class="header ' + title + '">' +
-            this.gettextCatalog_.getString(title) + '</div>';
+        return '<div class="header" dataset="' + type + '">' + this.gettextCatalog_.getString(type) + '</div>';
       }).bind(this),
+      footer: function() {
+        return '<p class="suggestion-more"><a href="/' + type + '" class="green-text">+ see more results</a></p>';
+      },
       suggestion: function(doc) {
-        return '<p>' + doc.label + '</p>';
-      }
-      // Note: The templates for `notFound` and `pending` are not set, because
-      // they would be shown for each dataset.
+        this.scope_['doc'] = doc;
+        return this.compile_('<app-suggestion  class="tt-suggestion"></app-suggestion>')(this.scope_);
+      }.bind(this),
+      empty: function(res) {
+        if ($('.header.empty').length === 0) {
+          return this.compile_(this.templatecache_.get('/static/partials/suggestionempty.html'))(this.scope_);
+        }
+      }.bind(this)
     }
   });
 };
-
 
 /**
  * @param {string} type The document type.
@@ -196,11 +218,10 @@ app.SearchController.prototype.createDataset_ = function(type, title) {
  * @private
  */
 app.SearchController.prototype.createAndInitBloodhound_ = function(type) {
-  var empty_template = $('#empty-results').html(); // in base.html
   var url = this.apiUrl_ + '/search?q=%QUERY';
 
   var bloodhound = new Bloodhound(/** @type {BloodhoundOptions} */({
-    limit: 10,
+    limit: 7,
     queryTokenizer: Bloodhound.tokenizers.whitespace,
     datumTokenizer: Bloodhound.tokenizers.obj.whitespace('label'),
     remote: {
@@ -216,28 +237,20 @@ app.SearchController.prototype.createAndInitBloodhound_ = function(type) {
           if (this.datasetLimit_.length === 1) {
             url = url + '&t=' + this.datasetLimit_;
           } else {
-            // you receive for ex 'wro' (waypoints, routes, users) -> split into w,r,o
+            // you receive for ex 'wru' (waypoints, routes, users) -> split into w,r,u
             this.datasetLimit_ = this.datasetLimit_.split('').join(',');
           }
         }
-
         settings['url'] = url.replace('%QUERY', encodeURIComponent(query));
         return settings;
       }).bind(this),
+
       filter: (function(/** appx.SearchResponse */ resp) {
         var documentResponse =
                 /** @type {appx.SearchDocumentResponse} */ (resp[type]);
-
         if (documentResponse) {
           var documents = documentResponse.documents;
           var currentLang = this.gettextCatalog_.currentLanguage;
-
-          if (documents.length === 0) {
-            $('.tt-dataset.empty').remove();
-            $('.tt-empty').append(empty_template);
-          } else {
-            $('.tt-dataset.empty').remove();
-          }
 
           return documents.map(function(/** appx.SearchDocument */ doc) {
             var locale = doc.locales[0];
@@ -278,6 +291,5 @@ app.SearchController.select_ = function(event, doc, dataset) {
     window.location.href = url;
   }
 };
-
 
 app.module.controller('AppSearchController', app.SearchController);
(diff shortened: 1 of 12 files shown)


c2corg_ui/static/js/search.js
```diff
@@ -177,14 +184,62 @@ app.SearchController.prototype.createDataset_ = function(type, title) {
...
+        return '<div class="header" dataset="' + type + '">' + this.gettextCatalog_.getString(type) + '</div>';
       }).bind(this),
+      footer: function() {
+        return '<p class="suggestion-more"><a href="/' + type + '" class="green-text">+ see more results</a></p>'
+      },
       suggestion: function(doc) {
-        return '<p>' + doc.label + '</p>';
+        var template;
+        // routes
+        if (doc.documentType === 'routes') {
```
asaunier: Do we really want to have a bunch of HTML code in the javascript code?
Can't we use a partial instead? https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials

asaunier:
Concerning the actual result in the interface, this looks awesome! :) 
Would it be possible to have the search word hilighted in the results (as currently)?
simplesearch

asaunier:
By the way: we don't need the attribute names (elevation) => just show the values.
For routes, the ratings and facings are also useful in addition to the elevation_max.
The regions are useful as well but could be done later if needed.

ginold:
I could add the other infos, but they are not returned by the api.
screenshot from 2016-04-15 12 46 31

asaunier:
OK, we'll fine tune the content later.

What about my question about the template/partial?

ginold:
as we stated - we have no clue how to work it around.
For now this returns a string (HTML) :
` if (doc.documentType === 'routes') {
          return that.createRouteSuggestion_(doc);
}`
But how could we use a partial? i don't know how to access it from this level, in addition there are variables that will not be compiled and understood by angular, nor making a directive solved the problem.

@gberaudo @tsauerwein  if you have any ideas it would be great!

gberaudo:
Not a real answer @ginold but it is done the same way in ngeo: https://github.com/camptocamp/ngeo/blob/ba4f2d98e159c757e118b1a30db50c25a3661171/contribs/gmf/src/directives/search.js#L379-L390.

tsauerwein:
You could create a new directive (`<search-suggestion document="...">`) with a partial and then use `$compile` in the template function. Similar to here: https://github.com/camptocamp/ngeo/blob/87cdea826e091a7447e41f36fd2ce3fe0e54701f/src/services/popup.js#L37-L57

But I leave it up to you if you want to change it or not.

ginold:
ok i finally managed to do it with a directive and it works nicely :) There was an issue with highlighting which didn't work anymore but i found a workaround. 

+ working redirect-to-protected-url buttons working on 'no results' template.

Ready to be re-reviewed!

tsauerwein:
Could you also switch to a maintained fork of Typeahead, see https://github.com/c2corg/v6_ui/issues/309?

c2corg_ui/static/js/search.js
```diff
@@ -163,44 +172,55 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
...
  * @return {Bloodhound} The bloodhound engine.
  * @private
  */
 app.SearchController.prototype.createAndInitBloodhound_ = function(type) {
-  var empty_template = $('#empty-results').html(); // in base.html
   var url = this.apiUrl_ + '/search?q=%QUERY';
 
   var bloodhound = new Bloodhound(/** @type {BloodhoundOptions} */({
-    limit: 10,
+    limit: 7,
```
asaunier: Why 7 instead of 10?

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,50 @@
...
+        return $sce.trustAsHtml(text.replace(new RegExp(search, 'i'), '<span class="tt-highlight">$&</span>'));
+      };
+
+      element.html(template(attrs.doctype, attrs.activities));
+      $compile(element.contents())(scope);
+    }
+  };
+}
+
+app.module.directive('suggestion', app.suggestionDirective);
```
asaunier: All other application-specific directive are prefixed with "app". Please keep the naming consistency and name your directive ``appSuggestion`` (``app-suggestion`` in the HTML).

asaunier:
> Ready to be re-reviewed!

Then don't forget to remove the "WIP" in the PR title.

I have just tested the PR in my instance, the rendering is good!

Concerning the outings:
* [ ] the date is missing and is essential => how can you figure out what is the outing you are looking for when the date is not shown?
* [ ] I'm not really sure outings should appear in the simple search. I will double check with the client.

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,50 @@
+goog.provide('app.suggestionDirective');
+
+/**
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {angular.$sce} $sce Angular Strict Contextual Escaping service.
+ * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
+ */
+app.suggestionDirective = function($compile, $sce) {
+  var header = '<span class="suggestion-text">' +
```
asaunier: Can't we use a partial for this?
This would be clearer to isolate the HTML for the JS code.

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,50 @@
...
+      for (var i = 0; i < acts.length; i++) {
+        activitiesHtml += '<span class="icon-' + acts[i] + '"></span>'
+      }
+      return header + '</span><span class="icon-container">' + activitiesHtml + footer;
+    }
+  }
+
+  return {
+    restrict: 'E',
+    scope: {'type': '@', 'label': '@', 'dataset': '@', 'elevation': '@', 'activities': '@', 'query': '@'},
```
asaunier: I feel a bit incomfortable with those scope params. They can be really different among document types. Can't we pass a single ``attr`` object that would contain the attribute to show in the partial depending on the document type?

ginold: Indeed that's what I told myself after there were a bunch of them. I will try to get rid of this when I come back to this pr.

c2corg_ui/static/js/search.js
ginold: it took too much space in my opinion, like 10 results took more than half of the screen height.

c2corg_ui/static/js/suggestiondirective.js
ginold: We have already discussed it - the partials are generated only when they are used in a templateUrl in a directive. It implies creating a different suggestion-directive for each type of document. If I create a dynamic templateUrl, it will not be compiled by mako. Surely there is a workaround but I would let it to the person that actually created those functions :P

c2corg_ui/static/js/search.js
```diff
@@ -163,44 +172,58 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
...
-      // Note: The templates for `notFound` and `pending` are not set, because
-      // they would be shown for each dataset.
+        var document = {
+          'query' : doc['_query'],
+          'elevation' : doc['elevation'] || doc['elevation_max'],
+          'waypoint_type' : doc['waypoint_type'],
+          'activities' : doc['activities'],
+          'date_end' : doc['date_end'],
+          'doctype' : doc['documentType']
+        };
```
asaunier: Why not directly passing ``doc`` to the ``app-suggestion`` directive instead of creating this intermediary ``document`` variable? Then we would leave the templates use the attributes they need.

Depending on the document type you may have really different attributes. ``waypoint_type`` is only relevant for waypoints. ``date_end`` only for outings.

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,54 @@
...
+  var html = '';
+  var header = '<span class="suggestion-text">' +
+                            '<span class="suggestion-title">' +
+                              '<span ng-bind-html="highlight(label, query)"></span>' +
+                              '<span ng-if="date_end">&nbsp;({{::date_end }})</span>' +
+                            '</span><br>' +
+                            '<span class="suggestion-info">' +
+                            '  <span ng-if="elevation">{{::elevation}}m</span>' +
+                            '</span>';
+  var footer =   '</span>';
```
asaunier: This "generic" header looks weird: we have no ``date_end`` or ``elevation`` for all document types.
What about testing the doctype first and then call functions dedicated to a given doctype that would return the corresponding template?

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,54 @@
...
+        activitiesHtml += '<span class="icon-' + doc['activities'][i] + '"></span>';
+      }
+      html = header + '</span><span class="icon-container">' + activitiesHtml + footer;
+    }
+    return html;
+  };
+
+  return {
+    restrict: 'E',
+    scope: {'doc': '@', 'label' : '@'}, // I left label because there were problems with escaping quotes in the title with json.parse
```
asaunier: What's the connection between E and the escaping problems?

asaunier:
I have noticed that a blank dropdown appears when the user first type a keyword until the results are actually displayed. When doing further searches, the blank dropdown appears already when the focus (mouse click) is in the input. See the capture below:
untitled

No such behaviour can be observed when using the simple search tool in master.

* [ ] Could you make sure the dropdown does not appear until the results are displayed?

c2corg_ui/static/js/suggestiondirective.js
ginold: no connection between E and the problems?
When i left the label within the doc (`doc['label']`) and then when I parsed it `JSON.parse(doc)`, JS could not parse it correctly when there were apostrophes in the title.

c2corg_ui/static/js/suggestiondirective.js
ginold: you are right, but checking every document type and having a different html for each will multiply the code.
Maybe a less "subtle" solution would be to have the same big template for every doc, but check with ng-ifs for the values in order to show them or not, as it's done now with date which does not appear in other docs than outings.
One way or another we will have to check for values and/or types of document.

c2corg_ui/static/js/search.js
ginold: maybe the one little reason for this was the 'elevation' property - some docs have elevation, elevation_max or elevation_min and I would have to check for each of them in the template, so I create a new object with the only property 'elevation'. Maybe I could simply rename the current object's property from elevation_whatever to just elevation.

ginold:
probably a firefox issue, i haven't noticed it on chrome before. Will check that out.

c2corg_ui/static/js/suggestiondirective.js
asaunier: Honestly I would prefer having separated template functions even though it makes the code bigger. Else the code is going to be a big mess, really hard to understand, with conditions everywhere.
I think you should have a switch on ``doctype`` calling dedicated template functions, with a default template that can be used by default for document types with no special layout (with only the document title).

c2corg_ui/static/js/suggestiondirective.js
asaunier: I see :)
But your comment was confusing. Do we really need this comment anyway?

asaunier:
> probably a firefox issue,

Indeed, it's OK with Chromium

c2corg_ui/static/js/suggestiondirective.js
ginold: If I didn't put it you would ask me why I did set 'label' apart from the 'doc' ? So I think it's relevant.

c2corg_ui/static/js/search.js
```diff
@@ -163,44 +172,58 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
...
+        return this.compile_('<app-suggestion label="' + doc['label'] + '" doc=\'' + JSON.stringify(document) + '\' class="tt-suggestion"></app-suggestion>')(this.scope_);
```
tsauerwein: Can't you set the document on `scope` instead of converting it to a JSON string and then back again to an object?

ginold: I have no idea how to pass an object directly to the directive's scope from this non-angular context.

tsauerwein: You have a scope with `this.scope_`. Either you use this scope or create a new one with [$new](https://docs.angularjs.org/api/ng/type/$rootScope.Scope#$new). Then set your values to this scope and access them again from your directive.

ginold: ok I see i can set scope: true in the directive and then the properties are accessible from the $parent. Will try that out!

ginold: I'm nearly there, but I still encounter a problem. I can access the passed document through scope.$parent['document'], I extend the directive's scope by the document from the parent, but the template doesn't seem to compile correctly and not a single value is evaluated.

My directive 

```
return {
    restrict: 'E',
    scope: true, 

    link: function(scope, element, attrs) {

      var document = scope.$parent['doc']
      angular.extend(scope, scope.$parent)
     // if i console.log(scope), I have all of the needed attributes of the document

      element.html(template(document));
      $compile(element.contents())(scope);  // I could use scope.$parent as well, but it doesn't change
    }
  }
```

and template like `... '<span ng-if="elevation_max">{{::elevation_max}}m</span>' ....`

ginold: I'm nearly there. Now I have access to the document in the directive's
scope, passed through the parent scope, but the compilation doesn't go well
because none of the values are evaluated.

      angular.extend(scope, scope['doc'])
     // here I have all of the doc properties in the scope needed for the
template
      element.html(template(scope['doc']));
      $compile(element.contents())(scope);

And none of the angular bindings show up. Any ideas?

On Wed, Apr 27, 2016 at 2:08 PM, Tobias Sauerwein <notifications@github.com>
wrote:

> In c2corg_ui/static/js/search.js
> <https://github.com/c2corg/v6_ui/pull/311#discussion_r61245941>:
>
> >        suggestion: function(doc) {
> > -        return '<p>' + doc.label + '</p>';
> > -      }
> > -      // Note: The templates for `notFound` and `pending` are not set, because
> > -      // they would be shown for each dataset.
> > +        var document = {
> > +          'query' : doc['_query'],
> > +          'elevation' : doc['elevation'] || doc['elevation_max'],
> > +          'waypoint_type' : doc['waypoint_type'],
> > +          'activities' : doc['activities'],
> > +          'date_end' : doc['date_end'],
> > +          'doctype' : doc['documentType']
> > +        };
> > +        return this.compile_('<app-suggestion label="' + doc['label'] + '" doc=\'' + JSON.stringify(document) + '\' class="tt-suggestion"></app-suggestion>')(this.scope_);
>
> You have a scope with this.scope_. Either you use this scope or create a
> new one with $new
> <https://docs.angularjs.org/api/ng/type/%24rootScope.Scope#%24new>. Then
> set your values to this scope and access them again from your directive.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/c2corg/v6_ui/pull/311/files/03c3a4bc5860d95e944bfd84e4fbea820a2007ba#r61245941>
>



-- 

adam.gina@camptocamp.com

Bâtiment A
1015 Lausanne

Tel + 41 21 619 10 10
http://www.camptocamp.com

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,87 @@
+goog.provide('app.suggestionDirective');
+
+/**
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {angular.$sce} $sce Angular Strict Contextual Escaping service.
+ * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
+ */
+app.suggestionDirective = function($compile, $sce) {
```
tsauerwein: Our original motivation to make a directive for the suggestions was to have a partials template to avoid having large parts of HTML in the JS code, right? Are you still planning to do that?

ginold: Yep, I will have a look on that. I've had a problem with partials - they didn't compile into templatecache until they were referenced in a directive with templateUrl (really!)  but now it seems to compile. I will check it and update if it works.

ginold:
omg that took me so long! 
I hope that now it's PERFECT :P it uses template cache with partials.

c2corg_ui/static/js/search.js
```diff
@@ -163,44 +180,49 @@ app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog
...
       }).bind(this),
-      suggestion: function(doc) {
-        return '<p>' + doc.label + '</p>';
-      }
-      // Note: The templates for `notFound` and `pending` are not set, because
-      // they would be shown for each dataset.
+      footer: function() {
+        return '<p class="suggestion-more"><a href="/' + type + '" class="green-text">+ see more results</a></p>';
+      },
+      suggestion: function(doc) { // escape apostrophes -> problem with json.parse
```
tsauerwein: I think now you can remove this comment.

c2corg_ui/static/js/suggestiondirective.js
```diff
@@ -0,0 +1,47 @@
...
+    return $templateCache.get('/static/partials/suggestion' + doctype + '.html');
+  };
+
+  return {
+    restrict: 'E',
+    scope: true,
+    /**@suppress {checkTypes} for JSON.parse */
+    link: function(scope, element, attrs) {
+
+      scope.highlight = function(text, search) {
```
tsauerwein: I am wondering if `highlight` might not get renamed during the compilation. Have you tested with the built version?

tsauerwein:
Nice, good job! :+1:

c2corg_ui/static/js/suggestiondirective.js
ginold: yep, it works correctly.

ginold:
Thanks! I've done soo many trial and errors and I still don't get what was wrong.
I had to put `scope.$apply()` to make it appear, if not - the html was not rendered/compiled. And I remember that when I did an apply after a compile it showed an error 'there is already a digest cycle' or something. In my previous method (with json stringify/parse) I didn't have to apply.

Another strange thing was that, just for fun, I appended 
`$compile(element.contents())(scope)` (without any scope.apply) to the sidemenu and it did append an uncompiled html, but when I clicked outside it suddenly compiled!

Any ideas?

tsauerwein:
In general you have to call `scope.$apply()` when events are triggered from outside, in your case by TypeAhead. Otherwise Angular would not know that something has changed.

> ...but when I clicked outside it suddenly compiled!

You probably have set up a click event listener, which triggered the Angular digest cycle.

asaunier:
Works fine in standard mode, but in debug mode the templates seem not to load as expected.

ginold:
as if $templateCache wasn't regenerated in debug mode, and the partials aren't used anywhere in any directive's templateUrl, so they are not referenced. Therefore $templateCache is empty in debug mode and the templates don't show up.

asaunier:
In debug mode the partials are loaded directly from ``static/partials``, ``templateCache.js`` is only used in prod mod.

ginold:
despite of that, ok for merging?

PR #315, opened 2016-04-15
Add support of line geometries when editing routes/outings

### LINKED ISSUE #303: Current document geometry editing: lines (routes, outings), including import of gpx track

asaunier:
Fixes #303 

TODO and open questions:
* [x] display and edit/draw line geometries in the editing form map
* [x] drag-drop a GPX file on the map to upload it
* [ ] Allow to reset the current geometry (for instance to draw a brand new one)? see https://github.com/c2corg/v6_ui/issues/330
* [x] don't send the geometry to the API if it has not changed (else it creates a new geom version in the DB?)
* [ ] Make sure 3D tracks are supported
 * [ ] Elevation/timestamp are removed from  GPX tracks see https://github.com/c2corg/v6_ui/issues/331
 * [ ] API seems to not support elevation/timestamp, see  https://github.com/c2corg/v6_api/issues/262
* [x] Simplify too big geometries
* [x] ~~``ol.interaction.DragAndDrop`` returns ``MultiLineString`` geometries whereas the API expects ``LineString`` and throw an error back. Is it possible to convert the ``MultiLineString`` to ``LineString`` (for instance by concatenating the ``LineString`` items of the ``MultiLineString``)? Or maybe we should change the API in order to support ``MultiLineString``? Which could be useful for instance for routes with variantes?~~
* [x] Support the "Add a GPS track" button (as for now it is a dummy one). ngeo has ``ngeo-filereader`` (http://camptocamp.github.io/ngeo/master/examples/importfeatures.html) but I don't know if it is that easy to combine with our ``app-map`` directive?

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -202,7 +202,6 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 app.DocumentEditingController.prototype.successRead_ = function(response) {
   var data = response['data'];
   var toCoordinates = (function(str) {
-    // FIXME handle lines and polygons
     var point = /** @type {ol.geom.Point} */
         (this.geojsonFormat_.readGeometry(str));
     return this.getCoordinatesFromPoint_(point);
@@ -210,7 +209,10 @@ app.DocumentEditingController.prototype.successRead_ = function(response) {
 
   if ('geometry' in data && data['geometry']) {
     var geometry = data['geometry'];
-    if ('geom' in geometry && geometry['geom']) {
+    // don't add lonlat for line or polygon geometries
+    // (point geometries have no 'geom_detail' attribute)
+    if (!('geom_detail' in geometry) &&
+        'geom' in geometry && geometry['geom']) {
       var coordinates = toCoordinates(geometry['geom']);
       data['lonlat'] = {
         'longitude': coordinates[0],
@@ -366,16 +368,31 @@ app.DocumentEditingController.prototype.handleMapFeatureChange_ = function(
     feature) {
   var geometry = feature.getGeometry();
   goog.asserts.assert(geometry);
+  var isPoint = geometry instanceof ol.geom.Point;
   var data = this.scope_[this.modelName_];
   // If creating a new document, the model has no geometry attribute yet:
   data['geometry'] = data['geometry'] || {};
-  data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(geometry);
-  if (geometry instanceof ol.geom.Point) {
-    var coords = this.getCoordinatesFromPoint_(geometry.clone());
+  if (isPoint) {
+    data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(geometry);
+    var coords = this.getCoordinatesFromPoint_(
+        /** @type {ol.geom.Point} */ (geometry.clone()));
     data['lonlat'] = {
       'longitude': coords[0],
       'latitude': coords[1]
     };
+  } else {
+    var center;
+    // For lines, use the middle point as point geometry:
+    if (geometry instanceof ol.geom.LineString) {
+      center = geometry.getCoordinateAt(0.5);
+    } else if (geometry instanceof ol.geom.MultiLineString) {
+      center = geometry.getLineString(0).getCoordinateAt(0.5);
+    } else {
+      center = ol.extent.getCenter(geometry.getExtent());
+    }
+    var centerPoint = new ol.geom.Point(center);
+    data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(centerPoint);
+    data['geometry']['geom_detail'] = this.geojsonFormat_.writeGeometry(geometry);
   }
 };
 
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -188,8 +188,8 @@ app.DocumentEditingController = function($scope, $element, $attrs,
     return;
   }
 
-  this.scope_.$root.$on('mapFeatureChange', function(event, feature) {
-    this.handleMapFeatureChange_(feature);
+  this.scope_.$root.$on('mapFeaturesChange', function(event, features) {
+    this.handleMapFeaturesChange_(features);
     this.scope_.$apply();
   }.bind(this));
 };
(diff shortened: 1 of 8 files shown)


asaunier:
> Simplify too big geometries

OL3 has a ``simplify`` tool: http://openlayers.org/en/v3.15.1/apidoc/ol.geom.LineString.html#simplify that might do the job fine.

I have also noticed Turf http://turfjs.org/static/docs/module-turf_simplify.html itself using simplify.js http://mourner.github.io/simplify-js/
Just curious if the difference is big?

@gberaudo Do you know what tool is used for simplifying uploaded GPX in Schweizmobil?

c2corg_ui/static/js/map/map.js
```diff
@@ -423,14 +423,54 @@ app.MapController.prototype.handleDraw_ = function(event) {
...
+
+
+/**
+ * @param {ol.Feature} feature Feature to process.
+ * @return {ol.Feature}
+ * @private
+ */
+app.MapController.prototype.processFeature_ = function(feature) {
+  var geometry = feature.getGeometry();
+  // convert MultiLineString to LineString
```
tsauerwein: Actually `MultiLineString` is ok, the API should also accept it. This is the same as in v5.

tsauerwein:
> I have also noticed Turf http://turfjs.org/static/docs/module-turf_simplify.html itself using simplify.js http://mourner.github.io/simplify-js/
> Just curious if the difference is big?

The algorithm of simplify.js is probably faster (I haven't tested it). But the ol3 implementation is also quite good. Just drag&drop a GPX file in this [example](http://openlayers.org/en/master/examples/drag-and-drop.html). The track is already being simplified when it is rendered on the map.

c2corg_ui/static/js/documentediting.js
```diff
@@ -359,16 +361,23 @@ app.DocumentEditingController.prototype.handleMapFeatureChange_ = function(
...
+    data['geometry']['geom'] = this.geojsonFormat_.writeGeometry(geometry);
+    var coords = this.getCoordinatesFromPoint_(
+        /** @type {ol.geom.Point} */ (geometry.clone()));
     data['lonlat'] = {
       'longitude': coords[0],
       'latitude': coords[1]
     };
+  } else {
+    var center = ol.extent.getCenter(geometry.getExtent());
+    var centerPoint = new ol.geom.Point(center);
```
tsauerwein: You could use [`getCoordinateAt(0.5)`](http://openlayers.org/en/master/apidoc/ol.geom.LineString.html#getCoordinateAt) to get a point in the middle of a line. This might be better than the center of the extent.

c2corg_ui/static/js/map/map.js
asaunier: Indeed. For some reason I thought the API only accepted ``LineString`` but I have just checked and ``MultiLineString``. Then no reason to make such a conversion, I have dropped it. Thanks.

c2corg_ui/static/js/documentediting.js
asaunier: True, but it only works with ``LineString`` I think. And since we get ``MultiLineString``... :P Or maybe we should use the middle point of the first ``LineString``?
I have added a fixup for this: 6d0dc2898f86b7d3f8b3434e1b1b8a5e8a55364a but I wonder if we should not use a more consistent approach for LineStrings/MultiLineStrings.

tsauerwein: > Or maybe we should use the middle point of the first LineString?

That's what I am doing during the migration.

asaunier: That sounds reasonable. I have pushed a fixup to get the middle point of the first LineString. Thanks!

gberaudo:
> @gberaudo Do you know what tool is used for simplifying uploaded GPX in Schweizmobil?

The GPX is simplified server side using postgis.

asaunier:
I think this PR is ready for reviewing.

Some points are still pending but could be addressed in some future PRs: for instance a button to reset the geometry (remove it before making a new drawing? cancel the changes if the user finally don't want to change the geometry?) => maybe that could be considered as an "extra" feature.

Concerning the 3D/4D:
* OL3's simplification of geometries only returns 2D features and drops the extra elevations/timestamps (see https://github.com/openlayers/ol3/blob/a81150508dd1f15ab4dd99ee1ef6f748a33a37fd/src/ol/geom/flat/simplifyflatgeom.js#L123-L126 => it returns only the lon/lat of the "flat" coordinates and ignore all other dimensions).
* Anyway the API seems not to support 3D/4D, see https://github.com/c2corg/v6_api/issues/262

That's a pity because having 3D geometries for routes and 4D geometries for outings would make sense. I don't know how it would combined with geometries modifying in the map though (we would have no elevation nor timestamp when adding/moving a point of the linestrings...). Perhaps it deserves a dedicated issue?

gberaudo:
@asaunier, If you want to copy and adapt the simplify code from OL3 to the UI, that is fine for me.

asaunier:
Well, since the 3D is currently not supported by the API, it is not worth making changes in the simplification tools now.

asaunier:
Issues have been created for the missing points (3D and "cancel" button).
OK for merging?

c2corg_ui/static/js/map/map.js
```diff
@@ -429,13 +443,43 @@ app.MapController.prototype.handleDraw_ = function(event) {
...
+
+/**
+ * @param {ol.Feature} feature Feature to process.
+ * @return {ol.Feature}
+ * @private
+ */
+app.MapController.prototype.simplifyFeature_ = function(feature) {
+  var geometry = feature.getGeometry();
+  // simplify geometry with a tolerance of 20 meters
+  geometry = geometry.simplify(20);
```
tsauerwein: 20 meters? I would probably have chosen a smaller value.

asaunier: 10m?

My GPS is set to save points every 50m and it's already quite enough since there are often errors and poor precision => it's not that important to be really precise.
See for instance one of my outings: http://www.camptocamp.org/outings/739447/fr/la-grande-lui-en-boucle-par-les-5-cols#map_container

tsauerwein:
That looks very good to me!

Can you edit the track after adding it to the map? And can you also draw a new line (without uploading a track)?

asaunier:
> Can you edit the track after adding it to the map?

Yes

> And can you also draw a new line (without uploading a track)?

Yes

What you can do:
* upload a GPX track either using the file button or by drag-dropping the file on the map. If a geom was already available it is replaced by the new one.
* edit an existing track or a track you have just uploaded
* draw a track from scratch if there was no geom and if you have no GPX file

asaunier:
Thanks for the review @tsauerwein !
We can change the simplification value later if needed.

tsauerwein:
Sounds good, thanks!

PR #316, opened 2016-04-19
404/400 page

### LINKED ISSUE #308: 404/500 page
We should create a simple page telling the user that the page is broken or cannot be found.

ginold:
screenshot from 2016-04-19 10 52 41

Page that's shown whenever there is a bad request, malformed url or document not found.
fixes #308


c2corg_ui/static/img/activities/falling.svg
```diff
@@ -0,0 +1,148 @@
...
+    </g>
+  </g>
+  <g
+     inkscape:groupmode="layer"
+     id="layer2"
+     inkscape:label="Layer 2"
+     transform="translate(0,-698.0315)">
+    <g
+       id="ice_climbing-5"
+       ng-click="editCtrl.pushToArray(route, 'activities', 'ice_climbing', $event)"
```
tsauerwein: ng-click?

less/404.less
```diff
@@ -0,0 +1,91 @@
...
+    margin-left: 13%;
+    @media  (max-width: @screen-xs-max) {
+      display: none;
+    }
+    img {
+      transform-origin: 49% -117%;
+      margin-left: 18%;
+      width: 47%;
+      &:hover {
+        -webkit-animation: swing 8s infinite;
```
tsauerwein: When using Less, doesn't Less take care of adding browser-specific rules like `-webkit-animation` or `-moz-animation`? E.g. is it not enough to use `animation: ...`?

c2corg_ui/static/js/map/map.js
```diff
@@ -209,14 +216,13 @@ app.MapController = function($scope, mapFeatureCollection) {
   }
 };
 
-
+console.log(app.MapController.zoom)
```
tsauerwein: debug output

c2corg_ui/views/document.py
```diff
@@ -221,3 +222,13 @@ def _diff(self):
             return self.template_input
 
         raise HTTPNotFound()
+
+    notFound = 'c2corg_ui:templates/404.html'
+
+    @view_config(context=HTTPInternalServerError, renderer=notFound)
+    @view_config(context=HTTPBadRequest, renderer=notFound)
+    @notfound_view_config(renderer=notFound)
+    def notfound(self):
```
tsauerwein: This view is not only used document pages, right? So maybe it would make sense to move this declaration from `views/document.py` to `views/__init__.py`?

c2corg_ui/views/document.py
```diff
@@ -221,3 +222,13 @@ def _diff(self):
             return self.template_input
 
         raise HTTPNotFound()
+
+    notFound = 'c2corg_ui:templates/404.html'
+
+    @view_config(context=HTTPInternalServerError, renderer=notFound)
+    @view_config(context=HTTPBadRequest, renderer=notFound)
```
tsauerwein: Do you also need to add a `@view_config` for `HTTPNotFound`?

tsauerwein: Ok, this is done with `@notfound_view_config`. :+1:

c2corg_ui/templates/waypoint/index.html
```diff
@@ -19,7 +19,7 @@
     from json import dumps, loads
     features = [];
     for waypoint in waypoints:
-        if waypoint['geometry']['geom'] is not None:
+        if waypoint.get('geometry'):
```
tsauerwein: Should this be:

     if waypoint.get('geometry') and waypoint.get('geometry').get('geom'): 

?

tsauerwein:
Cool idea!

c2corg_ui/templates/waypoint/index.html
ginold: maybe it's even better, but get('geometry') already fixed the problem.

less/404.less
ginold: no, it does not. I saw a js library that did take care of  adding all the vendor prefixes up to date, but less doesn't do it by itself.

c2corg_ui/views/document.py
ginold: maybe it would make more sense indeed. But where should i put it ? copy pasting it into init.py gives me this error : `Unbound method calls are not supported, please set the class as your view and the method as your attr
`

tsauerwein: There are two ways you can declare views with Pyramid, either as [a method of a class](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/viewconfig.html#view-defaults-class-decorator) or as a [distinct function](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/viewconfig.html#adding-view-configuration-using-the-view-config-decorator). The error message sounds like you tried to mix the two.

c2corg_ui/__init__.py
```diff
@@ -66,3 +71,18 @@ def _add_static_view(config, name, path):
     )
     config.add_cache_buster(path, version_cache_buster)
     CACHE_PATH.append(name)
+
+class NotFound():
```
tsauerwein: Please move this into `c2corg_ui/views/__init__.py`. All views defined in the `views` package.

c2corg_ui/views/document.py
```diff
@@ -101,8 +100,8 @@ def _get_archived_document(self, id, lang, version_id):
         if resp['status'] == '404':
             raise HTTPNotFound()
         elif resp['status'] != '200':
-            raise HTTPInternalServerError(
-                "An error occured while loading the document")
+            raise HTTPInternalServerError("An error \
```
tsauerwein: Does the string now contain the line-break?

c2corg_ui/templates/waypoint/index.html
```diff
@@ -19,7 +19,7 @@
     from json import dumps, loads
     features = [];
     for waypoint in waypoints:
-        if waypoint['geometry']['geom'] is not None:
+        if waypoint.get('geometry') and waypoint.get('geometry').get('geom'):
```
asaunier: Why do you need to do this in a 404 page PR?
Is it really related? If not please move the commit in a separated PR.

c2corg_ui/static/js/map/map.js
```diff
@@ -59,7 +59,14 @@ app.module.directive('appMap', app.mapDirective);
...
  */
-app.MapController = function($scope, mapFeatureCollection) {
+app.MapController = function($scope, mapFeatureCollection, $attrs) {
+
+
+  /**
+   * @type {number}
+   * @private
+   */
+  this.zoom_ = $attrs['zoom'];
```
asaunier: Why do we need this?
Is it really related to the 404 page?

asaunier: Maybe you need it for showing the map in the 404 page?

But do we really need to show a map in the 404 page?
I would prefer that we have no map in the 404 page (use for what?) rather than adding stuff in the JS code, useful only for it(?).

c2corg_ui/templates/404.html
```diff
@@ -0,0 +1,56 @@
...
+<header class="page-main-title">
+  <h3><span translate>unknown page</span></h3>
+</header>
+
+<section class="not-found-section">
+  <h1 class="green-text text-center text-uppercase">oops</h1>
+  <h4 translate class="text-center">The page you are looking for does not exist or is broken.</h4>
+  <h4 class="text-center"><span translate>You can go to the</span> <a href="javascript:history.go(-1)" translate>previous page</a> <span translate>or try to go to other pages:</h4>
+  <ul class="not-found-pages">
+    <a class="list-item text-center" href="/"><span translate>Home</span><br><span class="glyphicon glyphicon-home"></span></a>
```
asaunier: Shouldn't you rather use the usual Pyramid route names with ``${request.static_url(...``?

asaunier:
I admit the resulting 404 page is really cool, including the bermuda triangle recentering and the swinging falling climber icon :P 

As far as I can see it also takes into account the 500 errors?

c2corg_ui/static/js/map/map.js
```diff
@@ -59,7 +59,14 @@ app.module.directive('appMap', app.mapDirective);
  * @export
  * @ngInject
  */
-app.MapController = function($scope, mapFeatureCollection) {
+app.MapController = function($scope, mapFeatureCollection, $attrs) {
```
asaunier: Isn't ``$attrs`` missing in the params?
Instead of using it, I think it would be more consistent to add ``zoom`` in the ``scope`` param of the map directive.

c2corg_ui/static/js/map/map.js
asaunier: Well, OK for this change but please keep the consistency of the map directive attributes (see my comment above about the scope param).

ginold: I proposed a map because I put there an easter egg (waypoint lost in the bermuda's triangle)  ;) 
I needed this change while I was working on this PR.

ginold:
Yes, it does take the 500 page as well.

c2corg_ui/__init__.py
```diff
@@ -68,3 +73,19 @@ def _add_static_view(config, name, path):
...
+    notFound = 'c2corg_ui:templates/404.html'
+
+    def __init__(self, request):
+        self.settings = request.registry.settings
+
+    @view_config(context=HTTPInternalServerError, renderer=notFound)
+    @view_config(context=HTTPBadRequest, renderer=notFound)
+    @notfound_view_config(renderer=notFound)
+    def notfound(self):
+        return {
```
asaunier: This view seems to always return a status code 200 even though the caught error is 400/404/500 => the tests fail because the expected code is not returned.

Should we change the tests or forward the status code retrieved in some response object?

tsauerwein: I think for "not found" the status code should really be 404, so that search engines correctly index the site. So, maybe we should forward the status code?

asaunier: Yep that sounds relevant.
Just not sure how one can get the response status code and forward it in this ``notfound`` view,

tsauerwein: Untested, but in the [docs](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/viewconfig.html#view-config-placement) it sounds like you can get the `context`:

> When the decorator is used against a method of a class, a view is registered for the class, so the class constructor must accept an argument list in one of two forms: either a single argument, request, or two arguments, context, request.

Something like:

```
    def __init__(self, context, request):
        self.context = context
        self.settings = request.registry.settings
```

asaunier:
I have just pushed 2 additional commits:
* b2a56a6 uses @tsauerwein 's suggestion to forward the status code (by the way it makes the tests pass)
* 87e4e3e use a more consistent approach to pass a default zoom to the map.

ginold:
looks good to me :relieved:

asaunier:
I have made a few minor changes (for instance displayed in the 404 page the ``detail`` attribute of the HTTPException if any) and rebased the commits. Will merge when Travis passes.

tsauerwein:
> for instance displayed in the 404 page the detail attribute of the HTTPException if any


For example what would be displayed? Would that information be relevant to the user? It would make more sense to log the error, and show a generic message to the user.

asaunier:
The message are already quite generic. For instance:
https://github.com/c2corg/v6_ui/blob/master/config/alex#L6

asaunier:
I don't think we should spend more time on this part now. Travis is OK. OK for merging too?

tsauerwein:
:+1:

asaunier:
Thanks :)

Issue #306, opened 2016-04-13
Tool for reporting a problem in a page (2d)


fjacon:
That would be for instance an entry in the "plus" menu tool.
It would open a textarea to type a message and send an email (or create a trello card?) to moderators with the link of the concerned document.

tsauerwein:
It could be simply a link to a Google form: https://www.google.com/forms/about/

fjacon:
Good Idea. @asaunier ?

asaunier:
Yep.
Please note the association has a trello board for monitoring task. What about adding cards to this Trello using the Trello API?

fjacon:
are you talking specifically for this issue ?

asaunier:
I mean the reporting tool could add cards to the association trello.

ginold:
the "plus" button is available only in the detail-view page, and we would like to have this tool visible on every page, don't we ?

asaunier:
> the "plus" button is available only in the detail-view page, and we would like to have this tool visible on every page, don't we ?

What would be the use of this "Plus" button in other pages? As for now it only contains detail-view-related actions that we don't need in other pages.

Adding a "report a problem" action to the "plus" button would make sense. Not sure we need such an action for pages other than detail-views.

stef74:
For the the reported possibility is very relevant on document details.

We can probably easy a link add on "credit and contact"

stef74:
@loicperrin @desnoes could you please discuss a simple solution ... the google form approach can  be good solution. Please also coordinate with forum team.

loicperrin:
As I understand this will be fixed with #317

dkocich:
What about trying something like https://www.hotjar.com/ ? Next to other things there is also an option to quickly setup various feedback polls and surveys (similar to what we were implementing).

asaunier:
> What about trying something like https://www.hotjar.com/ 

Isn't it a tool to figure out how people use an interface? Here we only need a simple way to make people report a problem with the current page and provide a few data (eg. page URL, message, user name and email) to moderators.

PR #317, opened 2016-04-19
[WIP] reporting tool

ginold:
a suggestion for the reporting tool, not yet functional. 
It can be accessed from the sidemenu at the bottom under "report a problem" or within the "more" button on the detail-view pages.


c2corg_ui/static/js/reportproblemdirective.js
```diff
@@ -0,0 +1,106 @@
...
+  }, 1000)
+}
+
+
+/**
+ * @param {Object} problem {text, subject}
+ * @export
+ */
+app.ReportProblemModalController.prototype.send = function(problem) {
+  console.log(problem)
```
asaunier: Please remove the console.log

asaunier: Where do you use the ``problem`` argument?

asaunier:
This looks like a good start. We just have to figure what to do with the report: send it by email to the moderators? By the way some data about the person that made the report would be needed to answer him! For instance his name and email address (could be detected automatically for auth users).
We also need the URL of the page where the problem has been reported.

I would move the "report a problem" action near "Print" or "Save" rather than near "Versions" or "Translate".

c2corg_ui/static/js/reportproblemdirective.js
ginold: you mean where the `problem` object is initialized? 
it's set  in `base.html  <div id="report-problem" report-problem>` as `ng-model="problem.text" `and `ng-model="problem.subject"`

ginold:
good idea. All those infos can be obtained from the authService and from window.location.

ginold:
do we want to merge this 'as is' without being functional (like search-filters), just the UI or 'proof of concept', or we want to wait until we decide how to make it fully work?

fjacon:
I think using a google form in the end would be a good idea.
See for example what we did for this project 
http://www.thinkhazard.org/
login : Frederic Jacon
passw : fjacon
Search for indonesia for example
then click on "provide feedback" on bottom of page
So we could prepare a simple form as a proof of concept.

stef74:
The simple solution : send mail to topo@camptocamp.org with form (as in v5)

asaunier:
@ginold Do you think this PR could still be integrated to add this report tool?

It would also require a server-side service (UI? API?) that will receive the data (user data - including email! -, message and page URL) and send an email to the moderators. Please note that the email address is not topo@camptocamp.org  but topo-fr@... (topo-it@ when the user interface is in italian).
Not that simple actually :P

asaunier:
Please note the related issue is tagged "6.2":
https://github.com/c2corg/v6_ui/issues/306
=> low prio

ginold:
Yes It could be integrated, it's just simple html + css, nearly no JS so there is all the place needed to complete the requirements.

Issue #246, opened 2016-03-23
Keep the documentediting controller as generic as possible


asaunier:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/documentediting.js contains some document_type specific parts (waypoint_type, date, etc.).

It would be better that specific data/parts are added for instance in inheriting controllers (possible?).

asaunier:
I feel a bit concerned about the ``DocumentEditingController`` because it seems to me it integrates more and more code that is document_type-specific.

I think we should have a generic ``DocumentEditingController`` that could be extended into ``OutingEditingController``, ``WaypointEditingController`` if needed.

We could define dedicated directives as we do now, since it would not duplicate much code.
Or maybe we can change the directive controller according to some directive attribute?

What do you think?

ginold:
Maybe simply create OutingService ,WpService and so on, where we would put every doc-specific functions and call them as needed?

Issue #198, opened 2016-02-25
Tool to create a new document from the currently displayed one


asaunier:
Related to https://github.com/c2corg/v6_ui/issues/191

As in v5, the users should be able to create routes from a WP page, or an outing from a route page.
Outing creation is a bit special since people generally know if they already have created an outing.

For creating a route, in v5 there is a "create a new route" from every summit page:
http://www.camptocamp.org/summits/37511/fr/tete-a-vincent#routes

In v6 we could have the same. But as suggested by @ginold it would make sense that users use the association tool first to figure out if the route does not already exist. If it doesn't, the association tool should provide a "create route" button that would open the route creating form (with the WP pre-associated).
We need to find a way to make this process obvious to users...

fjacon:
to be closed ?
see also #375

asaunier:
Clearly not.

This is a global issue.
For instance how can a user create a new route?
From the advanced search page ? From the simple search widget? From a "create a new route associated to this WP" in the WP detail view pages?

stef74:
@loicperrin @desnoes could you discuss on this point internally ?

loicperrin:
We propose to add a button in the WP detail view pages to create a new route associated to this WP.

asaunier:
> We propose to add a button in the WP detail view pages to create a new route associated to this WP.

Actually we have not placed such a button because we thought that people should search if the route does not already exist, using the association tool.
When searching for an association, if there is no result, the association tool offers buttons to create new associated documents:
<img width="602" alt="capture d ecran 2016-11-05 a 09 43 27" src="https://cloud.githubusercontent.com/assets/1192331/20028838/96e1b618-a33c-11e6-9843-5162ffbf7205.png">

There is still in issue though, because such buttons are only shown if no result can be found, which does not happen that often because the search engine finds a bit too many results I think and because for routes, if the name is quite common and used for other unrelated routes (eg. "face N"), you'll find a bunch of unrelevant results.
Perhaps the "create a new ..." tool should appear after each set of results anyway?

I have created a dedicated issue: https://github.com/c2corg/v6_ui/issues/849

We could also add an explicit button in the list of associated routes (eg. "Add a new route to this WP"), a bit like for routes/outings:
<img width="707" alt="capture d ecran 2016-11-05 a 09 41 09" src="https://cloud.githubusercontent.com/assets/1192331/20028880/69347682-a33d-11e6-97ab-69b3fd20ec8c.png">
[...]

PR #318, opened 2016-04-20
Add new outing to existing route from the details-view

### LINKED ISSUE #269: "Add an outing" button in route detail view
When clicking on "Add an outing" in route detail view, the opened URL should contain a parameter giving the document_id of the route.
This route should then be directly listed in the associated routes of the outing.

Related to https://github.com/c2corg/v6_ui/issues/227

ginold:
We're halfway through! 
On a route view page, if you click on 'add new outing to this document' it will redirect to outings/add with a modified URL that has additional values, for example `outings/add?docToAssociateId=736485?doctype=routes`.
Then it will try to get it from the API and push into [associations][routes] and it will show up under route taken - > associated routes.
It would work for any kind of document, but for now we only want to let associate a route to a new outing (as I understood).

Now, the other half of the job is that the API should manage this case of associating a doc to an inexistent outing. 

fixes #269

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -207,6 +207,21 @@ app.Api.prototype.updateDocument = function(module, id, json) {
   return promise;
 };
 
+/**
+ * @export
+ * @param {number} id Document id.
+ * @param {string} doctype the first letter of document type (o, w, r...)
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.getDocumentByIdAndDoctype = function(id, doctype) {
+  var alerts = this.alerts_;
+  var promise = this.getJson_('/search?q=' + id + '&t=' + doctype);
+  promise.catch(function(response) {
+    alerts.addError(response);
+  });
+  return promise;
+};
+
 
 /**
  * @param {Object} response Response from the API server.
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -46,6 +46,7 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @param {angular.JQLite} $element Element.
  * @param {angular.Attributes} $attrs Attributes.
  * @param {app.Authentication} appAuthentication
+ * @param {ngeo.Location} ngeoLocation ngeo Location service.
  * @param {app.Alerts} appAlerts
  * @param {app.Api} appApi Api service.
  * @param {string} authUrl Base URL of the authentication page.
@@ -54,7 +55,14 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @export
  */
 app.DocumentEditingController = function($scope, $element, $attrs,
-    appAuthentication, appAlerts, appApi, authUrl) {
+    appAuthentication, ngeoLocation, appAlerts, appApi, authUrl) {
+
+  /**
+   * @type {ngeo.Location}
+   * @private
+   */
+  this.ngeoLocation_ = ngeoLocation;
+
 
   /**
    * @type {angular.Scope}
@@ -147,6 +155,11 @@ app.DocumentEditingController = function($scope, $element, $attrs,
    */
   this.max_steps;
 
+ /**
+  * @type {boolean}
+  * @private
+  */
+  this.submit_ = false;
 
   /**
    * Waypoint init
@@ -164,6 +177,12 @@ app.DocumentEditingController = function($scope, $element, $attrs,
    * @private
    */
   this.api_ = appApi;
+
+  // allow association only new outing  to existing route
+  if (this.ngeoLocation_.hasParam('routes')) {
+    this.urlParams_ = {'routes': this.ngeoLocation_.getParam('routes')};
+    this.pushDocToAssociations_();
+  }
 
   // When creating a new document, the model is not created until
   // the form is touched. At least create an empty object.
@@ -257,10 +276,9 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
   }
 
   if (!this.auth_.isAuthenticated()) {
-    this.alerts_.addError('You have no permission to modify this document');
+    this.alerts_.addError('You must log in to edit this document.');
     return;
   }
-
   // push to API
   var data = angular.copy(this.scope_[this.modelName_]);
   if (!goog.isArray(data['locales'])) {
@@ -295,6 +313,8 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
     delete data['read_lonlat'];
   }
 
+  this.submit_ = true;
+
   if (this.id_) {
     // updating an existing document
     if ('available_langs' in data) {
@@ -317,6 +337,16 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
   } else {
     // creating a new document
     this.lang_ = data['locales'][0]['lang'];
+    if (this.modelName_ === 'outing') {
+      // adapt the Object for what's expected at the API side + format the outing
+      this.formatOuting_(data);
+      data = {
+        'outing': data,
+        'route_id': data['associations']['routes'][0]['document_id'],
+        'user_ids': [this.auth_.userData.id]
+      };
+    }
+
     this.api_.createDocument(this.module_, data).then(function(response) {
       this.id_ = response['data']['document_id'];
       window.location.href = app.utils.buildDocumentUrl(
@@ -377,6 +407,18 @@ app.DocumentEditingController.prototype.handleMapFeatureChange_ = function(
       'latitude': coords[1]
     };
   }
+};
+
+
+/**
+ * It doesn't associate, it just pushes it into associations array.
+ * @private
+ */
+app.DocumentEditingController.prototype.pushDocToAssociations_ = function() {
+  var doctype = Object.keys(this.urlParams_)[0];
+  this.api_.getDocumentByIdAndDoctype(this.urlParams_[doctype], doctype[0]).then(function(doc) {
+    this.scope_[this.modelName_]['associations'][doctype].push(doc.data[doctype].documents[0]);
+  }.bind(this));
 };
 
 
@@ -524,7 +566,6 @@ app.DocumentEditingController.prototype.animateBar_ = function(step, direction) 
  */
 
 app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType) {
-
   if (app.constants.STEPS[waypointType]) {
     this.max_steps = app.constants.STEPS[waypointType];
   } else {
@@ -536,31 +577,52 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType) 
  * @param {Object} outing
  * @private
  */
-
 app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
-  // creating a new outing -> init locales
+  if (this.submit_) {
+    if (outing['locales'][0]['conditions_levels'][0]['level_place'].length > 0) {
+      // transform condition_levels to a string
+      outing['locales'][0]['conditions_levels'] = JSON.stringify(outing['locales'][0]['conditions_levels']);
+    } else {
+      delete outing['locales'][0]['conditions_levels'][0];
+    }
+    // if no date end -> make it the same as date start
+    if (!outing['date_end'] && outing['date_start'] instanceof Date) {
+      outing['date_end'] = outing['date_start'];
+    }
+  }
+  // creating a new outing -> init locales and associations
   if (!outing['locales']) {
     outing['locales'] = [{}];
   }
-
-  outing['date_start'] = app.utils.formatDate(outing['date_start']);
-  outing['date_end'] = app.utils.formatDate(outing['date_end']);
-
-  // conditions_levels -> to JSON and snow -> INT
-  if (outing['locales'][0]['conditions_levels']) {
-    outing['locales'][0]['conditions_levels'] = JSON.parse(outing['locales'][0]['conditions_levels']);
-    for (var i = 0; i < outing['locales'][0]['conditions_levels'].length; i++) {
-      var cond = outing['locales'][0]['conditions_levels'][i];
-      cond['level_snow_height_soft'] = parseInt(cond['level_snow_height_soft'], 10);
-      cond['level_snow_height_total'] = parseInt(cond['level_snow_height_total'], 10);
-    }
-  } else {
-    // init empty conditions_levels for ng-repeat
-    outing['locales'][0]['conditions_levels'] = [{'level_snow_height_soft': '', 'level_snow_height_total': '', 'level_comment': '', 'level_place': ''}];
+  if (!outing['associations']) {
+    outing['associations'] = {'routes': [], 'waypoints': [], users: []};
   }
 
+  // convert existing date from string to a date object
+  if (outing['date_end'] && typeof outing['date_end'] === 'string') {
+    outing['date_end'] = app.utils.formatDate(outing['date_end']);
+  }
+
+  if (outing['date_start'] && outing['date_start'] === 'string') {
+    outing['date_start'] = app.utils.formatDate(outing['date_start']);
+  }
+
+  var conditions = outing['locales'][0]['conditions_levels'];
+  // conditions_levels -> to Object, snow_height -> to INT
+  if (conditions && typeof conditions === 'string') {
+    conditions = JSON.parse(conditions);
+    for (var i = 0; i < conditions.length; i++) {
+      conditions[i]['level_snow_height_soft'] = parseInt(conditions[i]['level_snow_height_soft'], 10);
+      conditions[i]['level_snow_height_total'] = parseInt(conditions[i]['level_snow_height_total'], 10);
+    }
+  } else {
+    if (!this.submit_) {
+      // init empty conditions_levels for ng-repeat
+      outing['locales'][0]['conditions_levels'] = [{'level_snow_height_soft': '', 'level_snow_height_total': '', 'level_comment': '', 'level_place': ''}];
+    }
+  }
   return outing;
-}
+};
 
 
 /**
@@ -573,7 +635,7 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
  */
 app.DocumentEditingController.prototype.pushToArray = function(object, property, value, event) {
   app.utils.pushToArray(object, property, value, event);
-}
+};
 
 
 /**
--- a/c2corg_ui/templates/route/view.html
+++ b/c2corg_ui/templates/route/view.html
@@ -12,6 +12,7 @@ from c2corg_ui.templates.utils import get_lang_lists
 
 <%
 route_id = route['document_id']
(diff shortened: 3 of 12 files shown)


tsauerwein:
> Now, the other half of the job is that the API should manage this case of associating a doc to an inexistent outing.

I am not sure I understand.

When you create a new outing with the API, you can specify a route id. The API will then create an association between the new outing and the route.

See: https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_outing.py#L324

ginold:
Ok! i thought it wasn't implemented. 
I now see it will change a little the createDocument function we use to send data to the API.

In contrast to waypoints or routes, outings have a second level that the others dont have: the API is awaiting `body : { outing: {}, route_id : xxx, user.ids: [] }`, the others `body : {route_content}` is enough.
Until now, we were always sending `body : content` and i was wondering why the API says there is no outing at all. 

Would it be possible to look for the route_id in [associations][routes][0] or create something like waypoint has -> main_route_id ? Same idea for users, there is already [associations][users], no need for a separate property.

tsauerwein:
> In contrast to waypoints or routes, outings have a second level that the others dont have: the API is awaiting body : { outing: {content}, route_id : xxx, user.ids: [] }, the others body : {content} is enough.

True, the API for outings expects a different input format.

> Would it be possible to look for the route_id in [associations][routes][0] or create something like waypoint has -> main_route_id ? Same idea for users, there is already [associations][users], no need for a separate property.

`associations` is not considered at all, so you don't even have to send it.

tsauerwein:
For the other document types, can you also add/remove associations from the edit page? Or is this only the case for user and route associations when creating an outing? (also when editing an outing?)

c2corg_ui/static/js/documentediting.js
```diff
@@ -374,6 +394,23 @@ app.DocumentEditingController.prototype.handleMapFeatureChange_ = function(
 
 
 /**
+ * It doesn't associate, it just pushes it into associations array.
+ * @private
+ */
+app.DocumentEditingController.prototype.pushDocToAssociations_ = function() {
+  var doctype = this.urlParams_['doctype'];
+  var docid = this.urlParams_['docToAssociateId'];
+  $.get(this.apiUrl_ + '/' + doctype + '/' + docid, function(doc) {
```
tsauerwein: Here you could use the search service. Otherwise you will load the full document including the geometry, association, ...

c2corg_ui/static/js/documentediting.js
```diff
@@ -551,10 +589,23 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
...
   }
-
   return outing;
 }
 
+/**
+ * @returns {Object} url parameters
+ * @private
+ */
+app.DocumentEditingController.prototype.parseUrl_ = function() {
```
tsauerwein: The Angular service $location could take care of that:
https://docs.angularjs.org/api/ng/service/$location#search

c2corg_ui/static/js/documentediting.js
ginold: You mean typeahead/bloodhound?

tsauerwein: I mean if you make a request to `/search?q={document_id}&t=r` instead of `/routes/{document_id}` you will not get the full document but only the information you need.

ginold: right. Where can i get the full list of possible arguments for filtering? 
(and btw can you tell me where is the full doc with steps for indexing elastic search? it is not https://github.com/c2corg/v6_api, maybe somewhere deeper in the directory but i don't remember)

tsauerwein: > right. Where can i get the full list of possible arguments for filtering?

https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/search.py#L41-L62

> and btw can you tell me where is the full doc with steps for indexing elastic search?

Do you mean this?
https://github.com/c2corg/v6_api/tree/master/c2corg_api/scripts/migration#initialize-elasticsearch

c2corg_ui/static/js/documentediting.js
ginold: it doesn't work for me, it gives back some wierd stuff like `add?docToAssociateId : 736688?doctype=routes`.

tsauerwein: What does your URL look like?

ginold: /outings/add?docToAssociateId=736688?doctype=routes

tsauerwein: The URL should be `/outings/add?docToAssociateId=736688&doctype=routes`.

ginold: yeah i have just realized :dizzy_face:

ginold: $location works as good as the function i had created, but now changes of URL don't redirect anymore. Clicking on the menu changes the url, but does nothing. It works on the other pages where we don't use $location.

tsauerwein: Ok, thanks for testing.

But then we have to fix your function.

c2corg_ui/static/js/documentediting.js
```diff
@@ -551,10 +589,23 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
...
+  var query = decodeURIComponent(location.search.substr(1));
+  var result = {};
+  query.split('?').forEach(function(part) {
```
tsauerwein: `split('&')`

c2corg_ui/static/js/documentediting.js
```diff
@@ -551,10 +589,23 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
...
+    part = part.split('=');
```
tsauerwein: This could be `var parts = ...` to make it clear that we are dealing with an array.

c2corg_ui/static/js/documentediting.js
ginold: True. So you are ok to keep my function in that case?

c2corg_ui/static/js/documentediting.js
```diff
@@ -147,6 +156,17 @@ app.DocumentEditingController = function($scope, $element, $attrs,
...
 
+  // allow association only new outing -> to -> existing route
+  if (this.parseUrl_() && this.parseUrl_()['doctype'] === 'routes') {
+    /**
+     * docToAssociateId=2332?doctype=routes...
+     * -> {docToAssociateId: 2332, doctype: 'routes'}
+     * @type {Object}
+     * @private
+     */
+    this.urlParams_ = this.parseUrl_();
```
tsauerwein: Could you avoid calling `this.parseUrl_()` 3-times?

c2corg_ui/static/js/documentediting.js
tsauerwein: Yes, that's ok. If there would have been a function in one of the libraries that we are already using, I would have preferred that function instead of a re-implementation.

tsauerwein: FYI: This is how the function, that is used by $location, is implemented in Angular:
https://github.com/angular/angular.js/blob/5d695e5566212d93da0fc1281d5d39ffee0039a3/src/Angular.js#L1321-L1350

But we can keep it simple and use your function, that works in our case.

c2corg_ui/static/js/documentediting.js
```diff
@@ -374,6 +389,25 @@ app.DocumentEditingController.prototype.handleMapFeatureChange_ = function(
 
 
 /**
+ * It doesn't associate, it just pushes it into associations array.
+ * @private
+ */
+app.DocumentEditingController.prototype.pushDocToAssociations_ = function() {
```
asaunier: You should rather use the ``apiService`` https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/apiservice.js

c2corg_ui/static/js/documentediting.js
```diff
@@ -551,12 +586,28 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
...
   return outing;
 }
 
 
 /**
+ * get parameters from the URL separated by &
+ * @returns {Object} url parameters
+ * @private
+ */
+app.DocumentEditingController.prototype.parseUrl_ = function() {
```
asaunier: Do you really need this?

When creating a route from a WP page the URL would be http://.../routes/add?waypoint=123456
When creating an outing from a route page the URL would be http://.../outings/add?route=123456

I also think we should consider extending the documentEditing directive depending on the document type. See https://github.com/c2corg/v6_ui/issues/246

asaunier:
> For the other document types, can you also add/remove associations from the edit page?

It would be more consistent to be able to do so for all types of documents.
If you can choose the associated routes while creating an outing, why not being able to choose the associated waypoints when creating a route. By the way this means one could create a route without having to find the main WP page first (of course we have to make sure somewhere that the route does not already exist).

> (also when editing an outing?)

Creating/updating an outing should work the same.

c2corg_ui/static/js/documentediting.js
ginold: I specified that I don't want to 'really' associate the doc, I just push it into the array of the not-yet-existing doc. Maybe I don't know how it works behind, on the API side.
If you tell me that a virtual document ID is created and even at this point I can already associate the existing doc to this non existing, then you are right - I could use apiservice.associateDocument

c2corg_ui/static/js/documentediting.js
ginold: I still have to parse the url for waypoint=12345, so yes.

asaunier: OK but I think it's not the right way to do it. In ``ngeo`` there is a ``ngeo.Location`` service that can be used to get/update/delete the URL parameters.
Of course here we are not in a map context, but the service is still available and will be used in other pages (in the advanced search pages I guess).
See https://camptocamp.github.io/ngeo/master/apidoc/ngeo.Location.html and http://camptocamp.github.io/ngeo/master/examples/permalink.html

c2corg_ui/static/js/documentediting.js
tsauerwein: I am not sure if I understand you, but creating the association will not directly be done. You open the wizard to create a new outing, and the route is simply added to the associations list. When you send the request to the API, the outing will be created and then the specified associations (this is not yet supported by the API).

asaunier: I mean that for interacting with the API it's recommended to use (or create if not already available) a function from the apiService instead of adding the function directly in your directive. Ask @gberaudo for precisions :)

fjacon:
Lié à https://github.com/c2corg/v6_ui/issues/198

asaunier:
> True, the API for outings expects a different input format.

Why this difference?
Unfortunately it requires to have a special case in the ``documentediting`` directive that must use a different format for the request when editing outings.
In addition outings are not that different from other documents. Of course we must handle associated users and routes, but it's quite the same for routes: we could also send WP to associate to the edited route.

As for now, the outing editing fails because the API replies that:
> * outing is missing
* route_id is missing
* user_ids is missing

=> what is ``route_id``? At least it should be ``route_ids`` because we may associate several routes to the outing.

I also wonder if it would not make sense that the POST/PUT request uses the ``associations`` attributes (as received in the GET request) to send the associations back, including routes and users.
Then we could use the same request format for all document types and manage the documents listed as to be associated during the editing.

tsauerwein:
> Why this difference?

The API for outings was made with the outing creation wizard of v5 in mind. You need a route and users to create an outing. You don't need that information for other documents. That's the difference.

But if we want that associations can be updated when updating a document, we can make it more consistent and accept an `associations` field when creating/updating a document.

asaunier:
> The API for outings was made with the outing creation wizard of v5 in mind. You need a route and users to create an outing.

Indeed

> You don't need that information for other documents.

Yes and no. For instance for routes, you need at least a WP, I think.

tsauerwein:
> Yes and no. For instance for creating a route, you need at least a WP, I think.

Which is given as the "main waypoint".

asaunier:
True :)

c2corg_ui/static/js/documentediting.js
gberaudo: True, and since `this.api_.getJson_` is private it should not even compile!

c2corg_ui/static/js/apiservice.js
```diff
@@ -88,7 +88,7 @@ app.Api.prototype.putJson_ = function(url, json) {
 /**
  * @param {string} url Url suffix
  * @return {!angular.$http.HttpPromise}
- * @private
+ * @export
```
gberaudo: :rage: @ginold, this method should stay private!

c2corg_ui/static/js/documentediting.js
```diff
@@ -374,6 +390,25 @@ app.DocumentEditingController.prototype.handleMapFeatureChange_ = function(
...
 
 /**
+ * It doesn't associate, it just pushes it into associations array.
+ * @private
+ */
+app.DocumentEditingController.prototype.pushDocToAssociations_ = function() {
+  var doctype = Object.keys(this.urlParams_)[0];
+  var url = '/search?q=' + this.urlParams_[doctype] + '&t=' + doctype[0];
+  // get the doc from the API like /search?q=12345&t=r  (routes)
+  this.api_.getJson_(url).then(function(doc) {
```
gberaudo: As explained by @asaunier, this function must be moved to apiservice.js.

c2corg_ui/static/js/documentediting.js
ginold: I made it public in another pr because I needed it a part of other functions. It was only accessed through `readDocument` and `readAccount`. ReadDocument sounds ok, but this function forces me to use a URL that I don't want. 
Is it ok to make getJson public ?

c2corg_ui/static/js/apiservice.js
ginold: then how do you want me to access it without passing through `readDocument` because it's not what I want? shall I create another readDocument2(URL) that calls getJson(url) ? :P

tsauerwein: That sounds like a good idea. Maybe call it `searchDocumentById`.

c2corg_ui/static/js/documentediting.js
ginold: I don't think that pushing into [assosiations] of a document being edited is the job of apiservice. Apiservice shall get the document, but it's not its job to push it, is it?

tsauerwein: I think that is not what @gberaudo means. Create a function `searchDocumentById(document_id, doc_type)` which is called from here. The idea is that calls to the API are made with this service.

c2corg_ui/static/js/apiservice.js
gberaudo: +1

c2corg_ui/static/js/documentediting.js
ginold: searchById doesn't imply that you have to send and ID AND a doctype, but yes, we know that it will make the search quicker :) andSearchByIDandDoctype is hideous :D

c2corg_ui/static/js/documentediting.js
```diff
@@ -54,7 +55,14 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @export
  */
 app.DocumentEditingController = function($scope, $element, $attrs,
-    appAuthentication, appAlerts, appApi, authUrl) {
+    appAuthentication, appAlerts, appApi, authUrl, apiUrl, ngeoLocation) {
```
gberaudo: `apiUrl` should be removed right?

c2corg_ui/static/js/documentediting.js
```diff
@@ -257,10 +276,9 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
   }
 
   if (!this.auth_.isAuthenticated()) {
-    this.alerts_.addError('You have no permission to modify this document');
+    this.alerts_.addError('You have no permission to edit this document because you are not identified');
```
gberaudo: Maybe: `You must log in to edit a document`.

c2corg_ui/static/js/documentediting.js
ginold: true, thank's for noticing

c2corg_ui/static/js/documentediting.js
```diff
@@ -536,31 +577,52 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
  * @param {Object} outing
  * @private
  */
-
 app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
-  // creating a new outing -> init locales
+  if (this.submit_) {
+    if (outing['locales'][0]['conditions_levels'][0]['level_place'].length > 0) {
```
gberaudo: I think the formatOuting method is quite confusing.
As discussed, I would better see these low level conversions in the apiservice.
It may be too much for this already big PR, so feel free to try it later.

c2corg_ui/templates/route/view.html
```diff
@@ -105,69 +106,76 @@
...
+          </section>
+        </span>
+      </div>
+    % endif
+
     <div class="view-details-associations col-xs-12 tab associations">
       <span class="lead">
         <section>
-          <div ng-show="userCtrl.auth.isAuthenticated()">
+          <div ng-show="userCtrl.auth.isAuthenticated()" style="float: left;">
```
gberaudo: Maybe better with a CSS class?

c2corg_ui/static/js/documentediting.js
gberaudo: The ordering of the parameters is also unconsistent. I am wondering why you don't get an error from closure.

c2corg_ui/static/js/documentediting.js
ginold: You are super right, but as you stated I would prefer to correct it in the future, when the API will accept conditions_levels as objects or null (this will greatly simplify the code already by itself) and when we decide to tidy up the code and create document-specific controllers, in order to make the documentediting directive more generic.

c2corg_ui/static/js/documentediting.js
ginold: I've never really taken care of the order of parameters nor it's alphabetical declaration in main.js and I never got an error, hopefully. Don't change that! :dancer:

ginold:
@gberaudo when you suggestions will have been addressed, will it be ok for merging?

gberaudo:
> @gberaudo when you suggestions will have been addressed, will it be ok for merging?

Yes, if there is no objection, please merge.

ginold:
thank's for your review. You can also check #311 witch is most likely 99% done, just waiting for an approval. It could be yours, so hurry up!

c2corg_ui/static/js/documentediting.js
asaunier: Would an issue be useful about this formatOuting method?

ginold: I don't think so, as I said this will surely be part of the PR where we will factorize the code into document-specific controllers, and there are already issues created by you about this.

PR #322, opened 2016-04-22
Check if user has rights to edit an outing (view/edit form)

### LINKED ISSUE #321: Outing edit rights
Users that are not (co)authors of an outing should not be able to edit, am I right?
I can go and change what i want in an outing but at the end I receive an error 'No permission to change this outing', so there is no use to even let user to be redirected to the editing page of an outing if he has no rights to edit it.
This should hide the edit button of the details-view page and redirect user back when he enters directly by typing the URL.

ginold:
fixes #321

### DIFF
--- a/c2corg_ui/static/js/auth/authservice.js
+++ b/c2corg_ui/static/js/auth/authservice.js
@@ -73,6 +73,34 @@ app.Authentication.prototype.isAuthenticated = function() {
 
 
 /**
+ * Checks if the current user has rights to access/edit the document.
+ * TODO: rights to personal documents only for current user.
+ * @return {boolean}
+ * @export
+ */
+app.Authentication.prototype.hasEditRights = function(users) {
+  var userid = this.userData.id;
+  var roles = this.userData.roles;
+  // moderator has rigths
+  if (roles.indexOf('moderator') > -1) {
+    return true;
+  }
+  // we are checking for editing rights of an outing
+  if (users) {
+    for (var i = 0; i < users.length; i++) {
+      if (userid === users[i].id) {
+        return true;
+      }
+    }
+    return false;
+  // by default, user has rights to every doc
+  } else {
+    return true;
+  }
+};
+
+
+/**
  * @param {appx.AuthData} data User data returned by the login request.
  * @return {boolean} whether the operation succeeded.
  */
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -169,7 +169,6 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   // the form is touched. At least create an empty object.
   this.scope_[this.modelName_] = {};
   if (this.auth_.isAuthenticated()) {
-
     if (this.id_ && this.lang_) {
      // Get document attributes from the API to feed the model:
       goog.asserts.assert(!goog.isNull(this.id_));
@@ -231,10 +230,18 @@ app.DocumentEditingController.prototype.successRead_ = function(response) {
   this.scope_[this.modelName_] = data;
 
   if (this.modelName_ === 'outing') {
-    this.scope_['outing'] = this.formatOuting_(this.scope_['outing']);
-    this.differentDates = app.utils.areDifferentDates(this.scope_['outing']['date_start'], this.scope_['outing']['date_end']);
+    var outing = this.scope_['outing'];
+    // check if user has right to edit -> the user is one of the associated users
+    if (this.auth_.hasEditRights(outing['associations']['users'])) {
+      this.scope_['outing'] = this.formatOuting_(outing);
+      this.differentDates = app.utils.areDifferentDates(outing['date_start'], outing['date_end']);
+    } else {
+      this.alerts_.addError('you have no rights to edit this document');
+      setTimeout(function() { // redirect to the details-view page
+        window.location = window.location.href.replace('/edit', '');
+      }, 3000);
+    }
   }
-
   this.scope_.$root.$emit('documentDataChange', data);
 };
 
--- a/c2corg_ui/static/js/user.js
+++ b/c2corg_ui/static/js/user.js
@@ -100,4 +100,12 @@ app.UserController.prototype.logout = function() {
 };
 
 
+/**
+ * @return {boolean}
+ * @export
+ */
+app.UserController.prototype.hasEditRights = function(users) {
+  return this.auth.hasEditRights(users);
+};
+
 app.module.controller('AppUserController', app.UserController);
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -326,12 +326,13 @@
 
 
 ## Float buttons
-<%def name="show_float_buttons(string, doc_id, doc_lang, other_langs, missing_langs)">\
+<%def name="show_float_buttons(outingUsers, doctype, doc_id, doc_lang, other_langs, missing_langs)">\
   <div class="float-buttons" app-view-details>
-    <div class="float-button float-edit" tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}" protected-url-btn url="'${request.route_url( string + '_edit' , id=doc_id, lang=doc_lang)}'">
-      <span class="glyphicon glyphicon-edit"></span>
-      <p class="float-button-text" translate>Edit</p>
-    </div>
+      <div ng-if="::userCtrl.hasEditRights(${outingUsers if outingUsers is not None else ''})" class="float-button float-edit" tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}"
+           protected-url-btn url="'${request.route_url( doctype + '_edit' , id=doc_id, lang=doc_lang)}'">
+        <span class="glyphicon glyphicon-edit"></span>
+        <p class="float-button-text" translate>Edit</p>
+      </div>
     <div class="float-button float-favorites" tooltip-placement="left" uib-tooltip="{{'Add to favorites' | translate}}">
       <span class="glyphicon glyphicon-heart"></span>
       <p class="float-button-text" translate>Favorites</p>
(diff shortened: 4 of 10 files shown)


c2corg_ui/static/js/documentediting.js
```diff
@@ -231,8 +230,18 @@ app.DocumentEditingController.prototype.successRead_ = function(response) {
...
-    this.scope_['outing'] = this.formatOuting_(this.scope_['outing']);
-    this.differentDates = app.utils.areDifferentDates(this.scope_['outing']['date_start'], this.scope_['outing']['date_end']);
+    // check if user has right to edit -> is one of the associated users
+    var hasRight = app.UserController.prototype.hasEditRightsOuting(this.scope_['outing']['associations']['users']);
+
+    if (!hasRight) {
+      this.alerts_.addError('you have no rights to edit this outing');
+      setTimeout(function() {
+        window.location.href = '/';
+      }.bind(this), 3000);
```
gberaudo: There is no need to bind this.

c2corg_ui/static/js/user.js
```diff
@@ -100,4 +100,31 @@ app.UserController.prototype.logout = function() {
 };
 
 
+/**
+ * @private
+ */
+app.UserController.prototype.getUserId_ = function() {
+  if (window.localStorage.getItem('userData')) {
```
gberaudo: You must retrieve the userData from the auth service, not from the local storage.

c2corg_ui/static/js/documentediting.js
```diff
@@ -231,8 +230,18 @@ app.DocumentEditingController.prototype.successRead_ = function(response) {
   this.scope_[this.modelName_] = data;
 
   if (this.modelName_ === 'outing') {
-    this.scope_['outing'] = this.formatOuting_(this.scope_['outing']);
-    this.differentDates = app.utils.areDifferentDates(this.scope_['outing']['date_start'], this.scope_['outing']['date_end']);
+    // check if user has right to edit -> is one of the associated users
+    var hasRight = app.UserController.prototype.hasEditRightsOuting(this.scope_['outing']['associations']['users']);
```
gberaudo: It is a bit strange to access the fonction this way. Maybe the `app.UserController` is not the right place to put it?

ginold: what is wrong with this way?

gberaudo: I think you would better put the logics in an Angular service and directly call the service here instead of going through the UserController prototype.

tsauerwein: You method is defined on `prototype`, this means the method should be called on an instance of `UserController`. As your function does not require a `UserController` instance, you could define it like `app.UserController.hasEditRightsOuting = ...`.

asaunier:
Please note that this user check is also needed for other "personal" documents such as personal images, personal articles and of course user profiles.

I feel more and more uncomfortable with all those document-type-specific pieces of codes in the documentediting directive. I really think specific parts should be added in extended directives. See https://github.com/c2corg/v6_ui/issues/246

ginold:
You are completely right. I think we should merge current prs 'as is' and then tidy everything up within a new PR where we would create the new controllers/services with document-specific functions.

c2corg_ui/static/js/documentediting.js
```diff
@@ -231,8 +230,17 @@ app.DocumentEditingController.prototype.successRead_ = function(response) {
...
-    this.differentDates = app.utils.areDifferentDates(this.scope_['outing']['date_start'], this.scope_['outing']['date_end']);
+    var outing = this.scope_['outing'];
+    // check if user has right to edit -> the user is one of the associated users
+    if (this.auth_.hasEditRightsOuting(outing['associations']['users'])) {
+      this.scope_['outing'] = this.formatOuting_(outing);
+      this.differentDates = app.utils.areDifferentDates(outing['date_start'], outing['date_end']);
+    } else {
+      this.alerts_.addError('you have no rights to edit this document');
+      setTimeout(function () {
+        window.location.href = '/';
```
asaunier: It would be better to redirect to the document view page rather than to the homepage.

c2corg_ui/static/js/user.js
```diff
@@ -100,4 +100,12 @@ app.UserController.prototype.logout = function() {
 };
 
 
+/**
+ * @export
+ */
+app.UserController.prototype.hasEditRightsOuting = function(users) {
+  this.auth.hasEditRightsOuting(users);
```
asaunier: That looks weird to me to have that kind of alias! Can't you directly use the function from the auth service?

c2corg_ui/templates/helpers/view.html
```diff
@@ -326,12 +326,19 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
...
+<%def name="show_float_buttons(doctype, doc_id, doc_lang, other_langs, missing_langs)">\
   <div class="float-buttons" app-view-details>
-    <div class="float-button float-edit" tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}" protected-url-btn url="'${request.route_url( string + '_edit' , id=doc_id, lang=doc_lang)}'">
+    % if doctype != 'outings':
+    <div class="float-button float-edit" tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}" protected-url-btn url="'${request.route_url( doctype + '_edit' , id=doc_id, lang=doc_lang)}'">
       <span class="glyphicon glyphicon-edit"></span>
       <p class="float-button-text" translate>Edit</p>
     </div>
+    % else:
+    <div ng-if="::userCtrl.hasEditRightsOuting(outingUsers)" class="float-button float-edit" tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}" protected-url-btn url="'${request.route_url( doctype + '_edit' , id=doc_id, lang=doc_lang)}'">
```
asaunier: Can't the user permission check be done directly in the ``protected-url-btn`` directive? All checks should be done at the same place.

c2corg_ui/templates/outing/view.html
```diff
@@ -45,6 +45,8 @@
   });
 </%block>
 
+<div class="ng-hide">{{outingUsers = ${outing['associations']['users']} }}</div>
```
asaunier: I'm a bit uncomfortable with that way of doing. Sounds a bit hackish to set global angular variables.
Can't we directly pass the associated users list as a directive attribute (protected-url-btn?) somewhere?

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -71,6 +71,30 @@ app.Authentication.prototype.isAuthenticated = function() {
...
+ * @private
+ */
+app.Authentication.prototype.getUserId_ = function() {
+  return this.userData['id'];
+};
+
+
+/**
+ * Checks if the current user ID is among the outing's associated users.
+ * @return {boolean}
```
asaunier: ``users`` should be added as ``@param``

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -71,6 +71,30 @@ app.Authentication.prototype.isAuthenticated = function() {
   return !!this.userData && !this.isExpired_();
 };
 
+/**
+ * @return {number}
+ * @private
+ */
+app.Authentication.prototype.getUserId_ = function() {
```
asaunier: Do you really need this function? Why not simply calling ``this.userData['id']``?

c2corg_ui/static/js/user.js
gberaudo: @asaunier, it is not possible to call services directly from directive templates.
So this looks consistent with the angularjs philosophy.
Do you know of another way?

asaunier: See my other comment below: can't the check be done directly in the ``protected-url-btn`` directive along with the "is auth" check?

c2corg_ui/static/js/auth/authservice.js
ginold: I don't _need_, but I thought it was a good practice to have getters/setters.

asaunier: At the moment it is used only once, I don't think it is worth having it: the least the best.

ginold: But if you want me to move the checkings to the protected-url-directive, where I will access authService, then I need to leave this function if I want to access the userId, I presume?

c2corg_ui/static/js/user.js
gberaudo: Sure, it would make sense to have the check directly in the directive.

c2corg_ui/static/js/auth/authservice.js
gberaudo: No, the `userData` property is exported so directly readable from everywhere.

c2corg_ui/static/js/user.js
tsauerwein: Shouldn't this method return something?

ginold: We decided with @asaunier to leave this as is for reasons : 

- there is no direct link between checking if an url is protected and showing or not the button. Protected-url-btn just checks if the user clicking is logged and if he can be redirected to the linked page, if not -> auth form. 
- Whereas auth.checkEditRights will decide whether or not the button will be displayed at all. There are documents that are collaborative, private or require special rights and all should be checked separately from the protected-url-btn.

It is true that when when checkEditrights = true, the button will show up and the protected-url-btn checking for the user authentication is useless, because he's already logged in. But I could, for example, log out in between and still have  access to the page, so it's better to double-check.

asaunier:
Looks good to me except that the commits order is not really relevant. For instance the first commit uses the ``hasEditRights`` that is only defined in the third commit!
I would simple merge all commits in a single one.

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -73,6 +73,34 @@ app.Authentication.prototype.isAuthenticated = function() {
 
 
 /**
+ * Checks if the current user has rights to access/edit the document.
+ * TODO: rights to personal documents only for current user.
+ * @return {boolean}
+ * @export
+ */
+app.Authentication.prototype.hasEditRights = function(users) {
+  var userid = this.userData['id'];
```
gberaudo: You should access `userData` attributes with the dot notation.

ginold:
rebased and corrected. Will merge if no objections after travis passes

PR #324, opened 2016-04-26
Add instructions for installing required software packages

asaunier:
I had to reinstall the UI on my new laptop. Here are the packages I had to install before the ``make install`` completed.

There are probably other packages to install since ``make check`` tells me that
> ImportError: No module named 'venusian'

and ``make serve`` complains that
> ImportError: No module named 'paste'

Do you remember how to install them? Should they be added to ``requirements.txt``?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@ Requirements
  * gettext (0.18 or higher)
  * virtualenv (1.7 or higher)
 
+On Debian/Ubuntu systems the required components may be installed using
+
+    sudo apt-get install virtualenv python3-dev nodejs-legacy npm openjdk-7-jdk gettext
+
 Checkout
 --------
 


tsauerwein:
> Do you remember how to install them? Should they be added to requirements.txt?

These seem to be dependencies of [Pyramid](https://github.com/Pylons/pyramid/blob/master/setup.py#L41) and should be installed automatically with Pyramid. Did you run `make install`?

asaunier:
> Did you run make install?

Yes I did.

I have just tried to run a ``cleanall install`` and now ``serve`` and ``check`` pass. So it seems nothing else should be added to the PR. OK for merging it?

tsauerwein:
:+1:

asaunier:
OK, thanks for your help.

PR #325, opened 2016-04-26
Fix waypoints index

ginold:
there was an error on waypoint index page. This fixes it.


asaunier:
Are you sure this PR should not be rebased? :P

asaunier:
Looks good.

Issue #326, opened 2016-04-27
Parent <-- document --> child relationship


ginold:
Associating new documents is easy : I define which one is the parent and which one is the child, and I know the possible valid associations.
But the other way is not that simple. Sometimes I want to dissociate two documents and I obtain an error 'association does not exist' because I have no way of knowing which one is the parent and which one the child. Or do I ? If so, please tell me how I can get to know this relationship.

If it's not the case, maybe one of the possible solutions would be to send two IDs like `dissociateDocuments(ID1, ID2)` and the API will try one way or the other, if the 1st one does not pass (we could do the same, calling a 2nd time the dissociate method if the 1st time is unsuccessful).
Or maybe you have any better ideas?

Sooner or later we would need to have this information UI-side in order to determine the parent and it's children and to show it in the interface accordingly.

tsauerwein:
> If it's not the case, maybe one of the possible solutions would be to send two IDs like dissociateDocuments(ID1, ID2) and the API will try one way or the other, if the 1st one does not pass

Yes, that doesn't sound too bad. Especially since you can not have two associations D1->D2 and D2->D1 at the same time. See https://github.com/c2corg/v6_api/issues/263

tsauerwein:
Done in https://github.com/c2corg/v6_api/pull/264

ginold:
fixed in #328

asaunier:
> fixed in #328

Reopening the issue since the PR is not merged yet (as a result the issue is still valid).

Issue #298, opened 2016-04-13
Association tool is broken


asaunier:
(perhaps an API issue?)

It seems no longer possible to associate for instance a WP to a route, the API returns an error.

For example:
* open http://c2corgv6-demo.gis.internal:6555/routes/45168/fr
* search for "mattes" and pick "Pointe des Mattes"
* a request to the API is sent:
```
http://c2corgv6-demo.gis.internal:6545/associations
POST
child_document_id: 43618
parent_document_id: 45168
``` 
* get an error:
> {"status": "error", "errors": [{"location": "body", "name": "association", "description": "invalid association type"}]}

tsauerwein:
The valid association types are listed here:
https://github.com/c2corg/v6_common/blob/master/c2corg_common/associations.py#L15

    (WAYPOINT_TYPE, ROUTE_TYPE)

The request should send:
```
parent_document_id: 43618 (waypoint)
child_document_id: 45168 (route)
```

ginold:
So apparently we can't associate wp with routes? 
So we have to remove route results from the association tool in the wp detail view.

tsauerwein:
> So apparently we can't associate routes with wp?

Why do you think so?

You only have to change the request from:

```
parent_document_id: 45168 (route)
child_document_id: 43618 (waypoint)
```
to
```
parent_document_id: 43618 (waypoint)
child_document_id: 45168 (route)
```

ginold:
How am i supposed to understand this ?
`    # associations with waypoints
    (WAYPOINT_TYPE, WAYPOINT_TYPE)`

  `  # associations with routes
    (ROUTE_TYPE, ROUTE_TYPE),
    (WAYPOINT_TYPE, ROUTE_TYPE)`

I get it like this : 
  - you can only associate a wp to another wp
  - you can associate a route to a route and a wp to a route

However in the app i get the opposite result : i can associate both docs to a wp, but when trying to associate a wp to a route I get an error, whereas route to route works.
I think i'm lost.

tsauerwein:
An association has a parent and a child document. [c2corg_common/associations.py](https://github.com/c2corg/v6_common/blob/master/c2corg_common/associations.py) lists the possible association types with (parent_document_type, child_document_type) tuples. For example `(ROUTE_TYPE, OUTING_TYPE)` means that you can create association between routes and outing, whereas as the route acts as the parent and the outing as the child. Does that make it clearer?

ginold:
[...]

PR #328, opened 2016-04-27
[WIP] Association tool fix 2

ginold:
#326 #298 

Now you can associate routes and waypoints in each direction, you can delete them and the cards will (dis)appear as well as the whole "associated-something" block.


c2corg_ui/static/js/addassociationdirective.js
```diff
@@ -69,7 +79,17 @@ app.AddAssociationController = function(appApi, $attrs) {
  * @export
  */
 app.AddAssociationController.prototype.associate = function(doc) {
-  this.api_.associateDocument(this.parentId, doc).then(function() {
+  var parentId, childId;
+  // we cannot directly associate a wp to route, so we have to
+  // do it the other way around -> route to wp. Same for outing.
+  if ((this.parentDocType === 'routes' && doc['waypoint_type']) || this.parentDocType === 'outings') {
```
asaunier: ``this.parentDocType === 'routes' && doc['waypoint_type']``

Is the test about ``doc['waypoint_type']`` a way to check that the "child"  doc is a waypoint? Don't we have a straight way to figure this out?

By the way it's not very clear what is the parent doc and what is the child doc!?
Which one is the current doc and which one is the doc to be associated?

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -9,23 +9,20 @@ goog.require('app');
 app.viewDetailsDirective = function() {
   return {
     restrict: 'A',
-    controller: 'AppViewDetailsController',
-    controllerAs: 'detailsCtrl',
-    bindToController: true,
+    controller: 'AppViewDetailsController as detailsCtrl',
+    bindToController: {'routes' : '=', 'waypoints': '=', 'orientations': '@'},
```
asaunier: How will the other document types be handled?
Do we really need to explicitly pass routes and waypoints? Can't we have something more generic?

asaunier:
By the way I notice the "association" files (``addassociationdirective.js``, ``associationcarddirective.js``, ``deleteassociationdirective.js``) contain "directive" in their filenames, which is not consistent with the naming of other directive files!
I think it would be more consistent to remove the "directive" keyword in those filenames.

ginold:
I added a few comments and made the associations more generic in the HTML as well as in the JS file, should be better now. But still I have this non blocking ugly error in the console after removing an association :P

c2corg_ui/static/js/addassociation.js
```diff
@@ -55,7 +55,25 @@ app.AddAssociationController = function(appApi, $attrs) {
   this.api_ = appApi;
 
 
+  /**
+   * @private
+   */
   this.dataset = $attrs['dataset'];
```
asaunier: If ``this.dataset`` is actually private, it should rather be named ``this.dataset_``.
By the way, don't we need to add ``@type`` jsdoc to this property and ``this.api_``?

c2corg_ui/static/js/addassociation.js
```diff
@@ -55,7 +55,25 @@ app.AddAssociationController = function(appApi, $attrs) {
...
+   * It is the currently viewed document type.
+   *
+   * For example, if you are looking at a waypoint ->
+   * parentType = waypoint, parentID = waypoint's_id.
+   *
+   * The child type and ID will depend on the document selected
+   * in the results of the app-search dropdown.
+   * @private
+   */
+  this.parentDocType = $attrs['parentDocType'];
```
asaunier: Same comment here about the name and type.

To be honest I don't understand what your comment means.

c2corg_ui/static/js/addassociation.js
```diff
@@ -55,7 +55,25 @@ app.AddAssociationController = function(appApi, $attrs) {
...
+
+
```
asaunier: Extra blank line here (and at line 57).

c2corg_ui/static/js/addassociation.js
```diff
@@ -65,11 +83,23 @@ app.AddAssociationController = function(appApi, $attrs) {
 
 
 /**
+ * Associate two documents. Possible associations are listed under http://bit.ly/1T15jxx
```
asaunier: Do we really need a bit.ly link here? Why not directly using the underlying github URL?

c2corg_ui/static/js/deleteassociation.js
```diff
@@ -42,6 +50,17 @@ app.DeleteAssociationController = function(appApi) {
   this.parentId;
 
   /**
+   * @private
+   */
+  this.parentDocType = $attrs['parentDocType'];
```
asaunier: Missing trailing "_" in this private property name. Ditto for the next one.

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -72,7 +70,16 @@ app.ViewDetailsController = function($uibModal, $compile, $scope) {
    */
   this.scope_ = $scope;
 
-}
+  this.scope_['document'] = {
+    'orientations' : this['orientations'],
```
asaunier: What is ``orientations``?
A route-specific attribute? If yes, can't we avoid it since the directive is not route-specific and should be generic.

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -83,6 +89,19 @@ app.ViewDetailsController.prototype.openModal = function(selector) {
   this.modal_.open({animation: true, size: 'lg', template: this.compile_(template)(this.scope_)});
 };
 
+/**
+ * @param {Object} doc
+ * @export
```
asaunier: This function looks private (only used in the controller and with a trailing "_" in the name): ``@export`` => ``@private``

asaunier:
Please note that PR #347 ("advanced search") also introduces angular partials for WP, routes and outings (and a ``cardDirective`` as well). See https://github.com/c2corg/v6_ui/tree/advanced-search-ui/c2corg_ui/static/partials (by the way the naming is like ``card_waypoints.html`` instead of ``cardwaypoints.html`` as in this PR, ditto for the "suggestion" partials):
* [ ] we'll have to use the same system for both the advanced search cards and the associations.

asaunier:
While testing the branch in Chromium and FF, I have noticed some problems with the association tool:
* [ ] in a WP detail page, the association tool suggests WP and routes (OK!) but also outings, which does not make sense. Here is a recap of the allowed associations (not to mention the other doc types not implemented yet):
  * WP page: WP, routes
  * route page: WP, routes
  * outing page: routes, users
  * image page: WP, routes, outings
* [ ] If the user starts typing a keyword, gets results but changes the keyword (for instance using the "delete" key), the tool reacts slower and slower (Chromium even crashes!) => there's probably a memory leak problem somewhere.
* [ ] The "see more results" link is not relevant in the association tool context (only for the simple search).
* [ ] A confirmation message is needed before actually removing an association.

> But still I have this non blocking ugly error in the console after removing an association

* [ ] The removal indeed works but having an error in the console cannot really be accepted. Perhaps it's something done a wrong way. Here is the error I get when clicking the "remove association" button of a WP card in some WP detail view page:

> angular.js:13236 Error: [$compile:nonassign] Expression '{'waypoints': [{'version': 1, 'available_langs': None, 'protected': False, 'locales': [{'lang': 'fr', 'version': 1, 'title': 'Col du Pertuis'}], 'elevation': 2365, 'document_id': 37881, 'waypoint_type': 'pass', 'type': 'w'}, {'version': 1, 'available_langs': None, 'protected': False, 'locales': [{'lang': 'fr', 'version': 1, 'title': 'Pierre-Pertuis'}], 'elevation': 1000, 'document_id': 102797, 'waypoint_type': 'climbing_outdoor', 'type': 'w'}, {'version': 1, 'available_langs': None, 'protected': False, 'locales': [{'lang': 'fr', 'version': 1, 'title': 'Pas de Pertuité'}], 'elevation': 2527, 'document_id': 108979, 'waypoint_type': 'pass', 'type': 'w'}], 'routes': [{'version': 1, 'available_langs': None, 'elevation_max': 2135, 'protected': False, 'locales': [{'title_prefix': 'Pic Rouge', 'lang': 'fr', 'version': 1, 'title': 'Versant NW, depuis Lhers'}], 'activities': ['skitouring'], 'document_id': 734034, 'elevation_min': 997, 'type': 'r'}], 'recent_outings': {'outings': [{'version': 1, 'date_start': '2016-03-18', 'available_langs': None, 'protected': False, 'locales': [{'lang': 'fr', 'version': 1, 'title': 'Pic Rouge\xa0: Versant NW, depuis Lhers'}], 'date_end': '2016-03-18', 'activities': ['skitouring'], 'document_id': 734143, 'author': {'name': 'EXSKIADOR', 'user_id': 214438, 'username': 'padovani alain'}, 'type': 'o'}], 'total': 1}}' in attribute 'associations' used with directive 'appViewDetails' is non-assignable!
http://errors.angularjs.org/1.5.0/$compile/nonassign?p0=%7B'waypoints'%3A%2…3A%20'o'%7D%5D%2C%20'total'%3A%201%7D%7D&p1=associations&p2=appViewDetails

And when I look at the HTML source of the page (not exactly the same than with the error above) I see

```
<section class="view-details-section" app-view-details 
         associations="{&#39;waypoints&#39;: [{&#39;version&#39;: 1, &#39;available_langs&#39;: None, &#39;protected&#39;: False, &#39;locales&#39;: [{&#39;lang&#39;: &#39;fr&#39;, &#39;version&#39;: 1, &#39;title&#39;: &#39;Col du Pertuis&#39;}], &#39;elevation&#39;: 1566, &#39;document_id&#39;: 272527, &#39;waypoint_type&#39;: &#39;pass&#39;, &#39;type&#39;: &#39;w&#39;}], &#39;routes&#39;: [{&#39;version&#39;: 1, &#39;available_langs&#39;: None, &#39;elevation_max&#39;: 2135, &#39;protected&#39;: False, &#39;locales&#39;: [{&#39;title_prefix&#39;: &#39;Pic Rouge&#39;, &#39;lang&#39;: &#39;fr&#39;, &#39;version&#39;: 1, &#39;title&#39;: &#39;Versant NW, depuis Lhers&#39;}], &#39;activities&#39;: [&#39;skitouring&#39;], &#39;document_id&#39;: 734034, &#39;elevation_min&#39;: 997, &#39;type&#39;: &#39;r&#39;}], &#39;recent_outings&#39;: {&#39;outings&#39;: [{&#39;version&#39;: 1, &#39;date_start&#39;: &#39;2016-03-18&#39;, &#39;available_langs&#39;: None, &#39;protected&#39;: False, &#39;locales&#39;: [{&#39;lang&#39;: &#39;fr&#39;, &#39;version&#39;: 1, &#39;title&#39;: &#39;Pic Rouge\xa0: Versant NW, depuis Lhers&#39;}], &#39;date_end&#39;: &#39;2016-03-18&#39;, &#39;activities&#39;: [&#39;skitouring&#39;], &#39;document_id&#39;: 734143, &#39;author&#39;: {&#39;name&#39;: &#39;EXSKIADOR&#39;, &#39;user_id&#39;: 214438, &#39;username&#39;: &#39;padovani alain&#39;}, &#39;type&#39;: &#39;o&#39;}], &#39;total&#39;: 1}}">
```

=> I am not sure having such complicated expressions is really a good thing. and seems to cause the problem...

asaunier:
> The "see more results" link is not relevant in the association tool context (only for the simple search).

* [ ] By the way, since the advanced search finally no longer uses URLs with a syntax like ``parameter1/value1/parameter2/value2`` but simply ``?parameter1=value1&parameter2=value2`` (at least at short term), one should use ``?q=<searched keywords>`` in the "see more results" link:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search.js#L200

c2corg_ui/templates/helpers/view.html
```diff
@@ -144,50 +139,44 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
...
+        </a>
+      </div>
+      % for outing in document['associations']['recent_outings']['outings']:
+      <div class="list-item outing">
+        ${outing_card(outing)}
+        <app-delete-association parent-doc-type="${doctype}" child-doc-type="outings" parent-id="${document['document_id']}" child-id="${outing['document_id']}" ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+      </div>
+      % endfor
+
+      % if len(document['associations']['recent_outings']['outings']) == 10 :
```
tsauerwein: You could check for `document['associations']['recent_outings']['total']`.

c2corg_ui/static/js/addassociation.js
```diff
@@ -65,11 +83,23 @@ app.AddAssociationController = function(appApi, $attrs) {
...
+ * For example, we cannot directly associate a waypoint to a route, but a route to a waypoint
+ * is possible, so we have to do it the other way around and change parent/child IDs accordingly. Same for outings.
  * @param {appx.SearchDocument} doc Document
  * @export
  */
 app.AddAssociationController.prototype.associate = function(doc) {
-  this.api_.associateDocument(this.parentId, doc).then(function() {
+  var parentId, childId;
+  // if the parentDoc is a Route and the childDoc is a Waypoint OR the parentDoc is an Outing, inverse the IDs.
+  if ((this.parentDocType === 'routes' && doc['waypoint_type']) || this.parentDocType === 'outings') {
```
asaunier: I think a ``doctype`` attribute is now available. That would clearer to use it (``doc['doctype'] === 'waypoints'``) instead of  ``doc['waypoint_type']`` to check that the current document is a WP.

c2corg_ui/templates/helpers/view.html
asaunier: I have done the change in https://github.com/c2corg/v6_ui/pull/366/files

asaunier:
* [ ] This PR could also be the opportunity to address this issue https://github.com/c2corg/v6_ui/issues/349 ?

ginold:
already addressed :heavy_check_mark:

c2corg_ui/static/js/addassociation.js
```diff
@@ -26,9 +26,8 @@ app.addAssociationDirective = function($compile) {
       'dataset' : '='
     },
     controllerAs: 'addCtrl',
-
     link: function(scope, element, attrs, ctrl) {
-      element.html(template(ctrl.dataset));
+      element.html(template(ctrl.dataset_));
```
asaunier: Just curious: if you access a property of the controller from within the directive itself, is this property really private? Shouldn't it be ``@export``ed?

asaunier:
This PR is ready. Or almost since:
* there will be some conflict-fixing with https://github.com/c2corg/v6_ui/pull/372
* some routes/waypoints associations handling is still needed. But I think the associations attribute structure is too complicated. I think we should simplify it. See https://github.com/c2corg/v6_api/issues/301

What the PR does in short:
* introduce a ``documentData`` module value that can be injected in controllers and used to pass data between the HTML (mako) and the directives instead of using a "ng-hidden" expression.
* the card directive and partials (created for the advanced search) are now used to build associated-documents lists and update them when adding or removing an association.
* the formerly-used other partials and mako helpers have been removed (no longer used).

asaunier:
The PR is almost ready.

The 2 last points to address:
* [ ] the title_prefix in the route editing form is broken
* [ ] depending on the parent document type, newly associated waypoints or routes may not be added in the associated docs cards list. Ditto when removing.

The latter issue is related to the fact that the associations names vary depending on the parent and child type (routes/all_routes, waypoints/waypoint_children/waypoint_parents). See https://github.com/c2corg/v6_api/issues/301

It's really a nightmare to deal with those special cases. It requires complicated pieces of code and sometimes it is not even feasible without big changes because all the doc types are not always available.
It would be really cool to have consistent lists of associated docs simply named according to the child-doc type. eg ``waypoints``, ``routes``, ``images``, etc.

asaunier:
I have fixed the title_prefix in the route editing form and made a demo of this branch available at http://c2corgv6-demo.gis.internal:6555/

There is still changes to do about the association types issue but else I think the PR can be reviewed as is. Thanks :)

ginold:
- [ ] "associations" tab does not appear on mobile (ex: http://c2corgv6-demo.gis.internal:6555/waypoints/726682/fr) 
- [ ] removing an association does not remove its card from associations (view)

asaunier:
Thanks for testing!

The association cards adding/removal issues are most likely related to the fact that the association type (for instance "routes"/"all_routes") is not the correct one.

asaunier:
> due to the fact that the href link is not displayed in the a tag, documents cannot be opened with "open in new tab" !

I have removed the standard href by an ``ng-click`` calling a function in the card controller: https://github.com/c2corg/v6_ui/blob/association-tool-fix-2/c2corg_ui/static/js/card.js#L104-L111

Perhaps I could replace it by an ``ng-href`` with a URL built using the card properties. Would it solve the "new tab" issue? It should also be possible to neutralize the link when the card is displayed in the editing form?!

ginold:
Yes it should fix the issue.

>  It should also be possible to neutralize the link when the card is displayed in the editing form?!

It would be best if it were so!

asaunier:
I have made some simplifications according to a discussion @tsauerwein and I had today https://github.com/c2corg/v6_api/issues/301#issuecomment-223602535
I think now the associations adding/removal work quite well, including in editing mode.

I have also updated the translations in FR. By the way there are too many duplicates (or near duplicates). See https://github.com/c2corg/v6_ui/issues/392

asaunier:
I have just push commit fd97b97 to take into account the recent changes about ``waypoint_parents`` in the API: https://github.com/c2corg/v6_api/pull/305

Actually, from the UI perspective it would have made more sense to keep ``waypoint_parents`` and drop ``waypoint_children`` because the association/deassociation tools deal with the ``waypoint_children`` and not the ``waypoint_parents``. Then always dealing with ``waypoints``, whatever the current document type, would have allow to have a more simple and intuitive UI code (see fd97b97). But we can live with it.

asaunier:
This PR is ready and should be merge as soon as possible.
Please review :)

ginold:
in the editing forms you are using the `app-add-association` tool which directly associates the document, but it's not how we want it. 

I had set it to push or remove the association only from the current JS Object (with `app-simple-search` and `app-select `function) and a  and when you save the form and sent it to the API it will be updated.

asaunier:
> Route-view : removing a waypoint from association doesn't make the card dissapear.

Thanks for noticing. I have fixed the bug and updated commit e05c15c (the PR has been rebased).

> Error on waypoint-view :

Simply an outdated API :)

ginold:
A priori it looks good to me.

asaunier:
Thanks @ginold for the review.
Is there anyone else reviewing the PR?

tsauerwein:
> Is there anyone else reviewing the PR?

If you are interested I can take a look this afternoon.

asaunier:
Feel free to review @tsauerwein if you have time.
This is pretty urgent since the PR should be part of the alpha1 release of... last friday :P

c2corg_ui/static/js/advancedsearch.js
```diff
@@ -126,8 +130,8 @@ app.AdvancedSearchController.prototype.successList_ = function(response) {
   }
   this.documents = response['data']['documents'];
   this.total = response['data']['total'];
-  if (this.resCounter_) {
-    $('#' + this.resCounter_).html(
+  if (this.resCounter) {
+    $('#' + this.resCounter).html(
```
tsauerwein: This doesn't look like the Angular way.

c2corg_ui/static/js/apiservice.js
```diff
@@ -147,8 +151,12 @@ app.Api.prototype.unassociateDocument = function(parentId, childId) {
   };
 
   var promise = this.deleteJson_('/associations', data);
-  promise.catch(function() {
-    var msg = alerts.gettext('Failed to unassociate document');
+  promise.catch(function(res) {
+    var error = '';
+    for (var i = 0; i < res['data']['errors'].length; i++) {
+      error = error + ' ' + res['data']['errors'][i]['description'];
+    }
```
tsauerwein: You might want to refactor this code with the error handling code in `associateDocument`.

c2corg_ui/static/js/deleteassociation.js
```diff
@@ -0,0 +1,172 @@
...
+
+app.module.directive('appDeleteAssociation', app.deleteAssociationDirective);
+
+
+/**
+ * @constructor
+ * @param {angular.Scope} $rootScope
+ * @param {!angular.Scope} $scope Scope.
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {Object} $uibModal modal from angular bootstrap
```
tsauerwein: Do you want to add a minimal extern for this?

c2corg_ui/static/js/documentservice.js
```diff
@@ -57,6 +54,35 @@ app.Document = function(appAuthentication) {
...
+    }
+  }
+};
+
+
+/**
+ * @param {appx.DocumentAssociations} associations
+ * @export
+ */
+app.Document.prototype.setAssociations = function(associations) {
```
tsauerwein: What is the purpose of this function?

c2corg_ui/static/js/documentservice.js
```diff
@@ -78,62 +104,71 @@ app.Document.prototype.hasAssociation = function(type, id) {
...
  * @export
  */
-app.Document.prototype.pushToAssociations = function(doc) {
-  var routesLength =  this.document.associations['routes'].length;
+app.Document.prototype.pushToAssociations = function(doc, doctype,
+    setOutingTitle) {
   var associations = this.document.associations;
-  var doctype = app.utils.getDoctype(doc['type']);
+  doctype = doctype || app.utils.getDoctype(doc['type']);
+  setOutingTitle = typeof setOutingTitle !== 'undefined' ?
```
tsauerwein: Isn't `setOutingTitle = setOutingTitle !== 'undefined' ? ... ` enough?

c2corg_ui/static/js/documentservice.js
```diff
@@ -78,62 +104,71 @@ app.Document.prototype.hasAssociation = function(type, id) {
...
+  // When creating an outing, the outing title defaults to the title
+  // of the first associated route.
+  if (setOutingTitle && doctype === 'routes' &&
+      !this.document.locales[0]['title'] &&
+      this.document.associations.routes.length == 1) {
+    var title = 'title_prefix' in doc.locales[0] &&
+      doc.locales[0]['title_prefix'] ?
+      doc.locales[0]['title_prefix'] + ' : ' : '';
+    title += doc.locales[0]['title'];
+    this.document.locales[0]['title'] = title;
```
tsauerwein: Can `this.document.locales[0]['title']` already have an user-provided value at this point? Should this value be overwritten?

c2corg_ui/static/js/simplesearch.js
```diff
@@ -238,8 +237,12 @@ app.SimpleSearchController.prototype.createDataset_ = function(type) {
         return '<div class="header" dataset="' + type + '">' + this.gettextCatalog_.getString(type) + '</div>';
       }).bind(this),
       footer: function(doc) {
-        return '<p class="suggestion-more"><a href="/' + type + '/keyword/' + encodeURI(doc['query']) + '" class="green-text" translate>More results</a></p>';
-      },
+        if (!this.associationContext_) { // don't add this if you're typing in an add-association-tool
+          return '<p class="suggestion-more"><a href="/' + type + '?q=' + encodeURI(doc['query']) + '" class="green-text">+ see more results</a></p>';
```
tsauerwein: Translatable?

c2corg_ui/static/js/viewdetails.js
```diff
@@ -133,6 +142,27 @@ app.ViewDetailsController.prototype.openModal = function(selector) {
...
+app.ViewDetailsController.prototype.removeAssociation_ = function(doc) {
+  var associations;
+
+  if (doc['type'] === 'outings') {
+    associations = this.scope_['document']['associations']['recent_outings']['outings'];
+  } else {
+    associations = this.scope_['document']['associations'][doc['type']];
+  }
+  if (associations) {
+    for (var i = 0; i < associations.length; i++) {
```
tsauerwein: Do you want to refactor this in a function?
https://github.com/c2corg/v6_ui/pull/328/files#diff-8f682851013ffc3282976f742de82930R140

c2corg_ui/templates/helpers/edit.html
```diff
@@ -0,0 +1,33 @@
...
+      <div class="list-item"
+           ng-repeat="wp in ${model}.associations.${type} | orderBy: 'waypoint_type'"
+           ng-class="{'main-waypoint': ${model}.main_waypoint_id === wp.document_id}">
+        <app-card app-card-doc="wp" app-card-disable-click="true"></app-card>
+        % if model == 'route':
+          <div ng-if="${model}.main_waypoint_id === wp.document_id" ng-init="${model}.main_waypoint_title = wp.locales[0].title" translate>main waypoint</div>
+          <button class="btn btn-sm btn-default" translate
+                  ng-click="${model}.main_waypoint_id = wp.document_id; ${model}.main_waypoint_title = wp.locales[0].title"
+                  ng-if="${model}.main_waypoint_id != wp.document_id">
+            set as main waypoint
```
tsauerwein: Capita letter? "Set ..."

c2corg_ui/templates/waypoint/view.html
```diff
@@ -152,37 +157,43 @@ <h3 class="text-center" translate>No description available</h3>
...
+
+    <div class="view-details-associations col-xs-12 tab associations">
+      <span class="lead">
+        <section>
+          ${show_associated_routes(waypoint, 'all_routes')}
+        </section>
+      </span>
+    </div>
+
+    % if 'recent_outings' in waypoint['associations'] and len(waypoint['associations']['recent_outings']['outings']) > 0 :
```
tsauerwein: You could also check for `total`.

tsauerwein:
I added a couple of comments. I let you decide if you want to address them before or after a merge.

c2corg_ui/static/js/documentservice.js
asaunier: It is used to fill the ``documentService``'s ``associations`` attribute from the module variable passed in the document view page. E.g. https://github.com/c2corg/v6_ui/blob/association-tool-fix-2/c2corg_ui/templates/waypoint/view.html#L66-L71

c2corg_ui/static/js/documentservice.js
asaunier: No idea, I found both tests on stackoverflow. Is it really that different?

c2corg_ui/static/js/documentservice.js
asaunier: This is the default title when creating an outing, based upon the title of the first associated route. If the title is already set, this block would not be run at all thanks to the condition ``!this.document.locales[0]['title']``.

c2corg_ui/static/js/simplesearch.js
asaunier: Indeed. I will change this.

c2corg_ui/static/js/viewdetails.js
asaunier: Good catch! I think this function should be replaced by the one in the ``documentService`` (perhaps with the test about ``recent_outings`` and maybe ``all_routes``). Since such a change might require intensive testing to make sure nothing is broke, I would prefer it is done in another PR. I will create an issue.

c2corg_ui/templates/helpers/edit.html
asaunier: It's translated anyway so we can keep it as is.

tsauerwein: The English versions are translated again?

c2corg_ui/static/js/deleteassociation.js
asaunier: Arghhhh :P

c2corg_ui/static/js/documentediting.js
```diff
@@ -185,21 +186,20 @@ app.DocumentEditingController = function($scope, $element, $attrs,
...
-  // allow association only new outing  to existing route
+  // allow association only new outing to existing route
   if (this.ngeoLocation_.hasParam('routes')) {
     this.urlParams_ = {'routes': this.ngeoLocation_.getParam('routes')};
     this.pushDocToAssociations_();
   }
 
-  // When creating a new document, the model is not created until
-  // the form is touched. At least create an empty object.
-  this.scope_[this.modelName_] = this.scope_['document'] = this.documentService.document;
```
ginold: removing this line causes the impossibility to select orientations in the SVG Orientation selector, because the SVG uses `document.orientation` syntax and now it throws an error `TypeError: Cannot read property 'orientations' of undefined`

c2corg_ui/static/js/deleteassociation.js
ginold: I think it's a NO :no_good:

c2corg_ui/static/js/apiservice.js
asaunier: You may be right. But I have the feeling that for a 3-line loop, it's not really worth creating a function.

tsauerwein: What *cost* is associated with creating a function?

c2corg_ui/templates/helpers/edit.html
asaunier: Yes.
Anyway we need the EN translations for instance for the attributes names.

tsauerwein: For constants, yes. But I don't see any sense in "translating" (copy&pasting) the English language strings again.

c2corg_ui/static/js/deleteassociation.js
tsauerwein: Someone has to take @gberaudo's place.

c2corg_ui/templates/helpers/edit.html
asaunier: > But I don't see any sense in "translating" (copy&pasting) the English language strings again.

OK

c2corg_ui/static/js/apiservice.js
asaunier: 2 blank lines
4 jsdoc lines
5 function code lines
= 11 lines (instead of 3)

c2corg_ui/templates/waypoint/view.html
asaunier: Indeed, I will do it.

c2corg_ui/static/js/documentediting.js
asaunier: Can't we use "this.modelName_" in the HTML instead of "document". Eg.``route.orientation`` instead of ``document.orientation``?
I have the feeling that the document is passed to the scope using duplicated ways.

asaunier: OK I misread that you said the problem was in the SVG file!
By the way do we really need to include the SVG file directly in the page HTML? Can't it be used as a separated file?

c2corg_ui/static/js/apiservice.js
tsauerwein: Ehm, 20 lines below in function `unassociateDocument` you have exactly the same code (but for 2 characters). You saw that?

c2corg_ui/static/js/simplesearch.js
asaunier: Can someone remind me how to have an i18n string in the JS docs extracted to the POT file? I have used ``this.gettextCatalog_.getString('see more results')`` but the make command keeps ignoring it.

c2corg_ui/static/js/apiservice.js
asaunier: > You saw that?

Of course :P 
OK I have made a change at 	851d16d

c2corg_ui/static/js/simplesearch.js
asaunier: I have made changes at commit 	58ef63d
To make sure the i18n string is extracted I have also added it to ``c2corgui/templates/i18n.html``.

asaunier:
@tsauerwein I have made most of the changes you suggested, except the ``removeAssociation`` refactoring, postponed to https://github.com/c2corg/v6_ui/issues/394 (we really need to have this PR merged soon and I don't want to make this in the hurry).

Will merge the PR on wednesday morning if no one objects.

c2corg_ui/locale/c2corg_ui-client.pot
```diff
@@ -1704,15 +1708,15 @@ msgid "routes_taken"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
-msgid "sep"
+msgid "see more results"
```
tsauerwein: Lowercase on purpose?

c2corg_ui/static/js/apiservice.js
tsauerwein: Thanks! I would have moved the whole 'promise.catch(...)` block into a function, but it's already better.

tsauerwein:
Thanks for the updates, feel free to merge!

c2corg_ui/static/js/documentediting.js
ginold: It is a separate file `<%include file="c2corg_ui:static/img/orientation.svg" />`

asaunier: In the final page it is embedded in the HTML. Which means it is downloaded every time the page is downloaded. I just wondered if there was a way to download it separately like an image. But perhaps it wouldn't give much gain?

c2corg_ui/locale/c2corg_ui-client.pot
asaunier: Whatever the case I think. It's not really a sentence so capitalizing the first letter is not really mandatory. If a uppercasing the first letter proves needed, we can still do it later.

c2corg_ui/static/js/apiservice.js
asaunier: I have considered doing it but didn't want to complicate the code for figuring out what message to return.

asaunier:
Thank you very much @tsauerwein and @ginold for spending a lot of time reviewing and testing this PR!!

c2corg_ui/static/js/apiservice.js
tsauerwein:
```js
this.setupErrorHandler(promise, 'Failed to associate document:');
```

Issue #257, opened 2016-03-28
Default position for maps with no geometry


asaunier:
As for now, maps with no geometry are recentered on 0°N 0°E (western Africa). Using western Europe as default extent would make more sense.

PR #340, opened 2016-05-02
Default map extent is on western Europe

asaunier:
This PR tries to fix https://github.com/c2corg/v6_ui/issues/257
Instead of giving an initial center point to the map view (which is fairly simple), it tries to use an initial extent containing the Pyrénées and the Alps.

I had several problems and am not sure if the solutions are good:
* fitting the map view on an extent requires to get the map size. Which seems to be unavailable until the map is actually rendered => I have used the map's ``postrender`` event. This works but it seems the event is launched many times whereas I only need to set the initial extent once... What would be the best event to use?
* for some reason, Closure compiler was unhappy with the ``fit`` options argument (including for the ``diffmap`` that already used it previously) because:
```
ERR! compile /home/alex/Desktop/v6_ui/c2corg_ui/static/js/map/diffmap.js:84: ERROR - actual parameter 3 of ol.View.prototype.fit does not match formal parameter
ERR! compile found   : {
ERR! compile   constrainResolution: (boolean|undefined), 
ERR! compile   maxZoom: (number|undefined), 
ERR! compile   minResolution: (number|undefined), 
ERR! compile   nearest: (boolean|undefined), 
ERR! compile   padding: (Array<number>|undefined)
ERR! compile }
ERR! compile required: (olx.view.FitOptions|undefined)
ERR! compile     this.map.getView().fit(app.MapController.DEFAULT_EXTENT, mapSize, {
```
=> I then used a casting for this argument but is it really necessary?

### DIFF
--- a/c2corg_ui/static/js/map/diffmap.js
+++ b/c2corg_ui/static/js/map/diffmap.js
@@ -73,19 +73,16 @@ app.DiffMapController = function(mapFeatureCollection) {
       new ol.layer.Tile({
         source: new ol.source.OSM()
       })
-    ]
+    ],
+    view: new ol.View({
+      center: ol.extent.getCenter(app.MapController.DEFAULT_EXTENT),
+      zoom: app.MapController.DEFAULT_ZOOM
+    })
   });
 
   var mouseWheelZoomInteraction = new ol.interaction.MouseWheelZoom();
   this.map.addInteraction(mouseWheelZoomInteraction);
   app.utils.setupSmartScroll(mouseWheelZoomInteraction);
-
-  if (!mapFeatureCollection) {
-    this.map.setView(new ol.View({
-      center: app.MapController.DEFAULT_CENTER,
-      zoom: app.MapController.DEFAULT_ZOOM
-    }));
-  }
 
   if (!goog.array.isEmpty(this.features_)) {
     // Recentering on the features extent requires that the map actually
@@ -165,11 +162,13 @@ app.DiffMapController.prototype.showFeatures_ = function(features) {
   var vectorLayer = this.getVectorLayer_();
   vectorLayer.getSource().addFeatures(features);
 
-  var mapSize = this.map.getSize() || null;
-  this.map.getView().fit(vectorLayer.getSource().getExtent(), mapSize, {
-    padding: [10, 10, 10, 10],
-    maxZoom: app.MapController.DEFAULT_POINT_ZOOM
-  });
+  var mapSize = this.map.getSize();
+  if (mapSize) {
+    this.map.getView().fit(vectorLayer.getSource().getExtent(), mapSize, {
+      padding: [10, 10, 10, 10],
+      maxZoom: app.MapController.DEFAULT_POINT_ZOOM
+    });
+  }
 };
 
 app.module.controller('AppDiffMapController', app.DiffMapController);
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -166,10 +166,15 @@ app.MapController = function($scope, mapFeatureCollection) {
       this.map.addInteraction(clickInteraction);
     }
   } else {
+    // no special feature displayed on the map => use the default extent
     this.map.setView(new ol.View({
-      center: app.MapController.DEFAULT_CENTER,
-      zoom: this.zoom_ || app.MapController.DEFAULT_ZOOM
+      center: ol.extent.getCenter(app.MapController.DEFAULT_EXTENT),
+      zoom: app.MapController.DEFAULT_ZOOM
     }));
+    this.map.once('change:size', function(event) {
+      var mapSize = this.map.getSize() || null;
+      this.map.getView().fit(app.MapController.DEFAULT_EXTENT, mapSize);
+    }.bind(this));
   }
 
   if (this.drawType) {
@@ -222,7 +227,8 @@ app.MapController = function($scope, mapFeatureCollection) {
  * @const
  * @type {Array.<number>}
  */
-app.MapController.DEFAULT_CENTER = [0, 0];
+app.MapController.DEFAULT_EXTENT = [-400000, 5200000, 1200000, 6000000];
+
 
 /**
  * @const


c2corg_ui/static/js/map/map.js
```diff
@@ -166,10 +166,14 @@ app.MapController = function($scope, mapFeatureCollection) {
       this.map.addInteraction(clickInteraction);
     }
   } else {
-    this.map.setView(new ol.View({
-      center: app.MapController.DEFAULT_CENTER,
-      zoom: this.zoom_ || app.MapController.DEFAULT_ZOOM
-    }));
+    // no special feature displayed on the map => use the default extent
+    this.map.on('postrender', function(event) {
```
tsauerwein: i think you could use: `this.map.once('change:size', ...)`

c2corg_ui/static/js/map/diffmap.js
```diff
@@ -81,9 +80,10 @@ app.DiffMapController = function(mapFeatureCollection) {
   app.utils.setupSmartScroll(mouseWheelZoomInteraction);
 
   if (!mapFeatureCollection) {
-    this.map.setView(new ol.View({
-      center: app.MapController.DEFAULT_CENTER,
-      zoom: app.MapController.DEFAULT_ZOOM
+    var mapSize = this.map.getSize() || null;
+    this.map.getView().fit(app.MapController.DEFAULT_EXTENT, mapSize,
```
tsauerwein: if `mapSize` is `null`, do you maybe don't want to call `fit`?

asaunier: Maybe not, indeed.
Actually it's already like this in the code you have written: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/diffmap.js#L168-L172

Perhaps we don't really need to have ``|| null`` at line 83, but I guess it's required by the Closure Compiler so that mapSize is ``ol.Size|null``?

c2corg_ui/static/js/map/map.js
asaunier: Thanks for the tip!

c2corg_ui/static/js/map/diffmap.js
gberaudo: It should also work with:
```
var mapSize = this.map.getSize();
if (mapSize) {
  this.map.getView().fit(app.MapController.DEFAULT_EXTENT, mapSize...
}
```

tsauerwein: Yes, this is how it should be. Calling `fit` with `null` as `size` is wrong.

c2corg_ui/static/js/map/map.js
gberaudo: It feels wrong to create a map without center. Couldn't you use the center of the extent?

asaunier:
PR updated according to @gberaudo and @tsauerwein 's suggestions.

asaunier:
If there are no other comments, I will merge shortly.

c2corg_ui/static/js/map/diffmap.js
```diff
@@ -81,10 +80,13 @@ app.DiffMapController = function(mapFeatureCollection) {
   app.utils.setupSmartScroll(mouseWheelZoomInteraction);
 
   if (!mapFeatureCollection) {
-    this.map.setView(new ol.View({
-      center: app.MapController.DEFAULT_CENTER,
-      zoom: app.MapController.DEFAULT_ZOOM
-    }));
+    var mapSize = this.map.getSize();
```
gberaudo: Are you going to restore a center here?

c2corg_ui/static/js/map/diffmap.js
```diff
@@ -81,10 +80,13 @@ app.DiffMapController = function(mapFeatureCollection) {
...
+    if (mapSize) {
+      this.map.getView().fit(app.MapController.DEFAULT_EXTENT, mapSize,
+      /** @type {olx.view.FitOptions} */ ({
```
gberaudo: Is this cast required? There is no cast in https://github.com/c2corg/v6_ui/pull/340/files#diff-2fb178a58cd1af945acd73ff3c2b5ca7R172

asaunier: I know but if I remove the cast the compiler complains that
```
ERR! compile /home/alex/Desktop/v6_ui/c2corg_ui/static/js/map/diffmap.js:86: ERROR - actual parameter 3 of ol.View.prototype.fit does not match formal parameter
ERR! compile found   : {
ERR! compile   constrainResolution: (boolean|undefined), 
ERR! compile   maxZoom: (number|undefined), 
ERR! compile   minResolution: (number|undefined), 
ERR! compile   nearest: (boolean|undefined), 
ERR! compile   padding: (Array<number>|undefined)
ERR! compile }
ERR! compile required: (olx.view.FitOptions|undefined)
```

c2corg_ui/static/js/map/diffmap.js
asaunier: That would indeed be more consistent with the changes in the map directive.
I just wonder if we should not do it directly in the ``ol.Map`` object creation instead?

By the way do we really need to display a map if there is no ``mapFeatureCollection`` in the diff page? What about skipping this whole block since the diff template only use the diffmap directive when geometries are available: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/document/diff.html#L52-L66 ?

c2corg_ui/static/js/map/diffmap.js
tsauerwein: I am wondering if the definition of `olx.view.FitOptions` is not wrong in ol3.

Should `padding` be a required attribute?
https://github.com/openlayers/ol3/blob/master/externs/olx.js#L7406-L7423
http://openlayers.org/en/master/apidoc/ol.View.html#fit

asaunier: > Should padding be a required attribute?

Maybe but the API doc also says the default value is ``[0, 0, 0, 0]`` so I guess it is not that required to specify it?

asaunier:
I have just pushed a fixup 2d7cde8 that set a default center to the diffmap when creating it, then removing the default extent fitting block.

c2corg_ui/static/js/map/diffmap.js
gberaudo: I created a branch with the change: https://github.com/gberaudo/ol3/tree/unrequire_padding
@asaunier, would you be able to test if it fixes the compiler error?

asaunier: Honestly I don't want to spend more time on this PR. We can't spend days on such a small change ("default map extent"). OK for merging the PR?

gberaudo: Sure, please merge.

asaunier: OK thanks @gberaudo 
I have rebased the fixup ciommit and will merge when Travis is OK.

PR #345, opened 2016-05-03
has rights fix

### LINKED ISSUE #341: Error with hasEditRights in detail view page
When opening a detail view page, for instance a route, one gets the following error when **not authenticated**:
```
angular.js:13236 TypeError: Cannot read property 'id' of null
    at app.Authentication.hasEditRights (authservice.js:82)
    at app.UserController.hasEditRights (user.js:108)
    at fn (eval at <anonymous> (angular.js:14086), <anonymous>:4:258)
    at oneTimeWatch (angular.js:15154)
    at Scope.$digest (angular.js:16655)
    at Scope.$apply (angular.js:16928)
    at done (angular.js:11266)
    at completeRequest (angular.js:11464)
    at XMLHttpRequest.requestLoaded (angular.js:11405)
```

ginold:
fixes #341

### DIFF
--- a/c2corg_ui/static/js/auth/authservice.js
+++ b/c2corg_ui/static/js/auth/authservice.js
@@ -79,6 +79,9 @@ app.Authentication.prototype.isAuthenticated = function() {
  * @export
  */
 app.Authentication.prototype.hasEditRights = function(users) {
+  if (!this.isAuthenticated()) {
+    return false;
+  }
   var userid = this.userData.id;
   var roles = this.userData.roles;
   // moderator has rigths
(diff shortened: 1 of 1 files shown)


asaunier:
Looks good but the simplest change would have been to add
```
if (!this.isAuthenticated()) {
  return false;
}
```
at the beginning of the function => just 3 lines changed and more readable.

ginold:
true, didn't think of that this way. I will change

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -93,10 +96,11 @@ app.Authentication.prototype.hasEditRights = function(users) {
       }
     }
     return false;
-  // by default, user has rights to every doc
+    // by default, user has rights to every doc
   } else {
     return true;
   }
+  // not logged -> return false
```
asaunier: This comment is most likely not useful, at least misplaced :p 
Please merge.

Issue #337, opened 2016-04-28
Change the list pages to interact with the API for the advanced search


asaunier:
Discussed with @tsauerwein 

For the advanced search page, we need advanced interactivity in order to take into account the criterias and map extent set while searching (and to add criterias to the URL).

As for now the advanced search / list page is rendered server-side (for SEO reasons).

To avoid mixing server-side rendering and angular rendering, we suggest that:
* the rendering of the results list (+ the criterias form?) is made by angular and fed by AJAX. even when loading the page (that, by the way, will load a bit faster, at least the base page with no results).
* the existing list page server-side rendering is reused as a "sitemap" view that can be used by search engines to index the site. The sitemap will not be displayed to end users.

asaunier:
related to https://github.com/c2corg/v6_api/pull/252

tsauerwein:
> the existing list page server-side rendering is reused as a "sitemap" view that can be used by search engines to index the site. The sitemap will not be displayed to end users.

I was rather thinking of using sitemap.xml files, see https://github.com/c2corg/v6_api/issues/205

asaunier:
OK, I didn't mean we would use the current list pages exactly as is, changing the templates to return an XML sitemap is indeed possible.

Issue #230, opened 2016-03-11
Use default geometry for outings and routes in listing views


tsauerwein:
With https://github.com/c2corg/v6_api/pull/179 outings and routes now have a `geom_detail` field which contains the full track. The field `geom` contains a point geometry which can be used in listing pages.

asaunier:
To summarize:
* for list pages: use the point geometry (``geom``) to place/filter outings and routes on the map
* for route detail pages:
 * use the detailed geometry if available to show it on the map. If not available, show no geom.
 * show the associated WP geoms anyway.
* for outing page
 * use the detailed geometry if available to show it on the map.
 * not sure if we should show the associated routes detailed geometries if available (that might be a bit confusing because visitors might not figure out the tracks are not the actual outing track). Or only on demand, for instance when hovering the associated route cards?
 * if no detailed geom is available, perhaps we could simply recenter the map on the point ``geom`` (default geometry). Or show no map at all as in v5 (but I think showing a map anyway is interesting to figure out where the outing took place).
* detail pages of WP:
 * show the ``geom`` of the the WP (always available) => map is centered on it
 * show the associated WP geom?
 * show the associated routes detailed geom if available? no need to show the point default geom if no detailed geom is available.

tsauerwein:
> for route detail pages:
> 
>  * use the detailed geometry if available to show it on the map. If not available, show no geom.

For routes and outings, if no detail geometry is available, we could show the point geometry, so that you have at least an idea where it is.

Note that currently the geometries for associated documents are not included.

asaunier:
> For routes and outings, if no detail geometry is available, we could show the point geometry, so that you have at least an idea where it is.

For outings it could make sense.
For routes, I think showing the associated WPs is more relevant (even though not available yet :P).

Issue #343, opened 2016-05-02
Sitemaps for SEO


asaunier:
Related to https://github.com/c2corg/v6_ui/issues/337 and https://github.com/c2corg/v6_api/issues/205

According to http://www.sitemaps.org/protocol.html
- a sitemap file cannot contain more than 50'000 URLs (and should be < 10MB)
- several sitemap files can be created and linked in a sitemapindex file (with the same URL number and weight limits)

What we could do:
- have sitemaps dedicated to a given document types (waypoints, routes, outings, etc.)
- for each document type, a sitemap document would contain X URLs (10? 30? 100? 1000?)
  - we would then have Y sitemaps for a given document type (`Y = ceil(number of document for this type / X)`)
- those `sitemaps` would be listed in one (or several?) `sitemapindex`

If listing `sitemapindex` files in a `sitemapindex` file is possible (not sure!) we could have the following structure:

```
main sitemapindex
  waypoints sitemapindex
    sitemap with waypoints from 1 to X
    sitemap with waypoints from X+1 to 2X
    ...
  routes sitemapindex
    sitemap with routes from 1 to X
    sitemap with routes from X+1 to 2X
    ...
```

This way we could have a main "sitemap" view/Mako template listing the "sitemapindex" for each document type.
Each document type view class would have a "sitemapindex" view/template that would point to as many sitemaps as needed to dispatch the total number of documents for this type.
Finally we would have a "sitemap" view/template that would be used to list the URLs.

This works only if it possible to have several levels of sitemapindex. Not sure...

Another issue about sitemaps:
- when a document is modified, its "lastmod" attribute in the sitemap must be updated accordingly (else the document might not be crawled again by the search engines). Which mean that parent sitemap's `lastmod` must be updated as well.
- deleted/created docs will also impact the content of the sitemaps/sitemapindex.

How can we make sure the sitemaps are kept up-to-date?

Can this be related to https://github.com/c2corg/v6_ui/issues/144 ?

Could we have some kind of event such as "document updated" (resp. created/deleted), with watchers that would update caches, sitemaps etc.?

tsauerwein:
> This works only if it possible to have several levels of sitemapindex. Not sure...

Or the main index contains the links for all document types.

> How can we make sure the sitemaps are kept up-to-date?
> 
> Can this be related to #144 ?
> 
> Could we have some kind of event such as "document updated" (resp. created/deleted), with watchers that would update caches, sitemaps etc.?

[...]

PR #347, opened 2016-05-04
Advanced search UI

asaunier:
TODO
* [x] save the current list pages as "sitemap" even though not a real sitemap (see https://github.com/c2corg/v6_ui/issues/343)
* [x] rename the existing ``searchDirective`` to ``simpleSearchDirective``
* [x] create a basic ``advancedSearchDirective`` to feed the list page using an AJAX request to the API instead of doing it through the UI server side (only the waypoints list page works as for now - with only the documents titles).
* [x] Display the number of results (in the title section above the map, as previously)
* URLs
 * [x] Take the URL filter parameters into account when first loading the page
 * [x] Update the URL when a criteria is applied
* [x] Manage pagination
* Cards:
 * [x] Use a template for cards and display all the required data
  * [ ] Outing card layout needs some fixing (+ the activity tooltips are not displayed)
 * [x] Show the best locale in the cards (not the first one)
 * [ ] Make it work in debug mode (same problem for the simple search)
 * [ ] ~~Use "JS" cards templates for cards created by the association tool (or even all cards listed in a detail view page with a SEO fallback?)~~
* interact with the map
 * [x] show results on the map
 * [x] use the map extent as a search criteria
 * [ ] ~~better map transition~~
* [x] Manage and interact with the filters forms (specific to each document type)
* [ ] ~~Hilight the feature on the map when hovering a card (and vice-versa?)~~

Side effects to fix:
* [ ] ~~feature icon is now clickable in the detail view page map: either disable this behaviour or at least make sure the URL is OK~~

Fixes https://github.com/c2corg/v6_ui/issues/337
Fixes https://github.com/c2corg/v6_ui/issues/230

API: see https://github.com/c2corg/v6_api/pull/252 & https://github.com/c2corg/v6_api/pull/267


tsauerwein:
FYI: The API for the advanced search is documented here:
https://github.com/c2corg/v6_api/blob/ae130e8/c2corg_api/views/waypoint.py#L57-L119

asaunier:
Cool, thanks!

stef74:
Great !

It is planned to have capability to search criteria based on associated
document ?

Ex : Search an outing with summit > 1000m. Search route with WP type =
parking and WP elevation > 1000m



2016-05-04 12:56 GMT+02:00 Tobias Sauerwein <notifications@github.com>:

asaunier:
When using the pagination buttons (or the filtering criterias), the list of features on the map changes, the map is then recentered on the extent of those features.

To make the transition smoother (ditto when loading a GPX file while editing a route for instance), I have added a pan+bounce animation:
https://github.com/c2corg/v6_ui/pull/347/commits/9d436590f22ec0a297859ac451419217fbdd9b0d

but the result is not really cool: it makes the map hop a little too much even though the features extent has not really changed.
If I remove the bounce and keep only the pan, it's a bit more natural except when the new extent make the map zoom before panning.

What would be a cool transition animation then?

tsauerwein:
> When using the pagination buttons (or the filtering criterias), the list of features on the map changes, the map is then recentered on the extent of those features.

I think it is a problem to change the extent when the features have been loaded. We said that we automatically want to adjust the search criteria when the map extent changes. E.g. you zoom in, we make a new request to get all features in the new extent.
But if you load the features and change the extent, you would have to make a new request to get the features in the new extent (and so on).

I would propose the following: When you load the page, you set a fixed extent (like you are already doing). Then you use this extent to request the features. The features that you get are all in the current map extent. So no need to recenter.

asaunier:
> I think it is a problem to change the extent when the features have been loaded. 

When you use the pagination buttons, a new request to the API is done, loading the new features.

Anyway, maybe you're right, recentering on the features is probably indeed conflicting with the map extent filtering.

> I would propose the following: When you load the page, you set a fixed extent (like you are already doing). Then you use this extent to request the features. The features that you get are all in the current map extent. So no need to recenter.

That makes sense. I'm starting working on taking the filtering criterias into account (including the map extent) and will think to a consistent approach, as you suggest it.

Anyway recentering on features is useful for some other map functionalities (for instance in the document detail view pages or when uploading a GPX file etc.), so we might still have to think to this animation/transition issue :)

asaunier:
With 9df9a98, as suggested by @tsauerwein, I have made the following changes:
* the map is no longer recentered on the search results features => the map extent remains unchanged
* if the user pans or zooms the map, a ``bbox`` GET parameter is added/updated in the URL and a ``searchFilterChange`` event is emitted (which tells the advanced search controller to trigger a new API request using the ``bbox`` GET parameter). 
* when loading the search page:
 * if a ``bbox`` GET parameter is available, the map is centered on this extent and the API request is triggered.
 * if no ``bbox``, the default map extent is used, ``bbox`` is set in the URL and the API request is triggered accordingly.

I have also improved the map controller code, regrouping some existing parts.

asaunier:
I still have some work to do with this PR but it's already possible to start reviewing it.
The PR has grown quite big, I would like to avoid adding many other changes right now.

A demo is available on my dev instance: http://c2corgv6-demo.gis.internal:6555/waypoints

What remains to do:
* [ ] make the sliderDirective interact with the ``searchFiltersDirective``
* [ ] support all changes in the checkbox panel (see waypoint_types) => as for now the refreshing of the model is not done every time the checkboxes selection is changed (actually only the first one)
* [ ] add the ``pl`` parameter to API requests
* [ ] full support of the advanced search permalink to restore the checkboxes and slider status
 * [ ] encoded permalink values must be converted back to the expected types (for instance ``wtyp="summit,pass"`` should be understood as ``wtype=["summit","pass"]``

What will be postponed to future issues/PRs:
* take "more filters" filters into account for waypoints. As for now only the first-level criterias are taken into account (title, waypoint_types, elevation).
* take filers into account for routes and outings (none are supported yet).
* cards are not rendered correctly in debug mode (same issue than for the simple search)
* outing cards layout need some CSS fixing
* better map transition (finally no longer used in advanced search since features are no longer recentered on). Maybe I should simply drop the animations added at https://github.com/c2corg/v6_ui/pull/347/commits/9d436590f22ec0a297859ac451419217fbdd9b0d ?
* Use the new cardDirective when associating a document (or even for all cards listed in the detail view pages?).
* Hilight the feature on the map when hovering a card (and vice-versa?)
* feature icon is now clickable in the detail view page map: either disable this behaviour or at least make sure the URL is OK
* allow to bypass the map extent filtering as in Airbnb?

## A brief explanation of the commits:

The first couple ones save the current "list pages" as "sitemap" + rename the current "search" directive to "simpleSearch" in order to avoid confusing simple and advanced search.

A cardDirective with partials for every document type has been created to render the cards in the results lists. Those lists are always retrieved using a direct API request from the user's browser (AJAX - there is no longer an API request made by the UI server).

Form criterias and map extent are used to add filtering parameters to the API request. The responsibility of the API requests goes to the ``advancedSearch`` directive. It receives the filtering params using the URL (thanks to ``ngeo.Location``). Those URL params are useful to restore the current search page when reloading the URL.

asaunier:
The PR is ready and can be reviewed. See my explanations above.
A demo is available at http://c2corgv6-demo.gis.internal:6555/waypoints

c2corg_ui/__init__.py
```diff
@@ -28,6 +28,8 @@ def main(global_config, **settings):
     config.add_route('waypoints_view', '/waypoints/{id}/{lang}')
     config.add_route('waypoints_index_default', '/waypoints')
     config.add_route('waypoints_index', '/waypoints/list*filters')
```
tsauerwein: Is the "remainder match" (`*filters`) required? Because normal query parameters (`?...`) are used, I think you don't need it.

c2corg_ui/static/js/card.js
```diff
@@ -0,0 +1,105 @@
...
+   * @private
+   */
+  this.doc_ = this['doc'];
+
+  /**
+   * @type {Object}
+   * @export
+   */
+  this.locale = this['doc']['locales'][0];
+  // TODO: use a better default locale than the first one?
```
tsauerwein: If you are requesting the documents with `?pl=...`, the first (and only) locale is already the "best" one.

c2corg_ui/static/js/map/map.js
```diff
@@ -126,58 +165,69 @@ app.MapController = function($scope, mapFeatureCollection) {
     this.addTrackImporter_();
   }
 
+  // advanced search mode
+  if (this.advancedSearch_) {
+    if (this.location_.hasParam('bbox')) {
+      var bbox = this.location_.getParam('bbox');
+      var extent = bbox.split(',');
+      if (extent.length == 4) {
+        // FIXME what if extent is not valid?
```
tsauerwein: Use the default extent that is used when no extent is given?

c2corg_ui/static/js/map/map.js
```diff
@@ -126,58 +165,69 @@ app.MapController = function($scope, mapFeatureCollection) {
...
+      }
+    }
+
+    this.scope_.$root.$on('searchFeaturesChange',
+        this.handleSearchChange_.bind(this));
+
+    this.view_.on('propertychange',
+      ngeoDebounce(
+        this.handleMapSearchChange_.bind(this),
+        500, /* invokeApply */ true));
```
tsauerwein: :+1:

c2corg_ui/static/js/searchfilters.js
```diff
@@ -61,76 +52,121 @@ app.module.directive('appSearchFilters', app.searchFiltersDirective);
...
+  // Fill the filters according to the loaded URL parameters
+  var keys = this.location_.getParamKeys().filter(function(x) {
+    return app.SearchFiltersController.IGNORED_FILTERS.indexOf(x) === -1;
+  });
+  for (var i = 0, n = keys.length; i < n; i++) {
+    this.getFilterFromPermalink_(keys[i]);
+  }
 
+  // Deep watch is used here because we need to watch the list filters as well
+  // which a simple $watch or $watchCollection does not. Might cause
```
tsauerwein: "Might cause perf/memory issues though..."

This is true for complex objects. But is `this.filters` complex? Cloning and comparing an object with max ~15 properties should be no problem.

c2corg_ui/static/js/searchfilters.js
```diff
@@ -61,76 +52,121 @@ app.module.directive('appSearchFilters', app.searchFiltersDirective);
...
-   * @type {Object}
-   * @export
-   */
-  this.elevation = {min : 0, max: 9999};
+/**
+ * @param {string} key Filter key.
+ * @private
+ */
+app.SearchFiltersController.prototype.getFilterFromPermalink_ = function(key) {
+  // TODO find a more generic way or at least factorize code
```
tsauerwein: On the API side I am doing something similar (create filters for ES from URL params):
https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/search_filters.py

c2corg_ui/static/js/utils.js
```diff
@@ -2,6 +2,26 @@ goog.provide('app.utils');
...
+    case 'w':
+      return 'waypoints';
+    case 'r':
+      return 'routes';
+    case 'o':
+      return 'outings';
+    case 'i':
+      return 'images';
+    default:
+      return '';
```
tsauerwein: Maybe use `goog.assert.fail` to early raise an error.

tsauerwein:
That is looking great! :+1:

c2corg_ui/__init__.py
asaunier: Right! At first I thought I would use a ``key/value`` syntax for URLs but it seems it is not possible with ``ngeo.Location`` and anyway I'm not sure this syntax would be useful since the page heavily relies on AJAX requests and thus can't really be cached. So I decided not to spend time using this special syntax.
Then this route is indeed no longer needed. I will remove it. Thanks for noticing :)

asaunier: Done at a39ded8
* removed the remainder match ``*filters``
* renamed routes ``*_index_default`` to ``*_index`` => URLs like ``waypoints/list/foo/bar`` (as in v5) are no longer supported, only ``waypoints?foo=bar``. In some way it hurts a bit the backward compatibility from v5 but I prefer we avoid having a bunch of "legacy" routes in the pyramid code.

c2corg_ui/static/js/card.js
asaunier: OK than I can remove this comment since ``pl=...`` is automatically added to the API requests.

c2corg_ui/static/js/map/map.js
asaunier: Sure, it means simply ignore the invalid extent.
But what I meant was: how can we figure out that the provided extent is not valid?
I could for instance check that xmin is actually < than xmax etc. but that's a bit laborious. Or that the final extent items are numbers. Perhaps there's something in OL3 that does that kind of check (I have not found)?

Or maybe we can assume it's not that a big deal: if the extent is invalid, the search will obviously fail and that's it!?

c2corg_ui/static/js/searchfilters.js
asaunier: OK, thanks for confirming.
I thought it was worth warning in case the list of criterias grows big and complex :)

c2corg_ui/static/js/map/map.js
tsauerwein: > Or maybe we can assume it's not that a big deal: if the extent is invalid, the search will obviously fail and that's it!?

Yep. In the worst case you will not get results or the bbox filter is ignored.

asaunier: OK, then I will remove the comment and leave the code as is. Thanks!

c2corg_ui/static/js/utils.js
asaunier: You mean:
```
    default:
      goog.assert.fail('Unrecognized type: ' + type);
      return '';
```
(not sure the final "return" is needed?)?

I have tried to add ``"goog.asserts.ENABLE_ASSERTS=true",`` to https://github.com/c2corg/v6_ui/blob/master/build.json#L32-L39
(as found in the doc of ``goog.assert.fail`` http://google.github.io/closure-library/api/namespace_goog_asserts.html) but I keep having the following build error:
> ERR! compile /home/alex/Desktop/v6_ui/c2corg_ui/static/js/utils.js:20: ERROR - Property assert never defined on goog
ERR! compile       goog.assert.fail('Unrecognized type: ' + type);

c2corg_ui/static/js/searchfilters.js
asaunier: OK I have just pushed some refactoring of this code (d448c98), a bit like what you have done in the API.
My main concern is that I have to explicitely test for some specific parameters (eg. "walt", "wtyp") to convert them using the right filters creating functions.
I have tried to pass the filter types in the filter forms themselves using for instance
```
ng-init="filtersCtrl.filterTypes.wtyp = 'list'"
```
but it's a bit complicated making sure the types info are actually loaded when running the permalink parsing.

Anyway I prefer postponing this issue to a future "advanced search step2" issue/PR.

asaunier:
Thanks @tsauerwein for reviewing.
I have made the suggested changes except https://github.com/c2corg/v6_ui/pull/347#discussion_r64526322 because I have found no way yet to make the compiler accept ``goog.assert.fail``.

@gberaudo do you have comments or is the PR OK for you?

c2corg_ui/static/js/utils.js
tsauerwein: `goog.asserts` is already used elsewhere (e.g. [here](https://github.com/c2corg/v6_ui/blob/e6a9a14398131a6e00993968d1c121d2088ffa93/c2corg_ui/static/js/map/diffmap.js#L161)), so you don't have to change the build config. But it is `goog.asserts` and not `goog.assert` (note the "s").

asaunier: Arf... there was a typo in the code example I copied from http://google.github.io/closure-library/api/namespace_goog_asserts.html :P 
Fixed! Thanks!

gberaudo:
I am having a look to it now.

c2corg_ui/static/js/advancedsearch.js
```diff
@@ -0,0 +1,78 @@
...
+
+  this.getResults_();
+};
+
+
+/**
+ * @private
+ */
+app.AdvancedSearchController.prototype.getResults_ = function() {
+  this.api_.listDocuments(this.doctype_).then(function(response) {
```
gberaudo: This could be simplified by having `listDocuments` directly return the data.
On the model of `appx.SearchDocumentResponse`, the data and `this.documents` could be typed.

c2corg_ui/static/partials/advancedsearch.html
```diff
@@ -0,0 +1,6 @@
+Nb of results: {{searchCtrl.total}}
```
gberaudo: Missing translation?

gberaudo: Hmm, seems to be gone in a later commit.

c2corg_ui/static/js/advancedsearch.js
```diff
@@ -0,0 +1,179 @@
...
+ */
+app.AdvancedSearchController.prototype.successList_ = function(response) {
+  if (!('data' in response)) {
+    return;
+  }
+  this.documents = response['data']['documents'];
+  this.total = response['data']['total'];
+  if (this.resCounter_) {
+    $('#' + this.resCounter_).html(
+      /** @type {string} */ (this.total.toString()));
```
gberaudo: Was the cast needed?

c2corg_ui/static/js/advancedsearch.js
```diff
@@ -0,0 +1,179 @@
...
+ * @return {Array.<ol.Feature>}
+ * @private
+ */
+app.AdvancedSearchController.prototype.getFeatures_ = function() {
+  var features = [];
+  var format = new ol.format.GeoJSON();
+  for (var i = 0, n = this.documents.length; i < n; i++) {
+    var doc = this.documents[i];
+    if ('geometry' in doc && doc['geometry']) {
+      var properties = this.getFeatureProperties_(doc);
```
gberaudo: Maybe rename to `createFeatureProperties_`?

c2corg_ui/static/js/advancedsearch.js
```diff
@@ -0,0 +1,179 @@
...
+};
+
+
+/**
+ * @param {Object} doc document data.
+ * @return {Object}
+ * @private
+ */
+app.AdvancedSearchController.prototype.getFeatureProperties_ = function(doc) {
+  // TODO choose the locale according to the UI lang and user prefs
```
gberaudo: Do you still plan to implement the TODO?

c2corg_ui/static/js/map/map.js
```diff
@@ -126,58 +165,68 @@ app.MapController = function($scope, mapFeatureCollection) {
...
+    var clickInteraction = new ol.interaction.Select({
+      condition: ol.events.condition.click
+    });
+    clickInteraction.on('select', function(e) {
+      /**
+       * @type {ol.Collection.<ol.Feature>}
+       */
+      var features = e.target.getFeatures();
+      if (features.getLength() > 0) {
+        var first = features.getArray()[0];
```
gberaudo: There is an `item` method you could use.

c2corg_ui/static/js/map/map.js
```diff
@@ -114,9 +143,19 @@ app.MapController = function($scope, mapFeatureCollection) {
...
   });
 
+  /**
+   * @type {ol.View}
+   * @private
+   */
+  this.view_ = this.map.getView();
+
   // editing mode
   if (this['edit']) {
```
gberaudo: Not your code, but the edit variable should not be accessed using the bracket notation but declared and accessed using the point notation. In addition, the constructor should be marked `@struct` to avoid accessing inexistant variables.

c2corg_ui/static/js/map/map.js
```diff
@@ -245,6 +284,20 @@ app.MapController.DEFAULT_POINT_ZOOM = 12;
 
 
 /**
+ * @param {ol.Extent} extent Extent to recenter to.
+ * @param {olx.view.FitOptions=} options Options.
+ * @private
+ */
+app.MapController.prototype.recenterOnExtent_ = function(extent, options) {
+  var mapSize = this.map.getSize();
+  if (mapSize) {
```
gberaudo: Is the test still useful now that `change:size` is listened to on the map?

c2corg_ui/static/js/searchfilters.js
```diff
@@ -17,35 +20,23 @@ app.searchFiltersDirective = function() {
...
-      // Update height diff slider values
-      heightFilter.on('slide', function(slideEvt) {
-        ctrl.height_diff_up.min = slideEvt.value[0];
-        ctrl.height_diff_up.max = slideEvt.value[1];
-        scope.$apply();
-      });
-      //on some screen sizes, the dropdown menu is too large and is hidden within documents-list-section
-      // because it's to small. Given that this menu is inside, it will not be show entirely...
+      // FIXME On some screen sizes, the dropdown menu is too large and is
+      // hidden within documents-list-section because it's to small. Given that
```
gberaudo: Not your prose but there is a missing o in `to small`.

c2corg_ui/static/js/searchfilters.js
gberaudo: Where is the deep watch in the expression below?

gberaudo: Argh, need to change my eyes. Would be good to add a new line.

c2corg_ui/static/js/advancedsearch.js
asaunier: It may looks odd indeed to cast the output of ``toString()`` to a string but I think I had a compiling error otherwise...

c2corg_ui/static/partials/card_outings.html
```diff
@@ -0,0 +1,15 @@
+<a ng-click="cardCtrl.openCardPage()" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <div class="outing-author text-center">
+    <div class="outing-date">{{ doc['date_end'] | date: 'dd/MM/yyyy'}}</div>
```
gberaudo: Is it using moment for translations?

c2corg_ui/static/partials/card_routes.html
```diff
@@ -0,0 +1,12 @@
+<a ng-click="cardCtrl.openCardPage()" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <span class="list-item-title"><span ng-if="locale.title_prefix">{{locale.title_prefix}}&nbsp;: </span>{{locale.title}}</span>
+  <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{locale.lang}})</span>
+  <br>
+  <div class="list-item-info">
+    Min {{doc['elevation_min']}} m / Max {{doc['elevation_max']}} m
```
gberaudo: Missing translation

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {},
   "license": "BSD",
   "dependencies": {
-    "closure-util": "1.10.0",
+    "closure-util": "1.13.1",
```
gberaudo: It is 1.14.0 already :P

gberaudo:
Thanks @asaunier, a great PR. I added a few comments. Most are minors.
There are several places where the code could have been cleaner and more safe if an extern had been used; that beeing said, time is not extensible so feel free to do as you think is best. +1 for merging.

c2corg_ui/static/partials/card_routes.html
asaunier: I know but it's a very first version of the route cards with "demo" attributes. More work is needed. I am not even sure those attributes are the ones to show. Height diffs or ratings would make more sense. But that should be done in another PR.
So I don't think it is worth adding i18n here now.

c2corg_ui/static/partials/card_outings.html
asaunier: No clue :)
It's just a port from a Mako helper:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/view.html#L310-L313

c2corg_ui/static/js/map/map.js
asaunier: Maybe not. But is it really safe to assume this when we don't really not where this method is called?
As for now it is also used in the ``showFeatures`` function: https://github.com/c2corg/v6_ui/blob/advanced-search-ui/c2corg_ui/static/js/map/map.js#L434-L448

c2corg_ui/static/partials/card_routes.html
gberaudo: It makes sense, did not know about that.

c2corg_ui/static/js/advancedsearch.js
asaunier: Please note this is quite outdated code. See https://github.com/c2corg/v6_ui/blob/advanced-search-ui/c2corg_ui/static/js/advancedsearch.js#L108-L116 instead (even though the way it works is close).

> This could be simplified by having listDocuments directly return the data.

I don't understand what you mean. The way ``listDocuments`` is consistent with the other ``apiService`` methods. The success callback is quite specific to the advanced search and cannot be outsourced to the api service.

> On the model of appx.SearchDocumentResponse, the data and this.documents could be typed.

You mean I should create a ``appx.SearchDocumentResponse`` precising the response types as for ``appx.SimpleSearchDocumentResponse`` or maybe reuse it?

c2corg_ui/static/js/advancedsearch.js
asaunier: OK

c2corg_ui/static/js/searchfilters.js
asaunier: > Would be good to add a new line.

OK

c2corg_ui/static/js/advancedsearch.js
gberaudo: Not exactly what I proposed but you are right, the API service should return a response.

Are simple search and advanced search responses the same type? If yes it is indeed just a matter of changing the type returned by the API function to a promise of `appx.SearchDocumentResponse`.

But if it is different, then a similar response type and document type should be created.

c2corg_ui/static/js/map/map.js
asaunier: Argh, the changes we have made together passes in the compiler but generates errors in the console. For instance:
> angular.js:13236 Error: [$compile:nonassign] Expression 'undefined' in attribute 'appMapEdit' used with directive 'appMap' is non-assignable!
http://errors.angularjs.org/1.5.0/$compile/nonassign?p0=undefined&p1=appMapEdit&p2=appMap
    at angular.js:68
    at parentSet (angular.js:9550)
    at parentValueWatch (angular.js:9563)
    at regularInterceptedExpression (angular.js:15214)
    at Scope.$digest (angular.js:16655)
    at Scope.$apply (angular.js:16928)
    at bootstrapApply (angular.js:1694)
    at Object.invoke (angular.js:4604)
    at doBootstrap (angular.js:1692)
    at bootstrap (angular.js:1712)

asaunier: Perhaps because of the "=" at https://github.com/c2corg/v6_ui/blob/advanced-search-ui/c2corg_ui/static/js/map/map.js#L42

gberaudo: Strange. Are you sure the error was triggered by our change?

asaunier: It seemed related to lines such as
```
this.edit |= false;
```
I have remove them (we don't really need them, if the matching directive attribute is not set, the property is ``undefined`` which evaluates the same way then ``false``) and the error disappeared.

asaunier:
Thanks @gberaudo for reviewing!
I have made most of the changes you suggested except the 2 following ones that I have added to the TODO list of step 2 (I will create a new ticket):
* add an extern entry for the listDocuments function in apiService.
* make sure the features on the map use the most appropriate locales.

I will merge the PR soon if no one objects and if Travis tests pass.

PR #359, opened 2016-05-17
Reload recaptcha on expiration and failure

### LINKED ISSUE #358: The captcha should be reloaded
There are two cases when the captcha should be reloaded:
- on expiration;
- on submission failure.

gberaudo:
Fix #358.


asaunier:
Looks good.
Just curious: what/who is VC?

gberaudo:
There are several implementations of the Recaptcha client. VC stands for Vivid Cortex which is the implementation we use for angular. See https://github.com/c2corg/v6_ui/blob/master/package.json.

gberaudo:

PR #360, opened 2016-05-18
Redirect when login out in some cases

### LINKED ISSUE #350: Redirect user when logging out from an authenticated-only page
When the current page is only relevant for authenticated users, if the user clicks on the "log out" button, he should be redirected to some public page (eg. home page or the auth page?) and not stay in the current page since he has no longer permission to see/use it. For instance:
* [x] account page
* [x] edit page

gberaudo:
Redirect when the path contains '/edit/' or '/account'.

Close #350.

### DIFF
--- a/c2corg_ui/static/js/user.js
+++ b/c2corg_ui/static/js/user.js
@@ -96,6 +96,12 @@ app.UserController.prototype.logout = function() {
     this.alerts_.addSuccess('You have been disconnected');
   }.bind(this)).finally(function() {
     this.auth.removeUserData();
+    var path = this.ngeoLocation_.getPath();
+    if (path.indexOf('/edit/') !== -1 || path.indexOf('/account') !== -1) {
+      // The user is editing a document or viewing the account configuration.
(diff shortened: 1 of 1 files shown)


c2corg_ui/static/js/user.js
```diff
@@ -96,6 +96,10 @@ app.UserController.prototype.logout = function() {
     this.alerts_.addSuccess('You have been disconnected');
   }.bind(this)).finally(function() {
     this.auth.removeUserData();
+    var path = this.ngeoLocation_.getPath();
+    if (path.indexOf('/edit/') !== -1 || path.indexOf('/account') !== -1) {
```
asaunier: Looks good but I wonder if a clearer syntax could be used (the kind of double negation makes the reading not that obvious). At least add a comment?

gberaudo: OK, done.

asaunier: Thanks, please merge.

gberaudo:

Issue #354, opened 2016-05-14
Geometry editing


asaunier:
* [x] When editing an existing geometry, the user should not be able to draw a new separated line. 
* [x] If no line geometry is available for a route or outing, the map should be recentered on the default point geometry of the document (not the global default map extent). Fixed by https://github.com/c2corg/v6_ui/pull/371
* [x] Allow to reset the current geometry

PR #371, opened 2016-05-26
Recenter on the default geometry when editing a document with no deta…

asaunier:
…il geom

See https://github.com/c2corg/v6_ui/issues/354


asaunier:
I will merge this PR after lunch if no one objects :)

Issue #363, opened 2016-05-23
WP 4 UI remaining errors/fixes


ginold:
I allowed myself to create a single bigger issue with all the others gathered here because we have so many of them that I have troubles navigating between them.
- [x] don't list outings in simple-search

- [x] If set, the main waypoint name (title_prefix) should appear in front of the title input field. It should not be editable but changes when the main waypoint changes.

- [x] ratings for activity "via ferrata" are missing.

- [x] h2 and h3 titles in the document content (markdown formatting) have the same size.

- [x]  users-associating tool crashes, see c2corg/v6_api#286

- [x]  routes-associating tool should not try to make the association directly and wait that the outing editing is done (ditto for the users tool in OUTING editing form)

- [x]  the "summary" page blocks because of "outing_type" and "elevation" attribute that do not exist in the editing form. For instance https://github.com/c2corg/v6_ui/search?utf8=%E2%9C%93&q=outing_type

- [x] When creating an outing, the outing title should be set by default to the name of the first associated route.

- [x]  the association tool should strip already associated docs out of the search results list.

- [x] Cancel and Save (or Preview) buttons should be available at every step

- [ ]  The dropdown list is cut off if not enough space below to render the full list of results => should display above in that case. (AWAITING INFO)

- [ ] ~~Show WP hierarchy ("sommet/sur-sommet", etc.) => how to do this with cards? See #348~~

- [x]  The "plus" button in the detail view page no longer reacts when clicking on it.

asaunier:
@fjacon Should we close the other issues since they are all grouped here?

> ratings for activity "via ferrata" are missing.

@ginold This is for route editing, not outing editing. See https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py#L174

fjacon:
Yes we should close the other issues
thanks!

asaunier:
An other problem I have just noticed:

* [ ] The "plus" button in the detail view page no longer reacts when clicking on it.

asaunier:
* [ ] When associating a document (for instance a route to the current WP), if the related associated-docs block is not shown yet, the newly associated-document card is not visible until the page is reloaded.

ginold:
solved in #328 while ago. 
BTW have a look at it, it should be mergeable.

asaunier:
> The dropdown list is cut off if not enough space below to render the full list of results => should display above in that case. (AWAITING INFO)

[...]

PR #372, opened 2016-05-26
Editing fixes

### LINKED ISSUE #380: Outings should not be returned when using the standard simple search tool

ginold:
fixes almost all of the #363 issues.
fixes #380

### DIFF
--- a/c2corg_ui/externs/appx.js
+++ b/c2corg_ui/externs/appx.js
@@ -84,3 +84,207 @@ appx.BootstrapSliderValues;
  * }}
  */
 appx.auth.RequestChangePassword;
+
+
+/**
+ * @typedef {{
+ *   total: number,
+ *   outings: Array.<appx.Outing>
+ * }}
+ */
+appx.RecentOutings;
+
+
+/**
+ * @typedef {{
+ *   total: number,
+ *   routes: Array.<appx.Route>
+ * }}
+ */
+appx.AllRoutes;
+
+
+/**
+ * @typedef {{
+ *   images: Array.<Object>,
+ *   users: ?Array.<Object>,
+ *   routes: ?Array.<appx.Route>,
+ *   waypoints: ?Array.<appx.Waypoint>,
+ *   waypoint_children: ?Array.<appx.Waypoint>,
+ *   waypoint_parents: ?Array.<appx.Waypoint>,
+ *   recent_outings: ?appx.RecentOutings,
+ *   all_routes: ?appx.AllRoutes
+ * }}
+ */
+appx.DocumentAssociations;
+
+
+/**
+ * @typedef {{
+ *   associations: appx.DocumentAssociations,
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   quality: string,
+ *   document_id: number
+ * }}
+ */
+appx.Document;
+
+
+/**
+ * @typedef {{
+ *   associations: appx.DocumentAssociations,
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   quality: string,
+ *   document_id: number,
+ *   
+ *   geometry: Object,
+ *   date_start: (string | Date),
+ *   date_end: (string | Date),
+ *   activities : Array.<Object>,
+ *   frequentation: string,
+ *   participant_count : (number | string),
+ *   elevation_min : number,
+ *   elevation_max : number,
+ *   elevation_up_snow: number,
+ *   elevation_down_snow: number,
+ *   snow_quantity: string,
+ *   snow_quality: string,
+ *   glacier_rating: string,
+ *   duration: number,
+ *   avalanche_signs: string,
+ *   elevation_access: number,
+ *   height_diff_up: number,
+ *   height_diff_down: number,
+ *   length_total: number,
+ *   partial_trip: boolean,
+ *   public_transport: string,
+ *   access_condition: string,
+ *   lift_status: string,
+ *   awesomeness: string,
+ *   condition_rating: string,
+ *   hut_status: string,
+ * }}
+ */
+appx.Outing;
+
+
+/**
+ * @typedef {{
+ *   associations: appx.DocumentAssociations,
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   quality: string,
+ *   document_id: number,
+ *   
+ *   height_diff_up: number,
+ *   height_diff_down: number,
+ *   height_diff_access: number,
+ *   height_diff_difficulties: number,
+ *   lift_access: boolean,
+ *   elevation_min : number,
+ *   elevation_max : number,
+ *   route_type: Array.<string>,
+ *   orientations: Array.<string>,
+ *   durations: number,
+ *   main_waypoint_id: number,
+ *   activities: Array.<Object>,
+ *   route_length: number,
+ *   configuration: Array.<string>,
+ *   ski_rating: string,
+ *   ski_exposition: string,
+ *   labande_ski_rating: string,
+ *   labande_global_rating: string,
+ *   difficulties_height: number,
+ *   engagement_rating: string,
+ *   equipment_rating: string,
+ *   global_rating: string,
+ *   ice_rating: string,
+ *   geometry: Object,
+ *   risk_rating: string,
+ *   rock_types: Array.<string>,
+ *   mixed_rating: string,
+ *   glacier_gear: string,
+ *   exposition_rock_rating: string,
+ *   rock_free_rating: string,
+ *   rock_required_rating: string,
+ *   aid_rating: string,
+ *   climbing_outdoor_type: string,
+ *   hiking_mtb_exposition: string,
+ *   hiking_rating: string,
+ *   snowshoe_rating: string,
+ *   mtb_up_rating: string,
+ *   mtb_down_rating: string,
+ *   mtb_length_asphalt: string,
+ *   mtb_length_trail: string,
+ *   mtb_height_diff_portages: string,
+ *   via_ferrata_rating: string
+ * }}
+ */
+appx.Route;
+
+/**
+ * @typedef {{
+ *   associations: appx.DocumentAssociations,
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   quality: string,
+ *   document_id: number,
+ *   
+ *   elevation: number,
+ *   elevation_min: number,
+ *   maps: Array.<Object>,
+ *   maps_info: string,
+ *   areas: Array.<Object>,
+ *   public_transportation_types: Array.<string>,
+ *   public_transportation_rating: string,
+ *   prominence: number,
+ *   height_max: number,
+ *   height_min: number,
+ *   height_median: number,
+ *   length: number,
+ *   geometry: Object,
+ *   slope: (number | string),
+ *   ground_types: Array.<string>,
+ *   paragliding_rating: string,
+ *   exposition_rating: string,
+ *   routes_quantity: number,
+ *   rain_proof: string,
+ *   children_proof: string,
+ *   access_time: string,
+ *   snow_clearance_rating: string,
+ *   lift_access: (boolean | null),
+ *   parking_fee: (boolean | string),
+ *   equipment_ratings: Array.<string>,
+ *   climbing_rating_min: string,
+ *   climbing_rating_median: string,
+ *   climbing_rating_max: string,
+ *   url: string,
+ *   climbing_styles: Array.<string>,
+ *   orientations: Array.<string>,
+ *   equipment_rating: string,
+ *   best_periods: Array.<string>,
+ *   climbing_outdoor_types: Array.<string>,
+ *   phone: (number | string),
+ *   capacity: number,
+ *   capacity_staffed: number,
+ *   phone_custodian: (number | string),
+ *   custodianship: string,
+ *   matress_unstaffed: (string | boolean),
+ *   blanket_unstaffed: (string | boolean),
+ *   gas_unstaffed: (string | boolean),
+ *   heating_unstaffed: (string | boolean)
+ * }}
+ */
+appx.Waypoint;
+
+/**
+ * @typedef {{
+ *   associations: appx.DocumentAssociations,
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   document_id:number
+ * }}
+ */
+appx.Image;
--- a/c2corg_ui/static/js/addassociationdirective.js
+++ b/c2corg_ui/static/js/addassociationdirective.js
@@ -14,7 +14,7 @@ goog.require('app.Api');
 app.addAssociationDirective = function($compile) {
 
   var template = function(dataset) {
-    return '<app-simple-search app-select="addCtrl.associate(doc)" app-dataset="' + dataset + '"></app-simple-search>'
+    return '<app-simple-search app-select="addCtrl.associate(doc)" dataset="' + dataset + '"></app-simple-search>';
   };
 
   return {
@@ -22,16 +22,15 @@ app.addAssociationDirective = function($compile) {
     controller: 'AppAddAssociationController',
     bindToController: {
       'parentId': '=',
-      'addedDocuments': '=',
-      'dataset' : '='
+      'addedDocuments': '='
     },
     controllerAs: 'addCtrl',
 
     link: function(scope, element, attrs, ctrl) {
-      element.html(template(ctrl.dataset));
+      element.html(template(attrs.dataset));
       $compile(element.contents())(scope);
     }
-  }
+  };
 };
 
 app.module.directive('appAddAssociation', app.addAssociationDirective);
@@ -41,10 +40,12 @@ app.module.directive('appAddAssociation', app.addAssociationDirective);
  * @constructor
  * @param {app.Api} appApi The API service
  * @ngInject
+ * @struct
  */
-app.AddAssociationController = function(appApi, $attrs) {
+app.AddAssociationController = function(appApi) {
+
   /**
-   * @type {number} // bound from directive
+   * @type {number}
    * @export
    */
   this.parentId;
@@ -55,7 +56,6 @@ app.AddAssociationController = function(appApi, $attrs) {
   this.api_ = appApi;
 
 
-  this.dataset = $attrs['dataset'];
   /**
    * @type {Array.<appx.SimpleSearchDocument>}
    * @export
--- a/c2corg_ui/static/js/auth/authservice.js
+++ b/c2corg_ui/static/js/auth/authservice.js
@@ -79,6 +79,7 @@ app.Authentication.prototype.isAuthenticated = function() {
  * @export
  */
 app.Authentication.prototype.hasEditRights = function(users) {
+  // not logged -> return false
   if (!this.isAuthenticated()) {
     return false;
   }
@@ -100,7 +101,6 @@ app.Authentication.prototype.hasEditRights = function(users) {
   } else {
     return true;
   }
-  // not logged -> return false
 };
 
 
--- a/c2corg_ui/static/js/documentdetailsview.js
+++ b/c2corg_ui/static/js/documentdetailsview.js
@@ -2,6 +2,7 @@ goog.provide('app.ViewDetailsController');
 goog.provide('app.viewDetailsDirective');
 
 goog.require('app');
+goog.require('app.Document');
 
 /**
  * @ngInject
@@ -20,13 +21,12 @@ app.viewDetailsDirective = function() {
         ctrl.initPhotoswipe_();
         $('.photos figure').each(function() {
           $(this).css('width', '');
-        })
+        });
       }
       ctrl.loadImages_(initGalleries);
 
-      $('.pswp__button.info').click(function(e) {
+      $('.pswp__button.info').on('touchend click', function(e) {
         $('.photoswipe-image-container .image-infos, .photoswipe-image-container img').toggleClass('showing-info');
-        //$compile($('.image-infos.showing-info'))(ctrl.scope_);
       });
 
       $('.pswp__button--arrow--left, .pswp__button--arrow--right').click(function() {
@@ -44,7 +44,7 @@ app.viewDetailsDirective = function() {
         if (mql.matches) {
           $('.tab, .accordion').show();
         }
-      })
+      });
 
       // show description tab by default or selected tab
       notPhone.addListener(function(mql) {
@@ -69,12 +69,19 @@ app.module.directive('appViewDetails', app.viewDetailsDirective);
  * @param {Object} $uibModal modal from angular bootstrap
  * @param {!angular.Scope} $scope Scope.
  * @param {app.Api} appApi Api service.
+ * @param {app.Document} appDocument service
  * @param {angular.$compile} $compile Angular compile service.
  * @constructor
  * @export
  * @ngInject
  */
-app.ViewDetailsController = function($uibModal, $compile, $scope, appApi) {
+app.ViewDetailsController = function($uibModal, $compile, $scope, appApi, appDocument) {
+
+  /**
+   * @type {app.Document}
+   * @export
+   */
+  this.documentService = appDocument;
 
   /**
    * @type {Object}
@@ -99,14 +106,18 @@ app.ViewDetailsController = function($uibModal, $compile, $scope, appApi) {
    * @private
    */
   this.scope_ = $scope;
-  this.scope_['document'] = { // hardcoded for testing
+
+  /**
+   * @type {appx.Document}
+   */
+  this.scope_['document'] = {// hardcoded for testing
     'associations': {
       'images': [
-        {'src': 'http://s.camptocamp.org/uploads/images/1463740758_235613570.jpg', date: '23-02-2015', activities: ['hiking, snow'], 'elevation': Math.random(), iso_speed: 100, filename: 'image2.jpg', camera_name: 'Nikon', locales: [{'title' : 'inizio del canale'}], date_time: new Date(), 'fnumber' : 2.3},
-        {'src': 'http://s.camptocamp.org/uploads/images/1463694562_1216719818.jpg', date: '23-22-2025', activities: ['hiking'], elevation: Math.random(), iso_speed: 230, filename: 'image1.jpg', camera_name: 'Sony', locales: [{'title' : 'never let me down'}], date_time: new Date(), 'fnumber' : 5.5},
-        {'src': 'http://s.camptocamp.org/uploads/images/1463694970_824684659.jpg', date: new Date(), activities: ['snow'], elevation: Math.random(), iso_speed: 400, filename: 'image231.jpg', camera_name: 'Canon', locales: [{'title' : 'my favorite place'}], date_time: new Date(), 'fnumber' : 4.2},
-        {'src': 'http://s.camptocamp.org/uploads/images/1463741192_488006925.jpg', date: '23-323', activities: ['hiking, snow, ski'], elevation: Math.random(), iso_speed: 1200, filename: 'image94.jpg', camera_name: 'Fuji', locales: [{'title' : 'superb view'}]},
-        {'src': 'http://s.camptocamp.org/uploads/images/1463694980_966569102.jpg', date: '099-02-2015', activities: ['paragliding'], elevation: Math.random(), iso_speed: 2300, filename: 'image55.jpg', camera_name: 'Sigma', locales: [{'title' : 'great view'}] , date_time: new Date(), 'fnumber' : 10}
+        {'src': 'http://s.camptocamp.org/uploads/images/1463740758_235613570.jpg', date: '23-02-2015', activities: ['hiking, snow'], 'elevation': Math.random(), iso_speed: 100, filename: 'image2.jpg', camera_name: 'Nikon', locales: [{'title': 'inizio del canale'}], date_time: new Date(), 'fnumber': 2.3},
+        {'src': 'http://s.camptocamp.org/uploads/images/1463694562_1216719818.jpg', date: '23-22-2025', activities: ['hiking'], elevation: Math.random(), iso_speed: 230, filename: 'image1.jpg', camera_name: 'Sony', locales: [{'title': 'never let me down'}], date_time: new Date(), 'fnumber': 5.5},
+        {'src': 'http://s.camptocamp.org/uploads/images/1463694970_824684659.jpg', date: new Date(), activities: ['snow'], elevation: Math.random(), iso_speed: 400, filename: 'image231.jpg', camera_name: 'Canon', locales: [{'title': 'my favorite place'}], date_time: new Date(), 'fnumber': 4.2},
+        {'src': 'http://s.camptocamp.org/uploads/images/1463741192_488006925.jpg', date: '23-323', activities: ['hiking, snow, ski'], elevation: Math.random(), iso_speed: 1200, filename: 'image94.jpg', camera_name: 'Fuji', locales: [{'title': 'superb view'}]},
+        {'src': 'http://s.camptocamp.org/uploads/images/1463694980_966569102.jpg', date: '099-02-2015', activities: ['paragliding'], elevation: Math.random(), iso_speed: 2300, filename: 'image55.jpg', camera_name: 'Sigma', locales: [{'title': 'great view'}], date_time: new Date(), 'fnumber': 10}
       ]
     }
   };
@@ -312,7 +323,7 @@ app.ViewDetailsController.prototype.loadImages_ = function(initGalleries) {
     $('.photos').append(element);
 
     var scope = this.scope_.$new(true);
-    scope['photo'] = photos[i]
+    scope['photo'] = photos[i];
     this.compile_($('.photos figure:last-of-type').contents())(scope);
   }
   initGalleries();
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -48,6 +48,7 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @param {app.Authentication} appAuthentication
  * @param {ngeo.Location} ngeoLocation ngeo Location service.
  * @param {app.Alerts} appAlerts
+ * @param {app.Document} appDocument
  * @param {app.Api} appApi Api service.
  * @param {string} authUrl Base URL of the authentication page.
  * @constructor
@@ -55,20 +56,20 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @export
  */
 app.DocumentEditingController = function($scope, $element, $attrs,
-    appAuthentication, ngeoLocation, appAlerts, appApi, authUrl) {
+    appAuthentication, ngeoLocation, appAlerts, appApi, authUrl, appDocument) {
+
+  /**
+   * @type {app.Document}
+   * @export
+   */
+  this.documentService = appDocument;
+
 
   /**
    * @type {ngeo.Location}
    * @private
    */
   this.ngeoLocation_ = ngeoLocation;
-
-
-  /**
-   * @type {angular.Scope}
-   * @private
-   */
-  this.scope_ = $scope;
 
   /**
    * @type {string}
@@ -106,6 +107,12 @@ app.DocumentEditingController = function($scope, $element, $attrs,
    * @private
    */
   this.lang_ = $attrs['appDocumentEditingLang'];
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
 
   /**
    * @type {ol.format.GeoJSON}
@@ -168,12 +175,6 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   this.submit_ = false;
 
   /**
-   * Waypoint init
-   * @private
-   */
-  this.scope_.waypoint = {};
-
-  /**
    * @export
    */
   this.differentDates;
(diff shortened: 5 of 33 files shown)


c2corg_ui/templates/waypoint/associated.html
```diff
@@ -7,13 +7,13 @@
          ng-style="document['main_waypoint_id'] === waypoint['document_id'] && { 'order' : '-1'}  ">
       
       <%include file="../waypoint/card.html"/>
-      <div translate ng-if="document.main_waypoint_id === waypoint.document_id">main waypoint</div>
+      <div translate ng-if="document.main_waypoint_id === waypoint.document_id" ng-init="document['main_waypoint_title'] = waypoint.locales[0].title">main waypoint</div>
```
asaunier: Using ``document.main_waypoint_title`` instead of ``document['main_waypoint_title']`` would be more consistent with the following lines.

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -260,6 +260,16 @@ app.Authentication.prototype.addAuthorizationToHeaders = function(url,
 
 
 /**
+ * @export
```
asaunier: The ``@return`` jsdoc is missing.

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -260,6 +260,16 @@ app.Authentication.prototype.addAuthorizationToHeaders = function(url,
...
+ */
+app.Authentication.prototype.getUserToken = function() {
+  if (this.userData) {
+    return this.userData.token;
+  }
```
asaunier: The function should probably return something when ``this.userData`` is not set.

c2corg_ui/static/js/simplesearch.js
```diff
@@ -94,7 +94,7 @@ app.module.directive('appSimpleSearch', app.simpleSearchDirective);
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @ngInject
  */
-app.SimpleSearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog, $templateCache) {
+app.SearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog, $templateCache, appAuthentication) {
```
asaunier: Please don't revert my changes :)
The "search" directive/controller are now named "simpleSearch".

c2corg_ui/static/js/simplesearch.js
```diff
@@ -152,21 +159,32 @@ app.SimpleSearchController = function($scope, $compile, $attrs, apiUrl, gettextC
     minLength: 3
   });
 
+
```
asaunier: Useless extra blank line. Ditto at line 145

c2corg_ui/static/js/simplesearch.js
```diff
@@ -152,21 +159,32 @@ app.SimpleSearchController = function($scope, $compile, $attrs, apiUrl, gettextC
...
   /**
    * @type {string}
    * @private
    */
-  this.datasetLimit_ = $attrs['appDataset'];
+  this.datasetLimit_ = $attrs['dataset'];
+
+  // create only given datasets
+  for (var i = 0; i < this.datasetLimit_.length; i++) {
+    if (this.datasetLimit_[i] === 'u') {
```
asaunier: What about using a switch here?

c2corg_ui/static/partials/suggestionusers.html
```diff
@@ -0,0 +1,6 @@
+  <span class="suggestion-text">
```
asaunier: In a recent PR I have introducted a "_" in the "suggestion" partials names.
To be consistent this partial should be renamed to ``suggestion_users.html``.
See https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials

c2corg_ui/templates/base.html
```diff
@@ -50,7 +50,7 @@
     <button class="btn btn-default menu-open-close header"><span class="glyphicon glyphicon-align-justify"></span></button>
     <div class="page-header">
       <div class="quick-search">
-        <app-simple-search app-loading></app-simple-search>
+        <app-search app-loading app-search-map="" dataset="wro"></app-search>
```
asaunier: As explained previously, the directive is now "app-simple-search" and no longer "app-search".

less/simplesearch.less
```diff
@@ -25,14 +25,17 @@
...
-.view-details-associations, .route-associations, .associations, .associated-documents {
+.view-details-associations, .route-associations, .associations, .associated-documents, #participants-group {
   .tt-open {
     max-height: 400px;
     overflow-y: scroll;
     @media @phone {
       max-height: 300px;
     }
   }
+  app-search {
```
asaunier: app-simple-search

c2corg_ui/templates/base.html
asaunier: Where does this ``app-search-map`` attribute come from?

c2corg_ui/templates/outing/associated.html
```diff
@@ -3,7 +3,7 @@
   <article class="associated-documents" >
     <div class="list-item outing" ng-repeat="outing in document['associations']['recent_outings']['outings']">
       <%include file="../outing/card.html"/>
-      <app-delete-association parent-id="document.document_id" child-id="outing.document_id" ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+      <span class="glyphicon glyphicon-trash" ng-click="editCtrl.removeAssociation(outing.document_id, 'outings')" ng-if="userCtrl.auth.isAuthenticated()"></span>
```
asaunier: Does this mean the ``app-delete-association`` directive is removed?

c2corg_ui/static/js/documentediting.js
```diff
@@ -351,10 +351,16 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
     if (this.modelName_ === 'outing') {
       // adapt the Object for what's expected at the API side + format the outing
       this.formatOuting_(data);
+      var userIds = [this.auth_.userData.id];
```
asaunier: Some more non-generic vars in the generic document editing controller :(

c2corg_ui/static/js/documentediting.js
```diff
@@ -679,5 +683,46 @@ app.DocumentEditingController.prototype.setOrientation = function(orientation, d
...
-    app.DocumentEditingController);
+/**
+ * @param {Object} doc
+ * @export
+ */
+app.DocumentEditingController.prototype.pushToAssociations = function(doc) {
+  var routesLength =  this.scope_[this.modelName_]['associations']['routes'].length;
+  var associations = this.scope_[this.modelName_]['associations'];
+
+  if (doc['type'] === 'r') {
```
asaunier: Again...
For me we're not going to the right direction introducing document-type specific code in the document editing controller.

c2corg_ui/static/js/documentediting.js
```diff
@@ -679,5 +683,46 @@ app.DocumentEditingController.prototype.setOrientation = function(orientation, d
...
+  return;
+};
+
+
+/**
+ * @param {number} id
+ * @param {string} type
+ * @export
+ */
+app.DocumentEditingController.prototype.removeAssociation = function(id, type) {
```
asaunier: The user must confirm he really wants to remove the association before the request is actually performed.

ginold: done in #328

c2corg_ui/static/js/documentediting.js
ginold: this one precisely has to bee discussed with @tsauerwein if we want to change the way that API expects an outing. For now it has to be modified on the UI-side in order to make it acceptable by the API.

c2corg_ui/templates/base.html
ginold: no idea, I removed it and it still works.

c2corg_ui/static/js/documentediting.js
ginold: I think that after we release the next alpha or beta version, we should create a special PR to clean up this mess and create controllers/factories for document-specific controls.

c2corg_ui/templates/outing/associated.html
ginold: it is removed from this template, which is used only in editing forms.

asaunier: I mean: if we no longer need the ``app-delete-association`` directive and if it is no longer used anywhere else, should we completely remove the directive file. It is still available at https://github.com/c2corg/v6_ui/blob/editing-fixes/c2corg_ui/static/js/deleteassociationdirective.js

ginold: we still use it elsewhere.

c2corg_ui/static/js/documentediting.js
tsauerwein: With https://github.com/c2corg/v6_api/pull/293 being merged, outings can now be created like this:

https://github.com/c2corg/v6_api/blob/9485e78/c2corg_api/tests/views/test_outing.py#L344-L367

ginold: :trophy: !

ginold:
just updated this PR. You can have a look.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+  }
+
+  if (this.associationsIDs_[type].indexOf(id) > -1) {
+    return true;
+  } else {
+    return false;
+  }
+};
```
asaunier: One-liner:
```
return (this.associationsIDs_[type].indexOf(id) > -1);
```

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -74,7 +76,13 @@ app.module.directive('appViewDetails', app.viewDetailsDirective);
  * @export
  * @ngInject
  */
-app.ViewDetailsController = function($uibModal, $compile, $scope, appApi) {
+app.ViewDetailsController = function($uibModal, $compile, $scope, appApi, appDocument) {
```
asaunier: The new param should be listed as ``@param`` in the JSdoc as well.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,51 @@
...
+   * @type {Object}
+   * @export
+   */
+  this.associations = {};
+
+  /**
+   * @type {Object}
+   * @private
+   */
+  this.associationsIDs_ = {};
```
asaunier: Why capitalize the D?

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,51 @@
...
+goog.require('app');
+
+
+/**
+ * This services is used to acces the document properties
+ * within other controllers, for example to set the properties in one 
+ * controller and then read them in another.
+ * @constructor
+ */
+app.Document = function() {
```
asaunier: Is this service name explicit enough?
Will it be used for something else than checking associations?

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -3,6 +3,9 @@ goog.provide('app.viewDetailsDirective');
 
 goog.require('app');
 
+/** @suppress {extraRequire} */
```
asaunier: Why do you need this ``@suppress``?
Perhaps you don't need it if you fixed the ``@param`` issue below?

c2corg_ui/static/js/documentdetailsview.js
```diff
@@ -74,7 +76,13 @@ app.module.directive('appViewDetails', app.viewDetailsDirective);
...
  * @ngInject
  */
-app.ViewDetailsController = function($uibModal, $compile, $scope, appApi) {
+app.ViewDetailsController = function($uibModal, $compile, $scope, appApi, appDocument) {
+
+  /**
+   * @type {app.Document}
+   * @export
+   */
+  this.documentService = appDocument;
```
asaunier: Isn't it weird to make a service exported?

c2corg_ui/static/js/simplesearch.js
```diff
@@ -276,20 +290,28 @@ app.SimpleSearchController.prototype.createAndInitBloodhound_ = function(type) {
...
             doc.documentType = type;
-            return doc;
-          });
+
+            // don't show already associated docs in the results, but only in the app-add-association
+            // -> everything should be shown in the main simple-search.
+            if (!this.associationContext_ || (this.associationContext_ && !hasAssociation)) {
+              return doc;
+            } else { // typehead error if returns null or undefined
+              return 'nothing';
```
asaunier: Is it a legal return value?

c2corg_ui/templates/outing/view.html
```diff
@@ -163,7 +163,8 @@ <h3 class="heading"><span translate>personal comments</span></h3>
   <button class="btn btn-default scroll-top-btn" onclick="window.scrollTo(0, 0);">&nbsp;<span class="glyphicon glyphicon-menu-up"></span></button>
 </div>
 
-<div class="ng-hide">
+<div class="ng-hide" ng-init="associations = ${outing.get('associations')}">
+  {{detailsCtrl.documentService.associations= associations}}
```
asaunier: Is it the good way to communicate with a service?

asaunier:
Travis fails because there are some trailing blank spaces at the end of some line in ``documentservice.js``.

It also complains about many ``console.log``. Do we need that many?

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -260,6 +260,19 @@ app.Authentication.prototype.addAuthorizationToHeaders = function(url,
...
 
 /**
+ * @export
+ * @return {string | boolean} token
+ */
+app.Authentication.prototype.getUserToken = function() {
+  if (this.userData) {
+    return this.userData.token;
+  } else {
+    return false;
```
asaunier: Returning a boolean does not really make sense here. You should rather return a string (empty string?) or else you'll get a JS error when trying to concatenate the result of the function in the ``Authorization`` header, lower in this PR.

c2corg_ui/static/js/documentservice.js
ginold: yep, it could be used for everything. You spoke about our controllers being too document-specific -> now we have a service for that. At least a service which gathers functions that all or most documents have, but then we could also create an outingservice, routeservice etc if needed.

c2corg_ui/static/js/documentservice.js
ginold: because Id is super ugly!

c2corg_ui/static/js/documentdetailsview.js
ginold: :+1:

c2corg_ui/static/js/documentdetailsview.js
ginold: yep it solved!

c2corg_ui/templates/outing/view.html
ginold: how do you want to do it better? it this case, we have to go and pass it through the template because we only receive data with mako. I could write something like `detailsCtrl.setDocumentInDocumentService(document)` but it will still do the same :P

c2corg_ui/static/js/documentdetailsview.js
ginold: It is exported only in the details-view-controller so I can access it through the template, where I assign the document values to the document service.

c2corg_ui/templates/outing/view.html
asaunier: I don't know. But using an hidden expression to set a variable in a service looks a bit like a hack to me.

ginold:
> It also complains about many console.log. Do we need that many?

the most of them are in @gberaudo 's authservice and apiservice, if you are ok with that I can delete them.

c2corg_ui/static/js/auth/authservice.js
gberaudo: Is this function needed?

c2corg_ui/static/js/documentdetailsview.js
gberaudo: It is OK if it is accessed from a template.

c2corg_ui/static/js/documentediting.js
```diff
@@ -351,10 +349,16 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
...
+
+      for (var i = 0; i < data['associations']['users'].length; i++) {
+        userIds.push(data['associations']['users'][i]['document_id']);
+      }
+
       data = {
         'outing': data,
-        'route_id': data['associations']['routes'][0]['document_id'],
-        'user_ids': [this.auth_.userData.id]
+        'route_id': data['associations']['routes'][0]['document_id'] || null,
```
gberaudo: Would be great to have an extern for the `associations` hierarchy.

c2corg_ui/static/js/documentediting.js
```diff
@@ -679,5 +681,50 @@ app.DocumentEditingController.prototype.setOrientation = function(orientation, d
 }
 
 
-app.module.controller('appDocumentEditingController',
-    app.DocumentEditingController);
+/**
+ * @param {Object} doc
```
gberaudo: appx.SimpleSearchDocument?

c2corg_ui/static/js/documentservice.js
gberaudo: Missing `@struct`.

c2corg_ui/static/js/documentservice.js
gberaudo: +1 for small `d`.

c2corg_ui/static/js/documentediting.js
asaunier: What's the connection between the editing and the simple search?
I don't think simple search results are the same than regular "get-document" responses.

c2corg_ui/static/js/simplesearch.js
```diff
@@ -230,16 +269,16 @@ app.SimpleSearchController.prototype.createAndInitBloodhound_ = function(type) {
...
         if (this.datasetLimit_) {
-          if (this.datasetLimit_.length === 1) {
-            url = url + '&t=' + this.datasetLimit_;
-          } else {
-            // you receive for ex 'wru' (waypoints, routes, users) -> split into w,r,u
-            this.datasetLimit_ = this.datasetLimit_.split('').join(',');
+          // add the Auth header if searching for users
+          if (this.datasetLimit_.indexOf('u') > -1) {
+            settings['headers'] = {
+              'Authorization': 'JWT token="' + this.auth_.getUserToken() + '"'
```
gberaudo: No need for a function, the property can be accessed directly.

c2corg_ui/static/js/simplesearch.js
```diff
@@ -251,20 +290,28 @@ app.SimpleSearchController.prototype.createAndInitBloodhound_ = function(type) {
         if (documentResponse) {
           var documents = documentResponse.documents;
           var currentLang = this.gettextCatalog_.currentLanguage;
+          var hasAssociation;
 
           return documents.map(function(/** appx.SimpleSearchDocument */ doc) {
+            hasAssociation = this.documentService_.hasAssociation(type,  doc['document_id']);
```
gberaudo: No need for brackets.

c2corg_ui/static/js/simplesearch.js
```diff
@@ -251,20 +290,28 @@ app.SimpleSearchController.prototype.createAndInitBloodhound_ = function(type) {
...
+
             var locale = doc.locales[0];
-            doc.label = type === 'routes' && locale.title_prefix ?
-                    locale.title_prefix + ' : ' : '';
+            doc.label = type === 'routes' && locale.title_prefix ? locale.title_prefix + ' : ' : '';
```
gberaudo: Would be more predictable by using parenthesis.

c2corg_ui/static/js/utils.js
```diff
@@ -129,6 +129,9 @@ app.utils.formatDate = function(date) {
  */
 app.utils.areDifferentDates = function(date1, date2) {
   if (date1 !== null && date2 !== null) {
+    if (typeof date1 === 'string') {
```
gberaudo: The function declares taking two `Date` objects. How can `date` now be a `string`?
And what about `date2`?

c2corg_ui/templates/outing/view.html
gberaudo: I agree this looks strange. @ginold, have you tried moving the expression to the `ng-init` just above?

c2corg_ui/templates/route/view.html
```diff
@@ -193,7 +193,8 @@ <h3 class="heading show-phone"><span translate>description</span></h3>
   <button class="btn btn-default scroll-top-btn" onclick="window.scrollTo(0, 0);">&nbsp;<span class="glyphicon glyphicon-menu-up"></span></button>
 </div>
 
-<div class="ng-hide">
+<div class="ng-hide" ng-init="associations = ${route.get('associations')}">
+  {{detailsCtrl.documentService.associations= associations}}
```
gberaudo: Idem https://github.com/c2corg/v6_ui/pull/372/files#r65093095
Do we have other cases like these?

c2corg_ui/templates/waypoint/view.html
```diff
@@ -198,7 +198,8 @@ <h3 class="text-center" translate>No description available</h3>
   <button class="btn btn-default scroll-top-btn" onclick="window.scrollTo(0, 0);">&nbsp;<span class="glyphicon glyphicon-menu-up"></span></button>
 </div>
 
-<div class="ng-hide">
+<div class="ng-hide" ng-init="associations = ${waypoint.get('associations')}">
+  {{detailsCtrl.documentService.associations= associations}}
```
gberaudo: https://github.com/c2corg/v6_ui/pull/372/files#r65093095

c2corg_ui/static/js/documentediting.js
gberaudo: Right @asaunier, so there should be a dedicated extern.

c2corg_ui/templates/outing/view.html
ginold: like so ? `ng-init="detailsCtrl.documentService.associations = ${outing.get('associations')}"`

c2corg_ui/templates/route/view.html
ginold: only in view.html

c2corg_ui/static/js/utils.js
ginold: It is a string when it first comes from the API, then I convert it to a date.

c2corg_ui/templates/outing/view.html
gberaudo: Yes, or with a `;`.

c2corg_ui/static/js/utils.js
gberaudo: In this case you should use `Date|string`.
And `date2`?

c2corg_ui/externs/appx.js
```diff
@@ -84,3 +84,186 @@ appx.BootstrapSliderValues;
...
+/**
+ * @typedef {{
+ *   routes : Array<Object>,
+ *   waypoints : Array<Object>,
+ *   images : Array<Object>,
+ *   users : Array<Object>,
+ *   recent_outings : Array<Object>,
+ * }}
+ */
+appx.DocumentAssociations;
```
asaunier: ``all_routes``, ``waypoint_children`` and ``waypoint_parents`` are missing.
Since we do not get all those association types for all documents, ``?`` tags are probably necessary for some of them?

c2corg_ui/externs/appx.js
```diff
@@ -84,3 +84,186 @@ appx.BootstrapSliderValues;
...
+ *   capacity_staffed: number,
+ *   phone_custodian: (number | string),
+ *   custodianship: string,
+ *   matress_unstaffed: (string | boolean),
+ *   blanket_unstaffed: (string | boolean),
+ *   gas_unstaffed: (string | boolean),
+ *   heating_unstaffed: (string | boolean)
+ * }}
+ */
+appx.Waypoint;
```
asaunier: Nice job, @gberaudo will appreciate I think.
I just wonder if we really needed to be that precise? If some attribute is added, removed, renamed that will be one more place where to propagate the change. Do we really want this? @gberaudo what do you think?

c2corg_ui/static/js/addassociationdirective.js
```diff
@@ -55,7 +55,13 @@ app.AddAssociationController = function(appApi, $attrs) {
   this.api_ = appApi;
 
 
-  this.dataset = $attrs['dataset'];
+  /**
+   * @private
+   * @type {string}
+   */
+  this.dataset_ = $attrs['dataset'];
```
asaunier: Isn't it a job for ``@struct`` and simple "dot" notation? As you did for ``parentId`` and ``addedDocuments``.

I think you should simply write
```
/**
 * @type {string}
 * @export
 */
this.dataset;
```

asaunier: @gberaudo For ``this.addedDocuments = [];`` a few line below, do we really need to specify an initial value (``[]``)?

For ``this.parentId;`` I think the ``// bound from directive`` comment is not needed (or we could add it at many other places...).

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,128 @@
...
+ * @constructor
+ * @struct
+ */
+app.Document = function() {
+
+  /**
+   * @type {appx.Document}
+   * @export
+   */
+  this.document = {
```
asaunier: See my comment in the externs file about the missing association types.

c2corg_ui/externs/appx.js
ginold: those properties you listed do not replace any of the existing properties?
`?` means Something or Null ?

c2corg_ui/externs/appx.js
ginold: this might be useful when we decide to create document-specific controllers or services.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * - waypoints, routes and books are not protected => collaborative
+ * - outings and books are personal and we have to check if the current user is the owner
+ * - images can be both and we have to check the image_type property
+ * @param {string} type
+ * @export
+ * @return boolean
+ */
+app.Document.prototype.isProtectedDocument = function(type) {
```
asaunier: As explained here https://github.com/c2corg/v6_ui/issues/383#issuecomment-222940822
``protected`` has no connection with collaborative or personal documents.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,128 @@
...
+  } else if (type === 'outings' || type === 'books') {
+    return !app.UserController.prototype.hasEditRights(this.document.associations['users']);
+
+  } else if (type === 'images') {
+    if (this.document['image_type'] === 'personal') {
+      // TODO: check if user is the owner of the image
+      return true;
+    } else if (this.document['image_type'] === 'collaborative') {
+      return false;
+    }
```
asaunier: A return value is missing when the image_type is neither personal or collaborative.

c2corg_ui/static/js/simplesearch.js
```diff
@@ -89,12 +93,19 @@ app.module.directive('appSimpleSearch', app.simpleSearchDirective);
...
  * @param {angular.$compile} $compile Angular compile service.
  * @param {angular.$templateCache} $templateCache service
+ * @param {app.Document} appDocument service
  * @param {angular.Attributes} $attrs Angular attributes.
  * @param {string} apiUrl Base URL of the API.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @ngInject
  */
-app.SimpleSearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog, $templateCache) {
+app.SimpleSearchController = function(appDocument, $scope, $compile, $attrs, apiUrl, gettextCatalog, $templateCache, appAuthentication) {
```
asaunier: Would be cool to avoid too long lines.

c2corg_ui/externs/appx.js
asaunier: Nope.

For instance for a waypoint document, you have ``all_routes`` but not ``routes``. For an outing document you have ``routes`` but not ``all_routes``.
For a waypoint document you have ``waypoint_parents`` and ``waypoint_children`` but not ``waypoints``. The contrary for a route document.

c2corg_ui/static/js/documentservice.js
ginold: so maybe Ill rename to 'isCollaborativeDocument' :P

c2corg_ui/static/js/documentservice.js
ginold: the last type is 'copyright'. So is it personal or collaborative?

c2corg_ui/externs/appx.js
asaunier: > ? means Something or Null ?

I am not sure but according to the other externs, it seems that an optional property would we described like this:
```
recent_outings: ?Array<Object>
```

Oh by the way: ``recent_outings`` and ``all_routes`` are not arrays! They look like:
```
recent_outings: {
  total: {number}
  routes: Array<Object>
}
```

Actually I was doing the same kind of externs in the ``association-tool-fix-2`` (not commited  yet, I'll probably have to use yours instead). It looks like
```
/**
 * @typedef {{
 *   associations: appx.Associations,
 *   orientations: ?Array.<string>
 * }}
 */
appx.DocumentData;


/**
 * @typedef {{
 *   images: Array.<Object>,
 *   users: ?Array.<Object>,
 *   routes: ?Array.<Object>,
 *   waypoints: ?Array.<Object>,
 *   waypoint_children: ?Array.<Object>,
 *   waypoint_parents: ?Array.<Object>,
 *   recent_outings: ?appx.RecentOutings,
 *   all_routes: ?appx.AllRoutes
 * }}
 */
appx.Associations;

/**
 * @typedef {{
 *   total: number,
 *   outings: Array.<Object>
 * }}
 */
appx.RecentOutings;

/**
 * @typedef {{
 *   total: number,
 *   routes: Array.<Object>
 * }}
 */
appx.AllRoutes;

```

c2corg_ui/static/js/documentservice.js
asaunier: None :) It's "copyright". https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L585-L589
It's a very special case for book covers. You may assimilate it to "personal".

c2corg_ui/static/js/documentservice.js
asaunier: OK

asaunier: Or simply "isCollaborative" since it's already in the document service.

c2corg_ui/static/js/simplesearch.js
asaunier: What about using ``this.userData.token`` instead as suggested by @gberaudo ?
What happens if the user is not authenticated? In that case it is not relevant to add an ``Authorization`` header, is it?

c2corg_ui/externs/appx.js
ginold: I will copy-paste yours and adapt them.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,128 @@
...
+ * @param {string} type
+ * @export
+ * @return boolean
+ */
+app.Document.prototype.isProtectedDocument = function(type) {
+  if (type === 'waypoints' || type === 'routes' || type === 'books') {
+    return false;
+
+  } else if (type === 'outings' || type === 'books') {
+    return !app.UserController.prototype.hasEditRights(this.document.associations['users']);
```
asaunier: Is it really possible to statically call this non static function??
See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/user.js#L114

asaunier:
This PR will probably impact a lot the changes I'm doing in https://github.com/c2corg/v6_ui/pull/328
That would be cool to fix and merge it as soon as possible.

ginold:
Indeed. I will correct following your suggestions and let's merge this superb PR!

less/simplesearch.less
asaunier: Don't you need to make this change since the element is now ``<app-simple-search>`` instead of ``<app-search>``?

c2corg_ui/static/js/addassociationdirective.js
```diff
@@ -55,7 +56,6 @@ app.AddAssociationController = function(appApi, $attrs) {
   this.api_ = appApi;
 
 
-  this.dataset = $attrs['dataset'];
```
asaunier: Don't you need to declare ``this.dataset``, as you did for ``parentId`` and ``addedDocuments``?

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -260,6 +260,19 @@ app.Authentication.prototype.addAuthorizationToHeaders = function(url,
 
 
 /**
+ * @export
+ * @return {string} token
+ */
+app.Authentication.prototype.getUserToken = function() {
```
asaunier: It seems this function is no longer used. As a result you can remote it.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,135 @@
...
+ * @param {string} type
+ * @export
+ * @return boolean
+ */
+app.Document.prototype.isCollaborative = function(type) {
+  if (type === 'waypoints' || type === 'routes' || type === 'books') {
+    return true;
+
+  } else if (type === 'outings' || type === 'books') {
+    return this.auth_.hasEditRights(this.document.associations['users']);
```
asaunier: Books are always collaborative.
Perhaps you mean "article"?

c2corg_ui/static/js/addassociationdirective.js
ginold: I noticed that I don't have to - the only place I use it is in the directive itself, where I can get it directly from attrs.dataset, so finally there is no use for this.dataset in the controller.

c2corg_ui/static/js/documentservice.js
```diff
@@ -0,0 +1,135 @@
...
+
+  } else if (type === 'images') {
+    return this.document['image_type'] === 'collaborative';
+  }
```
asaunier: I think we need a default return value. I would say ``true`` as default.

Please note there are other document types such as articles or incident reports:
* books: collaborative
* incident reports (don't know the document type yet. In v5 it's "xreports"): personal
* articles: personal|collaborative

c2corg_ui/static/js/simplesearch.js
```diff
@@ -89,12 +93,21 @@ app.module.directive('appSimpleSearch', app.simpleSearchDirective);
...
+ * @param {app.Document} appDocument service
+ * @param {app.Authentication} appAuthentication service
  * @param {angular.Attributes} $attrs Angular attributes.
  * @param {string} apiUrl Base URL of the API.
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @ngInject
  */
-app.SimpleSearchController = function($scope, $compile, $attrs, apiUrl, gettextCatalog, $templateCache) {
+app.SimpleSearchController = function(appDocument, $scope, $compile, $attrs, apiUrl,
+                                                                            gettextCatalog, $templateCache, appAuthentication) {
```
asaunier: Come on! :P 
You don't need that much indent here. I would say that with a 4-space indent, the linter is happy.

c2corg_ui/static/js/addassociationdirective.js
asaunier: OK

c2corg_ui/externs/appx.js
gberaudo: @asaunier, if we are sure we will never have a use of some properties it is possible to omit them.
However if we set a non declared property there will be no warning and it will be minified.
Don't know what is worst.

c2corg_ui/static/js/simplesearch.js
ginold: whoa, it looks strange on github, on my computer it's like 
screenshot from 2016-06-02 09 07 09

asaunier: OK, but anyway we don't need that much indent. I think that for all other line breaks in arguments lists we used a 4-space indent like here: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/user.js#L39-L40

c2corg_ui/externs/appx.js
```diff
@@ -84,3 +84,207 @@ appx.BootstrapSliderValues;
...
+appx.Route;
+
+/**
+ * @typedef {{
+ *   associations: appx.DocumentAssociations,
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   quality: string,
+ *   document_id: number,
+ *
```
asaunier: Useless blank line

ginold: I have put this blank line in every document to show that those properties above are the same as found in appx.document.
At first I wanted to create an appx.document and appx.waypoint, so that waypoint inherits from document, but after discussing with @gberaudo it doesn't seem possible, so he suggested that I copy those lines from appx.document and separate them with a blank line.

asaunier: OK then

asaunier:
Looks good to me!
Thanks for the nice job :)

Issue #369, opened 2016-05-26
Blank cards in debug mode


asaunier:
When debug mode is on (``?debug``), the advanced search cards + the simple search results cards ("suggestion") remain blank.

A similar code is used in both cases and the issue (and solution!) is most likely the same. Perhaps because the partials are not included as expect when using
```
  var template = function(doctype) {
    return $templateCache.get('/static/partials/suggestion_' + doctype + '.html');
  };
```
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/suggestiondirective.js#L11-L13
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js#L17-L19

fjacon:
can this be closed ?

asaunier:
No, the problem is still valid => not possible to work in debug mode in the advanced search pages for instance, which is not very handy :(

ginold:
In the other project on EPFL, in prod mode it works with `angular.templateCache` as well, but when in dev mode, the actual HTML templates are loaded. We should have the same behavior.

asaunier:
Could you point where this is done in EPFL code?

asaunier:
> In the other project on EPFL, in prod mode it works with angular.templateCache as well, but when in dev mode, the actual HTML templates are loaded. We should have the same behavior.

I am not sure EPFL use the templateCache the same way we do.

I mean, the partials we pass as ``templateUrl`` in our directives actually work fine in debug mode. But not the dynamic template paths, like https://github.com/c2corg/v6_ui/blob/e1724c8c11fa5483bd78954a9c042f54a1132bc1/c2corg_ui/static/js/card.js#L18 and I don't see such system in https://github.com/camptocamp/epfl_authgeoportail/search?utf8=%E2%9C%93&q=templateCache&type=Code

ginold:
@sbrunner should know!

asaunier:
Actually it is no surprise that code like https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js#L17-L19
```
return $templateCache.get('/static/partials/cards/' + doctype + '.html');
```
does not work in debug mode. Because it tries to retrieve some template cache that has never been create (with ``$templateCache.put()``), except in the "prod" mode that explicitely creates those template cache https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/templatecache.js#L36

[...]

Issue #368, opened 2016-05-26
Advanced search UI (step 2)


asaunier:
TODO see https://github.com/c2corg/v6_ui/pull/373

asaunier:
The work on this issue is in progress. I have several questions though:

To avoid changing the way the advanced search works with level1-filters (a new search is triggered whenever a filter is changed), I wonder if could not simply do the same for the filters shown when clicking on the "more filters" button? The list of result cards is no (fully?) visible when having "More filters" but the user can still see the items changing on the map.
Then the "cancel and close" or "apply filters" buttons would no longer be useful and could be replaced by a simple "Close" or "See the results" button.

The sliders are currently (easily) used to filter a given numeric attribute to some range (eg. elevation).
That would be cool to be able to use the sliders for non numeric attributes, for instance the ratings (users would for instance use the sliders to filter on routes rated from 6a to 6c or PD+ to AD+). But I don't really know if this is possible because:
* the sliders expect numeric values, eg. 1, 2, 3, 10, 100, not strings such as 6a, 6c, PD+ or AD+
* can the API really compare ratings and return routes whose ratings would be in a given range of ratings?

I was also wondering if we could use the sliders approach to avoid multiplicating the min/median/max attributes. For example with climbing_outdoor_site waypoints, we have height_min, heigh_max, height_median
https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/waypoint_mapping.py#L37-L42
I don't know how we could do that, but as a user I would expect to use the sliders to provide the min/max height and that the API gives me back the list of WP whose height_min/height_max (height_median?) fit the range. That's a bit tricky because the filtering would be done on several attributes...

tsauerwein:
> can the API really compare ratings and return routes whose ratings would be in a given range of ratings?

The API doesn't know that for example PD+ < AD+. If you want routes with a rating between PD and ED, you would have to set a filter with `PD-,PD,PD+,...,ED-,ED,ED+`. Maybe not very practical.

> I was also wondering if we could use the sliders approach to avoid multiplicating the min/median/max attributes.

Wouldn't filtering on `heigh_max` be enough?

asaunier:
> The API doesn't know that for example PD+ < AD+. If you want routes with a rating between PD and ED, you would have to set a filter with PD-,PD,PD+,...,ED-,ED,ED+. Maybe not very practical.

That's not really handy, indeed.

And note very logical.
[...]

PR #373, opened 2016-05-26
Advanced search step2

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/368

TODO

* [x] take "more filters" filters into account for waypoints. As for now only the first-level criterias are taken into account (title, waypoint_types, elevation).
* [ ] filters for routes
* [ ] filters for outings
* ~~cards are not rendered correctly in debug mode (same issue than for the simple search)~~ => see https://github.com/c2corg/v6_ui/issues/369
* [x] outing cards layout need some CSS fixing
* [x] feature icon is now clickable in the detail view page map: either disable this behaviour or at least make sure the URL is OK
* [x] add an extern entry for the ``listDocuments`` function in ``apiService``.
* [x] make sure the features on the map use the most appropriate locales.
* [ ] Skip the map extent when filtering on a WP or a route ID? (with permalink like ``outings?r=123456``)
* [x] Handle orientations
* [x] Use sliders for ratings (ie. non-numeric values) as well?
* [ ] Handle date filters

TOFIX
* [ ] slider handles are not fully reset when clearing filters
* [x] loading spinner no longer shows up when changing the filters!?

EXTRA

* [ ] Hilight the feature on the map when hovering a card (and vice-versa?)
* [ ] allow to bypass the map extent filtering as in Airbnb?

Some changes in the API are needed: https://github.com/c2corg/v6_api/issues/311


c2corg_ui/static/js/card.js
```diff
@@ -61,17 +60,23 @@ app.CardController = function(gettextCatalog) {
...
+  this.type = app.utils.getDoctype(this.doc['type']);
 
   /**
    * @type {Object}
    * @export
    */
-  this.locale = this['doc']['locales'][0];
-  for (var i = 0, n = this['doc']['locales'].length; i < n; i++) {
-    var l = this['doc']['locales'][i];
+  this.locale = this.doc['locales'][0];
```
gberaudo: `locales` is a property of the document which is defined in the extern so you can also remove these brackets.

c2corg_ui/static/js/card.js
```diff
@@ -95,9 +100,7 @@ app.CardController.prototype.translate = function(str) {
  */
 app.CardController.prototype.openCardPage = function() {
   window.location.href = app.utils.buildDocumentUrl(
-    app.utils.getDoctype(this.doc_['type']),
-    this.doc_['document_id'],
-    this.doc_['locales'][0]['lang']);
+    this.type, this.doc['document_id'], this.doc['locales'][0]['lang']);
```
gberaudo: Idem, the `document_id` brackets can be removed.

c2corg_ui/static/js/card.js
ginold: so I  too was doing it wrong my whole life ...

asaunier: It seems to work indeed. But I am not sure to understand why because I don't really see the typedef for documents in the externs https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js

gberaudo: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js#L38-L47

asaunier:
This PR is growing really big again :P 
Actually it contains most of the change for the advanced search pages.

The outings and routes pages are not ready yet (not the right filters) but at least they work as in alpha1. I could add the filters in some other PR. There are also a few other remaining points in the TODO/TOFIX list but not critical.

Feel free to review, thanks :)

c2corg_ui/static/js/card.js
```diff
@@ -67,12 +65,18 @@ app.CardController = function(gettextCatalog) {
...
+  /**
    * @type {Object}
    * @export
    */
-  this.locale = this.doc['locales'][0];
-  for (var i = 0, n = this.doc['locales'].length; i < n; i++) {
-    var l = this.doc['locales'][i];
+  this.locale = this.doc.locales[0];
+  for (var i = 0, n = this.doc.locales.length; i < n; i++) {
+    var l = this.doc.locales[i];
```
tsauerwein: When using `.locales`, you should type `this.doc`.

c2corg_ui/static/js/search/searchfilters.js
```diff
@@ -189,4 +223,42 @@ app.SearchFiltersController.prototype.selectOption = function(prop, val, event)
...
+ * @export
+ */
+app.SearchFiltersController.prototype.setOrientation = function(orientation,
+    ctrl, e, filterName) {
+  if (this.orientations.indexOf(orientation) === -1) {
+    this.orientations.push(orientation);
+  } else {
+    this.orientations = this.orientations.filter(function(val) {
+      return val !== orientation;
+    });
```
tsauerwein: I don't understand what you are doing with the filter. You are creating a new array that contains all orientations in `this.orientations` that are not `orientation`. But in the `if` above you have already checked that `orientation` is not in `this.orientations`?

c2corg_ui/static/js/utils.js
```diff
@@ -26,14 +26,17 @@ app.utils.getDoctype = function(type) {
...
  * @param {string|number} documentId The document id.
  * @param {string} lang Lang.
  * @return {string} Url.
  */
-app.utils.buildDocumentUrl = function(document_type, documentId, lang) {
-  return '/{document_type}/{id}/{lang}'
-    .replace('{document_type}', document_type)
+app.utils.buildDocumentUrl = function(documentType, documentId, lang) {
+  if (!documentType || !documentId || !lang) {
+    return '';
```
tsauerwein: I would rather return `null` in this case.

tsauerwein:
I left some minor comments, feel free to merge if you want to do the rest in follow-up PRs.

c2corg_ui/static/js/card.js
asaunier: You're definitely right. I have just made a change with 55d2eb4

It seemed to work anyway since I previously had no warning when compiling and the cards locales were correctly fed.

c2corg_ui/static/js/search/searchfilters.js
asaunier: When using this ``setOrientation`` method (triggered when clicking on one of the directions of the SVG compass, eg. at https://github.com/c2corg/v6_ui/blob/advanced-search-step2/c2corg_ui/templates/helpers/orientations.svg?short_path=31b8504#L62):
* either the orientation is not already in the orientations list and we push it into the list
* or it is already in the list and we remove it using this "filter" call that make sure that all instances (in case several were - by mistake? - available) of this orientation are removed (we keep only the values that are not equal to the orientation we want to remoev).

tsauerwein: So `setOrientation` removes or adds an orientation? The name might be confusing. What about `toggleOrientation`?

c2corg_ui/static/js/utils.js
asaunier: Doing so implies changing the "return" type of the function + the type of all vars that are set with this function. That probably makes sense but this PR already does that many changes, I would rather avoid adding one more. To do in another PR?

tsauerwein: > To do in another PR?

Sure. Returning `null` also has the advantage that you are correctly handling the case that no url is returned wherever this function is called.

c2corg_ui/static/js/search/searchfilters.js
asaunier: Right! Done at dfc2490

c2corg_ui/static/js/utils.js
asaunier: I agree, but I would prefer not having to also do that kind of changes in this PR.

asaunier:
> I left some minor comments, feel free to merge if you want to do the rest in follow-up PRs.

I have addressed most of your suggestions. I have also added most of the filters for routes. I think I will merge when Travis passes and continue with another PR for the rest of the route filters + the outing filters.

PR #379, opened 2016-05-30
Fix associated docs blocks

### LINKED ISSUE #367: Add a "show more" button for associated routes blocks if len(routes)>30
See https://github.com/c2corg/v6_api/pull/294/files/e8db0a635a88691e13ea6b8c89e3681f319dbaed..4f0413c8db7f06076a812b090ab67d5021965b30#r64564087

### LINKED ISSUE #376: Associated routes are no longer displayed in the WP detail view pages
Only the "recent outings" are.

asaunier:
Fixes #376 
Fixes #367 

Done:
* take into account the new ``associations`` format of the API
* split the ``get_associated_documents`` mako helper into 3 helpers since it was getting a bit too big.
* some indent and too-long-lines fixing
* add a "show more" button for the list of routes cards as well (as for outings). The generated URL is currently not supported by the API, see https://github.com/c2corg/v6_api/issues/297

Questions:
* the margin below the association tool is incorrect (same with the master branch). @ginold do you know how we could fix this? I think it's ok in some other branches.
sans titre
* not sure the ``documentType`` filtering is OK for displaying cards of the newly associated documents (but that could be done in another PR since we should work on the cards directive/partials for the association tools).
* I have noticed that the outings detail view calls the associated-documents helper for associated WP as well but it seems the API always returns an empty ``waypoints`` list in ``associations``. Either there is something missing in the API or we don't need this helper call here?

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -125,6 +125,15 @@
               ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
           </div>
         % endfor
+        % if type == 'all_routes' and associations['all_routes']['total'] > len(routes):
+          <%
+              _query = {}
+              _query[document['type']] = document['document_id']
+          %>
+          <div class="list-item vertical-align" onclick="window.location='${request.route_url('routes_index', _query=_query)}';">
+            <button class="gray-btn btn show-more" translate>show more</button>
+          </div>
+        % endif
         <article class="list-item" ng-repeat="doc in added | filter:{'documentType': 'routes'}">
           <app-association-card parent-id="${document['document_id']}" doc="doc" added-documents="added"></app-association-card>
         </article>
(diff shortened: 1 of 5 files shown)


c2corg_ui/templates/helpers/view.html
```diff
@@ -125,6 +125,15 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
               ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
           </div>
         % endfor
+        % if type == 'all_routes' and associations['all_routes']['total'] > len(routes):
+          <%
+              _query = {}
+              _query[document['type']] = document['document_id']
+          %>
+          <div class="list-item vertical-align" onclick="window.location='${request.route_url('routes_index', _query=_query)}';">
+            <button class="gray-btn btn show-more">show more</button>
```
gberaudo: Missing translation?

ginold:
@asaunier the margin is ok on #363 for example, so maybe don't fix it and it will be ok when we merge the two branches.

asaunier:
Thanks @gberaudo @ginold for the comments, I have added the missing ``translate``. Concerning the margin issue, since it is fixed in another PR, we can ignore it here.

OK for merging?

gberaudo:
+1

asaunier:
@ginold OK for you?

ginold:
+1

Issue #377, opened 2016-05-30
Link of document badge


asaunier:
In the document detail view pages, a badge with the document type is shown near the document type.
As for now it redirects to the advanced search page with the default filtering.

It would be interesting that a few filtering parameters are added to the URL. For instance:
- [x] bbox around the current document
- [x] the waypoint type for WP?

asaunier:
Waypoint type has been added by #397 but not bbox I think.

asaunier:
> It would be interesting that a few filtering parameters are added to the URL. For instance:
* bbox around the current document
* the waypoint type for WP?

Done in some recent PR. Closing the issue.

PR #397, opened 2016-06-08
Card fix

### LINKED ISSUE #393: Enable opening cards in a new tab
As for now the cards use an ``ng-click`` to open the bound document:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/card_waypoints.html#L1

A drawback of this method is that it is not possible to open the bound document in a new tab.
Using an ``ng-href`` should make it possible. Please note that in editing mode, the card should not open the bound document (the link should be disabled).

ginold:
- added href to cards to let the user open them using 'open in new tab' (not available on editing pages)
- added animations for adding and removing association-cards
- added waypoint type to badge (waypoint view, #377)

fixes #393

### DIFF
--- a/c2corg_ui/static/js/card.js
+++ b/c2corg_ui/static/js/card.js
@@ -24,8 +24,7 @@ app.cardDirective = function($compile, $templateCache) {
     controllerAs: 'cardCtrl',
     bindToController: true,
     scope: {
-      'doc': '=appCardDoc',
-      'disableClick': '=appCardDisableClick'
+      'doc': '=appCardDoc'
     },
     link: function(scope, element, attrs) {
       var doc = scope['cardCtrl']['doc'];
@@ -68,12 +67,6 @@ app.CardController = function(gettextCatalog) {
   this.doc;
 
   /**
-   * @type {boolean}
-   * @export
-   */
-  this.disableClick;
-
-  /**
    * @type {Object}
    * @export
    */
@@ -100,13 +93,14 @@ app.CardController.prototype.translate = function(str) {
 
 /**
  * @export
+ * @return {string | undefined}
  */
-app.CardController.prototype.openCardPage = function() {
-  if (!this.disableClick) {
-    window.location.href = app.utils.buildDocumentUrl(
-      app.utils.getDoctype(this.doc['type']),
-      this.doc['document_id'],
-      this.doc['locales'][0]['lang']);
+app.CardController.prototype.createURL = function() {
+  if (window.location.pathname.indexOf('edit') === -1) {
+    return app.utils.buildDocumentUrl(
+            app.utils.getDoctype(this.doc['type']),
+            this.doc['document_id'],
+            this.doc['locales'][0]['lang']);
   }
 };
 
--- a/c2corg_ui/static/js/deleteassociation.js
+++ b/c2corg_ui/static/js/deleteassociation.js
@@ -20,11 +20,11 @@ app.deleteAssociationDirective = function() {
       'childDoctype': '@'
     },
     link: function(scope, element, attrs, controller) {
-      $(element).on('click', function() {
+      $(element).on('click', function(e) {
         var modal = controller.openModal_();
         modal.result.then(function(res) {
           if (res) {
-            controller.unassociateDocument_();
+            controller.unassociateDocument_(e);
           }
         });
       });
@@ -122,14 +122,12 @@ app.DeleteAssociationController.prototype.openModal_ = function() {
 
 /**
  * Unassociate the document and remove the card.
+ * @param {goog.events.Event | jQuery.Event} [event]
  * @private
  */
-app.DeleteAssociationController.prototype.unassociateDocument_ = function() {
+app.DeleteAssociationController.prototype.unassociateDocument_ = function(event) {
   this.api_.unassociateDocument(this.parentId, this.childId).then(function() {
-    this.documentService_.removeAssociation(this.childId, this.childDoctype);
-    this.rootscope_.$broadcast('unassociateDoc', {
-      'id': this.childId, 'type': this.childDoctype
-    });
+    this.documentService_.removeAssociation(this.childId, this.childDoctype, event);
   }.bind(this));
 };
 
--- a/c2corg_ui/static/js/documentservice.js
+++ b/c2corg_ui/static/js/documentservice.js
@@ -9,11 +9,18 @@ goog.require('app');
  * controller and then read them in another.
  * Also to do all functions common to all documents.
  * @param {app.Authentication} appAuthentication
+ * @param {angular.Scope} $rootScope Scope.
  * @constructor
  * @ngInject
  * @struct
  */
-app.Document = function(appAuthentication) {
+app.Document = function(appAuthentication, $rootScope) {
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.rootScope_ = $rootScope;
 
   /**
    * @type {app.Authentication}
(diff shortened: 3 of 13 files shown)


c2corg_ui/static/js/card.js
```diff
@@ -31,6 +31,11 @@ app.cardDirective = function($compile, $templateCache) {
       var doc = scope['cardCtrl']['doc'];
       var type = app.utils.getDoctype(doc['type']);
       element.html(template(type));
+      
+      if (type === 'outings') {
```
asaunier: I would prefer we avoid adding doc-specific code.
I have a similar problem with the advanced search => see the ``ng-class`` at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/advancedsearch.html#L2
By the way I had renamed the "outing" class to "outings" so that we can simply use the document type.

c2corg_ui/static/partials/card_routes.html
```diff
@@ -1,9 +1,11 @@
-<a ng-click="cardCtrl.openCardPage()" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+<a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
   <span class="list-item-title"><span ng-if="locale.title_prefix">{{locale.title_prefix}}&nbsp;: </span>{{locale.title}}</span>
   <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{locale.lang}})</span>
   <br>
   <div class="list-item-info">
-    Min {{doc['elevation_min']}} m / Max {{doc['elevation_max']}} m
+    <span ng-show="doc['elevation_min']">Min {{doc['elevation_min']}} m</span>
+    <span ng-show="doc['elevation_min'] && doc['elevation_max']">/</span>
+    <span ng-show="doc['elevation_max']">Max {{doc['elevation_max']}} m</span>
```
asaunier: OK but don't spend too much time on changes like this. Actually we don't even know if those data are really the ones to display in the cards!

c2corg_ui/templates/waypoint/view.html
```diff
@@ -73,7 +73,7 @@
 
 <header class="view-details-title">
   <h1>${locale['title']}
-    <label class="badge waypoint" onClick="window.location='${request.route_url('waypoints_index')}' ">
+    <label class="badge waypoint" ng-click="window.location='${request.route_url('waypoints_index')}&wt=${waypoint['waypoint_type']}&bbox=' ">
```
asaunier: bbox?

asaunier: In addition the ``wt`` parameter does not exist. I think you want to use ``wtyp`` (see the URL parameters in the waypoints advanced search page).

c2corg_ui/static/js/deleteassociation.js
```diff
@@ -122,14 +122,12 @@ app.DeleteAssociationController.prototype.openModal_ = function() {
...
+ * @param {goog.events.Event | jQuery.Event} [event]
  * @private
  */
-app.DeleteAssociationController.prototype.unassociateDocument_ = function() {
-  this.api_.unassociateDocument(this.parentId, this.childId).then(function() {
-    this.documentService_.removeAssociation(this.childId, this.childDoctype);
-    this.rootscope_.$broadcast('unassociateDoc', {
-      'id': this.childId, 'type': this.childDoctype
-    });
+app.DeleteAssociationController.prototype.unassociateDocument_ = function(event) {
```
asaunier: Why do you need the event param?

c2corg_ui/static/js/documentservice.js
```diff
@@ -134,16 +141,21 @@ app.Document.prototype.pushToAssociations = function(doc, doctype,
...
  */
-app.Document.prototype.removeAssociation = function(id, type) {
+app.Document.prototype.removeAssociation = function(id, type, event) {
+  event.currentTarget.closest('.list-item').className += ' remove-item';
   var associations = this.document.associations[type];
-  for (var i = 0; i < associations.length; i++) {
-    if (associations[i]['document_id'] === id) {
-      associations.splice(i, 1);
-      break;
+  setTimeout(function () {
```
asaunier: Why is a timeout needed?

ginold: if you remove it instantly, it will disappear instantly because of two-way angular binding, which will automatically remove the object from ng-repeat.

c2corg_ui/static/js/deleteassociation.js
ginold: to add class to the closest .list-item

c2corg_ui/static/partials/card_routes.html
ginold: I had to because I saw a lot of cards with "Min   m / max 2000m" :(

ginold: It should be the time to know :P....

c2corg_ui/static/js/card.js
```diff
@@ -100,13 +93,14 @@ app.CardController.prototype.translate = function(str) {
...
-  if (!this.disableClick) {
-    window.location.href = app.utils.buildDocumentUrl(
-      app.utils.getDoctype(this.doc['type']),
-      this.doc['document_id'],
-      this.doc['locales'][0]['lang']);
+app.CardController.prototype.createURL = function() {
+  if (window.location.pathname.indexOf('edit') === -1) {
+    return app.utils.buildDocumentUrl(
+            app.utils.getDoctype(this.doc['type']),
+            this.doc['document_id'],
```
asaunier: I think you can use the "dot" notation for those doc attributes.

c2corg_ui/templates/waypoint/view.html
```diff
@@ -73,7 +73,7 @@
 
 <header class="view-details-title">
   <h1>${locale['title']}
-    <label class="badge waypoint" onClick="window.location='${request.route_url('waypoints_index')}' ">
+    <label class="badge waypoint" onClick="window.location='${request.route_url('waypoints_index')}&wtyp=${waypoint['waypoint_type']}' ">
```
asaunier: By the way it is not the pyramid way to add URL parameters. The right is:
```
request.route_url('waypoints_index', _query={'wtyp': waypoint['waypoint_type']})
```

c2corg_ui/static/js/documentservice.js
```diff
@@ -134,18 +141,25 @@ app.Document.prototype.pushToAssociations = function(doc, doctype,
...
-    this.document.associations.recent_outings.outings :
-    this.document.associations[type];
-  for (var i = 0; i < associations.length; i++) {
-    if (associations[i]['document_id'] === id) {
-      associations.splice(i, 1);
-      break;
+          this.document.associations.recent_outings.outings :
+          this.document.associations[type];
+
+  event.currentTarget.closest('.list-item').className += ' remove-item';
```
asaunier: Perhaps add a short comment explaining why you need a timeout here.

c2corg_ui/static/js/addassociation.js
```diff
@@ -96,6 +96,7 @@ app.AddAssociationController.prototype.associate = function(doc) {
   }
 
   this.api_.associateDocument(parentId, childId).then(function() {
+    doc['new'] = true;
```
asaunier: Do you need this? Isn't it already added in ``app.Document.prototype.pushToAssociations``?

c2corg_ui/static/js/card.js
ginold: I tried by adding a @type in the declaration, but then I would have to initialize ALL the properties (as for document in documentService) because if I only do `@type {appx.Document} this.doc; `  then it will tell me `Property xxx never defined on app.CardController.doc`...

asaunier:
OK, looks good to me (perhaps the "timeout" comment has a too long line but I won't bother you more :P)

ginold:
Ok thank's for reviewing. I will merge soon if no one opposes.

asaunier:
Please do, thanks!

PR #399, opened 2016-06-08
fix add-new-outing button url

### LINKED ISSUE #396: Add an outing from a route doesnt work if this is the first outing
Click on "Add a new one on this document"
(BTW the text should probably be moidified)

The page you are looking for does not exist or is broken.

/routes/736963/'http:/c2corgv6.demo-camptocamp.com/outings/add' + '

ginold:
fixes #396

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -196,12 +196,14 @@
         <span translate>Last outings</span><span class="glyphicon glyphicon-menu-down"></span>
       </h3>
       <div class="associated-documents collapse in" id="associated-outings">
-        <div class="list-item vertical-align" protected-url-btn
-             url="${request.route_url('outings_add') + '?' + document['doctype'] + '=' + str(document['document_id'])}">
-          <a>
-            <button class="btn orange-btn" translate>add a new outing to this document</button>
-          </a>
-        </div>
+        % if document['doctype'] == 'routes':
+          <div class="list-item vertical-align" protected-url-btn
+               url="${request.route_url('outings_add') + '?r=' + str(document['document_id'])}">
+            <a>
+              <button class="btn orange-btn" translate>add a new outing to this document</button>
+            </a>
+          </div>
+        % endif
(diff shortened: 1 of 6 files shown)


c2corg_ui/templates/route/view.html
```diff
@@ -162,7 +162,7 @@ <h3 class="heading show-phone"><span translate>description</span></h3>
             <div class="add-outing">
               <p class="text-center"><strong translate>no outings yet?</strong><p>
                 <button type="button" class="btn gray-btn" protected-url-btn
-                        url="'${request.route_url('outings_add')}' + '?routes=${route_id}'"
+                        url="${request.route_url('outings_add', _query={'routes': route_id})}"
```
asaunier: Is the ``routes`` param actually accepted?
In the advanced search I think we rather use ``r``. If possible it would be cool to use the same param name for consistency reasons.

ginold: why wouldn't it be ?
there is a check for 'routes' param in viewdetails.js, but you are completely right we should use the same everywhere.

c2corg_ui/templates/helpers/view.html
```diff
@@ -196,12 +196,14 @@ <h3 class="heading show-phone outings" ng-click="mainCtrl.animateHeaderIcon($eve
...
-        <div class="list-item vertical-align" protected-url-btn
-             url="${request.route_url('outings_add') + '?' + document['doctype'] + '=' + str(document['document_id'])}">
-          <a>
-            <button class="btn orange-btn" translate>add a new outing to this document</button>
-          </a>
-        </div>
+        % if document['doctype'] == 'routes':
+          <div class="list-item vertical-align" protected-url-btn
+               url="${request.route_url('outings_add') + '?r=' + str(document['document_id'])}">
+            <a>
```
asaunier: Are the ``<a>`` tags really needed?

asaunier:
Just a minor question else looks good to me.

c2corg_ui/templates/helpers/view.html
ginold: `<a>` is a part of .list-item styling, so most likely yes.

PR #400, opened 2016-06-09
Associated images

ginold:
Uses the real associated images of the document.

- Images that are too small will be shown in its original size, they will not be stretched
- Fixes the problem where the last two images were repeated twice

See for example /outings/736666/fr

### DIFF
--- a/c2corg_ui/static/js/viewdetails.js
+++ b/c2corg_ui/static/js/viewdetails.js
@@ -84,6 +84,7 @@ app.ViewDetailsController = function($scope, $compile, $uibModal, appApi,
    * @export
    */
   this.documentService = appDocument;
+  this.documentService.setDocument(documentData);
 
   /**
    * @type {Object}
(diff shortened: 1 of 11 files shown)


c2corg_ui/static/js/viewdetails.js
```diff
@@ -108,24 +108,6 @@ app.ViewDetailsController = function($scope, $compile, $uibModal, appApi,
...
-    {'src': 'http://s.camptocamp.org/uploads/images/1463694970_824684659.jpg', date: new Date(), activities: ['snow'], elevation: Math.random(), iso_speed: 400, filename: 'image231.jpg', camera_name: 'Canon', locales: [{'title' : 'my favorite place'}], date_time: new Date(), 'fnumber' : 4.2},
-    {'src': 'http://s.camptocamp.org/uploads/images/1463741192_488006925.jpg', date: '23-323', activities: ['hiking, snow, ski'], elevation: Math.random(), iso_speed: 1200, filename: 'image94.jpg', camera_name: 'Fuji', locales: [{'title' : 'superb view'}]},
-    {'src': 'http://s.camptocamp.org/uploads/images/1463694980_966569102.jpg', date: '099-02-2015', activities: ['paragliding'], elevation: Math.random(), iso_speed: 2300, filename: 'image55.jpg', camera_name: 'Sigma', locales: [{'title' : 'great view'}] , date_time: new Date(), 'fnumber' : 10}
-  ];
-
-  this.documentService.setDocument(documentData);
```
asaunier: I think you should not remove this line. It is needed to fill the documentService with the data passed to the mako templates.

asaunier:
Except my previous comment, the PR looks good to me.

c2corg_ui/static/js/viewdetails.js
ginold: `documentService.document` nor` scope[document] `were initialized anywhere. Therefore (and maybe you have noticed) there were no associated documents showing in any view page, because .

Now I have set `documentService.document = ${document}` in every view page and everything works perfectly.

asaunier: > Now I have set documentService.document = ${document} in every view page and everything works perfectly.

I don't see where this is done.

ginold: In view.html, for ex:
`ng-init="detailsCtrl.documentService.document = ${outing}"`

asaunier: I don't think this is still available. For me it's soooo ugly :P 
In the "association tools" PR I have removed it and replaced it by a module value:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L66-L71
which is then provided to the viewdetails directive and the documentservice:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/viewdetails.js#L80
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/viewdetails.js#L121

ginold: i saw the documentData thing but I didn't notice you initialized it within the script tag! I can try to reuse what you did if you find it sooo ugly :P but mine is only 1 line long.. :heart_decoration:

asaunier: I see that you have move the ``this.documentService.setDocument(documentData);`` line a little bit earlier in the code. Thanks!
So for me there is no more objection against merging the PR.

Issue #414, opened 2016-06-13
Responsive design improvement


stef74:
The full menu appear on too small windows (about 1000 pixel ?). Could you change it to appear arround 1600 pixels.

An other change can be also the 2° colonum creation arround 1400px ... it's too "early" with full menu. 

The title document size is too big with the full menu, in mobile the title si also too big. 

( ginold: solution already suggested
On mobile the snow condition table is unreadable ...
Ex : http://c2corgv6.demo-camptocamp.com/outings/594600/fr the table of snow condition have truncated text due to the two small width available. )

ginold:
Could you please attach a screenshot? i'm not sure what menu you are speaking of (1000-1600px is pretty enormous).

Ok for the snow condition tables :+1:

stef74:
I'm speaking of left menu the pixel is the size of windows

stef74:
![capture](https://cloud.githubusercontent.com/assets/3962997/16039675/768c67b4-322b-11e6-9a0d-6d05922cf5d8.PNG)

asaunier:
By the way the left menu is hidden by the ad banner when the window height is too small. The ad banner should be displayed behind the menu (or not be displayed at all?).

![issue403](https://cloud.githubusercontent.com/assets/19718055/17263606/2a845856-55e2-11e6-91e6-b1991b649e99.jpg)

stef74:
@asaunier can we close ?

asaunier:
The ad banner issue has been fixed.

Issue #412, opened 2016-06-13
route modification


desnoes:
**All kind of activities**

- Impossible to change the language, 
- the duration value is lost (checkbox is unchecked). 
- When we modify the activity or a criteria in the Configuration section (e.g. orientation), it is impossible to save the document. We can save modifications only if we modify a number or a text with the keyboard.

ginold:
this is normal that you cannot change the language : the document that has been created in one language will remain in this language and cannot be modified. Imagine you wrote your description in french and someone changes it to german? In order to modify the language, you have to click on the right button "more" -> translate into another lang.

desnoes:
I understand your point. This means however that when a route is created, no error is allowed on the selection of the language.

----- Mail original -----
De: "ginold" <notifications@github.com>
À: "c2corg/v6_ui" <v6_ui@noreply.github.com>
Cc: "desnoes" <mathieu.desnoes@free.fr>, "Author" <author@noreply.github.com>
Envoyé: Mardi 14 Juin 2016 12:53:03
Objet: Re: [c2corg/v6_ui] route modification (#412)



this is normal that you cannot change the language : the document that has been created in one language will remain in this language and cannot be modified. Imagine you wrote your description in french and someone changes it to german? In order to modify the language, you have to click on the right button "more" -> translate into another lang.

asaunier:
> This means however that when a route is created, no error is allowed on the selection of the language.

I don't understand what this means. You still have to select a language when creating a language, don't you?

desnoes:
> I don't understand what this means. You still have to select a language when creating a language, don't you?

Yes you have to select a language, but if you made a mistake you cannot correct it. For instance, I create a route in english and this was an error i wanted to select french. In this case, I cannot modify the language, i have to create the route in french and then cancel the route in english. We are facing this kind of problem regularly in V5.

tsauerwein:
For me it would make sense that you can change the language when you create a new document (but not when editing a document).

asaunier:
> In this case, I cannot modify the language, i have to create the route in french and then cancel the route in english. We are facing this kind of problem regularly in V5.

Yep that's the case in v5. But that's for a good reason. With the wiki system you cannot change the old version, you can only create a new one. That would create a huge mess and complications if the "lang" attribute could be changed between 2 versions.

> For me it would make sense that you can change the language when you create a new document (but not when editing a document).

+1

We won't change this behaviour.

fjacon:
I think this ticket can be closed.

PR #416, opened 2016-06-14
better edit-form checking before saving

### LINKED ISSUE #410: Smartphone : problems when creating an outing

ginold:
- save button will always be available if all default required fields are set
- if not, it will look like 'disabled' but clicking on it will list all remaining missing fields (without unnecessary sending the object to the API)
- added 'table-responsive' to conditions_levels table so it is better rendered on mobile.

Related to #410
Related partly to #414 (conditions levels)
Related partly to #412 (impossible to save the document)

### DIFF
--- a/c2corg_ui/static/js/alertservice.js
+++ b/c2corg_ui/static/js/alertservice.js
@@ -120,11 +120,11 @@ app.Alerts.prototype.formatErrorMsg_ = function(response) {
   if (len > 1) {
     var msg = '<ul>';
     for (var i = 0; i < len; i++) {
-      msg += '<li>' + this.filterStr_(errors[i]['description']) + '</li>';
+      msg += '<li>' + this.filterStr_(errors[i]['description']) + ' : ' + this.filterStr_(errors[i]['name']) + '</li>';
     }
     return msg + '</ul>';
   }
-  return this.filterStr_(errors[0]['description']);
+  return this.filterStr_(errors[0]['description']) + ' : ' + this.filterStr_(errors[0]['name']);
 };
 
 
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -223,7 +223,6 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   }
 
   this.scope_[this.modelName_] = this.scope_['document'] = this.documentService.document;
-
   if (this.auth_.isAuthenticated()) {
     if (this.id_ && this.lang_) {
      // Get document attributes from the API to feed the model:
@@ -629,6 +628,55 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType) 
 
 /**
  * @param {appx.Outing} outing
+ * Check what properties are missing and tell the user.
+ * @export
+ * @return {boolean | undefined}
+ */
+app.DocumentEditingController.prototype.hasMissingProps = function(doc, showError) {
+  var type = doc.type ? app.utils.getDoctype(doc.type) : this.module_;
+  var fields = app.constants.REQUIRED_FIELDS[type];
+  // understandable structure by alertService
+  var missing = {'data': {'errors': []}};
+  var error;
+  var field;
+  for (var i = 0; i < fields.length; i++) {
+    field = fields[i];
+    error = false;
+
+    if (field === 'title' || field === 'lang') {
+      error = (!doc['locales'] || !doc['locales'][0][field]);
+    } else if (field === 'activities') {
+      error = (!doc['activities'] || doc['activities'].length === 0);
+    } else if (field === 'routes') {
+      error = (!doc['associations'] || doc['associations']['routes'].length === 0);
+    } else if (field === 'latitude' || field === 'longitude') {
+      error = (!doc['lonlat'] || (doc['lonlat'][field] === null || doc['lonlat'][field] === undefined));
+    } else if (field === 'date_start') {
+      error = (doc['date_start'] === null || doc['date_start']  === undefined);
+    } else {
+      error = (!doc[field] || doc[field] === null || doc[field] === undefined);
+    }
+    if (error) {
+      if (showError) {
+        missing['data']['errors'].push({
+          'description': 'Missing field',
+          'name': field
+        });
+      } else {
+        return true;
+      }
+    }
+  }
+  if (missing['data']['errors'].length > 0) {
+    this.alerts_.addError(missing);
+  }
+  return error;
+};
+
+
+/**
+ * @param {Object} outing
+ * better edit-form checking before saving
  * @private
  */
 app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
--- a/c2corg_ui/templates/waypoint/edit.html
+++ b/c2corg_ui/templates/waypoint/edit.html
@@ -273,21 +273,21 @@
             </div>
 
             ## ACCESS TIME
-            <div class="form-group data half" ng-if="type == 'climbing_outdoor'">
+            <div class="form-group data half" ng-if="type == 'climbing_outdoor'" ng-class="{'has-success': waypoint.access_time}">
               <label translate>access_time</label>
               <select class="form-control" ng-model="waypoint.access_time" ng-options="time as time for time in ${access_times}"><option value=""></option></select>
               <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.access_time"></span>
             </div>
 
             ## RAIN PROOF
-            <div class="form-group data half" ng-if="type == 'climbing_outdoor'" ng-class="{ 'has-error': editForm.rain_proof.$touched && editForm.rain_proof.$invalid, 'has-success': editForm.rain_proof.$valid }">
+            <div class="form-group data half" ng-if="type == 'climbing_outdoor'" ng-class="{'has-success': waypoint.rain_proof}">
               <label><span translate>rain_proof</span></label>
               <select class="form-control " ng-model="waypoint.rain_proof" ng-options="type as mainCtrl.translate(type) for type in ${rain_proof_types}"><option value=""></option></select>
               <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.rain_proof"></span>
             </div>
 
             ## CHILDREN PROOF
-            <div class="form-group data half" ng-if="type == 'climbing_outdoor'">
+            <div class="form-group data half" ng-if="type == 'climbing_outdoor'" ng-class="{'has-success': waypoint.children_proof}">
               <label translate>children_proof</label>
               <select ng-options="type as type for type in ${children_proof_types}" ng-model="waypoint.children_proof" class="form-control "><option value=""></option></select>
               <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.children_proof"></span>
@@ -482,7 +482,7 @@
 
             ## PARAGLIDING RATING
             <div class="form-group data half rating" ng-if="type == 'paragliding_takeoff' || type == 'paragliding_landing'"
-                 ng-class="{ 'has-error': editForm.paragliding_rating.$touched && editForm.paragliding_rating.$invalid, 'has-success': editForm.paragliding_rating.$valid }">
+                 ng-class="{'has-success': waypoint.paragliding_rating }">
               <label><span translate>paragliding_rating</span></label><br>
               <select class="form-control" ng-options="rat as rat for rat in [1, 2, 3, 4, 5]" ng-model="waypoint.paragliding_rating"><option value=""></option></select>
               <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.paragliding_rating"></span>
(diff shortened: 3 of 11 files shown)


c2corg_ui/static/js/alertservice.js
```diff
@@ -49,11 +49,8 @@ app.Alerts.prototype.add = function(data) {
   var msg = data['msg'];
   msg = msg instanceof Object ? this.formatErrorMsg_(msg) :
       this.filterStr_(msg);
-  this.alerts_.push({
-    type: data['type'] || 'warning',
-    msg: msg,
-    timeout: timeout
-  });
+  data['type'] = data['type'] || 'warning';
+  this.alerts_.push(data);
```
asaunier: This change is not correct: ``timeout`` and ``msg`` changed a few lines above are no longer propagated. Why do you need to do that change?

c2corg_ui/static/js/alertservice.js
```diff
@@ -120,11 +117,11 @@ app.Alerts.prototype.formatErrorMsg_ = function(response) {
   if (len > 1) {
     var msg = '<ul>';
     for (var i = 0; i < len; i++) {
-      msg += '<li>' + this.filterStr_(errors[i]['description']) + '</li>';
+      msg += '<li>' + this.filterStr_(errors[i]['description'] + ' at ' + errors[i]['name']) + '</li>';
```
asaunier: I think this will cause i18n problems.

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,39 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+    } else {
+      missing[fields[i]] = doc[fields[i]] == (null || undefined) ? true : false;
+    }
+  }
+  for (var field in missing) {
+    if (missing[field]) {
+      error.push(field);
+    }
+  }
+  this.alerts_.addError('Missing fields: ' + error.join(', '));
```
asaunier: How can we translate this?

c2corg_ui/static/js/alertservice.js
ginold: oO i must have probably misinterpreted something. I will revert this change.

c2corg_ui/static/js/documentediting.js
ginold: hm, maybe i should create an` ul-li `with separate lines and `spans` that have `translate` for each missing field.

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,39 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+    if (fields[i] === 'title' || fields[i] === 'lang') {
+      missing[fields[i]] = (!doc['locales'] || doc['locales'][0][fields[i]].length === 0) ? true : false;
+    } else if (fields[i] === 'activities') {
+      missing[fields[i]] = (!doc['activities'] || doc['activities'].length === 0) ? true : false;
+    } else if (fields[i] === 'routes') {
+      missing[fields[i]] = (!doc['associations'] || doc['associations']['routes'].length === 0) ? true : false;
+    } else if (fields[i] === 'latitude' || fields[i] === 'longitude') {
+      missing[fields[i]] = (!doc['lonlat'] || doc['lonlat'][fields[i]] == (null || undefined)) ? true : false;
+    } else {
+      missing[fields[i]] = doc[fields[i]] == (null || undefined) ? true : false;
```
asaunier: ``? true : false`` is a redundant notation. It should rather be something like that:
```
missing[fields[i]] = (doc[fields[i]] == null || undefined);
```

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,39 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+ * @export
+ */
+app.DocumentEditingController.prototype.checkMissingProps = function(doc) {
+  var type = doc.type ? app.utils.getDoctype(doc.type) : this.modelName_ + 's'; // missing 's' in modelName if creating a new document
+  var fields = app.constants.REQUIRED_FIELDS[type];
+  var missing = {};
+  var error = [];
+
+  for (var i = 0; i < fields.length; i++) {
+    if (fields[i] === 'title' || fields[i] === 'lang') {
```
asaunier: All those ``fields[i]`` make the code hard to read. You should use an intermediate ``var field = fields[i];``.

ginold:
#updated

c2corg_ui/static/js/alertservice.js
```diff
@@ -120,11 +120,16 @@ app.Alerts.prototype.formatErrorMsg_ = function(response) {
   if (len > 1) {
     var msg = '<ul>';
     for (var i = 0; i < len; i++) {
-      msg += '<li>' + this.filterStr_(errors[i]['description']) + '</li>';
+      if (errors[i]['description'] && errors[i]['name']) {
+        msg += '<li><span translate>' + this.filterStr_(errors[i]['description']) +
+                        '</span><span translate> ' + this.filterStr_(errors[i]['name']) +
```
asaunier: I don't think the "translate" directive are needed here since ``this.filterStr_`` already translates the strings:
https://github.com/c2corg/v6_ui/blob/7c50d4b7b4fff12a8eb4b4d2b49efd9d18a87e70/c2corg_ui/static/js/alertservice.js#L141-L144

c2corg_ui/static/js/alertservice.js
```diff
@@ -120,11 +120,16 @@ app.Alerts.prototype.formatErrorMsg_ = function(response) {
...
+                        '<span>' +
+                      '</li>';
+      }
     }
     return msg + '</ul>';
   }
-  return this.filterStr_(errors[0]['description']);
+  return this.filterStr_(errors[0]['description'] + ' at ' + errors[0]['name']);
```
asaunier: This is still not possible to translate using this syntax. And do we need the "at"? It makes things even more complicated. What about something like:
```
return this.filterStr_(errors[0]['description']) + ' (' +  this.filterStr_(errors[0]['name']) + ')';
```
or
```
return '{description} ({error})'
  .replace('{description}', this.filterStr_(errors[0]['description']))
  .replace('{error}', this.filterStr_(errors[0]['error']));
```

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,46 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
 
 
 /**
+ * Check what properties are missing and tell the user.
+ * @param {Object} doc
+ * @export
+ */
+app.DocumentEditingController.prototype.checkMissingProps = function(doc) {
+  var type = doc.type ? app.utils.getDoctype(doc.type) : this.modelName_ + 's'; // missing 's' in modelName if creating a new document
```
asaunier: > this.modelName_ + 's'; // missing 's' in modelName if creating a new document

What a hack :P 

You should rather use ``this.module_`` which contains the correct type name with the "s".
See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/documentediting.js#L91
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/edit.html#L28

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,46 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+  var fields = app.constants.REQUIRED_FIELDS[type];
+  // understandable structure by alertService
+  var missing = {'data': {'errors': []}};
+  var error;
+  for (var i = 0; i < fields.length; i++) {
+    var field = fields[i];
+    error = missing['data']['errors'][i] = {};
```
asaunier: I don't understand this code.
Doesn't this mean that a "missing" error is added for all required fields?
Shouldn't you add the error only after checking that the field is not set?

c2corg_ui/static/js/alertservice.js
ginold: it is still complicated, we should create something like <span translate>{description}</span> <span translate>{name}</span>

c2corg_ui/static/js/alertservice.js
ginold: I didn't notice, you're right.

ginold:
updated pt 2

c2corg_ui/static/js/alertservice.js
```diff
@@ -120,11 +120,13 @@ app.Alerts.prototype.formatErrorMsg_ = function(response) {
   if (len > 1) {
     var msg = '<ul>';
     for (var i = 0; i < len; i++) {
-      msg += '<li>' + this.filterStr_(errors[i]['description']) + '</li>';
+      if (errors[i]['description'] && errors[i]['name']) {
```
asaunier: Do you really need to check that ``description`` and ``name`` are set? Is it possible they are not?

asaunier:
A minor comment else looks good to me.

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,45 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+    if (field === 'title' || field === 'lang') {
+      pass = (!doc['locales'] || !doc['locales'][0][field]);
+    } else if (field === 'activities') {
+      pass = (!doc['activities'] || doc['activities'].length === 0);
+    } else if (field === 'routes') {
+      pass = (!doc['associations'] || doc['associations']['routes'].length === 0);
+    } else if (field === 'latitude' || field === 'longitude') {
+      pass = (!doc['lonlat'] || doc['lonlat'][field] === (null || undefined));
+    } else if (field === 'date_start') {
+      pass = doc['date_start'] === (null || undefined);
```
tsauerwein: This is maybe no doing what you are expecting. The expression `val === (null || undefined)` first evaluates the part in the brackets `(null || undefined)`, so you end up with `val === undefined`. Is that what you want to check for?

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,6 +594,45 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+    }
+    if (pass) {
```
tsauerwein: Is `pass` a good name? Maybe `hasError`?

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -780,7 +780,14 @@ <h2 class="heading show-phone"><span translate>summary</span></span></h2>
...
+          <button type="submit" class="btn orange-btn btn-lg float-button" 
+            ng-if="!editForm.$invalid && (waypoint.locales[0].title && waypoint.waypoint_type && waypoint.locales[0].lang && waypoint.elevation && waypoint.lonlat.latitude && waypoint.lonlat.longitude)" tooltip-placement="left" uib-tooltip="{{'Save' | translate}}">
+            <span class="glyphicon glyphicon-ok"></span>
+          </button>
+          <button type="button" class="btn btn-lg orange-btn float-button" 
+                  ng-style="(editForm.$invalid || (!waypoint.locales[0].title || !waypoint.waypoint_type || !waypoint.locales[0].lang || !waypoint.elevation || !waypoint.lonlat.longitude || !waypoint.lonlat.latitude)) && {'opacity': 0.5}" 
+                  tooltip-placement="left" uib-tooltip="{{'What is missing?' | translate}}" 
+                  ng-if="editForm.$invalid || (!waypoint.locales[0].title || !waypoint.waypoint_type || !waypoint.locales[0].lang || !waypoint.elevation || !waypoint.lonlat.longitude || !waypoint.lonlat.latitude)" ng-click="editCtrl.checkMissingProps(waypoint)">
```
tsauerwein: Can't you move this check if all required fields are given into a method? And can't you use `checkMissingProps` for that purpose?

c2corg_ui/static/js/documentediting.js
ginold: I would like to literally check if my value == null or undefined. I'd rather write `val == null || val == undefined`?

tsauerwein: If that's what you want to check, yes. But note the difference between `==` and `===` in JavaScript.

c2corg_ui/templates/waypoint/edit.html
ginold: Indeed checkMissingProps works for this checking as well. I first wrote the code above and forgot about it, as it worked :P

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -273,21 +273,21 @@ <h2 class="heading show-phone" >
...
-            <div class="form-group data half" ng-if="type == 'climbing_outdoor'" ng-class="{ 'has-error': editForm.rain_proof.$touched && editForm.rain_proof.$invalid, 'has-success': editForm.rain_proof.$valid }">
+            <div class="form-group data half" ng-if="type == 'climbing_outdoor'" ng-class="{'has-success': waypoint.rain_proof}">
               <label><span translate>rain_proof</span></label>
               <select class="form-control " ng-model="waypoint.rain_proof" ng-options="type as mainCtrl.translate(type) for type in ${rain_proof_types}"><option value=""></option></select>
               <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.rain_proof"></span>
             </div>
 
-            ## CHILDREN PROOF
-            <div class="form-group data half" ng-if="type == 'climbing_outdoor'">
+            ## CHILDREN PROOFrou
```
asaunier: Typo

asaunier:
:+1:

Issue #406, opened 2016-06-11
Outing creation: problems with dates


loicperrin:
See issue #401 
But there are other problems with dates:
- [x] When I select 04/06/2016 (with calendar tool), the date 03/06/2016 is saved in the outing.
- [x] Then I try to modify this starte date, and replace the date 03/06/2016 by 11/06/2016, but 11/06/2016 is put in start date (that's OK) and 03/06/2016 is saved in end date (that's KO)
- [x] I can select a start date after the end date (the tool should control that end date is after start date)
- [x] New problem with date: it is no more displayed in the outing (I think it was in Alpha1).

loicperrin:
New problem with date: it is no more displayed in the outing (I think it was in Alpha1).

shortoland:
Display is OK.
Pb with end date in case of several days outing: the displayed day is the day before the entered one.

stef74:
@fjacon please fix the lastest point

stef74:
@loicperrin please test

loicperrin:
OK

Issue #404, opened 2016-06-11
Problem after saving an outing 


loicperrin:
I try to create an outing. I have filled several fields. When I save it, the following error message displays:
---
Internal Server Error

The server encountered an internal error or misconfiguration and was unable to complete your request.

Please contact the server administrator at [no address given] to inform them of the time this error occurred, and the actions you performed just before this error.

More information about this error may be available in the server error log.
---
The URL displayed is outings/736971/fr, so it seems that the outing has been created, but if I try to display it (copy/paste in URL address), I get the same error message.
This outing is not visible on the topoguide page.

ginold:
does this happen every time you create an outing or is it occasional? If it happens sometimes, please try to write down the steps you make in order to reproduce this bug.

loicperrin:
I have this problem once, and I don't succeed to reproduce it today. 
But if I try to display the outing outings/736971/fr, I have still Error 500. Maybe there is more information on this "wrong" outing in the log files or in database?

ginold:
`Failed to load resource: the server responded with a status of 404 (Not Found)` @tsauerwein looks like you are the best to fix it up!

tsauerwein:
> Failed to load resource: the server responded with a status of 404 (Not Found) @tsauerwein looks like you are the best to fix it up!

When are you getting this error? The API correctly returns the document for this outing: http://c2corgv6-api.demo-camptocamp.com/outings/737037?l=fr

asaunier:
It's probably not an API issue, rather a UI template error. The apache logs of the UI say that

> [Mon Jun 20 12:59:29.912019 2016] [wsgi:error] [pid 28435] [remote 10.27.71.4:576]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/outing/view.html", line 138, in render_body
[Mon Jun 20 12:59:29.912025 2016] [wsgi:error] [pid 28435] [remote 10.27.71.4:576]     ${get_document_locale_text(locale, 'weather')}
[Mon Jun 20 12:59:29.912032 2016] [wsgi:error] [pid 28435] [remote 10.27.71.4:576]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/helpers/view.html", line 261, in render_get_document_locale_text
[...]

PR #417, opened 2016-06-14
make manual date writing impossible

### LINKED ISSUE #401: Impossible to save an outing if date is manually filled
I try to create an outing (either from a waypoint or with "Ajouter une sortie" button).
I have filled some flelds but not all the fields. For the start date, I fill the fiel manually "04/06/2016".
When trying to save, I have the following message:
"    date_end is missing
    date_start is missing
    Required
    Required"
If the start date is filled using the button (calendar), it is possible to save.

ginold:
- inputs where the date appear are disabled and can be changed only with the calendar button.
- don't let selecting end_date before start_date
- don't let selecting start_date after end_date
- don't let selecting a date posterior to today
- minor fix for conditions_levels

Related to #404 (conditions_levels)
Related to #429
Related to #401 
Related to #406

### DIFF
--- a/c2corg_ui/locale/c2corg_ui-client.pot
+++ b/c2corg_ui/locale/c2corg_ui-client.pot
@@ -162,11 +162,11 @@ msgid "Describe here the waypoint"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
-msgid "Document has been deprotected"
+msgid "Document has been protected"
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
-msgid "Document has been protected"
+msgid "Document has been unprotected"
 msgstr ""
 
 #: c2corg_ui/static/partials/imageuploader.html
@@ -243,6 +243,10 @@ msgstr ""
 
 #: c2corg_ui/templates/auth.html
 msgid "Forgot password?"
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "Form is not valid"
 msgstr ""
 
 #: c2corg_ui/templates/sidemenu.html
@@ -584,6 +588,14 @@ msgstr ""
 
 #: c2corg_ui/static/partials/imageuploader.html
 msgid "You can add geolocation info by typing in the input field."
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "You have no rights to edit this document."
+msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "You must log in to edit this document."
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
--- a/c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
+++ b/c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
@@ -158,12 +158,12 @@ msgid "Describe here the waypoint"
 msgstr "Décrivez ici l'itinéraire"
 
 #: c2corg_ui/templates/i18n.html
-msgid "Document has been deprotected"
-msgstr "Le document a été déverrouillé."
-
-#: c2corg_ui/templates/i18n.html
 msgid "Document has been protected"
 msgstr "Le document a été verrouillé."
+
+#: c2corg_ui/templates/i18n.html
+msgid "Document has been unprotected"
+msgstr ""
 
 #: c2corg_ui/static/partials/imageuploader.html
 msgid "Drop images here or click to upload"
@@ -238,6 +238,10 @@ msgstr "Filtres"
 #: c2corg_ui/templates/auth.html
 msgid "Forgot password?"
 msgstr "Mot de passe oublié?"
+
+#: c2corg_ui/templates/i18n.html
+msgid "Form is not valid"
+msgstr ""
 
 #: c2corg_ui/templates/sidemenu.html
 msgid "Forum"
@@ -570,6 +574,14 @@ msgstr "Vous consultez une ancienne version de ce document."
 #: c2corg_ui/static/partials/imageuploader.html
 msgid "You can add geolocation info by typing in the input field."
 msgstr ""
+
+#: c2corg_ui/templates/i18n.html
+msgid "You have no rights to edit this document."
+msgstr "Vous n'avez pas les droits pour éditer ce document."
+
+#: c2corg_ui/templates/i18n.html
+msgid "You must log in to edit this document."
+msgstr "Vous devez être loggué pour éditer ce document."
 
 #: c2corg_ui/templates/i18n.html c2corg_ui/templates/outing/edit.html
 #: c2corg_ui/templates/waypoint/create_edit_summary.html
@@ -1282,7 +1294,7 @@ msgstr "Remontées mécaniques"
 #: c2corg_ui/templates/outing/edit.html
 #: c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
 msgid "lift_status"
-msgstr ""
+msgstr "Remontées mécaniques"
 
 #: c2corg_ui/templates/i18n.html
 msgid "local_product"
@@ -1294,7 +1306,7 @@ msgstr "Lieu-dit"
 
 #: c2corg_ui/templates/helpers/view.html c2corg_ui/templates/outing/edit.html
 msgid "location"
-msgstr ""
+msgstr "Situation"
 
 #: c2corg_ui/templates/waypoint/edit.html
 msgid "longitude"
@@ -1417,7 +1429,7 @@ msgstr "novembre"
 
 #: c2corg_ui/templates/route/edit.html
 msgid "numbers"
-msgstr ""
+msgstr "Nombres"
 
 #: c2corg_ui/templates/i18n.html
 msgid "oct"
@@ -1604,7 +1616,7 @@ msgstr "aller-retour"
 #: c2corg_ui/templates/outing/edit.html c2corg_ui/templates/route/edit.html
 #: c2corg_ui/templates/waypoint/edit.html
 msgid "review"
-msgstr ""
+msgstr "Revue"
 
 #: c2corg_ui/templates/route/create_edit_summary.html
 #: c2corg_ui/templates/route/edit.html
@@ -1673,7 +1685,7 @@ msgstr "Quantité des itinéraires"
 
 #: c2corg_ui/templates/outing/edit.html
 msgid "routes_taken"
-msgstr ""
+msgstr "Itinéraires empruntés"
 
 #: c2corg_ui/templates/i18n.html
 msgid "see more results"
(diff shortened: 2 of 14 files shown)


c2corg_ui/static/js/utils.js
```diff
@@ -123,6 +123,26 @@ app.utils.formatDate = function(date) {
 
 
 /**
+ * check that the start date if before end date.
+ * Return true is start > end
+ * @export
+ * @param {Object} outing
+ * @return {(boolean | undefined)}
```
tsauerwein: Why is the type `boolean | undefined`? Isn't `boolean` enough? You just have to make sure that your function always returns a boolean value.

ginold: We had a function where we actually wanted to return in one case a `string` and else nothing. So we wrote `string | undefined`, in this function actually i may want `true` or nothing, but `false` may be interpreted as such :P

tsauerwein: > in this function actually i may want true or nothing, but false may be interpreted as such :P

What do you mean?

ginold: I mean that in this case i'm never interested in `false`, only `true` interests me. That's why I wanted true or nothing. I have already corrected it to return false as well, but it's of no use in this particular case.

tsauerwein: > ... but it's of no use in this particular case.

In this function you want to check if the start date is before the end data. You can answer this question with `true` or `false`. I don't see why you would need a third value `undefined`.

c2corg_ui/static/js/utils.js
```diff
@@ -123,6 +123,28 @@ app.utils.formatDate = function(date) {
...
+ * @return {boolean}
+ */
+app.utils.checkDates = function(outing) {
+  var start = new Date(outing['date_start']);
+  var end = new Date(outing['date_end']);
+
+  if (!end) {
+    return false;
+  }
+  if (start > end) {
```
tsauerwein: This could be: `return start > end`.

ginold: I always forget about those shorthands !

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,25 +605,40 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
+ * @return {boolean}
+ */
+app.DocumentEditingController.prototype.checkDates = function() {
+  var check = app.utils.checkDates(this.scope_['outing']);
+  if (check) {
+    this.alerts_.addError(check);
+    return false;
+  } else {
+    return true;
+  }
```
asaunier: You can save one line writing it like this:
```
if (check) {
  this.alerts_.addError(check);
  return false;
}
return true:
```
:P

c2corg_ui/static/js/documentediting.js
```diff
@@ -595,25 +605,40 @@ app.DocumentEditingController.prototype.updateMaxSteps = function(waypointType)
...
     }
-  }
-  // creating a new outing -> init locales and associations
-  if (!outing['locales']) {
-    outing['locales'] = [{}];
-  }
-  if (!outing['associations']) {
-    outing['associations'] = {'routes': [], 'waypoints': [], users: []};
+    for (var i = 0; i < outing['associations']['users'].length; i++) {
+      outing['associations']['users'][i]['id'] = outing['associations']['users'][i]['document_id'] || outing['associations']['users'][i]['id'];
```
asaunier: Perhaps you can use the dot notation here and elsewhere to save the brackets and make the lines a little shorter? Also use an intermediate ``var associations = outing['associations'];``?

c2corg_ui/static/js/utils.js
```diff
@@ -123,6 +123,30 @@ app.utils.formatDate = function(date) {
...
+ */
+app.utils.checkDates = function(outing) {
+  var start = new Date(outing['date_start']);
+  var end = new Date(outing['date_end']);
+
+  if (!end) {
+    return;
+  }
+  if (start > end) {
+    return 'The starting date is after the ending date.';
```
asaunier: Would it be possible to have those strings detected by the i18n extractor? If not we need to add them in the translations files.

c2corg_ui/static/js/documentediting.js
```diff
@@ -205,6 +205,11 @@ app.DocumentEditingController = function($scope, $element, $attrs,
           this.successRead_.bind(this)
       );
     } else if (this.modelName_ === 'outing') {
+      this.scope_['outing']['associations']['users'].push({
```
asaunier: Some more outing-specific code in the documentediting controller...

ginold: this is nothing new, just updating...

c2corg_ui/static/js/utils.js
ginold: I don't think it's too easy to detect it in .js files, how would the extractor know which are translatable or not?
I will add those to i18n file.

asaunier: I had the same problem recently but couldn't find the right way to do it. But it should be possible, see https://github.com/c2corg/v6_ui/blob/master/Makefile#L104-L105

To mark the string as extractable, one should use something like ``this.gettextCatalog_.getString('some i18n string')`` as in https://github.com/c2corg/v6_ui/commit/58ef63d8135cac5aa1bc3c9993ef34f8ff70901e but that's probably not the correct syntax.

@tsauerwein @gberaudo do you know the right way to do it?

ginold:
#updated

c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
```diff
@@ -571,6 +579,18 @@ msgstr "Vous consultez une ancienne version de ce document."
 msgid "You can add geolocation info by typing in the input field."
 msgstr ""
 
+#: c2corg_ui/templates/i18n.html
+msgid "You cannot select a date later than today."
+msgstr "Vous ne pouvez pas séléctionner une date postérieure à aujourd'hui."
```
asaunier: sélectionner

c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
```diff
@@ -1282,7 +1302,7 @@ msgstr "Remontées mécaniques"
 #: c2corg_ui/templates/outing/edit.html
 #: c2corg_ui/templates/outing/helpers/detailed_outings_attributes.html
 msgid "lift_status"
-msgstr ""
+msgstr "Etat d'ascenseur"
```
asaunier: WTF? :P
This translation does not make sense. In v5 it's simply translated as "Remontées mécaniques".

c2corg_ui/static/js/documentediting.js
asaunier: OK, we'll do something with https://github.com/c2corg/v6_ui/pull/431

c2corg_ui/static/js/utils.js
```diff
@@ -123,6 +123,30 @@ app.utils.formatDate = function(date) {
 
 
 /**
+ * check that the start date if before end date.
+ * Returns an error message if there is an error with dates.
+ * @export
+ * @param {Object} outing
+ * @return {(undefined | string)}
+ */
+app.utils.checkDates = function(outing) {
```
asaunier: Instead of displaying errors when the user select incorrect dates, would it be possible to prevent him or her from doing so?
Isn't it possible to disable future dates in the date picker and to disable dates in the "end" date picker that are anterior to the selected date in the "start" date picker?

c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
ginold: well, gotta know that "lift (ascenseur) status (etat)" is translated to rémontées mécaniques :D

c2corg_ui/static/js/utils.js
ginold: It should be possible, i havent looked for it yet.

c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
asaunier: :)
Anyway it's no big deal. It's not your job to make the translations.

c2corg_ui/static/js/utils.js
asaunier: I think it would be better.
"mieux vaut prévenir que guérir" "an ounce of prevention is worth a pound of cure"

ginold: you mean Vorbeugen ist besser als Heilen?

asaunier: genau :)

ginold: ok I got it, i will update the pr soon.

ginold:
updated pt 2

c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
asaunier: Please apply this spelling fix.

c2corg_ui/locale/fr/LC_MESSAGES/c2corg_ui-client.po
asaunier: Please replace by "Remontées mécaniques"

c2corg_ui/locale/c2corg_ui-client.pot
```diff
@@ -486,6 +490,10 @@ msgstr ""
 msgid "The page you are looking for does not exist or is broken."
 msgstr ""
 
+#: c2corg_ui/templates/i18n.html
+msgid "The starting date is after the ending date."
```
asaunier: This translation seems no longer used and should be removed.

c2corg_ui/locale/c2corg_ui-client.pot
```diff
@@ -587,6 +595,18 @@ msgid "You can add geolocation info by typing in the input field."
 msgstr ""
 
 #: c2corg_ui/templates/i18n.html
+msgid "You cannot select a date later than today."
```
asaunier: This translation seems no longer used and should be removed.

c2corg_ui/templates/helpers/view.html
```diff
@@ -263,12 +263,12 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
                 % for condition in json.loads(locale['conditions_levels']):
                 <tr>
                   <td>${condition['level_place']}</td>
-                  % if condition['level_snow_height_soft'] != '':
+                  % if 'level_snow_height_soft' in condition and condition['level_snow_height_soft'] != '':
```
asaunier: Does this solve https://github.com/c2corg/v6_ui/issues/404?

asaunier:
Looks good to me. OK for merging when Travis passes.

PR #418, opened 2016-06-14
Clean up externs files

tsauerwein:
Currently there are several places where external libraries are used without providing an externs file (e.g. [here](https://github.com/c2corg/v6_ui/blob/1a7fafe/c2corg_ui/static/js/deleteassociation.js#L114) or [here](https://github.com/c2corg/v6_ui/blob/32f052b04b7246cffd5577ea8cbfdf1c4237e1da/c2corg_ui/static/js/imageuploader.js#L311)).

When calling a method of an external library or when referencing a property using the *dot-syntax* (e.g. `foo.bar()` or `foo.bar`), the Closure Compiler will rename these methods or properties if no externs file is provided. It is only by chance that the methods are not renamed in the two cases linked above (there are methods in other externs that by accident have the same names).

So, when using an external library, you can either use the *string-syntax* (e.g. `foo['bar']()`, see also [Inconsistent Property Names](https://developers.google.com/closure/compiler/docs/api-tutorial3#propnames)) or use a basic externs file. Before writing your own externs file, do a quick search if no-one wrote one before (for example there was already one for angular-bootstrap-ui). And if you have to write your own, you only need to include the methods that you are actually using.

### DIFF
--- a/build.json
+++ b/build.json
@@ -16,6 +16,10 @@
       "c2corg_ui/externs/slick.js",
       "c2corg_ui/externs/photoswipe.js",
       "c2corg_ui/externs/loadImage.js",
+      "c2corg_ui/externs/ui-bootstrap.js",
+      "c2corg_ui/externs/angular-1.5.js",
+      "c2corg_ui/externs/angular-1.5-q_templated.js",
+      "c2corg_ui/externs/angular-1.5-http-promise_templated.js",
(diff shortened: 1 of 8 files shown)


asaunier:
Looks good to me, thanks!
@ginold OK for you?

ginold:
if it still works, yeah :D looks superb.

Issue #407, opened 2016-06-11
Problems with outing modifications


loicperrin:
* [x] It is not possible to modify "Conditions d'enneigement par zones": the table is empty and when clicking on "Ajouter une ligne", nothing happens

* [x] When I select a "Transports publics", it is displayed in the outing. But if I modified this outing, it is not selected in the form. Please notice that "Transport publics" have to be replaced by a checkbox (already mentionned in another issue).

(see outings/edit/736972/fr to reproduce this problems)

asaunier:
> In step 4 (Review/Résumé): it seems to be possible to modify the fields Titre, Commentaires personnels and Conditions. In fact only the modifications in the title are saved.

The "review/résumé" should not be tested. It's a temporary page that should be replaced by a real preview page. See https://github.com/c2corg/v6_ui/issues/168

loicperrin:
I have suppressed the bug on "review/résumé", but added another one.

asaunier:
> It is not possible to modify "Conditions d'enneigement par zones": the table is empty and when clicking on "Ajouter une ligne", nothing happens

I have noticed that when opening the editing page of an already existing outing, the JS console complains that:
> angular.js:13642 Error: [ngRepeat:dupes] http://errors.angularjs.org/1.5.6/ngRepeat/dupes?p0=cond%20in%20outing.locales%5B0%5D.conditions_levels&p1=string%3Ae&p2=e

which is probably the reason why the "conditions_levels" table is not displayed correctly.

https://docs.angularjs.org/error/ngRepeat/dupes?p0=cond%20in%20outing.locales%5B0%5D.conditions_levels&p1=string:e&p2=e
says that
> Error: ngRepeat:dupes
Duplicate Key in Repeater
Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: cond in outing.locales[0].conditions_levels, Duplicate key: string:e, Duplicate value: e

ginold:
conditions_levels problem is due to the fact that it happens to be hardcoded for empty strings, if it has not been set (required by API). See  https://github.com/c2corg/v6_api/issues/266

ginold:
public transport issue -> see #333

ginold:
> "Transport publics" have to be replaced by a checkbox 

it could be a checkbox, but I will replace it by a selection tool with values yes, no, and nothing because it is more coherent to the design, a single checkbox in the middle of selection tools is not nice.

[...]

PR #431, opened 2016-06-16
Move the outing-specific editing to a dedicated controller

### LINKED ISSUE #246: Keep the documentediting controller as generic as possible

asaunier:
Fixes #246 

This PR introduces the possibility to select an inherited controller when using the ``appDocumentEditingDirective``. This way, code specific to outings may be moved to an ``AppOutingEditingController`` controller.

See explanations at http://stackoverflow.com/a/23647720

For class inheritance with Closure, see also
* http://stackoverflow.com/questions/11122608/why-is-goog-basethis-necessary-in-addition-to-goog-inherits
* http://bolinfest.com/javascript/inheritance.php
* http://bolinfest.com/essays/googbase.html

~~The PR is almost ready even though I have to fix some error that occurs when editing an outing (but the editing seems to work anyway, intensive testing is still to be done).~~

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -32,7 +32,8 @@ app.documentEditingDirective = function() {
   return {
     restrict: 'A',
     scope: true,
-    controller: 'appDocumentEditingController',
+    controller: '@',
+    name: 'appDocumentEditingControllerName',
     controllerAs: 'editCtrl',
     bindToController: true
   };
@@ -80,9 +81,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 
   /**
    * @type {app.Authentication}
-   * @private
+   * @public
    */
-  this.auth_ = appAuthentication;
+  this.auth = appAuthentication;
 
   /**
    * @type {string}
@@ -93,9 +94,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 
   /**
    * @type {string}
-   * @private
+   * @public
    */
-  this.modelName_ = $attrs['appDocumentEditingModel'];
+  this.modelName = $attrs['appDocumentEditingModel'];
 
   /**
    * @type {number}
@@ -111,9 +112,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 
   /**
    * @type {angular.Scope}
-   * @private
+   * @export
    */
-  this.scope_ = $scope;
+  this.scope = $scope;
 
   /**
    * @type {ol.format.GeoJSON}
@@ -135,9 +136,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 
   /**
    * @type {app.Alerts}
-   * @private
+   * @public
    */
-  this.alerts_ = appAlerts;
+  this.alerts = appAlerts;
 
   /**
    * first next step would be 2 by default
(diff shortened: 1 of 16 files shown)


asaunier:
> The PR is almost ready even though I have to fix some error that occurs when editing an outing (but the editing seems to work anyway, intensive testing is still to be done).

Actually this a problem that also occurs on the demo instance and is not related to the proposed changes. See https://github.com/c2corg/v6_ui/issues/407#issuecomment-226717233

I have made a few usage tests and noticed no problems with editing of waypoints/routes/outings.
@ginold could you have a look at the code splitting I propose and make sure I have not made mistakes with your code? Thanks!

c2corg_ui/static/js/documentediting.js
```diff
@@ -80,9 +81,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
 
   /**
    * @type {app.Authentication}
-   * @private
+   * @export
```
tsauerwein: Do all these properties have to be exported? Or do you just want that they can be accessed from the child controllers?

c2corg_ui/static/js/documentediting.js
```diff
@@ -374,6 +341,18 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
 
 
 /**
+ * @param {appx.Document} data Document attributes.
+ * @return {appx.Document}
+ * @export
```
tsauerwein: Same question here: Why `export`? `@export` makes sure that a property/method is not renamed by the compiler so that you can call if from a template. But that doesn't seem to be the case?

tsauerwein:
Nice move! That looks good!

c2corg_ui/static/js/documentediting.js
asaunier: These properties are indeed accessed by the child controllers and cannot be tagged as private. They do not really need to be exported. Should I simply remove the ``@private`` tag without adding an ``@export`` one?
If yes is the properties renaming appropriate (no trailing "_")?

c2corg_ui/static/js/documentediting.js
asaunier: Indeed, it's only that the method can be overridden in the child controllers.
Same fix: removing the ``@export`` tag?

c2corg_ui/static/js/documentediting.js
tsauerwein: You could use `@public`. The underscore is only required for private attributes/methods.

asaunier: OK, I'll make the change. Thanks for the tip again :)

asaunier:
I have just updated the PR to use ``@public`` instead of ``@export`` as suggested by @tsauerwein 

@ginold Does the PR look good to you? Do you prefer we merge your changes from https://github.com/c2corg/v6_ui/pull/416 and https://github.com/c2corg/v6_ui/pull/417 first (there will probably be conflicts)? Are you almost done with those PRs?

ginold:
Finally someone did it :D it looks good, but we'd better first merge my 2 prs that have some additional functions and then your pr. I will have finished soon.

asaunier:
I have rebased and adapted this PR to take into account changes from #416 and #417. The PR is ready for reviewing.

The editing of waypoints/routes/outings seems to still work but that would be cool if you could double-check my changes @ginold Thanks :)

ginold:
looks and works fine for me.

asaunier:
Cool, thanks for reviewing @ginold!

Issue #125, opened 2016-01-27
Maps : background layer selector


asaunier:
Here are a few examples in ngeo:
http://camptocamp.github.io/ngeo/master/examples/backgroundlayer.html
http://camptocamp.github.io/ngeo/master/examples/backgroundlayerdropdown.html

I have also noticed some work has been done in GMF. For instance:
http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/mobilebackgroundlayerselector.html
http://adube.github.io/ngeo/gmf-background-selector-mobile/examples/contribs/gmf/apps/mobile/index.html
even though the example above might be a little overkill => we need a simple way to switch among background layers by clicking on an icon or a dropdown on the map.

@tsauerwein I have noticed you have worked in GMF mobile background selector, what do you recommend?

Related to https://github.com/c2corg/v6_ui/issues/67

tsauerwein:
The ngeo can be of use. To we have more detailed specifications on how the background-selector and the layer selector should look like?

asaunier:
Not really. I think that having a background selector dropdown (only one background at a time, no need to handle opacity) is OK. Please note that when changing the background layer we may have to update the map projection (for instance if using the swisstopo WMTS).

Not much instructions about the layer selector either. We don't really need a tree (?) and have not that many layers to display: the geographical document types such as WP (not sure if we need to be able to show WP types separately), routes, outings, images, areas, users, accidents, etc. + some extra layers such as the ">30° slopes" layer http://www.camptocamp.org/map?map_x=737760.64227729&map_y=5740502.0377603&map_zoom=14&tree_layers=slopes

Issue #67, opened 2015-12-18
Map background layers


asaunier:
In v5, maps are based upon Gmaps relief (by default), OSM, Swisstopo (for Switzerland) and IGN (for France).

In v6, we use OpenLayers3. Which does not come with GMaps layers, because of some conflicts with GMaps terms of use.

If we don't use Gmaps, we need to find an alternate default background layer. Either free such as OSM (but their are limitations concerning the number of tiles that can be loaded) or paying ones (the fee depending on the number of tiles that can be loaded).

Free (?) layers:
* OSM
* Esri maps: http://arcg.is/1TRMeJJ / http://blogs.esri.com/esri/arcgis/2011/02/22/use-arcgis-online-basemaps-for-free-in-your-gis-projects-and-web-applications/
* Bing maps (not sure "relief" maps are available)
* ?

Paying options:
* https://www.mapbox.com/blog/mapbox-outdoors/
* ??

Using an extension of OL3 for using Gmaps layers:
* http://www.mapgears.com/fr/blogue/archive/2015-10-23-ol3-google-maps

Please note that when switching among background layers, we might have to change the map projection (for instance when using Swisstopo WMTS).

tsauerwein:
ESRI base maps seem to be free for non-commercial sites:
https://github.com/leaflet-extras/leaflet-providers/issues/22#issuecomment-28003372

> Development and Testing Plans are generally intended for developing and testing your applications for ArcGIS Online. **However, you can also make your applications powered by your Development and Testing Plan available to the public, so long as you don’t monetize your application in any way**, and it's not used for the benefit of a business or government agency.
https://developers.arcgis.com/terms/

fjacon:
@asaunier  : Should we ask for confirmation ? What is the position of the client ?

asaunier:
I don't really know. I can ask.

asaunier:
Not much input from the client...

Any way we need the same IGN and Swisstopo baselayers than in v5 in addition to the OSM layer already available in v6:
* we need a baselaer selector (available in ngeo)
* we need to support reprojection (Swisstopo baselayers do not use EPSG:3857 as OSM, IGN, GMaps...)

Having Google Maps would be much appreciated but would require ol3-google-maps:
https://github.com/mapgears/ol3-google-maps/
Do you think it would cause problems to use it instead of a standard OL3?

tsauerwein:
[...]

PR #439, opened 2016-06-17
Baselayer selector

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/67 and https://github.com/c2corg/v6_ui/issues/125

Supported baselayers:
* [x] OSM
* [x] Stamen (for testing - will probably be removed)
* [x] Esri World Topo Map
 * [x] Handle capabilities loading error
* [x] IGN WMTS
 * [x] Maps
 * [x] Ortho?
 * [x] Use c2corg key
* [x] Swisstopo WMTS
 * [ ] ~~Reproject overlays because of the different projection~~

Also TODO:
* [ ] i18n
* [ ] permalink?
* [x] Make swisstopo baselayer only available to auth users (for quota reasons)

### DIFF
--- a/build.json
+++ b/build.json
@@ -3,6 +3,7 @@
     "node_modules/openlayers/src/**/*.js",
     "node_modules/openlayers/build/ol.ext/**/*.js",
     "node_modules/ngeo/src/**/*.js",
+    "node_modules/ngeo/contribs/gmf/src/**/*.js",
     "c2corg_ui/static/js/**/*.js",
     "c2corg_ui/static/build/templatecache.js"
   ],
@@ -43,7 +44,8 @@
     "js": [
       "node_modules/openlayers/externs/olx.js",
       "node_modules/openlayers/externs/oli.js",
-      "node_modules/ngeo/externs/ngeox.js"
+      "node_modules/ngeo/externs/ngeox.js",
+      "node_modules/ngeo/contribs/gmf/externs/gmfx.js"
     ],
     "jscomp_error": [
       "*"
--- a/c2corg_ui/static/js/map/baselayerselector.js
+++ b/c2corg_ui/static/js/map/baselayerselector.js
@@ -1,0 +1,263 @@
+goog.provide('app.BaselayerSelectorController');
+goog.provide('app.baselayerSelectorDirective');
+
+goog.require('app');
+goog.require('ngeo.BackgroundLayerMgr');
+goog.require('ol.Attribution');
+goog.require('ol.format.WMTSCapabilities');
+goog.require('ol.layer.Tile');
+goog.require('ol.source.OSM');
+goog.require('ol.source.WMTS');
+goog.require('ol.tilegrid.WMTS');
+
+
+/**
+ * This directive is used to display a map baselayer selector.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.baselayerSelectorDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppBaselayerSelectorController',
+    controllerAs: 'bglayerCtrl',
+    bindToController: true,
+    scope: {
+      'map': '=appBaselayerSelectorMap'
+    },
+    templateUrl: '/static/partials/map/baselayerselector.html'
+  };
+};
+
+
+app.module.directive('appBaselayerSelector', app.baselayerSelectorDirective);
+
+
+/**
+ * @param {angular.$http} $http
+ * @param {ngeo.BackgroundLayerMgr} ngeoBackgroundLayerMgr Background layer
+ *     manager.
+ * @param {string} ignApiKey IGN API key.
+ * @param {app.Alerts} appAlerts
+ * @param {app.Authentication} appAuthentication
+ * @constructor
+ * @export
+ * @struct
+ * @ngInject
+ */
+app.BaselayerSelectorController = function($http, ngeoBackgroundLayerMgr,
+  ignApiKey, appAlerts, appAuthentication) {
+
+  /**
+   * @type {angular.$http}
+   * @private
+   */
+  this.http_ = $http;
+
+  /**
+   * @type {ngeo.BackgroundLayerMgr}
+   * @private
+   */
+  this.backgroundLayerMgr_ = ngeoBackgroundLayerMgr;
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.ignApiKey_ = ignApiKey;
+
+  /**
+   * @type {app.Alerts}
+   * @private
+   */
+  this.alerts_ = appAlerts;
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map;
+
+  /**
+   * @type {Object}
+   * @private
+   */
+  this.cachedLayers_ = {};
+
+  /**
+   * @type {ol.format.WMTSCapabilities}
+   * @private
+   */
+  this.wmtsParser_;
+
+  /**
+   * @type {Object}
+   * @export
+   */
+  this.currentBgLayer;
+
+  /**
+   * @type {Array.<Object>}
+   * @export
+   */
+  this.bgLayers = app.BaselayerSelectorController.BACKGROUND_LAYERS.filter(
+    function(layer) {
+      return !('auth' in layer) || !layer['auth'] ||
+        (layer['auth'] && appAuthentication.isAuthenticated());
+    }
+  );
+
+  this.setLayer(this.bgLayers[0]);
+};
+
+
+/**
+ * @const
+ * @type {Array.<Object>}
+ */
+app.BaselayerSelectorController.BACKGROUND_LAYERS = [{
+  'name': 'osm'
+}, {
+  'name': 'esri'
+}, {
+  'name': 'ign maps'
+}, {
+  'name': 'ign ortho'
+}, {
+  'name': 'swisstopo',
+  'auth': true
+}];
+
+
+/**
+ * Function called when the user selects a new background layer in the
+ * dropdown. Called by the ng-click directive used in the partial.
+ * @param {Object} layerSpec Layer specification object.
+ * @export
+ */
+app.BaselayerSelectorController.prototype.setLayer = function(layerSpec) {
+  this.currentBgLayer = layerSpec;
+  var layer = this.createLayer_(layerSpec['name']);
+  if (layer) {
+    this.backgroundLayerMgr_.set(this.map, layer);
+  }
+};
+
+
+/**
+ * @param {string} layerName Layer name.
+ * @return {ol.layer.Tile} The layer.
+ * @private
+ */
+app.BaselayerSelectorController.prototype.createLayer_ = function(layerName) {
+  if (layerName in this.cachedLayers_) {
+    return this.cachedLayers_[layerName];
+  }
+  var source;
+  switch (layerName) {
+    case 'esri':
+      this.createWMTSLayerFromCapabilities_(
+        'https://server.arcgisonline.com/arcgis/rest/services/' +
+        'World_Topo_Map/MapServer/WMTS/1.0.0/WMTSCapabilities.xml', {
+          'layer': 'World_Topo_Map',
+          'matrixSet': 'GoogleMapsCompatible'
+        }, 'esri');
+      // layer creation is asynchronous
+      return null;
+    case 'ign maps':
+      source = this.createIgnSource_('GEOGRAPHICALGRIDSYSTEMS.MAPS');
+      break;
+    case 'ign ortho':
+      source = this.createIgnSource_('ORTHOIMAGERY.ORTHOPHOTOS');
+      break;
+    case 'swisstopo':
+      this.createWMTSLayerFromCapabilities_(
+        'http://wmts10.geo.admin.ch/EPSG/3857/1.0.0/WMTSCapabilities.xml', {
+          'layer': 'ch.swisstopo.pixelkarte-farbe',
+          'matrixSet': '3857',
+          'projection': 'EPSG:3857',
+          'requestEncoding': 'REST'
+        }, 'swisstopo');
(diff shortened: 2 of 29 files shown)


build.json
```diff
@@ -39,11 +44,6 @@
       "ol.ENABLE_DOM=false",
       "ol.ENABLE_WEBGL=false"
     ],
-    "js": [
-      "node_modules/openlayers/externs/olx.js",
-      "node_modules/openlayers/externs/oli.js",
-      "node_modules/ngeo/externs/ngeox.js"
```
asaunier: OK, I wasn't aware of that. I made the change because I had noticed it wasn't the case in agridea: https://github.com/camptocamp/agridea_geoacorda/blob/master/jsapi-buildtools/geoacorda.json
I will revert the change.

tsauerwein: In that example the symbol names were kept, because it is an API.

asaunier:
I have rebased the PR to make it a bit clearer.

@tsauerwein I still have problems with compiling the ``gmf``files despite the changes you suggested in ``build.json``. The ``make build`` fails saying that

> /home/alex/v6_ui/c2corg_ui/static/js/appmodule.js
  18:3  error  'gmf' is not defined  no-undef
/home/alex/v6_ui/c2corg_ui/static/js/map/baselayerselector.js
  173:20  error  'gmf' is not defined  no-undef

tsauerwein:
> The make build fails saying that ...

I tried your branch locally and it builds without problem. Can you try a `make clean` maybe?

tsauerwein:
Ah no, I see on Travis that it is ESLint that complains. Add `gmf` to the globals here:
https://github.com/c2corg/v6_ui/blob/master/.eslintrc#L3

asaunier:
OK the build passes now, thanks!

asaunier:
I had to revert some of my changes with 96d2563
Else the UI JS app fails in my browser saying that
> angular.js:38Uncaught Error: [$injector:modulerr] http://errors.angularjs.org/1.5.6/$injector/modulerr?p0=app&p1=Error%3A%20%…gis.internal%3A6555%2Fnode_modules%2F%2Fangular%2Fangular.min.js%3A21%3A19)

Since the "gmf" changes, closure compiles from 474 sources instead of 448. I guess that's the "gmf" files that get included in the built file? It's a bit a pity since I only need the ``gmf.source.Swisstopo`` file. Is it a big deal? Is there another way? Copying the ``gmf.source.Swisstopo`` file in the project files?

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,260 @@
...
+goog.require('ol.tilegrid.WMTS');
+
+
+/**
+ * This directive is used to display a map baselayer selector.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.baselayerSelectorDirective = function($compile, $templateCache) {
```
asaunier: Both arguments are not useful.

tsauerwein:
> Since the "gmf" changes, closure compiles from 474 sources instead of 448.

Your build will be compiled using more source files. But the build will only include what you are actually using. That's the nice thing about the Closure Compiler and one of the reasons we are doing it like that.

asaunier:
> Your build will be compiled using more source files. But the build will only include what you are actually using. That's the nice thing about the Closure Compiler and one of the reasons we are doing it like that.

Great! :)

asaunier:
Swisstopo tiles are also available with projection EPSG:3857:
https://api3.geo.admin.ch/services/sdiservices.html#other-projections
=> which means we would not have to handle a projection switch (all baselayers would be in EPSG:3857).

A few limitations though:
* Swisstopo 3857 tiles "are generated on-the-fly with MapProxy. If you plan to heavily use this service, please inform us in advance."
* We can no longer use ``gmf.source.Swisstopo`` to create the Swisstopo source since it relies on EPSG:21781. But I had already created a method for Esri maps that creates a source using the capabilities files.

asaunier:

What it does:
* add a baselayer selector ("dropup") in all maps
* available baselayers: OSM (default), Esri World Topo Map, IGN maps and orthos, Swisstopo maps
* Esri and Swisstopo maps are loaded using a capabilities file
* IGN baselayers requires a key that depends on the referer (for dev configs the "localhost" key is used by default - there's a key for the demo URL as well)
* Swisstopo baselayer is only available to auth users (because there is a yearly quota).

A few improvements/fix that would be nice (but I don't have much time to work more on this tool now) and that could be done in a additional PR:
* the Swisstopo EPSG:3857 baselayer is loaded parsing the capabilities file at https://wmts10.geo.admin.ch/EPSG/3857/1.0.0/WMTSCapabilities.xml Perhaps we could extend ``gmf.source.Swisstopo`` to "hard-code" the WMTS data. I'm just not sure what to use for instance for the tilegrid https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/source/swisstoposource.js#L22-L23 => reproject the extent?
* the baselayer names in the dropup are not correctly translated despite the translation are available in the PO and the dropup items are "x-translated": https://github.com/c2corg/v6_ui/blob/8694c39f00d56d4672c0d8a3c975c95f7d4ce886/c2corg_ui/static/partials/map/baselayerselector.html#L7
* add a permalink param to set the initial baselayer at page loading time? Anyway it's only a nice to have.

Please review :)

tsauerwein:
> the Swisstopo EPSG:3857 baselayer is loaded parsing the capabilities file at https://wmts10.geo.admin.ch/EPSG/3857/1.0.0/WMTSCapabilities.xml Perhaps we could extend gmf.source.Swisstopo to "hard-code" the WMTS data. I'm just not sure what to use for instance for the tilegrid https://github.com/camptocamp/ngeo/blob/master/contribs/gmf/src/source/swisstoposource.js#L22-L23 => reproject the extent?

Yes, it would make sense not to make an additional request to get the capabilities. If you want I can give it a try. Maybe we can do the same for the ESRI layer.

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+
+
+/**
+ * @const
+ * @type {Array.<Object>}
+ */
+app.BaselayerSelectorController.BACKGROUND_LAYERS = [{
+  'name': 'osm'
+}, {
+  'name': 'esri'
```
tsauerwein: If you are not working in the geospatial domain, you probably have no idea what esri is. What about 'Topographic Map' or 'Topo'?

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+
+  return new ol.source.WMTS({
+    url: '//wxs.ign.fr/' + this.ignApiKey_ + '/wmts',
+    layer: layer,
+    matrixSet: 'PM',
+    format: 'image/jpeg',
+    projection: 'EPSG:3857',
+    tileGrid: tileGrid,
+    style: 'normal',
+    attributions: [new ol.Attribution({
```
tsauerwein: Are there also attributions for the other layers?

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+  this.alerts_ = appAlerts;
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map;
+
+  /**
+   * @type {Object}
```
tsauerwein: The type could be `{Object<string, ol.layer.Base>}`.

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+  this.cachedLayers_ = {};
+
+  /**
+   * @type {ol.format.WMTSCapabilities}
+   * @private
+   */
+  this.wmtsParser_;
+
+  /**
+   * @type {Object}
```
tsauerwein: The type should be `ol.layer.Base`.

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+   * @export
+   */
+  this.currentBgLayer;
+
+  /**
+   * @type {Array.<Object>}
```
tsauerwein: `{Array.<ol.layer.Base>}`

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+   * @export
+   */
+  this.bgLayers = app.BaselayerSelectorController.BACKGROUND_LAYERS.filter(
+    function(layer) {
+      return !('auth' in layer) || !layer['auth'] ||
+        (layer['auth'] && appAuthentication.isAuthenticated());
```
tsauerwein: Couldn't this simply be:

    return layer['auth'] && appAuthentication.isAuthenticated();

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+app.BaselayerSelectorController.BACKGROUND_LAYERS = [{
+  'name': 'osm'
+}, {
+  'name': 'esri'
+}, {
+  'name': 'ign maps'
+}, {
+  'name': 'ign ortho'
+}, {
+  'name': 'swisstopo',
```
tsauerwein: What about swisstopo ortho?

c2corg_ui/static/js/map/baselayerselector.js
tsauerwein: Ah no, maybe the variable should be named `currentBgLayerSpec`.

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -0,0 +1,263 @@
...
+      var options = ol.source.WMTS.optionsFromCapabilities(result, config);
+      var layer = new ol.layer.Tile({
+        source: new ol.source.WMTS(options)
+      });
+      this.cachedLayers_[name] = layer;
+      this.backgroundLayerMgr_.set(this.map, layer);
+    }.bind(this),
+    // failure
+    function(response) {
+      this.alerts_.addError('Failed loading this map background');
```
tsauerwein: i18n?

c2corg_ui/static/js/map/baselayerselector.js
asaunier: I don't think so because you could have:
* no 'auth' property set => your test would evaluate to false I guess whereas if there is no 'auth' it means the layer is public
* it could be set to 'false' (meaning it's public)

c2corg_ui/static/js/map/baselayerselector.js
asaunier: Sure! But that's the job of the i18n to display human readable name. Here we use short labels to manipulate in the code.

c2corg_ui/static/js/map/baselayerselector.js
tsauerwein: Ah ok, I see:

    return appAuthentication.isAuthenticated() || !layer['auth']

c2corg_ui/static/js/map/baselayerselector.js
asaunier: It would make sense indeed (as the ign ortho). But since there is a yearly quota (when it's reached it's no longer possible to display swisstopo maps until the end of the year!) we have to be very careful and avoid having too many swisstopo layers. As a result I prefer having only the maps (as in v5), we'll see later if we can afford proposing the orthos.

c2corg_ui/static/js/map/baselayerselector.js
asaunier: Maybe. Actually I have used the same variable name than in ngeo's example at http://camptocamp.github.io/ngeo/master/examples/backgroundlayerdropdown.js

c2corg_ui/static/js/map/baselayerselector.js
tsauerwein: Talking about ortho fotos, didn't we want to use Bing Maps for global ortho fotos?

c2corg_ui/static/js/map/baselayerselector.js
asaunier: OK I'll do the change and rebase the PR to avoid the current conflicts.

asaunier:
> Yes, it would make sense not to make an additional request to get the capabilities. If you want I can give it a try. Maybe we can do the same for the ESRI layer.

That would be cool indeed. Feel free to do it if you have time. But that could also be done later.

c2corg_ui/static/js/map/baselayerselector.js
asaunier: Actually ``this.bgLayers`` is not an array of ``ol.layer.Base``, only of basic objects. See ``app.BaselayerSelectorController.BACKGROUND_LAYERS``.
Perhaps the names are confusing? I didn't want to use complicated names.

tsauerwein: `bgLayerSpecs`?

tsauerwein: > Perhaps the names are confusing? I didn't want to use complicated names.

Either you use a name that makes it clear if you are dealing with an ol layer or a layer spec. Or you type it properly. ;)

c2corg_ui/static/js/map/baselayerselector.js
asaunier: Yep, I have just rebased the PR, made most of the changes you have suggested (not the attributions yet) and added a Bing Maps bglayer.

asaunier:
I have added the attributions. Thanks for the reviewing @tsauerwein !

I totally agree the Esri and Swisstopo sources could be improved to avoid using the capabilities files but I don't have more time to improve the PR. This could be in a future PR.
OK for merging?

tsauerwein:
:+1:

PR #443, opened 2016-06-19
Files structure

asaunier:
"suggestion" and "card" partials are grouped in dedicated directories.

``make clean`` may be needed to remove the cached partials.


asaunier:
Any objections?

tsauerwein:
:+1: 

Thanks for cleaning up!

asaunier:
Thanks @tsauerwein for the reviews!

Issue #374, opened 2016-05-26
Duplicated property in appx.SimpleSearchDocument?


asaunier:
I see that type `appx.SimpleSearchDocument` has both properties `document_type` and `documentType`:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js#L38-L47

`document_type` seems not used in the simple search controller:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/simplesearch.js

Perhaps it can be removed?

gberaudo:
Keeping the version with the underscore would be consistent with the other properties.

asaunier:
Perhaps. But the first question to answer: do we really need to have both properties?

gberaudo:
@asaunier, obviously we only want one.

asaunier:
I guess so :)
OK reformulating my question: why do we have those 2 properties ? what can we do to have only one? where should we made changes? who will take care of this?

gberaudo:
These two properties were introduced by @tsauerwein in f06fcca8.
I looked quickly to the commit and it might be a mistake since only one property is used.
@tsauerwein do you have some input about this?

tsauerwein:
If I check the actual response of the simple search both definitions are wrong:
http://c2corgv6-api.demo-camptocamp.com/search?q=mon&pl=en

There is only `type`.

gberaudo:
What a `coup de théâtre`!

asaunier:
I have tried to remove ``document_type`` and ``documentType`` and to replace them by ``type`` but it doesn't work (no compilation error but the simple search returns blank results). Ditto if only ``document_type``. It works if only ``documentType``. No idea why. See https://github.com/c2corg/v6_ui/pull/444

stef74:
@asaunier this question is allready relevant ?

ginold:
I thought i had already answered this question: `documentType` was set in the times when `type` was not returned by the API yet, so we have to track it down and remove it from it's origin!

PR #444, opened 2016-06-19
Remove duplicated property in simple search externs

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/374


asaunier:
Any objections?

tsauerwein:
The simple search [returns](http://c2corgv6-api.demo-camptocamp.com/search?q=mon&pl=en) `type` so your change does not look correct?

asaunier:
Perhaps, but as I said it doesn't seem to work with ``type``, only with ``documentType``!?

ginold:
very very likely `documentType` has been created long time ago by me, when `type` wasn't returned yet and probably there are some functions that still use `documentType`.

PR #454, opened 2016-06-24
Avoid parsing capabilities for background layers

tsauerwein:
Instead of requesting the capabilities document for the Swisstopo and Esri layer, the config is hard coded to avoid the request.

### DIFF
--- a/c2corg_ui/static/js/map/baselayerselector.js
+++ b/c2corg_ui/static/js/map/baselayerselector.js
@@ -4,11 +4,11 @@ goog.provide('app.baselayerSelectorDirective');
 goog.require('app');
 goog.require('ngeo.BackgroundLayerMgr');
 goog.require('ol.Attribution');
-goog.require('ol.format.WMTSCapabilities');
 goog.require('ol.layer.Tile');
 goog.require('ol.source.BingMaps');
 goog.require('ol.source.OSM');
 goog.require('ol.source.WMTS');
+goog.require('ol.source.XYZ');
 goog.require('ol.tilegrid.WMTS');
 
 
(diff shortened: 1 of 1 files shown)


c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -285,4 +232,42 @@ app.BaselayerSelectorController.prototype.createIgnSource_ = function(layer) {
...
+ */
+app.BaselayerSelectorController.prototype.createSwisstopoSource_ = function(layer) {
+  return new ol.source.XYZ({
+    attributions: [
+      new ol.Attribution({
+        html: '<a target="_blank" href="http://www.swisstopo.admin.ch/' +
+        'internet/swisstopo/en/home.html">swisstopo</a>'
+      })
+    ],
+    url: 'http://wmts20.geo.admin.ch/1.0.0/' + layer + '/default/current' +
```
asaunier: We'd better use https tiles (the UI is supposed to work with https URLS).

tsauerwein: Good point!

c2corg_ui/static/js/map/baselayerselector.js
```diff
@@ -285,4 +232,44 @@ app.BaselayerSelectorController.prototype.createIgnSource_ = function(layer) {
...
+ */
+app.BaselayerSelectorController.prototype.createSwisstopoSource_ = function(layer) {
+  return new ol.source.XYZ({
+    attributions: [
+      new ol.Attribution({
+        html: '<a target="_blank" href="http://www.swisstopo.admin.ch/' +
+        'internet/swisstopo/en/home.html">swisstopo</a>'
+      })
+    ],
+    urls: ['10', '20'].map(function(i) {
```
asaunier: The doc says that 
> You have to use the hosts wmts{10-14}.geo.admin.ch. This is done to avoid parsing every requests to determine which are using native tiles and which are using reporjected tiles.
See https://api3.geo.admin.ch/services/sdiservices.html#other-projections

asaunier:
Was it that simple!? :P 

Except my comment/question about the swisstopo subdomains (?), it looks really good to me, thanks for the great improvement! :)

c2corg_ui/static/js/map/baselayerselector.js
tsauerwein: Thanks, I missed that in the docs!

tsauerwein:
> Was it that simple!? :P

After all, yes. But it also took me a moment to realize... ;)

PR #384, opened 2016-06-01
[WIP] integrate image backend

gberaudo:
In this first part, upload images to image backend:
- send files to backend;
- have them thumbnailed and uploaded to S3 incoming;
- receive filename.

TODO:
- prevent http service spinner to block UI;
- have a dedicated hostname for backend;
- configure proxy to 8480?;

PR #455, opened 2016-06-24
Post images to API service

arnaud-morvan:
Superseed #384

TODO:
* check required fields before post to API (image_type)
* cancel all upload requests when user manually close the dialog

### DIFF
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -84,18 +84,6 @@ app.ImageUploaderController = function($scope, $uibModal, $compile, $q, appAlert
   this.q_ = $q;
 
   /**
-   * @type {Array.<angular.$q.Promise>}
-   * @export
-   */
-  this.uploading = [];
-
-  /**
-   * @type {Array.<!angular.$q.Deferred>}
-   * @export
-   */
-  this.cancellers = [];
-
-  /**
    * @type {Array.<File>}
    * @export
    */
@@ -204,8 +192,8 @@ app.ImageUploaderController.prototype.upload_ = function() {
         file['progress'] = 100 * progress;
         console.log('progress', 100 * progress, file['metadata']['id']);
       }.bind(this, file));
-      this.uploading[i] = promise;
-      this.cancellers.push(canceller);
+      file.uploading = promise;
+      file.canceller = canceller;
 
       promise.then(function(resp) {
         console.log('100% uploaded! ' + file['metadata']['title'] + ' ' + i);
@@ -224,8 +212,12 @@ app.ImageUploaderController.prototype.upload_ = function() {
  * @private
  */
 app.ImageUploaderController.prototype.areAllUploadedCheck_ = function(interval) {
-  this.q_.all(this.cancellers).then(function(res) {
-    if (this.files.length > 0 && (this.files.length === this.cancellers.length || this.cancellers.length >= this.cancellers.length)) {
+  var uploading = this.files.map(function(file) {
+    return file.uploading;
+  });
+
+  this.q_.all(this.uploading).then(function(res) {
+    if (this.files.length > 0) {
       this.areAllUploaded = true;
       clearInterval(interval);
     } else {
@@ -278,10 +270,7 @@ app.ImageUploaderController.prototype.save = function() {
  * @export
  */
 app.ImageUploaderController.prototype.abortUploadingImage = function(index) {
-  this.uploading.splice(index, 1);
-  this.files.splice(index, 1);
-  var canceller = this.cancellers.splice(index, 1)[0];
-  canceller.reject(); // cancel
+  this.files[index].canceller.resolve();
 };
 
 
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -40,13 +40,12 @@ app.module.directive('appImageUploader', app.imageUploaderDirective);
  * @param {angular.$q} $q promise
  * @param {app.Api} appApi Api service.
  * @param {app.Alerts} appAlerts
- * @param {Object} Upload ng-file-upload Upload service.
  * @param {app.Document} appDocument service
  * @constructor
  * @struct
  * @ngInject
  */
-app.ImageUploaderController = function($scope, Upload, $uibModal, $compile, $q, appAlerts, appApi, appDocument) {
+app.ImageUploaderController = function($scope, $uibModal, $compile, $q, appAlerts, appApi, appDocument) {
 
   /**
    * @type {app.Document}
@@ -83,11 +82,6 @@ app.ImageUploaderController = function($scope, Upload, $uibModal, $compile, $q, 
    * @private
    */
   this.q_ = $q;
-
-  /**
-   * @private
-   */
-  this.Upload_ = Upload;
 
   /**
    * @type {Array.<angular.$q.Promise>}
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -41,11 +41,12 @@ app.module.directive('appImageUploader', app.imageUploaderDirective);
  * @param {app.Api} appApi Api service.
  * @param {app.Alerts} appAlerts
  * @param {app.Document} appDocument service
+ * @param {String} imageUrl URL to the image backend.
  * @constructor
  * @struct
  * @ngInject
  */
-app.ImageUploaderController = function($scope, $uibModal, $compile, $q, appAlerts, appApi, appDocument) {
+app.ImageUploaderController = function($scope, $uibModal, $compile, $q, appAlerts, appApi, appDocument, imageUrl) {
 
   /**
    * @type {app.Document}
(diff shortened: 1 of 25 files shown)


c2corg_ui/templates/outing/view.html
```diff
@@ -49,6 +49,9 @@
   });
 
   module.value('documentData', {
+    "document_id": ${dumps(outing.get('document_id')) | n},
```
tsauerwein: Do you need to call `dumps`?

c2corg_ui/static/js/imageuploader.js
```diff
@@ -204,16 +203,22 @@ app.ImageUploaderController.prototype.upload_ = function() {
...
       this.getImageMetadata_(file);
-      this.uploading[i] = this.api_.uploadImage(file);
-      this.promises.push(this.uploading[i]);
 
-      this.uploading[i].then(function(resp) {
+      var canceller = this.q_.defer();
+      var promise = this.api_.uploadImage(file, canceller.promise, function(file, event) {
+        var progress = event.loaded / event.total;
+        file['progress'] = 100 * progress;
+        console.log('progress', 100 * progress, file['metadata']['id']);
```
tsauerwein: debug stmt

c2corg_ui/static/js/imageuploader.js
```diff
@@ -204,16 +203,22 @@ app.ImageUploaderController.prototype.upload_ = function() {
...
+      this.cancellers.push(canceller);
+
+      promise.then(function(resp) {
         console.log('100% uploaded! ' + file['metadata']['title'] + ' ' + i);
       }.bind(this), function(resp) {
         this.alerts_.addError('error while uploading the image ' + resp);
-      }.bind(this), function(evt) { // handled by defer.notify()
-          //var progressPercentage = parseInt(100.0 * evt.loaded / evt.total, 10);
-      });
+        this.files.splice(this.files.indexOf(file), 1);
```
tsauerwein: Do you also need to remove the entry in `this.uploading` and `this.cancellers`? Or what is the purpose of removing the file from `this.files`?

c2corg_ui/static/js/imageuploader.js
```diff
@@ -226,8 +231,8 @@ app.ImageUploaderController.prototype.upload_ = function() {
  * @private
  */
 app.ImageUploaderController.prototype.areAllUploadedCheck_ = function(interval) {
-  this.q_.all(this.promises).then(function(res) {
-    if (this.files.length > 0 && (this.files.length === this.promises.length || this.promises.length >= this.promises.length)) {
+  this.q_.all(this.cancellers).then(function(res) {
+    if (this.files.length > 0 && (this.files.length === this.cancellers.length || this.cancellers.length >= this.cancellers.length)) {
```
tsauerwein: `this.cancellers.length >= this.cancellers.length` is always true?!

c2corg_ui/static/js/imageuploader.js
```diff
@@ -270,21 +275,23 @@ app.ImageUploaderController.prototype.save = function() {
...
- * TODO: REALLY cancel an image being sent? promise.resolve?
  * @param {number} index
  * @export
  */
 app.ImageUploaderController.prototype.abortUploadingImage = function(index) {
   this.uploading.splice(index, 1);
   this.files.splice(index, 1);
-  this.promises.splice(index, 1);
-  this.api_.abortUploadingImage(index);
+  var canceller = this.cancellers.splice(index, 1)[0];
```
tsauerwein: Do you also need to remove it from `this.promises`?

c2corg_ui/static/js/imageuploader.js
arnaud-morvan: Here I remove the file from files because we don't want to post an image on api that is not correctly uploaded on image backend.

less/images.less
```diff
@@ -75,7 +75,15 @@
         }
       }
     }
-
+    .dropdown-menu > li > div.radio {
+      display: block;
+      padding: 3px 20px;
+      clear: both;
+      font-weight: normal;
+      line-height: 1.42857143;
```
ginold: that's what i call swiss precision :watch:

arnaud-morvan: This comes from a copy paste from css in sibling element ( .dropdown-menu > li > a ).

arnaud-morvan:
@ginold : I've rebased and make some finishes on imageuploader, can you make a last review to merge this for alpha2.

c2corg_ui/static/js/imageuploader.js
```diff
@@ -204,16 +192,27 @@ app.ImageUploaderController.prototype.upload_ = function() {
...
+      promise.then(function(resp) {
+        file['metadata']['filename'] = resp['data']['filename'];
       }.bind(this), function(resp) {
-        this.alerts_.addError('error while uploading the image ' + resp);
-      }.bind(this), function(evt) { // handled by defer.notify()
-          //var progressPercentage = parseInt(100.0 * evt.loaded / evt.total, 10);
-      });
+        if (resp.status == -1) {
+          if (!file['manuallyAborted']) {
+            this.alerts_.addError('error while uploading the image : Timeout');
```
tsauerwein: @asaunier or @ginold: Must i18n be handled here?

asaunier: The message probably has to be added to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/i18n.html#L132-L135 in order to be extracted and added to the PO files.

c2corg_ui/templates/outing/view.html
```diff
@@ -50,6 +50,9 @@
   });
 
   module.value('documentData', {
+    "document_id": ${outing.get('document_id')},
+    "lang": ${dumps(lang) | n},
```
arnaud-morvan: Yes ``lang`` is a string and it should be escaped using ``dumps``, not ``html_entities`` which is the default from the template engine.

asaunier: On the other hand, "lang" is always a 2-letter code. Do we really need to
escape it?

2016-06-30 11:10 GMT+02:00 Arnaud Morvan <notifications@github.com>:

arnaud-morvan: I agree that this is not really necessary for "lang". IMHO, it is a good practice to always use the good escaping method, if dumps has nothing to do, it will do nothing. HTML templating always do html entities escaping, even if it is not visible in the template, this is the default. I don't see any filter for json in mako documentation, but it should be possible to create one, example : ``${lang | json}``.

arnaud-morvan:
@ginold : do you think we could merge now this for the alpha2.

c2corg_ui/views/document.py
```diff
@@ -32,7 +32,9 @@ def __init__(self, request):
             'debug': 'debug' in self.request.params,
             'api_url': self.settings['api_url'],
             'ign_api_key': self.settings['ign_api_key'],
-            'bing_api_key': self.settings['bing_api_key']
+            'bing_api_key': self.settings['bing_api_key'],
+            'image_backend_url': self.settings['image_backend_url'],
+            'image_url': self.settings['image_url']
```
asaunier: Perhaps you have to add those 2 mako variables also to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/__init__.py#L95-L100 for the 404 page. You don't really need them in the 404 page but ``templates/404.html`` also inherits from ``base.html`` https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/404.html#L1

Try to add whatever string to your instance URL and check that the 404 page is not broken.
Normally ``make check`` should report the problem as well I think.

arnaud-morvan: These variables are already passed to template in notfound view. 404 page works as expected.

asaunier: My mistake, it's at the beginning of the diff!

ginold:
sorry for the late review. Here are some remarks: 

- The progress just appears for a moment and then you have to wait until the spinning gif disappears (which doesn't let to click 'cancel' because of the gray/semi-opaque modal). Would be better if the progress bar was as long as the real waiting time.

- with the new `imageUlr`, the images don't appear anymore (but this is normal because the images haven't been migrated yet?)

- activities and categories values  and other metadatas (aperture, shutter speed etc) are lost after reloading the page. They appear first in the template, but on a reload they are lost. 

- I cannot post multiple images at once, only one image at a time ```
(POST 500 Internal Server Error, File "/home/agina/v6_api/c2corg_api/views/image.py", line 73, in validate_list_image_create
    for image in request.validated['images']:
KeyError: 'images' )
```

- even though upload fails, the images will appear in the gallery.

Other than that, It looks like an awesome piece of work, keep going! :sunglasses:

arnaud-morvan:
I've make a pull request to fix the post multiple image bug on the api side.

arnaud-morvan:
The progress bar appears when the file is uploading to the backend. After that, the request is threated on server side which create thumbnails, etc. At the end we receive the response.
But this is not really related to this pull request.
Can we open a ticket for dicussion about this.

arnaud-morvan:
Concerning activities, categories, this is not related to this pull request. Can we open a ticket for that.

ginold:
So, if now you can upload multiple images and fields + progress bar are not related to this PR, you can look at the 2 last comments of mine and we're good to go.

PR #457, opened 2016-06-25
[WIP] Loading indicator for auth actions

asaunier:
As for now when authenticating/registering/resetting/changing the account data, there is nothing indicating something is going on in the background.

This PR adds the `app-loading` directive to the forms of those actions. It works well for login and account changing but causes an error with registering:

> Cannot set property 'lang' of undefined
>     at app.AuthController.register (auth.js:211)

and also for password resetting:

> Cannot read property 'email' of undefined
>     at app.AuthController.requestPasswordChange (auth.js:232)

No pb if I remove `app-loading` from the related forms... 
@ginold @gberaudo would you have an idea where it comes from.

Adding a loading indicator when logging out would also make sense but I cannot use the `app-loading` directive since the logout button directly calls a controller method. Perhaps we should add some code to this method as in the `app-loading` directive https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/loading.js#L25-L31 ?

### DIFF
--- a/c2corg_ui/templates/account.html
+++ b/c2corg_ui/templates/account.html
@@ -6,7 +6,7 @@
   <section>
     <div ng-controller="appAccountController as accountCtrl">
       <h1 translate>Change account parameters</h1>
-      <form name="accountForm" novalidate ng-submit="accountCtrl.save()">
+      <form name="accountForm" novalidate ng-submit="accountCtrl.save()" app-loading>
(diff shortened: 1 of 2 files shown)


ginold:
as @gberaudo once stated to me : the `app-loading `directive uses the `$http` service which is global, therefore it should not be needed to write this attribute on every element where a request can happen - it should be enough to declare it once, anywhere.

asaunier:
See also https://github.com/c2corg/v6_ui/issues/593#issuecomment-258136632

PR #458, opened 2016-06-27
[WIP] get document-related comments from discourse

ginold:
So what I have achieved until now is the possibility to get a forum thread with its `document_id` and `lang` (`topic_slug` in JSON), for example getting :
http://c2corgv6-discourse.demo-camptocamp.com/t/262782-fr 
will redirect to :
http://c2corgv6-discourse.demo-camptocamp.com/t/262782-fr/108763 
with the last part being the topic ID.

If there is no data returned for the combination `document_id`-`lang`, the user is suggested to create a new thread. Some problems emerge:
- You cannot change the topic title, which is used to look for the contents and get the topic id. This is not viable - users cannot have some numbers as topic title !
- The first main comment created does not appear in the appended comments, although it is returned in the JSON. Only the responses for the main comment will appear.

We should be able to save the topic id for each document-view-page on create the thread so that I can directly get the thread without first getting `document_id`-`lang` and then the thread, if it exists. 
This operation takes a while imo.

### DIFF
--- a/c2corg_ui/static/js/constants.js
+++ b/c2corg_ui/static/js/constants.js
@@ -37,5 +37,6 @@ app.constants = {
     FORM_PROJ: 'EPSG:4326',
     DATA_PROJ: 'EPSG:3857'
   },
+  discourseURL: 'http://c2corgv6-discourse.demo-camptocamp.com/',
   imagesURL : 'http://s.camptocamp.org/uploads/images/'
 };
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -325,7 +325,8 @@
       <p class="float-button-text" translate>Favorites</p>
     </div>
     
-    <div class="float-button float-comments" tooltip-placement="left" uib-tooltip="{{'Comments' | translate}}">
+    <div class="float-button float-comments" tooltip-placement="left" uib-tooltip="{{'Comments' | translate}}" ng-if="!detailsCtrl.createNewTopic"
+         ng-click="detailsCtrl.scrollToComments()">
       <span class="glyphicon glyphicon-comment"></span>
       <p class="float-button-text" translate>Comments</p>
     </div>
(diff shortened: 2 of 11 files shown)


c2corg_ui/static/js/constants.js
```diff
@@ -37,5 +37,6 @@ app.constants = {
     FORM_PROJ: 'EPSG:4326',
     DATA_PROJ: 'EPSG:3857'
   },
+  discourseURL: 'http://c2corgv6-discourse.demo-camptocamp.com/',
```
asaunier: You cannot hardcode URLs in the JS files (ditto for ``imagesURL`` below) because you cannot change those URLs without making a new commit. You have to use a ``module.value`` and inject it in your JS controllers. See for instance ``apiUrl`` https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L143

c2corg_ui/templates/helpers/view.html
```diff
@@ -572,3 +573,24 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
...
+    <span class="lead">
+      <section>
+        <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#discourse-comments">
+          <span class="glyphicon glyphicon-comment"></span>
+          <span translate>Comments</span>
+          <span class="glyphicon glyphicon-menu-down"></span>
+        </h3>
+        <section id="discourse-comments" class="collapse in" ng-init="detailsCtrl.getComments('${id}-${lang}')">
+          <div ng-if="detailsCtrl.createNewTopic" class="add-outing">
+            <p class="text-center">No thread yet?</p>
```
asaunier: Missing ``translate``?

c2corg_ui/templates/helpers/view.html
```diff
@@ -572,3 +573,24 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
...
+            <button protected-url-btn url="http://c2corgv6-discourse.demo-camptocamp.com/new-topic?title=${id}_${lang}&category=commentaires du topoguide/comments" class="btn gray-btn"  translate>Post the first comment</button>
```
asaunier: Hardcoded URL!

asaunier:
> You cannot change the topic title, which is used to look for the contents and get the topic id. This is not viable - users cannot have some numbers as topic title !

I am not sure this is the real problem :P 
In v5 this works the same but no one complains because... no one this the comments in the forum interface.

Isn't the real problem that people are reading comments embedded in the document itself and when they want to reply, they are redirected to the forum? Then how do they come back to the document?

ginold:
>     I am not sure this is the real problem :P 
Well, in v5 it wasn't any problem because people didn't see the this combination, but here - they do and it it's a problem, probably the major problem and reason of not wanting to implement it this way.

According to Discourse, it's not possible to respond to a thread from embedded comments.

 How do they come back to the document? they will have to click "back" :P I don't know any better options yet.

asaunier:
> How do they come back to the document? they will have to click "back" :P I don't know any better options yet.

I think we should make the client aware of this because it's not really user-friendly.

asaunier:
It seems this PR can be closed (and its branch deleted) since it has been replaced by https://github.com/c2corg/v6_ui/pull/486

@ginold @arnaud-morvan can you confirm?

PR #460, opened 2016-06-27
Advanced search step 3

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/368
Sequel of https://github.com/c2corg/v6_ui/pull/373

I have merged PR https://github.com/c2corg/v6_ui/pull/373 because it was getting too big and will create a new branch/PR.

Still to be addressed:

TODO
* [x] add the missing route filters
* [x] outing filters
* [ ] ~~Skip the map extent when filtering on a WP or a route ID? (with permalink like ``outings?r=123456``)~~ see https://github.com/c2corg/v6_ui/issues/471
* [x] Handle date filters
* [x] Handle simple checkbox
* [x] show orientations filter for mobile screens as well

TOFIX
* [ ] slider handles are not fully reset when clearing filters

EXTRA
* [ ] Hilight the feature on the map when hovering a card (and vice-versa?)
* [ ] allow to bypass the map extent filtering as in Airbnb?

### DIFF
--- a/c2corg_ui/templates/outing/filters.html
+++ b/c2corg_ui/templates/outing/filters.html
@@ -26,6 +26,28 @@ from c2corg_common.attributes import default_langs, quality_types, activities, f
     <div class="row collapse" id="moreFilters">
 
       <div class="col-xs-12 col-sm-6 filter">
+        <label translate>date_start</label><br>
+        <div class="input-group" ng-model="filtersCtrl.dates[0]"
+             uib-datepicker-popup="dd MM yyyy" is-open="openDateStart" ng-change="filtersCtrl.setDate('date')">
+          <input type="text" disabled class="form-control" value="{{filtersCtrl.dates[0] | amDateFormat:'Do MMMM YYYY'}}" required/>
+          <span class="input-group-btn">
+            <button type="button" class="btn btn-default" ng-click="openDateStart = true"><span class="glyphicon glyphicon-calendar"></span></button>
+          </span>
+        </div>
+      </div>
+
+      <div class="col-xs-12 col-sm-6 filter">
+        <label translate>date_end</label><br>
+        <div class="input-group" ng-model="filtersCtrl.dates[1]"
+             uib-datepicker-popup="dd MM yyyy" is-open="openDateEnd" ng-change="filtersCtrl.setDate('date')">
+          <input type="text" disabled class="form-control" value="{{filtersCtrl.dates[1] | amDateFormat:'Do MMMM YYYY'}}" required/>
+          <span class="input-group-btn">
+            <button type="button" class="btn btn-default" ng-click="openDateEnd = true"><span class="glyphicon glyphicon-calendar"></span></button>
+          </span>
+        </div>
+      </div>
+
+      <div class="col-xs-12 col-sm-6 filter">
         <label translate>elevation_max</label><br>
         <app-slider filter="oalt" filters-list="filtersCtrl.filters" max="8850"></app-slider>
       </div>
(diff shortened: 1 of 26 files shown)


asaunier:

Some points are still open:
* Hilight the feature on the map when hovering a card (and vice-versa?)
￼* allow to bypass the map extent filtering as in Airbnb?

as well as 
> Skip the map extent when filtering on a WP or a route ID? (with permalink like outings?r=123456)

for which I have created a dedicated issue: https://github.com/c2corg/v6_ui/issues/471

But I won't have time to work on them before alpha2.

There is also a small bug I have not been able to fix:
> slider handles are not fully reset when clearing filters

It's reproducible with master:
* open http://c2corgv6.demo-camptocamp.com/waypoints?bbox=-676234%2C3425519%2C1476233%2C7774480&walt=1200%2C8850
* press the "Clear filters" button => filters are actually removed, the displayed values of the "elevation" slider are actually reset but the handles are not moved to the ends of the slider.

In https://github.com/c2corg/v6_ui/blob/advanced-search-step3/c2corg_ui/static/js/search/slider.js#L224-L227 I have tried to set by hand the values of the slider widget:
```
this.element_.find('.range-between').slider({'values': this.boundaries});
```
but with no effect!

Anyway it's a non-critical bug and it could be fixed later.

ginold:
If you select all the activities, those boxes will go outside the screen.
screenshot from 2016-06-30 11 03 18

You could simply move them up.

asaunier:
Thanks for reporting. I don't think the language is the most important criteria, I prefer to keep where it is now placed. You generally don't have to select all activities and have all the criterias displayed. We'll see later if this criteria (or others) should be moved at some other place.

c2corg_ui/templates/outing/filters.html
```diff
@@ -1,104 +1,145 @@
...
-        <span class="slider-min-max">
-          <p class="min-value">{{searchFiltersCtrl.elevation.min}} m</p>
-          <p class="max-value">{{searchFiltersCtrl.elevation.max}} m</p>
-        </span>
+
+      <div class="col-xs-12 col-sm-6 filter">
+        <label translate>public_transport</label><br>
+        <a ng-click="filtersCtrl.toggleCheckbox('owpt')">
+          <input type="checkbox" ng-checked="filtersCtrl.filters.owpt">
+          <span></span>
```
ginold: what a hack :P I thought you had adapted the style!

asaunier: Nope I have done it the standard way :)

asaunier:
Any objection against merging or tip about my slider resetting problem?

c2corg_ui/templates/outing/filters.html
asaunier: Do you think it is really a problem to use this way to display a lone checkbox with no text? Since there is no text, the ``<span>`` is logically empty and is anyway required by the CSS selector for the checkbox styles https://github.com/c2corg/v6_ui/blob/master/less/c2corg_ui.less#L899-L928

tsauerwein: Does it look like in the the [screenshot](https://github.com/c2corg/v6_ui/issues/407#issuecomment-229602341) you posted? The box of the checkbox looks a bit lost for my taste. Either put box and label on the same line. Or add a description right of the box, something like "Uses public transport".

asaunier: Yes, it's like in the screenshot.
I would prefer keeping the checkbox as is. I have already tried to put the checkbox and label on the same line but it looks weird since this criteria is misaligned with the others. I am not sure adding a new text is needed (and it will add one more string to translate among millions). Perhaps the current label should be updated to something more explicit.

asaunier:
I have rebased the PR and fixed a conflict.
I will merge the PR soon if no one objects in the meantime (I'll be away for 2 weeks in a couple of hours so it's quite urgent).

tsauerwein:
Go for it! Thanks :)

asaunier:
Yoohoo :)
Thanks!

Issue #422, opened 2016-06-14
Association user 


celinestern:
- [x] Impossible to save modification with adding a other user to an outing (even with activate champ "Courte description des changements", error message:
    Required
    Required
at least one user required 
at least one route required"

ginold:
see #409

loicperrin:
Another problem with user association in outing: in the list of proposed users, if I click on "Plus de résultats", a page "oops" appears.

Issue #420, opened 2016-06-14
Text formatting issue


stef74:
http://c2corgv6.demo-camptocamp.com/routes/265151/fr

List are not well formatted ( no return chart on list item)

V5 reference : http://www.camptocamp.org/routes/265151/fr/route-du-soleil-col-du-simplon-realp

fjacon:
You just need to add a line break (done on demo server)
is that OK ?

stef74:
@fjacon yes for new document ... but for old document how to manage (migration ?) For me there too many document impacted

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/131

asaunier:
I close this issue since it's already part of #131

PR #463, opened 2016-06-28
outing fixes #410 #420

### LINKED ISSUE #410: Smartphone : problems when creating an outing

ginold:
- language + search field take 100% width on phone
- don't show 'show more' btn if searching for users in outing editing

related to #410
related to #422


ginold:
@arnaud-morvan, could you have a look on this PR plase?

less/documentedit.less
```diff
@@ -596,3 +610,7 @@ input[type="radio"] {
   color: @C2C-orange !important;
   transition: 0.2s ease-in-out;
 }
+
+#participants-group {
+  margin-top: 10px;
+}
```
arnaud-morvan: empty line is missing at end of file

arnaud-morvan:
LGTM

Issue #433, opened 2016-06-16
Display Outing / route / Waypoints


loicperrin:
- [x] The table "Enneigement par zone" is displayed even if it is empty and even in activity for which it is not relevant (see outings/737006/fr or outings/737036/fr).
- [x] Smartphone : All blocs should be displayed without having to clic on "Détails", "Association", ...
See wireframes : https://projects.invisionapp.com/share/MK53LLARJ#/screens/108058201. It should be better to open each bloc by default.
- [x] Smartphone : On the route page, the associated outings are visible only in a small bloc in which it is difficult to scroll.

desnoes:
> toto

desnoes:
## OUTINGS

- [ ] sections (Itinéraire emprunté, Commentaires sur le refuge...) should not be hidden when the document is opened. See http://c2corgv6.demo-camptocamp.com/outings/737032/fr.

- [ ] Associated WP should be classified and grouped accoridng to their type (parking, summit,...)

- [ ] Display on the map the pictos of the associated WP with the possibility to open this WP by clicking on the picto (cf. V5)

- [ ] List of participants: increase spacing between names

- [ ] Title: reduce the font size. For instance in the table for snow conditions, titles are written in 2 lines because the font size is too large (ex: http://c2corgv6.demo-camptocamp.com/outings/736972/fr)

- [ ] The link with weather forecast is missing

- [ ] Cotations should be displayed (at the top just under the title of the outing)

stef74:
- [ ] In information block remove activity "name" and replace by pinto

stef74:
## ROUTE

- [ ] place abstract in dedicated block

stef74:
## waypoints

- [ ] please

 show the waypoints 'name' not 'point de passage' on the title of WP.

asaunier:
> Unit: no space between the altitude and 'm': write '545m' instead of '545 m'

There is a space between the units, at least in french:
https://fr.wikipedia.org/wiki/Wikip%C3%A9dia:Conventions_typographiques#Unit.C3.A9s_de_mesure
> Les symboles d’unités de mesure ne comportent pas de point abréviatif, sont habituellement écrits en minuscules et sont séparés du nombre par une espace insécable.

> The link with weather forecast is missing

What weather forecast? Where is this point described?
What about routes? It's far more useful to have a weather forecast for routes document then for outings (since the latter can be outdated and not matching the weather forecast).

> Cotations should be displayed (at the top just under the title of the outing)

[...]

PR #464, opened 2016-06-28
Remove tab system on mobile (detail-view)

ginold:
related to #433 

> All blocs should be displayed without having to clic on "Détails", "Association",


asaunier:
I have not tested the PR in my instance but the changes look good to me.

ginold:
merging if no one minds.

Issue #413, opened 2016-06-13
Prevent zoom on map


stef74:
On topoguide,  document have map on top ... this map are zoomed too often until instead go down on the document. It' little bit annoying
How to prevent it ? with click on map before zooming ? no full large screen map ? other ?

ginold:
We have implemented "smart scrolling" because of this problem : when you scroll from the top to the bottom and your mouse cursor will go through the map, it would unintentionally zoom in, which was annoying. But not it will not zoom until it stays for 0.5 sec on the map. So just make sure your pointer is not on the map while scrolling, if not it is actually supposed to zoom in.

stef74:
I think 0.5s is too small. Could you try with 2s.

ginold:
I changed to 1s, 2 seconds is super long.

PR #466, opened 2016-06-29
longer timeout for map smart-scroll

ginold:
changed from 500ms to 1000ms.
2 seconds suggested by the client is way too long. 
related to #413


tsauerwein:
Looks good, please merge when Travis agrees.

PR #468, opened 2016-06-29
conditions levels + public_transport fix

ginold:
- conditions_levels will not show up if they are empty or null. 
- Edition and addition works in editing mode.
- Changed the style a little.

- added the correct type for public_transport in outing editing
related to #407

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -29,7 +29,7 @@
 <%def name="show_other_langs_links(module, document, other_langs)">\
   % if other_langs:
     <div id="other-langs">
-      <div class="modal-header"><h3 class="text-center" translate>View in other lang</h3></div>  
+      <div class="modal-header"><h3 class="text-center" translate>View in other lang</h3></div>
       <ul>
       % for l in other_langs:
         <li>
@@ -45,7 +45,7 @@
 <%def name="show_missing_langs_links(module, document, missing_langs)">\
 % if missing_langs:
   <div  id="missing-langs">
-    <div class="modal-header"><h3 class="text-center" translate>Translate into an other lang</h3></div>  
+    <div class="modal-header"><h3 class="text-center" translate>Translate into an other lang</h3></div>
     <ul>
       % for lang in missing_langs:
       <li protected-url-btn url="${request.route_url(module + '_edit', id=document['document_id'], lang=lang)}">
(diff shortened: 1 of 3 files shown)


c2corg_ui/templates/helpers/view.html
```diff
@@ -226,9 +226,19 @@ <h3 class="heading show-phone outings" ng-click="mainCtrl.animateHeaderIcon($eve
...
   ## WEATHER AND CONDITIONS
   % if text == 'weather':
-    % if locale['weather'] or locale['conditions'] or locale['conditions_levels']:
+  <%
+      cond = json.loads(locale['conditions_levels'])
+      conditions_levels_exist = (len(cond) > 0) and \
+          ( ('level_snow_height_total' in cond[0] and cond[0].get('level_snow_height_total') != '' and cond[0].get('level_snow_height_total') is not None) or \
+          ('level_snow_height_soft' in cond[0] and  cond[0]['level_snow_height_soft'] != '' and cond[0].get('level_snow_height_soft') is not None) or \
+          ('level_place' in cond[0] and cond[0].get('level_place')  != '') or \
+          ('level_comment' in cond[0] and cond[0].get('level_comment') != '') )
```
asaunier: Isn't it enough to test if ``cond`` is not empty?
If no condition was set, has ``cond`` an initial item with blank ``level_snow_height_total`` etc.?

c2corg_ui/templates/helpers/view.html
```diff
@@ -226,9 +226,19 @@ <h3 class="heading show-phone outings" ng-click="mainCtrl.animateHeaderIcon($eve
...
 
 <%def name="get_document_locale_text(locale, text)">\
+
+
   ## WEATHER AND CONDITIONS
   % if text == 'weather':
-    % if locale['weather'] or locale['conditions'] or locale['conditions_levels']:
+  <%
+      cond = json.loads(locale['conditions_levels'])
+      conditions_levels_exist = (len(cond) > 0) and \
```
asaunier: I think ``(len(cond) > 0)`` could simply be replaced by ``cond``.

asaunier:
Just a question for a possible (?) simplification, else looks good to me!

c2corg_ui/templates/outing/edit.html
```diff
@@ -94,11 +94,12 @@ <h5 translate ng-show="outing.associations.routes.length == 0">choose at least o
             </div>
 
             ## PARTIAL TRIP
-            <div class="form-group data half" ng-class="{'has-success': outing.partial_trip }">
+            <div class="form-group data half" ng-class="{'has-success': outing.partial_trip == false || outing.partial_trip == true }">
```
asaunier: Oops, I forgot to have a look at this commit :P

What about ``outing.partial_trip !== undefined`` instead of ``outing.partial_trip == false || outing.partial_trip == true``?

ginold: I will try that.

c2corg_ui/templates/helpers/view.html
ginold: I have seen a lot of possible cases: when conditions_levels = [], or "[ ]" (string), or has one or two or all of the fields, depending on when and how you delete them (save an outing without conditions_levels, save with conditions_levels and then delete, fill only one field etc).
I hope this big check is bullet-proof.

asaunier: OK then

Issue #437, opened 2016-06-17
Incomplete or problematic build requirement


LaurentBonnaud:
README.md has a list of requirements but this one has a problem:

` * Python 3.4`

because it does not specify "or higher".  Either python 3.5 is OK and it should be written down, or it is not and this is a big problem since recent/upcoming Linux distribution (Ubuntu 16.04, Debian stretch) ship python 3.5 by default.

tsauerwein:
Python 3.4 is the Python version that we are using. But it should also work with Python 3.5, it's just that we haven't tested. Feel free to report back if you are using 3.5 so that we could adjust the requirements.

LaurentBonnaud:
On my Ubuntu 16.04 system with Python 3.5 I could:

 - finish the build
 - launch the HTTP server
 - display the application

But running the tests gives the following error:

```
$ make -f config/dev test
.build/venv/bin/nosetests
Trying `CDLL(libgeos_c.so.1)`
Library path: 'libgeos_c.so.1'
DLL: <CDLL 'libgeos_c.so.1', handle 2e374f0 at 0x7f49e3524b00>
Trying `CDLL(libc.so.6)`
Library path: 'libc.so.6'
DLL: <CDLL 'libc.so.6', handle 7f49e5377a88 at 0x7f49e3af2e48>
......F......
======================================================================
FAIL: test_api_call (c2corg_ui.tests.views.test_route.TestRouteUi)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/donnees/v6_ui/c2corg_ui/tests/views/test_route.py", line 16, in test_api_call
    self._test_api_call()
  File "/donnees/v6_ui/c2corg_ui/tests/views/__init__.py", line 39, in _test_api_call
    self.assertEqual(resp['status'], '200')
AssertionError: '500' != '200'
- 500
? ^
+ 200
? ^

-------------------- >> begin captured logging << --------------------
c2corg_ui.views.document: DEBUG: API: GET http://c2corgv6-api.demo-camptocamp.com/routes
--------------------- >> end captured logging << ---------------------

----------------------------------------------------------------------
Ran 13 tests in 6.960s

FAILED (failures=1)
Makefile:66: recipe for target 'test' failed
make: *** [test] Error 1

```

LaurentBonnaud:
However, this one does not fail:

```
$ .build/venv/bin/nosetests -s
Trying `CDLL(libgeos_c.so.1)`
[...]

PR #469, opened 2016-06-29
Add Python 3.5 note in the requirements doc

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/437

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ UI Application for camptocamp.org v6
 Requirements
 ------------
 
- * Python 3.4
+ * Python 3.4 (3.5 seems OK as well)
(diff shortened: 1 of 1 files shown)


asaunier:
Any objection?

tsauerwein:
:+1:

asaunier:
Thanks @tsauerwein :)

mfournier:
@asaunier, @tsauerwein, would it be an idea to test both python versions in travis-ci, the same way nodejs 4 and 6 are currently tested in parallel ?

asaunier:
No clue.
As for now we only use python 3.4. Is it really useful to test with 3.5 if anyway we have no infrastructure to make the required changes for 3.5?

Issue #304, opened 2016-04-13
Map recentering tools


fjacon:
user position, simple search, geonames/OSM?, predefined C2C regions

fjacon:
According to offer : geonames only

asaunier:
For geonames see 
https://github.com/camptocamp/agridea_geoacorda/blob/master/jsapi/src/searchcontrol.js
https://github.com/camptocamp/agridea_geoacorda/blob/master/jsapi/src/partials/main.html#L46

asaunier:
The "recenter on my position" directive has been pushed to master. But it only works with localhost or in an HTTPS page. Currently on the demo, pressing the button (in the advanced search maps) does nothing except displaying in the console:
> getCurrentPosition() and watchPosition() are deprecated on insecure origins. To use this feature, you should consider switching your application to a secure origin, such as HTTPS. See https://goo.gl/rStTGz for more details.

The geonames tool is still WIP.

PR #475, opened 2016-06-30
Add map-search directive

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/304

Relies on the geonames service.

### DIFF
--- a/c2corg_ui/static/js/map/search.js
+++ b/c2corg_ui/static/js/map/search.js
@@ -1,0 +1,189 @@
+/**
+ * Adapted from https://github.com/camptocamp/agridea_geoacorda/blob/master/jsapi/src/searchcontrol.js
+ */
+goog.provide('app.MapSearchController');
+goog.provide('app.mapSearchDirective');
+
+goog.require('app');
+goog.require('app.constants');
+/** @suppress {extraRequire} */
+goog.require('ngeo.searchDirective');
+goog.require('ol.Feature');
+goog.require('ol.geom.Point');
+
+
+/**
+ * @return {angular.Directive} Directive Definition Object.
+ */
+app.mapSearchDirective = function() {
+  return {
+    restrict: 'E',
+    scope: {
+      'map': '=appMapSearchMap'
+    },
+    controller: 'AppMapSearchController',
+    bindToController: true,
+    controllerAs: 'searchCtrl',
+    templateUrl: '/static/partials/map/search.html',
+    link:
+        /**
+         * @param {angular.Scope} scope Scope.
+         * @param {angular.JQLite} element Element.
+         * @param {angular.Attributes} attrs Atttributes.
+         */
+        function(scope, element, attrs) {
+          // Empty the search field on focus and blur.
+          element.find('input').on('focus blur', function() {
+            $(this).val('');
+          });
+        }
+  };
+};
+
+app.module.directive('appMapSearch', app.mapSearchDirective);
+
+
+/**
+ * @constructor
+ * @param {angular.Scope} $rootScope Angular root scope.
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
+ * @ngInject
+ * @ignore
+ */
+app.MapSearchController = function($rootScope, $compile, gettextCatalog) {
+
+  /**
+   * @type {ol.Map}
+   * @export
+   */
+  this.map;
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.gettextCatalog_ = gettextCatalog;
+
+  /** @type {Bloodhound} */
+  var bloodhoundEngine = this.createAndInitBloodhound_();
+
+  /**
+   * @type {TypeaheadOptions}
+   * @export
+   */
+  this.options = /** @type {TypeaheadOptions} */ ({
+    highlight: true,
+    hint: undefined,
+    minLength: undefined
+  });
+
+  /**
+   * @type {Array.<TypeaheadDataset>}
+   * @export
+   */
+  this.datasets = [{
+    source: bloodhoundEngine.ttAdapter(),
+    display: function(feature) {
+      return feature.get('name');
+    },
+    limit: Infinity,
+    identify: function(feature) {
+      return feature.getId();
+    },
+    templates: {
+      suggestion: function(feature) {
+        return '<p>' + feature.get('name') + '</p>';
+      },
+      notFound: function() {
+        return this.gettextCatalog_.getString('No result');
+      }.bind(this)
+    }
+  }];
+
+  /**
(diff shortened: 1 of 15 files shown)


c2corg_ui/static/js/map/search.js
```diff
@@ -0,0 +1,165 @@
...
+  return {
+    restrict: 'E',
+    scope: {
+      'map': '=appMapSearchMap'
+    },
+    controller: 'AppMapSearchController',
+    bindToController: true,
+    controllerAs: 'searchCtrl',
+    template:
+        '<input type="text" placeholder="{{\'Recenter on...\' | translate}}" ' +
```
tsauerwein: Is "Recenter on" extracted correctly?

tsauerwein:
:+1:

c2corg_ui/static/js/map/search.js
asaunier: I think so.

asaunier:
Thanks for reviewing and for the original code @tsauerwein 
Actually the PR is not ready (WIP): the request to geonames is triggered and I can see a valid response in my console but no results are displayed in the dropdown.
Perhaps I should specify somewhere how to read the response? Like here: https://github.com/camptocamp/cgxp/blob/master/core/src/script/CGXP/plugins/Geonames.js#L103-L115 ?

tsauerwein:
> Perhaps I should specify somewhere how to read the response?

`CreateGeoJSONBloodhound` expects GeoJSON. You will have convert the response similar to what is done for the simple search: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/simplesearch.js#L299

asaunier:
I see. The geonames response is not at all a GeoJSON response:
> {"totalResultsCount":5,"geonames":[{"adminCode1":"24","lng":"1.44006","geonameId":3038744,"toponymName":"Abondant","countryId":"3017382","fcl":"P","population":1807,"countryCode":"FR","name":"Abondant","fclName":"city, village,...","countryName":"France","fcodeName":"populated place","adminName1":"Centre","lat":"48.7859","fcode":"PPL"},{"adminCode1":"25","lng":"21.5159","geonameId":245752,"toponymName":"Abonday","countryId":"2434508","fcl":"P","population":0,"countryCode":"TD","name":"Abonday","fclName":"city, village,...","countryName":"Tchad","fcodeName":"populated place","adminName1":"Région du Sila","lat":"11.8618","fcode":"PPL"},{"adminCode1":"84","lng":"6.72026","geonameId":3038745,"toponymName":"Abondance","countryId":"3017382","fcl":"P","population":1498,"countryCode":"FR","name":"Abondance","fclName":"city, village,...","countryName":"France","fcodeName":"populated place","adminName1":"Auvergne-Rhône-Alpes","lat":"46.28031","fcode":"PPL"},{"adminCode1":"MQ","lng":"-60.88308","geonameId":3571154,"toponymName":"Abondance","countryId":"3570311","fcl":"P","population":0,"countryCode":"MQ","name":"Abondance","fclName":"city, village,...","countryName":"Martinique","fcodeName":"populated place","adminName1":"Martinique","lat":"14.53733","fcode":"PPL"},{"adminCode1":"RE","lng":"55.67338","geonameId":936193,"toponymName":"Abondance","countryId":"935317","fcl":"P","population":0,"countryCode":"RE","name":"Abondance","fclName":"city, village,...","countryName":"Réunion","fcodeName":"populated place","adminName1":"Réunion","lat":"-21.03465","fcode":"PPL"}]}

> You will have convert the response similar to what is done for the simple search: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/simplesearch.js#L299

OK thanks for the tip!

asaunier:
> CreateGeoJSONBloodhound expects GeoJSON. You will have convert the response similar to what is done for the simple search

Does this mean that I cannot use CreateGeoJSONBloodhound here since the response is not a GeoJSON? The ``filter`` function seems to be called on an already geojson-formatted response:
https://github.com/camptocamp/ngeo/blob/master/src/services/creategeojsonbloodhound.js#L78

Should I use a standard ``Bloodhound`` instead, as in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/simplesearch.js#L271 ?

asaunier:
Other options:
* create a proxy service that will reformate the geonames reponses to GeoJSON
* use another service with a GeoJSON response. Perhaps Bing or Google geocoding services, but it seems both return JSON, not GeoJSON. https://developers.google.com/maps/documentation/geocoding/start https://msdn.microsoft.com/en-us/library/ff701711.aspx

tsauerwein:
We don't have to use `CreateGeoJSONBloodhound`.

asaunier:
> We don't have to use CreateGeoJSONBloodhound.

Yes, I have removed it and used a standard ``Bloodhound`` object.
The PR is almost ready:
* it seems there is a styling conflict with the simple search tool => the results are shown with a not really adapted style. In addition the country name is not visible (should be for instance ``Abondance (France)``). See
sans titre
* when selecting a result, the recentering is not done to the correct place. There is probably a reprojection problem around here https://github.com/c2corg/v6_ui/blob/4a36a5d2fafd6ebb2270d069804155f03bf11279/c2corg_ui/static/js/map/search.js#L150-L155

Unfortunately I don't have time to fix the remaining problems. Feel free to fix them (especially the recentering issue) and to merge the PR, thanks :)

asaunier:
The PR may be tested in my dev instance at http://c2corgv6-demo.gis.internal:6555/waypoints

asaunier:
I have fixed the recentering problem and improved the layout of the suggestions dropdown even though it still needs to be fixed:
sans titre

@ginold I need your help because I am lost among all those typeahead styles!

May be tested live at http://c2corgv6-demo.gis.internal:6555/routes

asaunier:
Thanks @ginold for the layout improvement, it looks really good. Just one thing: do you know how we can get rid of the white left margin when the focus is on a result (blue line)?
sans titre

asaunier:
> do you know how we can get rid of the white left margin when the focus is on a result (blue line)?

OK I found it. Will propose a small change...

asaunier:
> Will propose a small change...

I have removed ``margin-left: 5px;`` in https://github.com/c2corg/v6_ui/blob/map-search/less/map.less#L40 and it fixes the problem.

I have also rebased the PR.

@ginold OK for merging?

ginold:
Maybe the last thing I would suggest is to change the color while a suggestion is selected. Orange on blue background is bof. 

When moving with arrows, the suggestion div will have a `tt-cursor` (like `<div class="tt-suggestion tt-cursor">`) class, but you have to add the same styling when hovering.

At the moment I see it should be done for every app-simple search too (it has only been done on hover, not when moving with arrows), so you can add this style to simplesearch.less.

asaunier:
In my opinion the current styling is fair enough.
If you think it can be improved easily and without spending too much time, feel free to do it :)

Issue #408, opened 2016-06-11
Problems with versions of a document


loicperrin:
- [x] Date/time is wrong:
  I modify an outing (see outings/history/736972/fr). The modification time is wrong: 06:mm instead of 16:mm
- [x] Pb to display outing
  In "Comparaison des documents" page (see outings/diff/736972/fr/1164793/1164795)
  When I click on "Version du....", I get an "Internal Server Error".
- [x] It should be possible to display the current version of the outing from this page.

asaunier:
> It should be possible to display the current version of the outing from this page.

Yep, I have also noticed that it's not easy to go back to the history page of a document when viewing diffs.

fjacon:
Display current version is a nice to have (extra)

tsauerwein:
> Display current version is a nice to have (extra)

If I understand correctly this is simply about adding a link to go back?

asaunier:
* [ ] When showing on old version of a document, the message warning that the current page is an old one is not displayed correctly (probably a CSS default introduced in the meantime). This the screeshot below:
![sans titre](https://cloud.githubusercontent.com/assets/1192331/16445697/2f443d54-3de4-11e6-8898-ed2c57466cfa.png)

asaunier:
> If I understand correctly this is simply about adding a link to go back?

When you see a diff page, for instance http://c2corgv6.demo-camptocamp.com/routes/diff/51225/fr/51910/335059 you have no direct way to see the document itself.
Well you could click on the link toward one of the diffed version and then you have links towards the last version of the document. I agree this is not critical.

asaunier:
* [ ] The "title_prefix" (main waypoint name) is missing for routes in the history pages and diff pages.

Concerning the versions/history page, see also the radio button issue at https://github.com/c2corg/v6_ui/issues/344

ginold:
@loicperrin are you still getting an internal server error when comparing 2 outings? I cannot access the two you posted and can't find any other two that give such an error.
I will mark this as resolved for now.

loicperrin:
I still have this pb: In "Comparaison des documents" page (see outings/diff/736972/fr/1164793/1164795)
When I click on "Version du....", I get an "Internal Server Error".
See http://c2corgv6.demo-camptocamp.com/outings/diff/736956/fr/1164743/1164744

And the Date/time is still wrong too.

asaunier:
What about the "main waypoint name" missing for routes in the history and diff pages?
[...]

Issue #344, opened 2016-05-03
Bad rendering of the radio buttons in the history page


asaunier:
![sans titre](https://cloud.githubusercontent.com/assets/1192331/14979021/19851b3c-1120-11e6-92a5-75de45ea3a34.png)

fjacon:
@asaunier @ginold looks good to me.
can we close ?

asaunier:
> can we close ?

Nope, the problem has not been fixed.
Try for instance to click on some other radio button in http://c2corgv6.demo-camptocamp.com/routes/history/45168/fr

stef74:
@shortoland could you please test

shortoland:
OK for me, I close

PR #482, opened 2016-07-06
history fixes

ginold:
- warning title in history page css fix
- radio buttons alignment
- 'version' variable added to outing history page
- correct date displayed in history table
- display the latest version of document from diff page

related to https://github.com/c2corg/v6_ui/issues/408 and  https://github.com/c2corg/v6_ui/issues/344

Check:
/waypoints/733992/en/1160836
/outings/diff/736707/fr/1164428/1164441
/outings/736707/fr/1164428
/routes/736728/fr/1164452

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -127,35 +127,35 @@
 <%def name="show_associated_routes(document, type='routes')">\
   ## Possible values for type: routes, all_routes
   <%
-      routes = document['associations']['all_routes']['routes'] if type == 'all_routes' else document['associations']['routes']
-      associations = 'detailsCtrl.documentService.document.associations.'
-      associations += 'all_routes.routes' if type == 'all_routes' else 'routes'
+    routes = document['associations']['all_routes']['routes'] if type == 'all_routes' else document['associations']['routes']
+    associations = 'detailsCtrl.documentService.document.associations.'
+    associations += 'all_routes.routes' if type == 'all_routes' else 'routes'
   %>
-  ${add_seo_association_links(routes, 'routes')}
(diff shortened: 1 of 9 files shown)


ginold:
@arnaud-morvan, could you have a look on this PR plase?

ginold:
@gberaudo you could also have a look at this after we merge the progressbar?

c2corg_ui/templates/helpers/view.html
```diff
@@ -184,8 +184,7 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
             ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
         % endif
       </div>
-    </div>
```
arnaud-morvan: Are you sure you should remove this closing tag

arnaud-morvan:
When I open an history page, I have the Plus flat button with Translate in other lang option. Is this normal ? When I click on it it seems to open an empty modal dialog.

ginold:
I think you are right, there should be no buttons at all because you're just consulting an old version and should not be able to do anything with it.

arnaud-morvan:
Missing update on locale files (.pot and .po)

ginold:
Locale files update is not very important for the moment because it's always changing and in the end the translations will be made by an external enterprise.

arnaud-morvan:
Yes, this is not very important, but this made my working dir dirty when I build your branch.

ginold:
updated

ginold:
sorry, not yet here!

ginold:
If I remember well, Alex removed the locale compiling when running make check or serve because it was so annoying and we never needed those locales.

PR #485, opened 2016-07-07
Move progressbar code to a dedicated directive

### LINKED ISSUE #450: Move the step bar code from the documentediting controller to a directive
This code is not really related to the processing of the edited document data and could be moved to a dedicated directive.

ginold:
- create progress bar directive and remove it's old code from documentediting.js
- update edit.html accordingly

fixes #450

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -141,39 +141,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
   this.alerts = appAlerts;
 
   /**
-   * first next step would be 2 by default
-   * @export
+   * @type {boolean}
+   * @public
    */
-  this.next_step = 2;
-
-  /**
-   * Current step is 1 by default
-   * @export
-   */
-  this.current_step = 1;
-
-  /**
-   * Waypoint type initialised in the edit.html
-   * @export
-   */
-  this.waypoint_type;
-
-  /**
-   * Previous step is 0 by default
-   * @export
-   */
-  this.previous_step = 0;
-
-  /**
-   * Max possible steps for creation/edition
-   * @export
-   */
-  this.max_steps;
-
- /**
-  * @type {boolean}
-  * @public
-  */
   this.submit = false;
 
   /**
@@ -354,8 +324,7 @@ app.DocumentEditingController.prototype.prepareData = function(data) {
  * @param {string} index_url URL of index page.
  * @export
  */
-app.DocumentEditingController.prototype.cancel = function(view_url,
-    index_url) {
+app.DocumentEditingController.prototype.cancel = function(view_url, index_url) {
   var url = !view_url || this.isNewLang_ ? index_url : view_url;
   window.location.href = url;
 };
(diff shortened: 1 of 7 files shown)


ginold:
@gberaudo, could you have a look at this pr please?

c2corg_ui/static/js/documentediting.js
```diff
@@ -141,39 +141,9 @@ app.DocumentEditingController = function($scope, $element, $attrs,
...
-   * Previous step is 0 by default
-   * @export
-   */
-  this.previous_step = 0;
-
-  /**
-   * Max possible steps for creation/edition
-   * @export
+   * @type {boolean}
+   * @public
```
gberaudo: The `@public` annotation does not exist. Do you mean `@export`?

c2corg_ui/static/js/style/progressbar.js
```diff
@@ -0,0 +1,191 @@
...
+ * The progress bar on the top of editing pages.
+ * The navigation buttons (right and left) are also managed
+ * by this directive.
+ * @return {angular.Directive} The directive specs.
+ */
+app.progressBarDirective = function() {
+  return {
+    restrict: 'EA',
+    controller: 'appProgressBarController as progBarCtrl',
+    bindToController: true
```
gberaudo: Shouldn't some properties be bound to the controller? Like `waypointType`?

c2corg_ui/static/js/style/progressbar.js
```diff
@@ -0,0 +1,191 @@
...
+ * @constructor
+ * @export
+ */
+app.ProgressBarController = function() {
+
+  /**
+   * first next step would be 2 by default
+   * @export
+   */
+  this.next_step = 2;
```
gberaudo: The properties should be camelcase.

c2corg_ui/static/js/style/progressbar.js
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Navigate through creation/editing steps
+ * @param {number} step
+ * @param {string} direction
+ * @export
+ */
+app.ProgressBarController.prototype.step = function(step, document, direction) {
+  switch (step) {
+    case 1:
+      $('.editing').animate({left: '0'});
```
gberaudo: The jQuery queries should not be global.

c2corg_ui/static/js/style/progressbar.js
```diff
@@ -0,0 +1,191 @@
...
+    willBe = (percent * (step - 1)) - 10;
+    nextPosition = (percent * step) - 10;
+  } else {
+    willBe = (percent * step) - 10;
+    nextPosition = (percent * (step + 1)) - 10;
+    stopBack = willBe;
+  }
+
+  // bar animation, timeout for a gradual filling
+  var timeout = setInterval(function() {
```
gberaudo: Shouldn't it be Angular timeout?

c2corg_ui/static/js/style/progressbar.js
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * The progress bar on the top of editing pages.
+ * The navigation buttons (right and left) are also managed
+ * by this directive.
+ * @return {angular.Directive} The directive specs.
+ */
+app.progressBarDirective = function() {
+  return {
+    restrict: 'EA',
+    controller: 'appProgressBarController as progBarCtrl',
```
gberaudo: A better name may be `progressBarCtrl`.

c2corg_ui/static/js/style/progressbar.js
gberaudo: Is the directive / controller limited to `editing`?

c2corg_ui/static/js/style/progressbar.js
ginold: If there is any advantage of using Angular instead of the normal `setInterval` that has worked just well until now, maybe yes?

c2corg_ui/static/js/style/progressbar.js
ginold: `waypoint_type` might be the only one.

c2corg_ui/static/js/style/progressbar.js
ginold: For now yes, it is only used in editing.

c2corg_ui/static/js/style/progressbar.js
gberaudo: If you are never reading the properties changed here from Angular then it is fine not using angular timeout.

c2corg_ui/static/js/documentediting.js
ginold: you will notice that nearly every @export has changed to @public since Alex created the outingediting.js that inherits from documentediting.js, I think there is a reason for that.

gberaudo: To prevent variable renaming by closure the `@export` annotation must be used. There is no mention of a `@public` annotation in closure documentation.

ginold: You should have said that when reviewing Alexe's PR ... :laughing: 

Someone on stackoverflow says:

> The access control annotations @private, @protected, and @public are directives for the Closure Compiler that help developers enforce the desired level of visibility for properties and functions.

But if you think it is needed, I can change them back to @export.

gberaudo: I see. The `@public` annotation is not listed on https://developers.google.com/closure/compiler/docs/js-for-compiler. I think it is unrelated to minification.

ginold:
so @gberaudo, ok for merging?

gberaudo:
OK

ginold:
Thank's, you can have a look on the 2 other prs I tagged you in and we are probably good to update the demo.

c2corg_ui/static/js/documentediting.js
asaunier: > But if you think it is needed, I can change them back to @export. However, if @asaunier changed it to be this way, there maybe is a reason.

@tsauerwein has recommended to me to use ``@public`` instead of ``@export`` for properties that should be accessed by inheriting classes but that do not need to be accessed in the HTML code.

PR #486, opened 2016-07-07
Show document-related comments from discourse

arnaud-morvan:
Superseed #458 

Add a comments section in the document (waypoint, outing and route) detail view with embeded comments from discourse.

For performance and technical problems, discourse topics ids are stored on guidebook side in documents_topics table.

If the topic do not exists, the comments section contains a button to create the first post (<=> topic/thread) in discourse. This button is only accessible to authenticated users, unauthenticated users get a "log in to create the first post" message.

For now the topic first post (<=> topic/thread description) only contains a link to the current document. Note that the first post is always created by the discourse system user and do not appears in the guidebook comments section, but only on the discourse side.

### DIFF
--- a/c2corg_ui/static/js/viewdetails.js
+++ b/c2corg_ui/static/js/viewdetails.js
@@ -124,6 +124,16 @@ app.ViewDetailsController.prototype.openModal = function(selector) {
 
 
 /**
+ * @export
+ */
+app.ViewDetailsController.prototype.scrollToComments = function() {
+  $('html, body').animate({
+    scrollTop: $('#discourse-comments').offset().top
+  }, 1000);
+};
+
+
+/**
  * @param {Event} tab the clicked tab
  * @export
  */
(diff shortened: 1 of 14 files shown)


c2corg_ui/static/js/viewdetails.js
```diff
@@ -299,6 +323,57 @@ app.ViewDetailsController.prototype.initSlickGallery_ = function() {
...
 /**
+ * @export
+ */
+app.ViewDetailsController.prototype.getComments = function() {
+  var topic_id = this.documentService.document['topic_id'];
+  if (topic_id === null) {
+    return;
+  }
+
+  // create DiscourseEmbed script tag. From a discourse tutorial.
```
tsauerwein: Do you still have the link to this tutorial?

ginold: https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963

asaunier: Please add the link to the comment in the code.

less/viewdetails.less
```diff
@@ -659,4 +658,10 @@ app-add-association {
     height: 110px;
     width: auto;
   }
+}
+
+#discourse-comments {
+  padding-top: 20px;
+  .flex();
```
arnaud-morvan: @ginold : .flex() breaks the collapse functionnality

arnaud-morvan:
I think we should merge this and give some time to c2corg administrators to describe precisely the content they want in the first post.

asaunier:
Could you precise what this PR does?
Embed the comments in the related documents and add a tool (link?) to post a new comment?
What happens when there is no comment yet? Is there always an initial post in the related thread?
What is this first post? An automatic post created by some automatic user? What is its current content?

arnaud-morvan:
I've update the pull request description.

asaunier:
Thanks!

> For now the topic first post (<=> topic/thread description) only contains a link to the current document. The first post is always created by the discourse system user.

Is this post displayed in the document-embedded comments section?

> I think we should merge this and give some time to c2corg administrators to describe precisely the content they want in the first post.

That sounds reasonable.

LGTM!

arnaud-morvan:
The first post do not appears on the document page, but on the discourse side.

PR #487, opened 2016-07-08
Finishes on images modal dialog

arnaud-morvan:


arnaud-morvan:
@ginold : can you review

ginold:
I encounter an error when uploading multiple images: when dropping for example 3 images, only the last one will upload. Same when I want to abort uploading images one by one - there are some errors.

Uploading and cancelling one image at a time works fine.

ginold:
and uploading one image, selecting the type and then uploading a second image, selecting the type and clicking on save also throws an error.

arnaud-morvan:
Bug fixed, merging

PR #491, opened 2016-07-11
Fix editing button ng-if expression

arnaud-morvan:
Fix editing button that do not display in routes because empty array evaluate to true.
https://github.com/c2corg/v6_ui/blob/7391fbf804e916fc55cdbb7eba51b3654bb49173/c2corg_ui/static/js/auth/authservice.js#L93

Note that IMHO all python variables passed to javacript part of HTML templates should be escaped with : 
```
${json.dumps(mavar) | n}
```
This pull request has been base on apha2 so it should ported in master after merge.


arnaud-morvan:
@ginold : Can you take a look here as a fix for cf6b5371e6bd35407053899cba0f2af95f5b1fba

arnaud-morvan:
As this fix a blocking issue, I merge it now.

PR #500, opened 2016-07-20
Fix typo in areas listing for route detail view

asaunier:


asaunier:
@ginold OK for merging?

Issue #498, opened 2016-07-19
Defaut map zoom


stef74:
For some route (climbing route ?)  ... the defaut zoom is to higher.

Ex : http://c2corgv6.demo-camptocamp.com/routes/736686/fr ... show only "green", we have to click more than 5 times to have other element like road on the map.

asaunier:
The problem seems to only occur for routes with no own geometry (GPS track => a default point geometry is used instead) since changes have been  made to (partially) add associated WP:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/route/view.html#L60-L75
=> associated WP geometries are missing and that probably causes the incorrect recentering/zooming (if the route has a GPS track, the problem is masked).
=> related to issue https://github.com/c2corg/v6_ui/issues/362

ginold:
In general I find that we should limit the zoom because the tiles are too pixelated and the half of the layers don't even appear at great zoom.

asaunier:
> In general I find that we should limit the zoom because the tiles are too pixelated and the half of the layers don't even appear at great zoom.

We use the default OSM etc. settings with the zoom levels defined for those layers. I don't think we should add extra restrictions, no one forces users to zoom to those zoomlevels :P

stef74:
Is fixed ? for me is not fixed for the moment (demo not updated ?)
![capture](https://cloud.githubusercontent.com/assets/3962997/17055739/0d029d20-500f-11e6-918e-11e8f03f568a.PNG)

asaunier:
> demo not updated ?

Indeed! I have just updated it.

stef74:
Great It's work !

Issue #362, opened 2016-05-18
Show the associated documents on the map in a document detail view page


asaunier:
See https://github.com/c2corg/v6_ui/issues/230#issuecomment-195326869

> * for list pages: use the point geometry (``geom``) to place/filter outings and routes on the map (we might show the ``geom_detailed`` when hovering/clicking the feature on the map)
* for route detail pages:
 * use the detailed geometry if available to show it on the map. If not available, show no geom.
 * show the associated WP geoms anyway.
* for outing detail pages:
 * use the detailed geometry if available to show it on the map.
 * not sure if we should show the associated routes detailed geometries if available (that might be a bit confusing because visitors might not figure out the tracks are not the actual outing track). Or only on demand, for instance when hovering the associated route cards?
 * if no detailed geom is available, perhaps we could simply recenter the map on the point ``geom`` (default geometry). Or show no map at all as in v5 (but I think showing a map anyway is interesting to figure out where the outing took place).
* detail pages of WP:
 * show the ``geom`` of the the WP (always available) => map is centered on it
 * show the associated WP geom?
 * show the associated routes detailed geom if available? no need to show the point default geom if no detailed geom is available.

PR #506, opened 2016-07-20
Fix showing associated WP on map

asaunier:
(not fully tested because the associated WP geometries are not provided by the API yet, see https://github.com/c2corg/v6_api/issues/342)

Related to https://github.com/c2corg/v6_ui/issues/498 and https://github.com/c2corg/v6_ui/issues/362

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -1,6 +1,6 @@
 <%!
     from c2corg_ui.templates.utils import get_attr
-    import json
+    from json import dumps, loads
 %>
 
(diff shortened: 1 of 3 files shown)


asaunier:
As I said above, this PR won't allow to show the associated WP on the map since the API does not provide their geometries yet, but at least it fixes the zoomlevel issue https://github.com/c2corg/v6_ui/issues/498 and is safe to merge I think.

OK for merging?

ginold:
ok for me

asaunier:
Thanks for reviewing @ginold

PR #507, opened 2016-07-20
Fix i18n duplicates

### LINKED ISSUE #392: Duplicated translations
There are many interface string translations that duplicated or very close (singular/plural, upper case / lower case, etc.).

We need to use the least translations as possible (lighter translation files + less work for translators) and should remove the duplicates.

asaunier:
Fixes #392 

I have noticed typos (missing or extra "s") for some attributes. For instance ``avalanche_sign`` instead of ``avalanche_signs``. Or ``orientation`` instead of ``orientations``. Please note that some attributes are almost homonyms, for instance ``duration`` for outings vs ``durations`` for routes.
I have then made changes in the editing forms. @ginold could you please double check/test that my changes are OK?

### DIFF
--- a/c2corg_ui/templates/i18n.html
+++ b/c2corg_ui/templates/i18n.html
@@ -28,13 +28,8 @@
 <span translate>lake</span>
 <span translate>bisse</span>
 <span translate>waterfall</span>
-<span translate>cliff</span>
 <span translate>cave</span>
-<span translate>pit</span>
 <span translate>locality</span>
-<span translate>confluence</span>
-<span translate>glacier</span>
-<span translate>bergschrund</span>
 <span translate>waterpoint</span>
 <span translate>climbing_outdoor</span>
 <span translate>climbing_indoor</span>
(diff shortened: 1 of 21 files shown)


asaunier:
I have tested (very fast) the creation/edition of WP, routes and outings and it works as expected. May be tested here: http://c2corgv6-demo.gis.internal:6555/

Looks good to you @ginold ?

c2corg_ui/templates/i18n.html
```diff
@@ -88,24 +83,20 @@
...
 ## home
 <span translate>Home</span>
 
 ## waypoint attributes
-<span translate>rain_proof</span>
 <span translate>very_safe</span>
 <span translate>exposed</span>
-<span translate>rock_types</span>
-<span translate>best_periods</span>
-<span translate>climbing_styles</span>
```
ginold: why did you delete all those lines?
I remember having `best_periods` or `climbing_styles` but they have disappeared oO.
Or you delete from here the strings that appear in the HTML directly ?

ginold:
after a few tests, I haven't encountered any errors :ok:

c2corg_ui/templates/i18n.html
asaunier: Yes, those strings are already available in HTML templates, thus are already extracted from those templates. We don't need them here.

asaunier:
Thanks for reviewing @ginold !

Issue #504, opened 2016-07-20
Filters interface


asaunier:
- [ ] ~~slider handles are not fully reset when clearing filters (see https://github.com/c2corg/v6_ui/pull/460#issuecomment-229400429)~~ => moved to https://github.com/c2corg/v6_ui/issues/655

See also https://github.com/c2corg/v6_ui/issues/471

Nice to have

- [x] Hilight the feature on the map when hovering a card (and vice-versa?) (done at https://github.com/c2corg/v6_ui/pull/521)
- [x] allow to bypass the map extent filtering as in Airbnb? (see https://github.com/c2corg/v6_ui/pull/648)

Issue #333, opened 2016-04-28
Show documents as popup on the map when clicking them


fjacon:

ginold:
Not sure to understand how it is supposed to work. Is it an extra enhancement ? has it been requested and described by the client?

asaunier:
I am not sure it is really relevant to show a popup on the map when clicking on a feature:
* as for now, clicking on a feature opens the related document page, having to open the popup and then click on some link would add an extra step/click
* what data should be displayed in the popup? A card? But cards are already displayed in the list of result for the advanced search page (or the list of associated WP etc. for detail view pages)

I suggest to skip the popups and to replace them by a better interactions between cards and map features in the advanced search pages. See https://github.com/c2corg/v6_ui/pull/521

tsauerwein:
I think this is relevant when you have multiple documents at the same position. If a click only returns a single document, you could jump to the document page. But if you get multiple documents, you would have to select the document you want. The popup could simply show a list of the document names.

ginold:
In order to not disturb the already implemented click functionality, the popup could appear on hover - at the same time it highlights the card from the list.

asaunier:
I am not sure we have that many cases where we have several features at the exact same place. It seems to me the hover/click interactions only "returns" one feature at a time.

The hover interactions already highlights and displays the name of the feature. With PR https://github.com/c2corg/v6_ui/pull/521 it also hilights the matching card in the list of results. For me it's already quite satisfaying and I don't think we have time to make more work on this part.

So I think we should restrict to what's suggested in https://github.com/c2corg/v6_ui/pull/521

Issue #285, opened 2016-04-11
Features icons on maps


asaunier:
* [x] When hovering a feature, its icon is magnified. But the standard-sized icon is still visible underneath.
* [ ] Features icons are too stacked on the map
* [ ] To avoid downloading every icon separately, we should use a font file containing the list of icons (in SVG). See maybe also https://github.com/c2corg/v6_ui/issues/152
* [x] Conflict between the "mouseover" and the "click" select interactions: when the icon is enlarged by the the mouseover, the feature is no longer found when clicking on it (see https://github.com/c2corg/v6_ui/pull/521#issuecomment-235838965)
* [x] When a feature style has been modified by the card mouseover in the advanced search, the feature icon is no longer updated/enlarged at feature mouseover (see https://github.com/c2corg/v6_ui/pull/521#issuecomment-235656710)

asaunier:
> Features icons are too stacked on the map

Moving here the related issue #338 that I will consequently close as duplicate.

@tsauerwein suggested:
> When clicking on the map, we could show a popup with a list of features at that point.

@asaunier replied:
> Would it actually solve the current issue (too many icons at the same place)? Or do you mean we could have a cluster layout + a popup listing the object at this point?

asaunier:
> ￼ When hovering a feature, its icon is magnified. But the standard-sized icon is still visible underneath.

Instead of showing a bigger image, a solution could be to keep the same size of icon but to change the color. This way the standard icon is not visible at all.

ginold:
Copy from #152 
It seems possible in OL2 according to this fiddle: http://jsfiddle.net/ldonnet/bdzz0ujk/ using font characters (icons) as label. Probably also in OL3?
According to @tsauerwein

You can set a label with a custom font:
http://stackoverflow.com/questions/27809479/use-semantic-ui-or-font-awesome-icons-as-markers-in-openlayers3

Or you create a single sprite image and then use offset+size when defining your ol.style.Icon:
http://openlayers.org/en/master/apidoc/ol.style.Icon.html

tsauerwein:
Because it is a bit tricky to use sprites with SVG files, we might consider generating a font with the SVG images.

asaunier:
Several points of this issue are already fixed by https://github.com/c2corg/v6_ui/pull/521

> Features icons are too stacked on the map

[...]

PR #521, opened 2016-07-22
Make map features and cards interact

asaunier:
* [x] When hovering a card in the advanced search page, hilight the feature on the map
* [x] When hovering a feature on the map, hilight the matching card

TOFIX
* [x] After hovering cards (features being then hilighted), the ``ol.interaction.Select`` no longer hilight the features on the map (style not updated?)
* [x] ~~Map features seem no longer clickable after hovering cards~~ => already the case in master
* [x] If a card has been hovered, the matching feature on the map no longer reacts to hover (perhaps because the feature style has been changed by the card hovering?).
* [x] Use ``ngClass`` to hilight the cards?

Related to https://github.com/c2corg/v6_ui/issues/333 and https://github.com/c2corg/v6_ui/issues/504

### DIFF
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -18,6 +18,7 @@ goog.require('ol.interaction.DragAndDrop');
 goog.require('ol.interaction.Draw');
 goog.require('ol.interaction.Modify');
 goog.require('ol.interaction.MouseWheelZoom');
+/** @suppress {extraRequire} */
 goog.require('ol.interaction.Select');
 goog.require('ol.layer.Vector');
 goog.require('ol.source.Vector');
@@ -171,6 +172,12 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
   this.isDrawing_ = false;
 
   /**
+   * @type {?number}
+   * @private
+   */
+  this.currentSelectedFeatureId_ = null;
+
+  /**
    * @type {ol.Map}
    * @export
    */
@@ -211,6 +218,12 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
 
     this.scope_.$root.$on('searchFeaturesChange',
         this.handleSearchChange_.bind(this));
+    this.scope_.$root.$on('cardEnter', function(event, id) {
+      this.toggleFeatureHighlight_(id, true);
+    }.bind(this));
+    this.scope_.$root.$on('cardLeave', function(event, id) {
+      this.toggleFeatureHighlight_(id, false);
+    }.bind(this));
 
     this.view_.on('propertychange',
       ngeoDebounce(
@@ -230,34 +243,9 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
 
   // add the features interactions
   if (this.features_.length > 0 || this.advancedSearch) {
-    this.getVectorLayer_().setStyle(this.createStyleFunction_(false));
-
-    var pointerMoveInteraction = new ol.interaction.Select({
-      style: this.createStyleFunction_(true),
-      condition: ol.events.condition.pointerMove
-    });
-    this.map.addInteraction(pointerMoveInteraction);
-
-    var clickInteraction = new ol.interaction.Select({
-      condition: ol.events.condition.click
-    });
-    clickInteraction.on('select', function(e) {
-      /**
-       * @type {ol.Collection.<ol.Feature>}
-       */
-      var features = e.target.getFeatures();
-      if (features.getLength() > 0) {
-        var first = features.item(0);
-        var module = /** @type {string} */(first.get('module'));
-        var id = first.get('documentId').toString();
-        var lang = /** @type {string} */(first.get('lang'));
-        var url = app.utils.buildDocumentUrl(module, id, lang);
-        if (url) {
-          window.location.href = url;
-        }
-      }
-    }.bind(this));
-    this.map.addInteraction(clickInteraction);
+    this.getVectorLayer_().setStyle(this.createStyleFunction_());
+    this.map.on('click', this.handleMapFeatureClick_.bind(this));
+    this.map.on('pointermove', this.handleMapFeatureHover_.bind(this));
   }
 
   if (this.edit && this.drawType) {
@@ -354,11 +342,10 @@ app.MapController.prototype.getVectorLayer_ = function() {
 
 
 /**
- * @param {boolean} highlight
  * @return {ol.style.StyleFunction}
  * @private
  */
-app.MapController.prototype.createStyleFunction_ = function(highlight) {
+app.MapController.prototype.createStyleFunction_ = function() {
   return (
       /**
        * @param {ol.Feature|ol.render.Feature} feature
@@ -369,12 +356,12 @@ app.MapController.prototype.createStyleFunction_ = function(highlight) {
         var module = /** @type {string} */ (feature.get('module'));
         switch (module) {
           case 'waypoints':
-            return this.createPointStyle_(feature, resolution, highlight);
+            return this.createPointStyle_(feature, resolution);
           case 'routes':
           case 'outings':
             return this.advancedSearch ?
-              this.createPointStyle_(feature, resolution, highlight) :
-              this.createLineStyle_(feature, resolution, highlight);
+              this.createPointStyle_(feature, resolution) :
+              this.createLineStyle_(feature, resolution);
           default:
             return null;
         }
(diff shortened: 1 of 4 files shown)


asaunier:
With 871d2d7 when hovering a feature on the map, the matching card in the list of results is hilighted (turns gray as if the mouse cursor was above the card itself).

I add to use jQuery's ``addClass`` and ``removeClass`` to change the CSS classes applied on the cards. I have first tried to do it using ``ngClass`` but with no success: the classes are not updated as expected (as if the ngClass expressions where only evaluated once for all). See https://github.com/c2corg/v6_ui/blob/871d2d7f6ff1b6171b52e5aabdae663241eb4e38/c2corg_ui/static/partials/advancedsearch.html#L2
Tips welcome :)

If this is not possible to use ``ngClass`` here, the ngClass-related code should be removed before merging the PR.

asaunier:
This PR is almost ready. 2 problems remain:

* it seems there's a kind of conflict between the mouveover and click interactions with map features. Both user ``ol.interaction.Select``. The problem can be reproduced with branch master (not related to my changes) See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/map.js#L235-L244
=> when a feature is hovered, its icon is enlarged but then it is no longer possible to click on it (and open the object page). It's only possible to click when the icon is not enlarged and the mouse is close enough (!?).
=> I won't fix it in this PR unless a simple solution is found.

* when the mouse cursor hovers a card, the style of the matching map feature is updated to enlarge its icon:  https://github.com/c2corg/v6_ui/blob/818416d178079be065fdd44aa9a2b66e77f831be/c2corg_ui/static/js/map/map.js#L563-L571
but then the map feature no longer reacts to mouseover events, as if the style defined for the ``ol.interaction.Select`` was lost!?https://github.com/c2corg/v6_ui/blob/818416d178079be065fdd44aa9a2b66e77f831be/c2corg_ui/static/js/map/map.js#L223-L226

tsauerwein:
> I had to use jQuery's addClass and removeClass to change the CSS classes applied on the cards. I have first tried to do it using ngClass but with no success: the classes are not updated as expected

The event is probably triggered by OpenLayers? In this case Angular does not know that something has changed (you have to call `$scope.apply()`).

> Both user ol.interaction.Select.

It could be that 2 select interactions do not work together very well. For the click, you could directly set up an event with `map.on('click', ...)` with `map.forEachFeatureAtPixel`. For example like here: http://openlayers.org/en/master/examples/icon.html

asaunier:
> The event is probably triggered by OpenLayers? In this case Angular does not know that something has changed (you have to call $scope.apply()).

Yes, that's what I did in commit 818416d and now the ng-class trick works.
Thanks!

> For the click, you could directly set up an event with map.on('click', ...) with map.forEachFeatureAtPixel.

OK I will try that.

asaunier:
> It could be that 2 select interactions do not work together very well. For the click, you could directly set up an event with map.on('click', ...) with map.forEachFeatureAtPixel. For example like here: http://openlayers.org/en/master/examples/icon.html

I have tried the following code to replace the "click" select interaction:
```
    this.map.on('click', function(evt) {
      var feature = this.map.forEachFeatureAtPixel(evt.pixel, function(feature) {
        return feature;
      }); 
      if (feature) {
        var module = /** @type {string} */(feature.get('module'));
        var id = feature.get('documentId').toString();
        var lang = /** @type {string} */(feature.get('lang'));
        var url = app.utils.buildDocumentUrl(module, id, lang);
        if (url) {
          window.location.href = url;
        }
      }   
    }.bind(this));
```
but the problem remains: when the icon is enlarged by the "pointermove" interaction, clicking on the icon has no effect. Actually the click callback is triggered but no feature is found.

I would like to move this problem (and the other described at point 2 of https://github.com/c2corg/v6_ui/pull/521#issuecomment-235656710) to https://github.com/c2corg/v6_ui/issues/285

Then I will "autosquash" the commits of this PR and merge it. Sounds reasonable?

asaunier:
PR rebased and issue https://github.com/c2corg/v6_ui/issues/285 updated to take into account the problems mentioned above.

Please review :)

tsauerwein:
> Actually the click callback is triggered but no feature is found.

I tried your branch and it seems to work. Is it only for some icons? Related to https://github.com/c2corg/v6_ui/issues/527?

asaunier:
> I tried your branch and it seems to work. Is it only for some icons? Related to #527?

Are you sure it works? It's easier to test on a small area with not too many map features at the same place.

* Hover a feature on the map => the icon is enlarged and the card is hilighted.
* Hover the matching card => the card is hilighted and the map feature is enlarged
* Hover again the map feature => the card is hilighted but the map feature icon is no longer enlarged. Other map feature icons are enlarged as expected until their card is hovered.

Don't know if it is related to the "flashing" icon (#527). I would say no.

asaunier:
I have pushed 3 new "fixup" commits. I will squash them before merging the PR. In the meantime I leave them as is since they might be useful to understand the changes I made.

It seems that updating the features style when hovering cards or features caused problem, including the one reported at https://github.com/c2corg/v6_ui/pull/521#issuecomment-236841161
Finally I have prefered having a single style definition and make it rely on a boolean ``highlight`` property that is set/updated when hovering cards or features. It fixes the problem mentioned above. See 903e558

In discussions found on the web I have read that having 2 ``ol.interaction.Select`` at the same time was a problem. I have decided to get rid of both of them and to replace them by ``click`` and ``pointermove`` map events. Now it is possible both to hover a feature and to click it. See 		￼	3f203b8

Not sure which one of the 2 changes above did the trick but 3 of the problems reported in issue https://github.com/c2corg/v6_ui/issues/285 are also fixed:
* when enlarging the hovered icon, one no longer sees a small one in the background
* the 2 last ones
In addition it seems that the "flashing" effect reported at https://github.com/c2corg/v6_ui/issues/527 is a lot attenuated (for me it is reasonable and the issue might be closed as well).

One last small "problem": since the ``ol.interaction.Select`` interactions are no longer used, I have tried to remove the ``goog.require``. But then I got the following closure error:
> ERR! compile node_modules/openlayers/externs/olx.js:3040: ERROR - Bad type annotation. Unknown type ol.interaction.SelectFilterFunction
ERR! compile  *     filter: (ol.interaction.SelectFilterFunction|undefined),
ERR! compile                 ^
ERR! compile 
ERR! compile node_modules/openlayers/externs/olx.js:3142: ERROR - Bad type annotation. Unknown type ol.interaction.SelectFilterFunction
ERR! compile  * @type {ol.interaction.SelectFilterFunction|undefined}
ERR! compile           ^

To avoid it, I have left the goog.require and add an ``/** @suppress {extraRequire} */`` comment. Anyone knows if I could do an other way?

Please review :)

tsauerwein:
> One last small "problem": since the ol.interaction.Select interactions are no longer used, I have tried to remove the goog.require. But then I got the following closure error: ...

This should be fixed when we update ol3 (it's related to https://github.com/openlayers/ol3/pull/5335).

c2corg_ui/static/js/map/map.js
```diff
@@ -619,6 +625,66 @@ app.MapController.prototype.handleMapSearchChange_ = function(event) {
...
 /**
+ * @param {ol.MapBrowserEvent} event
+ * @private
+ */
+app.MapController.prototype.handleMapFeatureClick_ = function(event) {
+  var feature = this.map.forEachFeatureAtPixel(event.pixel, function(feature) {
+    return feature;
+  }, this, function(layer) {
+    // test only features from the current vector layer
+    return layer === this;
```
tsauerwein: Could you make this:
```
   return layer === this.getVectorLayer_();
}, this));
```

This would be less confusing.

c2corg_ui/static/js/map/map.js
```diff
@@ -619,6 +625,66 @@ app.MapController.prototype.handleMapSearchChange_ = function(event) {
...
+ * @private
+ */
+app.MapController.prototype.handleMapFeatureHover_ = function(event) {
+  if (event.dragging) {
+    return;
+  }
+  var pixel = this.map.getEventPixel(event.originalEvent);
+  var hit = this.map.hasFeatureAtPixel(pixel, function(layer) {
+    // test only features from the current vector layer
+    return layer === this;
```
tsauerwein: Same remark here.

tsauerwein:
That looks good, thanks!

c2corg_ui/static/js/map/map.js
asaunier: Good catch! Actually I have hesitated using this syntaxe but I thought passing ``this.getVectorLayer_()`` as the ``this`` argument of the function would make ``getVectorLayer_()`` be called only once, not for every tested layer.
But perhaps it's not that significant (not such a big deal and not that many layers to test) and as you said, it's a bit clearer.
What do you recommend?

asaunier:
> This should be fixed when we update ol3

Should we do the upgrade now (+ ngeo?), in a separated PR. Anyway I think snyk recommends to do so too. See https://github.com/c2corg/v6_ui/pull/529#issuecomment-235631144

c2corg_ui/static/js/map/map.js
tsauerwein: If you are worried about that, you could cache the results of the function before calling `forEachFeatureAtPixel`. But I think that is not a problem.

asaunier:
Thanks @tsauerwein for reviewing!

I have made the change you have suggested and squashed the commits. I will merge when Travis passes.

Issue #516, opened 2016-07-21
Outing: pb with field 'Nombre de participants'


desnoes:
When i write some letters (e.g. 'ehf') in the field 'Nombre de participants', there is no warning saying that only integers are allowed. On the other hand, it is impossible to save the outing if we let these letters. It is possible to save the outing when they are erased.

stef74:
@desnoes please test

desnoes:
the bug has been corrected

Issue #514, opened 2016-07-21
'Matériel glacier' is not displayed


desnoes:
The selection of a filed in 'Materiel glacier'is not displayed
http://c2corgv6.demo-camptocamp.com/routes/edit/736977/it

ginold:
yes it is : 3rd step under "materiel". Probably it's not the best place for this field. Where do you think it should be moved to?

desnoes:
When you save the new route, this field is not displayed. You can check with the example i gave above.

stef74:
@desnoes please test

desnoes:
the bug has been corrected

Issue #509, opened 2016-07-21
No rating block for via ferrata


stef74:
On http://c2corgv6.demo-camptocamp.com/routes/edit/342031/fr 
I have no rating block.

If I add an other activity rating block appear with good rating field (for both activities)

Previsoulys this route is multiple activity I have changed to via ferrata activity.

stef74:
@ginold not fixed on my example.

ginold:
I see, I'm surprised something went wrong because I had fixed it. 
Refixing it right now.

ginold:
Ok I understood the problem !
@stef74, I think we didn't add the property materiel glacier for activity via ferrata and therefore it doens't appear, but when you select another activity that does, it works.
Should we add materiel glacier for via ferrata?

stef74:
we don't have to add materiel de glacier for via ferrata activity

2016-07-29 11:41 GMT+02:00 ginold <notifications@github.com>:

ginold:
Ok, so I remove the field "materiel glacier" when only via ferrata is selected.

loicperrin:
In fact there is no rating block when only one activity is selected. When 2 activities, the rating blocks appears (and mixed for the 2 activities).

asaunier:
> In fact there is no rating block when only one activity is selected. When 2 activities, the rating blocks appears (and mixed for the 2 activities).

I think this has been fixed last week by https://github.com/c2corg/v6_ui/pull/545/files
I have upgraded the demo, please test again :)

shortoland:
Looks good to me, I close the ticket

Issue #403, opened 2016-06-11
Minor problems on outing creation


loicperrin:
- [x] In the first step, the fields are not aligned (Durée, Longueur totale)
- [x] In the first step, "Parcours partiel" should be a checkbox

(- [x] Step 2: date should be pre-filled with today date )

- [x] Step 2: the field "Commentaires sur l'accès" should be located with other fields related to "Accès"
- [x] Step 2: "Transports publics" should be a checkbox
- [x] Step 3: language should be pre-filled with the currently selected language for the interface
- [x] Step 3 : the default title of the outing does not begin with the name of the main waypoint => all new outings will be named "Versant W" or "Face N"

( - [ ] In the first step, the fields should be pre-filled either from the associated route or from the GPS track (when associated) )
( - [ ] Step 2: the "Glacier" field should not appear if the associate route(s) have no "matériel glacier" )

asaunier:
> Step 2: date should be pre-filled with today date

I am not sure this is a good idea. In v5 outings dates are often incorrectly set because people don't realize that the default date (the current date) is not the good one. If no default date is set, people will have to pay attention to the date of the outing. Please note that the calendar is automatically set to today's date when opened the first time.

> Step 3: language should be pre-filled with the language of the navigator

You mean the currently selected language for the interface I guess.

loicperrin:
> You mean the currently selected language for the interface I guess.

Yes ! (I have modified the issue).

For the pre-filled date, it is to discussed indeed.

stef74:
Please add also 

- [ ] Snow information ("quantité de neige", "qualité de la neige", "altitude de la neige à la montée",...) should not appear in activities for which it is not relevant (climbing, hiking, ...). See the google sheet "Outings" for more details.

- [ ] The table "Enneigement par zones" does not appear in creation form for theses activities but appears when displaying the outing.

ginold:
@stef74, "Enneigement par zones" is a default field for every outing, no matter what activity it has? 
according to this sheet https://docs.google.com/spreadsheets/d/110umbpgS1CUDtzeJ62fNBvn9UBHl_wSbXx5xhFlQTng/edit#gid=792641820 and https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_outing.py, it is the case.

ginold:
Snow information does not appear for activities it is not relevant: it only appears when you have selected two activities that have uncommon fields, for example for hiking AND snowshoeing selected, snow information will appear.

[...]

PR #525, opened 2016-07-25
Route and outing editing fixes

### LINKED ISSUE #510: Document creation: incorrect default lang
When creating a new document (route/outing/...), the language selected by default in the editing form is "english" whereas my interface is in french => the default language should be the same than the interface language.

ginold:
related to  #403 - via ferrata glacier gear showing up
related to #403  - associated routes stlye
related to #509 - rating block via ferrata not showing up
fixes #510
related to #514 - glacier_gear not displayed in route view pages
related to  #516 - participant_count field
related to #531

`partial_trip` and `public_transportation` checkboxes left align

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -47,7 +47,7 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @param {angular.Scope} $scope Scope.
  * @param {angular.JQLite} $element Element.
  * @param {angular.Attributes} $attrs Attributes.
- * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
+ * @param {angular.$cookies} $cookies Cookies service.
  * @param {app.Authentication} appAuthentication
  * @param {ngeo.Location} ngeoLocation ngeo Location service.
  * @param {app.Alerts} appAlerts
@@ -58,8 +58,9 @@ app.module.directive('appDocumentEditing', app.documentEditingDirective);
  * @ngInject
  * @export
  */
-app.DocumentEditingController = function($scope, $element, $attrs, gettextCatalog,
-    appAuthentication, ngeoLocation, appAlerts, appApi, authUrl, appDocument) {
+app.DocumentEditingController = function($scope, $element, $attrs, $cookies,
+        appAuthentication, ngeoLocation, appAlerts, appApi, authUrl, appDocument) {
+
 
   /**
    * @type {app.Document}
@@ -108,7 +109,7 @@ app.DocumentEditingController = function($scope, $element, $attrs, gettextCatalo
    * @type {string}
    * @private
    */
-  this.lang_ = $attrs['appDocumentEditingLang'] || gettextCatalog.currentLanguage;
+  this.lang_ = $attrs['appDocumentEditingLang'];
 
   /**
    * @type {angular.Scope}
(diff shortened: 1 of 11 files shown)


asaunier:
I feel a bit uncomfortable about the cookies approach but if you think we cannot use a ``gettextCatalog`` solution, than go for it and merge.

ginold:
Clearly if we can figure out how to do it with `gettextCatalog` it would be better, but I didn't manage to do so. At the end, how does `gettextCatalog` work? isn't it getting the value somewhere from cookies too?

ginold:
I will add 2 small commits for two issues created and we can merge.

tsauerwein:
> At the end, how does gettextCatalog work? isn't it getting the value somewhere from cookies too?

Take a look at [lang.js](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/lang.js#L95-L96):

```
 var lang = this.cookies_.get('interface_lang') ||
      ngeoGetBrowserLanguage(this['langs']) || 'fr';
```

The problem is that the language controller seems to be initialized after you try to access the language. It would be better to wrap the logic which manages the locale into a service, and use this service in the language controller and here.

ginold:
@asaunier I added 2 commits and for me it's ok to be reviewed and merged.

c2corg_ui/static/js/documentediting.js
```diff
@@ -74,6 +75,12 @@ app.DocumentEditingController = function($scope, $element, $attrs, gettextCatalo
   this.ngeoLocation_ = ngeoLocation;
 
   /**
+   * @type {angular.$cookies}
+   * @private
+   */
+  this.cookies_ = $cookies;
```
asaunier: This property is not needed since used in the constructor => you can use ``$cookies`` directly.

c2corg_ui/templates/route/helpers/detailed_route_attributes.html
```diff
@@ -211,3 +211,21 @@
   </div>
   % endif
 </%def>
+
+## GLACIER GEAR
+<%def name="get_route_glacier_gear(route)">\
+  % if route.get('glacier_gear'):
+  <div class="name-icon-value"
+       ng-click="detailsCtrl.openTab($event)">
+      <h4><span translate>gear</span><span class="glyphicon glyphicon-menu-right"></span></h4>
```
asaunier: Indentation is not correct here.

asaunier:
I agree using a lang service would be an appropriate solution but could be done separately so that we don't wait too much to merge the fixes of this PR.
OK for me to merge the PR when the minor comments I just reported are fixed.

c2corg_ui/static/js/documentediting.js
ginold: If I write `$cookies_.get('interface_lang');` i get  `ERROR - variable $cookies_ is undeclared
`

c2corg_ui/static/js/documentediting.js
```diff
@@ -164,7 +165,7 @@ app.DocumentEditingController = function($scope, $element, $attrs, gettextCatalo
       );
     } else if (!this.id_) {
       // new doc lang = user interface lang
-      this.scope[this.modelName]['locales'][0]['lang'] = this.lang_;
+      this.scope[this.modelName]['locales'][0]['lang'] = $cookies_.get('interface_lang');
```
asaunier: Should be ``$cookies``, not ``$cookies_``.

ginold:
Ok, I will be merging pretty soon. I just added one more commit (i promise it's the last one :P) for the issue that has just been created #531  - it was not a big deal to solve it.

asaunier:
:+1:

PR #532, opened 2016-07-27
Move geolocation button near the geonames service in maps

asaunier:
As for now the geolocation button was in the left bottom corner of maps (next to the baselayer selector) whereas the geonames recentering input was at the top of the map.
This PR groups both recentering tools in the top left corner.


tsauerwein:
Looks good.

asaunier:
Thanks :)

PR #537, opened 2016-07-27
Add a lang service

asaunier:
as suggested by @tsauerwein at https://github.com/c2corg/v6_ui/pull/525#issuecomment-235191318

### DIFF
--- a/c2corg_ui/static/js/langservice.js
+++ b/c2corg_ui/static/js/langservice.js
@@ -67,6 +67,11 @@ app.Lang = function($cookies, gettextCatalog, ngeoGetBrowserLanguage, langs,
    * @private
    */
   this.langUrlTemplate_ = langUrlTemplate;
+
+  this.updateLang(
+    this.cookies_.get('interface_lang') ||
+    this.ngeoGetBrowserLanguage_(this.langs_) || 'fr'
+  );
 };
 
 
@@ -81,9 +86,8 @@ app.Lang.prototype.getAvailableLangs = function() {
 /**
  * @return {string}
  */
-app.Lang.prototype.detectLang = function() {
-  return this.cookies_.get('interface_lang') ||
-    this.ngeoGetBrowserLanguage_(this.langs_) || 'fr';
+app.Lang.prototype.getLang = function() {
+  return this.gettextCatalog_.currentLanguage;
 };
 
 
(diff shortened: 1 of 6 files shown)


c2corg_ui/static/js/langservice.js
```diff
@@ -0,0 +1,121 @@
...
+  return this.langs_;
+};
+
+
+/**
+ * @return {string}
+ */
+app.Lang.prototype.detectLang = function() {
+  return this.cookies_.get('interface_lang') ||
+    this.ngeoGetBrowserLanguage_(this.langs_) || 'fr';
```
tsauerwein: Can you do it like in the previous `LangController`: In the constructor get the language (from the cookie/browser/default) and set it on `gettextCatalog` as current language. Then if you need to know the current language you call `langService.getLang()` which returns `gettextCatalog.getCurrentLanguage()`.

asaunier: Do you mean something like 030ff9a ?

tsauerwein:
Thanks for updating, that looks good!

c2corg_ui/static/js/langservice.js
```diff
@@ -0,0 +1,125 @@
...
+  // server side.
+  var d = new Date();
+  d.setFullYear(d.getFullYear() + 1); // today + 1 year
+  this.cookies_.put('interface_lang', lang, {
+    'path': '/',
+    'expires': d
+  });
+
+  if (this.appAuthentication_.isAuthenticated()) {
+    this.api_.updatePreferredLanguage(lang);
```
asaunier: I feel a bit uncomfortable about this code (ported from the lang controller): for authenticated users an "updatePreferredLanguage" request is made to the API for every single page opened by the user, even though the language has not been changed.

tsauerwein: At which point do we load the preferred language from the server?

asaunier: I don't remember. @gberaudo what do you think?

gberaudo: Isn't it retrieved on login?

asaunier:
I have merged the PR. The "updatePreferredLang" issue has been moved at https://github.com/c2corg/v6_ui/issues/540

Issue #511, opened 2016-07-21
Add error message if no WP is associated to a new route


shortoland:
It seems it's mandatory to associate a WP when creating a new route, so could it be possible to display an error message (requesting to add a WP) when this is not done ?

shortoland:
#508

loicperrin:
A route cannot be created with no associated  WP.
When saving such a route, an error message should be displayed (similar to routes without activity), and the route is not saved, until a WP is associated.

stef74:
@loicperrin @shortoland please test

shortoland:
OK, error message added.

loicperrin:
It is OK when saving a route without associated WP: there is an error message.
But once saved, if I remove the WP, no error message is displayed and the route has no more associated WP.

ginold:
you mean - when you go to the view page, you can delete the last associated waypoint ? (not the editing page)

loicperrin:
> you mean - when you go to the view page, you can delete the last associated waypoint ? (not the editing page)

Yes exactly. An error message should be added in this case (as for the main waypoint for example).

stef74:
@loicperrin @shortoland  please test

shortoland:
There is now an error message when deleting the last WP of a route in edition mode.
I haven't been able to remove a WP in a view mode.
OK for me, I close the ticket. Loic, feel free to reopen if needed.

loicperrin:
OK for this. But there is another problem: I create another issue.

PR #538, opened 2016-07-27
add error msg when no wp is associated to route (creation)

ginold:
related to #511

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -446,8 +446,8 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
       hasError = (!doc['locales'] || !doc['locales'][0][field]);
     } else if (field === 'activities') {
       hasError = (!doc['activities'] || doc['activities'].length === 0);
-    } else if (field === 'routes') {
-      hasError = (!doc['associations'] || doc['associations']['routes'].length === 0);
+    } else if (field === 'routes' || field === 'waypoints') {
+      hasError = (!doc['associations'] || doc['associations'][field].length === 0);
     } else if (field === 'latitude' || field === 'longitude') {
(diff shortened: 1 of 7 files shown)


c2corg_ui/static/js/documentediting.js
```diff
@@ -445,17 +445,21 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
...
     } else if (field === 'date_start') {
       hasError = (doc['date_start'] === null || doc['date_start']  === undefined);
     } else {
       hasError = (!doc[field] || doc[field] === null || doc[field] === undefined);
     }
+
     if (hasError) {
       if (showError) {
+        if (field === 'routes' || field === 'waypoints') {
+          field = 'no associated ' + field;
```
asaunier: How do we translate this?
Currently the message is probably never added to the locale files.

ginold: well, we have to add "no associated waypoints" and "no associated routes". Or "no associated" xxx.

asaunier: "no associated" is not possible because the order of the words changes according to the language.

See also https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/i18n.html

c2corg_ui/templates/i18n.html
```diff
@@ -118,6 +118,9 @@
 <span translate>You have no rights to edit this document.</span>
 <span translate>Form is not valid</span>
 <span translate>You must log in to edit this document.</span>
+<span translate>no associated routes</span>
```
asaunier: It's probably better to use a capital initial letter for those messages (in the translations as well).

ginold: in this case I think it's ok - the message will appear as "Missing field: no associated route" and looks like a normal sentence. In contrast to the other messages with capital letters that are ugly, like "Missing field: Date start".

asaunier:
Looks good. Please merge.

PR #541, opened 2016-07-28
Outing date

ginold:
related to #406

using moment.js all across the app.

### DIFF
--- a/c2corg_ui/externs/moment.js
+++ b/c2corg_ui/externs/moment.js
@@ -1,0 +1,28 @@
+/**
+ * @fileoverview Externs for moment
+ *
+ * @externs
+ */
+
+
+/**
+ * @param {Object | string} date
+ */
+function moment(date) {};
+
+
+/**
+ * Formats the date in given format, for ex. 'YYYY-MM-DD'
+ * @param {string} format
+ */
+moment.prototype.format = function(format) {};
+
+/**
+ * Difference in miliseconds between two dates
+ * @param {Object | string} date
+ * @return {Number}
+ */
+moment.prototype.diff = function(date) {};
+
+/** @return {(Date | string)} */
+moment.prototype.toDate = function() {};
--- a/c2corg_ui/static/js/outingediting.js
+++ b/c2corg_ui/static/js/outingediting.js
@@ -5,7 +5,6 @@ goog.require('app.DocumentEditingController');
 goog.require('app.Alerts');
 goog.require('app.Document');
 goog.require('app.Lang');
-goog.require('app.utils');
 
 
 /**
@@ -96,7 +95,7 @@ app.OutingEditingController.prototype.successRead = function(response) {
   // check if user has right to edit -> the user is one of the associated users
   if (this.auth.hasEditRights(outing['associations']['users'])) {
     outing = this.formatOuting_(outing);
-    this.differentDates = app.utils.areDifferentDates(outing['date_start'], outing['date_end']);
+    this.differentDates = window.moment(outing['date_start']).diff(outing['date_end']) !== 0;
     if (!this.differentDates) {
       outing['date_end'] = undefined;
     }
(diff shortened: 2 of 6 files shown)


c2corg_ui/externs/moment.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Object | string} date
+ */
+function moment(date) {};
+
+
+/**
+ * Formats the date in given format, for ex. 'YYYY-MM-DD'
+ * @param {string} format
+ */
+moment.prototype.format= function(format){};
```
asaunier: blank space before the = sign (ditto below)

c2corg_ui/static/js/outingediting.js
```diff
@@ -133,37 +132,39 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
...
+    // convert existing date from string to a date object
+    if (outing.date_end && typeof outing.date_end === 'string') {
+      outing.date_end = window.moment(outing.date_end)._d;
+      this.dateMaxStart = outing.date_end;
+    }
+    if (outing.date_start && typeof outing.date_start === 'string') {
+      outing.date_start = window.moment(outing.date_start)._d;
+      this.dateMinEnd = outing.date_start;
+    }
+    console.log(outing.date_start, outing.date_end);
```
asaunier: remove debug (the "make lint" command should tell you)

c2corg_ui/static/js/search/outingfilters.js
```diff
@@ -64,7 +63,7 @@ app.OutingFiltersController.prototype.getFilterFromPermalink = function(key) {
     }
     var dates = val.split(',');
     dates.forEach(function(date) {
-      this.dates.push(app.utils.formatDate(date));
+      this.dates.push(window.moment(date)._d);
```
tsauerwein: `_d`? Are you accessing some internal property of moment.js?

ginold: yes, it's the Date object.

as defined in extern

```
/** @type {(Date | string)} */
moment.prototype._d;
```

tsauerwein: You could do it some other way? Relying on an internal property is always fragile.

ginold: why is it fragile? 

I could use `moment.toDate()`

tsauerwein: `_d` is not exposed as API property/method from the authors of moment.js. So, they can simply rename/remove/change this property in a next release.

ginold:
@asaunier could you have a look at this PR? I would need it for the one i'm currently working on.

c2corg_ui/static/js/outingediting.js
```diff
@@ -96,7 +95,7 @@ app.OutingEditingController.prototype.successRead = function(response) {
   // check if user has right to edit -> the user is one of the associated users
   if (this.auth.hasEditRights(outing['associations']['users'])) {
     outing = this.formatOuting_(outing);
-    this.differentDates = app.utils.areDifferentDates(outing['date_start'], outing['date_end']);
+    this.differentDates = (window.moment(outing['date_start']).diff(outing['date_end']) === 0) ? false : true;
```
asaunier: What about simply writing
```
this.differentDates = window.moment(outing['date_start']).diff(outing['date_end']);
```
?

asaunier:
A minor code optimization suggestion (do as you think is the best) else looks good to me.

c2corg_ui/static/js/outingediting.js
ginold: I first have to check out what `!!` means and i will tell you :P

asaunier: It's a way to convert some value to a boolean value.

ginold: if !! means a double negation, and `window.moment(outing['date_start']).diff(outing['date_end']);` == `!!window.moment(outing['date_start']).diff(outing['date_end']);`, what's the deal?

asaunier: It's a common way in javascript to make sure something is actually a boolean.
Of course ``!!false`` will evaluate to ``false``.
But it also means that ``!!''`` (empty string) or ``!!0`` will evaluate to ``false``.

What does ``window.moment(outing['date_start']).diff(outing['date_end'])`` return? If it is already a boolean, you don't need the ``!!``. Else it depends on the rest of your code, especially of the type of ``this.differentDates``. Note sure Closure compiler will be happy if ``this.differentDates`` is a boolean and you set it to a number or a string.

The thing is to avoid having expression like ``condition ? true : false`` or ``condition ? false : true`` which are a bit of a non-sense since ``condition`` is already a boolean.

ginold:
updated and corrected, I will merge when travis passes.

Issue #365, opened 2016-05-25
Upgrade of components?


asaunier:
Pyramid 1.7 went out a few days ago:
http://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-1.7.html
Would it be worth upgrading (so that c2corg UI and API stick to the latest versions)?

Same question for AngularJS whose latest version if 1.5.5 whereas we use 1.5.0.

gberaudo:
We might be affected by the change regarding exception views. +1 for trying the update.

asaunier:
Pyramid has not been upgraded yet but we now use AngularJS 1.5.6 (see https://github.com/c2corg/v6_ui/pull/451)

fjacon:
What is the status of this issue ?
Should we freeze now the components ?

tsauerwein:
Can this one be closed now?

asaunier:
Yes

PR #542, opened 2016-07-28
Upgrade to pyramid 1.7

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/365

As far as I can see, I have not observed broken pages or features.

In https://github.com/c2corg/v6_ui/issues/365#issuecomment-221582770 @gberaudo expected that
> We might be affected by the change regarding exception views. +1 for trying the update.

Not sure what exception views are. At least the 404 page still works.


tsauerwein:
Thanks for testing, please merge.

PR #543, opened 2016-07-28
Fix templatecache issue in debug mode

### LINKED ISSUE #369: Blank cards in debug mode

asaunier:
With the precious help of @sbrunner

A synchronous request is made in debug mode to get the partial in the link function. The browser complains synchronous requests are evil:
> Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.

but it's only in debug mode. Prod mode stays (almost) the same.

Fixes #369 

Fixes both the cards and the simple search (suggestions) in debug mode.

### DIFF
--- a/c2corg_ui/static/js/card.js
+++ b/c2corg_ui/static/js/card.js
@@ -15,7 +15,8 @@ goog.require('app.utils');
  */
 app.cardDirective = function($compile, $templateCache) {
   var template = function(doctype) {
-    return $templateCache.get('/static/partials/cards/' + doctype + '.html');
+    var path = '/static/partials/cards/' + doctype + '.html';
+    return app.utils.getTemplate(path, $templateCache);
   };
 
   return {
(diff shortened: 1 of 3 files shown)


tsauerwein:
Thanks for fixing! :+1:

Issue #471, opened 2016-06-29
Handle waypoint/route/user id params in advanced search results


asaunier:
(extracted from https://github.com/c2corg/v6_ui/pull/460)

As for now when opening the advanced search pages, a ``bbox`` param is automatically appended to the URL.

When filtering on a document id (eg. to get the list of routes associated to a given waypoint), the map should be recentered on the document id (or on the resulting features extent) and only then the bbox param may be added.

PR #544, opened 2016-07-28
Recenter map when searching using a doc id

asaunier:
When searching using a document id (eg. ``/routes?w=37355``), the initial map extent should not be taken into account unless explicitely provided via a ``bbox`` parameter and the map should be recentered on the search results.

Map extent changes resulting from zooming/panning are then taken into account as filter as usual. Ditto if the initial page URL contains a ``bbox`` param => the map extent is taken into account as filter as usual and no recentering on results is performed.

Fixes https://github.com/c2corg/v6_ui/issues/471

### DIFF
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -321,7 +321,10 @@ app.MapController.DEFAULT_POINT_ZOOM = 12;
  */
 app.MapController.prototype.recenterOnExtent_ = function(extent, options) {
   var mapSize = this.map.getSize();
-  if (mapSize) {
+  if (!mapSize || !ol.extent.getWidth(extent) || !ol.extent.getHeight(extent)) {
+    this.view_.setCenter(ol.extent.getCenter(extent));
+    this.view_.setZoom(this.zoom || app.MapController.DEFAULT_POINT_ZOOM);
+  } else {
     options = options || {};
     this.view_.fit(extent, mapSize, options);
   }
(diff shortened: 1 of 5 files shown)


asaunier:
The PR is ready to be reviewed.

asaunier:
I have noticed a bug when recentering on a set of features with exactly the same geometry (eg. outings inheriting the same route/WP point geom). In that case, the features extent is 0-sized and the recentering on it fails. I have changed so that the point recentering fallback is used then.

c2corg_ui/static/js/map/map.js
```diff
@@ -313,7 +321,10 @@ app.MapController.DEFAULT_POINT_ZOOM = 12;
  */
 app.MapController.prototype.recenterOnExtent_ = function(extent, options) {
   var mapSize = this.map.getSize();
-  if (mapSize) {
+  if (!mapSize || !ol.extent.getWidth(extent) || !ol.extent.getHeight(extent)) {
```
tsauerwein: [ol.extent.isEmpty](http://openlayers.org/en/master/apidoc/ol.extent.html#.isEmpty) might be useful.

tsauerwein:
LGTM

c2corg_ui/static/js/map/map.js
tsauerwein: Ah no, forget the last comment. You want to check if you have a point.

asaunier:
Thanks @tsauerwein for the review!

PR #549, opened 2016-08-03
Update of image-slides

### LINKED ISSUE #478: Images slideshow defaults
* [x] activities and categories values and other metadatas (aperture, shutter speed etc) are lost after reloading the page. They appear first in the template, but on a reload they are lost.

ginold:
fixes partly  #478

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -439,17 +439,31 @@ app.Api.prototype.createImages = function(files, document) {
   var associations = {};
   associations[documentType] = [{'document_id': document.document_id}];
   var images = [];
+
   for (var i = 0; i < files.length; i++) {
-    var image = files[i]['metadata'];
-    image['size'] = files[i]['size'];
+    var file = files[i];
+    var meta = file['metadata'];
+    var date = meta['DateTime'] ? meta['DateTime'].substring(0, 10).split(':') : null; // convert "2015:12:31 20:56:09"
+    var image = meta;
+
+    image['categories'] = meta['categories'];
+    image['activities'] = meta['activities'];
+    image['file_size'] = file['size'];
+    image['exposure_time'] = meta['ExposureTime'];
     image['associations'] = associations;
     image['locales'] = [{'lang': document.lang, 'title': image['title']}];
+    image['date_time'] = date ? date[0] + '-' + date[1] + '-' + date[2] : window.moment(new Date()).format('YYYY-MM-DD');
+    image['iso_speed'] = meta['PhotographicSensitivity'];
+    image['focal_legth'] = meta['FocalLengthIn35mmFilm'];
+    image['fnumber'] = meta['FNumber'];
+    image['camera_name'] = (meta['Make'] && meta['Model']) ? (meta['Make'] + ' ' + meta['Model']) : null;
+
     images.push(image);
   }
-  var json = {'images': images};
 
-  var promise = this.postJson_('/images/list', json);
+  var promise = this.postJson_('/images/list', {'images': images});
   promise.catch(this.errorSaveDocument_.bind(this));
+  promise['images'] = images;
   return promise;
 };
 
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -216,28 +216,33 @@ app.ImageUploaderController.prototype.uploadFile_ = function(file) {
   angular.extend(file, {
     'src': app.utils.getImageFileBase64Source(file),
     'progress': 0,
+    'processed': false,
     'metadata': {
-      'title': file['name'],
-      'id': file['name'] + '-' + new Date().toISOString()
+      'id': file['name'] + '-' + new Date().toISOString(),
+      'activities': [],
+      'categories': []
     }
   });
-
   this.getImageMetadata_(file);
 
-  file['processed'] = false;
   var canceller = this.q_.defer();
   var promise = this.api_.uploadImage(file, canceller.promise, function(file, event) {
     var progress = event.loaded / event.total;
     file['progress'] = 100 * progress;
   }.bind(this, file));
+
   file['uploading'] = promise;
   file['canceller'] = canceller;
 
   promise.then(function(resp) {
+    var image = new Image();
+    image['src'] = file['src'];
+
     file['metadata']['filename'] = resp['data']['filename'];
     file['processed'] = true;
+
   }.bind(this), function(resp) {
-    if (resp.status == -1) {
+    if (resp.status === -1) {
       if (!file['manuallyAborted']) {
         this.alerts_.addError(this.alerts_.gettext('Error while uploading the image:') + ' Timeout');
       }
@@ -278,33 +283,24 @@ app.ImageUploaderController.prototype.save = function() {
   var defer = this.q_.defer();
 
   this.api_.createImages(this.files, this.documentService.document)
-  .then(function() {
-    var meta;
+  .then(function(data) {
     var id;
+    var images = data['config']['data']['images'];
+    var imageIds = data['data']['images']; // newly created document_id
 
-    $('.img-container').each(function(i, image) {
-      meta = this.files[i]['metadata'];
+    $('.img-container').each(function(i) {
       id = 'image-' + (+new Date());
-      this.files[i]['document_id'] = id;
+      images[i]['image_id'] = id;
 
-      var element = app.utils.createImageSlide(this.files[i], this.imageUrl_);
-      $('.photos').slick('slickAdd', element);
+      var element = app.utils.createImageSlide(images[i], this.imageUrl_);
+      $('.photos').slick('slickAdd', element, true);
 
       var scope = this.scope_.$new(true);
-      scope['photo'] = {
-        'filename' : meta['filename'],
-        'locales' : [{'title': meta['title']}],
-        'date_time': meta['DateTime'],
-        'activities' : meta['activities'],
-        'iso_speed' : meta['PhotographicSensitivity'],
-        'image_type' : meta['image_type'],
-        'fnumber' : meta['FocalLength'],
-        'camera_name' : meta['Make'] + ' ' + meta['Model'],
-        'categories': meta['categories'],
-        'document_id': id
-      };
+      scope['photo'] = images[i];
+      scope['photo']['image_id'] = id;
+      scope['photo']['edit_url'] = '/images/edit/' + imageIds[i]['document_id'] + '/' + images[i]['locales'][0]['lang'];
       this.documentService.document.associations['images'].push(scope['photo']);
-      this.compile_($('#image-' + id).contents())(scope);
+      this.compile_($('#' + id).contents())(scope);
     }.bind(this));
 
     defer.resolve();
--- a/c2corg_ui/static/js/slideinfo.js
+++ b/c2corg_ui/static/js/slideinfo.js
@@ -4,9 +4,11 @@ goog.require('app');
 
 
 /**
+ * @param {angular.$compile} $compile Angular compile service.
  * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
  */
-app.slideInfoDirective = function() {
+app.slideInfoDirective = function($compile) {
   return {
     restrict: 'E',
     controller: 'AppSlideInfoController as slideCtrl',
@@ -15,6 +17,7 @@ app.slideInfoDirective = function() {
     templateUrl: '/static/partials/slideinfo.html',
     link: function(scope, el) {
       angular.extend(scope, scope.$parent['photo']);
+      $compile(el.contents())(scope);
     }
   };
 };
@@ -43,37 +46,6 @@ app.SlideInfoController = function(appApi, $scope) {
    * @private
    */
   this.scope_ = $scope;
-
-  /**
-   * @type {boolean}
-   * @export
-   */
-  this.editing = false;
 };
-
-
-/**
- * @export
- */
-app.SlideInfoController.prototype.edit = function() {
-  this.editing = true;
-};
-
-
-/**
- * @export
- */
-app.SlideInfoController.prototype.save = function() {
-  this.editing = false;
-};
-
-
-/**
- * @export
- */
-app.SlideInfoController.prototype.cancel = function() {
-  this.editing = false;
-};
-
 
 app.module.controller('AppSlideInfoController', app.SlideInfoController);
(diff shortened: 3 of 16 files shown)


c2corg_ui/static/js/utils.js
```diff
@@ -163,13 +163,17 @@ app.utils.createImageSlide = function(file, imageUrl) {
...
+
     ahref = '<a href="' + imageUrl + file.filename + '" data-info-id="' + file['document_id'] + '">';
-    img = '<img src="' + imageUrl + file.filename + '">';
+    img = '<img src="' + imageUrl + smallImage + '">';
   } else {
     ahref = '<a href="' + file['src'] + '" data-info-id="' + file['document_id'] + '">';
     img = '<img src=" ' + file['src'] + '">';
   }
-  return '<figure id="image-' + file['document_id'] + '">' +
+  return '<figure id="image-' + file['document_id'] + '")">' +
```
asaunier: Where is the opening ")" ?

ginold:
this pr is ready, only 2 upgrades that could be done (but i think they require another pr to do so) are :
- save image dimensions (height x width pixels) so I can display correctly the width and height, adapted to the viewport (I could be done by the plugin, but we are using our custom template...). 
When you first load the view page, the small thumbnails are loaded, then when you click on the image, it will download the big one, but the function that gets the width and height of the image is too late... it would be clearly better if this value was stored.

As from now, I calculate the `image.naturaWidth` and `image.natural height` on the client side but it's a bit tricky and expensive, if it could be done on the server side, I think it would be better.

- there is a problem when uploading more than one image at a time :
```

document = self._create_document(document_in, schema_image)
File "/home/agina/v6_api/c2corg_api/views/document.py", line 253, in _create_document
Display the sourcecode for this frameOpen an interactive python shell in this framecreate_associations(document, document_in['associations'], user_id)
File "/home/agina/v6_api/c2corg_api/models/association.py", line 315, in create_associations
Display the sourcecode for this frameOpen an interactive python shell in this framefor doc in docs:
TypeError: 'NoneType' object is not iterable
```

asaunier:
> the function that gets the width and height of the image is too late... it would be clearly better if this value was stored.

> As from now, I calculate the image.naturaWidth and image.natural height on the client side but it's a bit tricky and expensive, if it could be done on the server side, I think it would be better.

Aren't the image width and height provided by the API? At least they are available in the model: https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/image.py#L43-L45
Those are the original image dimensions but you can compute the dimensions of the resized images knowing the maximal length of the biggest image dimension and applying the ratio to the stored dimensions.

c2corg_ui/static/js/utils.js
```diff
@@ -135,25 +135,28 @@ app.utils.getImageFileBase64Source = function(file) {
  * @return {string}
  * @export
  */
-app.utils.createImageSlide = function(file, imageUrl) {
+app.utils.createImageSlide = function (file, imageUrl) {
```
asaunier: This blank space is not needed, see the other functions of the file :)

c2corg_ui/static/js/slideinfo.js
```diff
@@ -4,17 +4,26 @@ goog.require('app');
...
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {angular.$templateCache} $templateCache service
  * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
  */
-app.slideInfoDirective = function() {
+app.slideInfoDirective = function($compile, $templateCache) {
+
+  var template = function() {
+    return app.utils.getTemplate('/static/partials/slideinfo.html', $templateCache);
```
asaunier: :+1:

ginold:
Well, that could be a solution. When we define the % of the resize, it could be probably done.
For now, the width and height are not required and if not provided, they are null. See /images/736998

c2corg_ui/static/js/slideinfo.js
asaunier: But just curious: why can't you directly use ``templateUrl`` here?

c2corg_ui/static/partials/imageuploader.html
```diff
@@ -20,8 +20,9 @@
         <!-- EXIF INFO-->
         <div class="exif" ng-style="file.metadata.FocalLength && {'opacity': 1} || {'opacity' : 0}">
           <p ng-if="file.metadata.DateTime"><span class="glyphicon glyphicon-calendar"></span><span>{{file.metadata.DateTime}}</span></p>
-          <p><span class="glyphicon glyphicon-certificate"></span><span>{{file.metadata.FocalLength}}f</span>&nbsp;
-            <span ng-if="file.metadata.ShutterSpeedValue">1/{{file.metadata.ShutterSpeedValue}}s</span> &nbsp; 
+          <p><span class="glyphicon glyphicon-certificate"></span><span>f/{{file.metadata.FNumber}}</span>&nbsp;
+            <span ng-if="file.metadata.ExposureTime">1/{{1/file.metadata.ExposureTime}}s</span> &nbsp;
```
asaunier: not breaking space before the unit

c2corg_ui/static/partials/imageuploader.html
```diff
@@ -20,8 +20,9 @@
...
+            <span ng-if="file.metadata.FocalLengthIn35mmFilm">{{file.metadata.FocalLengthIn35mmFilm}}mm</span> &nbsp;
```
asaunier: not breaking space before the unit

c2corg_ui/static/partials/imageuploader.html
```diff
@@ -20,8 +20,9 @@
...
             <span><b>ISO:</b> <span>{{file.metadata.PhotographicSensitivity}}</span></span>
```
asaunier: I would rather put the unit after the value:
```
{{file.metadata.PhotographicSensitivity}}&nbsp;ISO
```
By the way why do you need the embedded ``<span>``?

asaunier: Also: not sure this ISO data is available (for instance if the image is a drawing) => don't display the info line if the info is not available.

c2corg_ui/static/partials/slideinfo.html
```diff
@@ -1,42 +1,46 @@
...
+      <ul>
+        <li ng-repeat="category in categories">{{category | translate}}</li>
+      </ul>
+    </div>
+
+    <div ng-show="iso_speed || focal_length || exposure_time || camera_name" class="settings">
+      <label translate>Settings</label>
+      <ul>
+        <li ng-show="camera_name"><span>{{camera_name}}</span></li>
+        <li ng-show="exposure_time" uib-tooltip="{{'exposure_time' | translate}}"><span>1/{{exposure_time}}s</span></li>
```
asaunier: blank (not breaking) space before units

asaunier:
> Well, that could be a solution. When we define the % of the resize, it could be probably done.
For now, the width and height are not required and if not provided, they are null. See /images/736998

Required by who?
It's should be automatically set by the image uploading system, not by the user. So the info should always be available.

It seems that the info is not available for all v5 images:
```
c2corg_alex=# select count(*) from guidebook.images where width is null or height is null;
 count  
--------
 224633
```
but perhaps we can make sure the migration script of images actually update these data?

c2corg_ui/static/js/slideinfo.js
ginold: I thought `getTemplate` function was kind of special and we should use it everywhere, because before it didn't work in debug mode.

ginold:
>  Required by who? 
By the API, it's an optional field. 
If it should be set on the server side, then it would greatly lighten the client processing.

>  but perhaps we can make sure the migration script of images actually update these data?
That's what i'm suggesting. In addition, we have to regenerate the thumbnails, so we could also add the real width and height to the DB, in the meantime.

c2corg_ui/static/js/slideinfo.js
asaunier: The recommended way is to use the usual ``template`` or ``templateUrl`` directive parameters whenever possible (standard angular way).
For cards and suggestions (simple search) it was not possible because the partial to use depends on a variable.
In the current case, it seems you don't need to do so. So it's probably better to use ``templateUrl``.

asaunier:
> If it should be set on the server side, then it would greatly lighten the client processing.

The user should not be prompted for it => it should be detected or computed by the upload system.

> we have to regenerate the thumbnails, so we could also add the real width and height to the DB, in the meantime.

I have created an issue here: https://github.com/c2corg/v6_images/issues/7

tsauerwein:
@ginold Why do you need the image sizes? For the thumbnails you know the exact size. And for the other images, can't you work with relative sizes (e.g. `100%`)?

ginold:
>   can't you work with relative sizes (e.g. 100%)?

Maybe something is wrong with my CSS, but when I use width: 100% or height: 100% it will stretch the image to fill the whole screen, so I have to use only one or the other, depending on the image dimensions. see https://github.com/c2corg/v6_ui/blob/8e702400fa347104f2413302c7bf342f925d2911/c2corg_ui/static/js/viewdetails.js#L153

c2corg_ui/static/js/imageuploader.js
```diff
@@ -278,31 +286,22 @@ app.ImageUploaderController.prototype.save = function() {
...
   this.api_.createImages(this.files, this.documentService.document)
-  .then(function() {
+  .then(function(data) {
     var meta;
     var id;
+    var images = data['config']['data']['images'];
 
     $('.img-container').each(function(i, image) {
-      meta = this.files[i]['metadata'];
+      meta = images[i]['metadata'];
```
asaunier: There's probably a problem with that change, or at least with this piece of code.

When uploading an image and pressing the "Save" button, the console warns me that
> angular.js:13708 TypeError: Cannot set property 'document_id' of undefined
    at .<anonymous> (imageuploader.js:304)

=> ``meta`` is undefined. It seems there is no ``metadata`` property in ``images[i]``.
Perhaps the correct code would rather be
```
meta = images[i]
```
?
At least it seems to work in my instance.

c2corg_ui/static/js/imageuploader.js
```diff
@@ -278,31 +286,22 @@ app.ImageUploaderController.prototype.save = function() {
...
       id = 'image-' + (+new Date());
-      this.files[i]['document_id'] = id;
+      images[i]['document_id'] = id;
```
asaunier: Is it correct to set a ``document_id`` as ``'image-' + (+new Date())``? It doesn't seem consistent with the API side document numbering? What is this ``document_id`` used for?

c2corg_ui/static/js/imageuploader.js
asaunier: > Perhaps the correct code would rather be

I have pushed a fixup commit about this issue, see 4cb360c
It seems to work but it would be cool if someone could doublecheck.

asaunier:
> there is a problem when uploading more than one image at a time :

I have not observed this error when uploading for instance 2 images at a time. Has it been fixed in the meantime?

Here are other remarks and questions:
* the thumbnails use the small image version ("SI") which is cool, but it is stretched to fill the square container.
* the slideshow shows the original images. Which is cool for the quality of the displayed images. On the other hand those images can be quite big (several MB), so the loading speed might be a bit slow, especially on a phone. In v5, the slideshow uses an intermediate size ("BI" = big image). Should we do the same in v6?
* what is the order of the images in the slideshow? First push first shown? After being uploaded I notice that the images are shown at the end of the slideshow but if I reload the page, they are shown at the beginning!?
* the first time they are displayed, portrait images are stripped/cropped to fit the page width (the bottom of the image is missing). OK if closing and reopening the slideshow.
* I don't think it's a good idea to use the original filename as the default image title. It's better to keep the title empty (and to not  make it a required attribute).

tsauerwein:
> the slideshow shows the original images. Which is cool for the quality of the displayed images. On the other hand those images can be quite big (several MB), so the loading speed might be a bit slow, especially on a phone. In v5, the slideshow uses an intermediate size ("BI" = big image). Should we do the same in v6?

Oh yes! People will upload 6MB photos, we definitely want to use a down-scaled version.

ginold:
> I have not observed this error when uploading for instance 2 images at a time. Has it been fixed in the meantime?

The error appears when you want to save multiple images at once. No problem with uploading and seeing them in the image uploader though.

ginold:
i have updated the PR - i think i found a way to resize images inside its container to their max-width and height accordingly, but it's hard to test because there are no images and I somehow cannot upload new ones. There is an error message 'could not save changes' but nothing returned in the response ;/

c2corg_ui/static/js/apiservice.js
```diff
@@ -439,17 +439,33 @@ app.Api.prototype.createImages = function(files, document) {
...
+    var meta = file['metadata'];
+    var date = meta['DateTime'] ? meta['DateTime'].substring(0, 10).split(':') : null; // convert "2015:12:31 20:56:09"
+    var image = meta;
+
+    image['categories'] = meta['categories'];
+    image['activities'] = meta['activities'];
+    image['file_size'] = file['size'];
+    image['width'] = file['width'];
+    image['height'] = file['height'];
+    image['exposure_time'] = parseInt(1 / meta['ExposureTime'], 10);
```
ginold: remove the division

c2corg_ui/static/js/imageuploader.js
```diff
@@ -216,28 +216,36 @@ app.ImageUploaderController.prototype.uploadFile_ = function(file) {
...
   file['uploading'] = promise;
   file['canceller'] = canceller;
 
   promise.then(function(resp) {
+    var image = new Image();
+    image['src'] = file['src'];
+
     file['metadata']['filename'] = resp['data']['filename'];
     file['processed'] = true;
+    file['width'] = image.naturalWidth;
```
ginold: no need?

c2corg_ui/static/js/slideinfo.js
```diff
@@ -4,9 +4,12 @@ goog.require('app');
 
 
 /**
+ * @param {angular.$compile} $compile Angular compile service.
+ * @param {angular.$templateCache} $templateCache service
  * @return {angular.Directive} Directive Definition Object.
+ * @ngInject
  */
-app.slideInfoDirective = function() {
+app.slideInfoDirective = function($compile, $templateCache) {
```
ginold: remove templatecache

c2corg_ui/static/js/imageuploader.js
ginold: well, in this case it has nothing to do with the 'real' document_id, it's a randomly generated on the fly just to render things correctly ui-side.

asaunier: It's a bit confusing to use ``document_id`` though. Couldn't you use another property name?

PR #550, opened 2016-08-03
Image editing

ginold:
image.author = should it be a text area or a username, searched with app-search tool?

should it be possible to associate routes and wps or just to list them?

### DIFF
--- a/c2corg_ui/static/js/constants.js
+++ b/c2corg_ui/static/js/constants.js
@@ -31,7 +31,8 @@ app.constants = {
   REQUIRED_FIELDS : {
     waypoints: ['title' , 'lang', 'waypoint_type', 'elevation', 'longitude', 'latitude'],
     routes : ['title' , 'lang', 'activities', 'waypoints'],
-    outings : ['title' , 'lang', 'date_start', 'routes', 'activities']
+    outings : ['title' , 'lang', 'date_start', 'routes', 'activities'],
+    images: ['image_type']
   },
   documentEditing: {
     FORM_PROJ: 'EPSG:4326',
--- a/c2corg_ui/templates/image/edit.html
+++ b/c2corg_ui/templates/image/edit.html
@@ -1,0 +1,312 @@
+<%!
+from c2corg_common.attributes import default_langs, activities, quality_types, image_types, image_categories \
+%>
+<%inherit file="../base.html"/>
+<%
+  updating_doc = image_id and image_lang
+%>
+<%namespace file="../helpers/common.html" import="show_title"/>
+
+<%namespace file="../helpers/edit.html" import="show_editing_associated_waypoints, show_editing_associated_routes"/>
+
+
+<%block name="pagetitle">
+<title ng-init="id = ${image_id if image_id else 0}" ng-bind="id ?
+          mainCtrl.page_title('Editing an image') : mainCtrl.page_title('Creating an image')">${show_title('Create/edit image')}
+</title>
+</%block>
+
+<%block name="metarobots"><meta name="robots" content="noindex,nofollow"></%block>
+
+<section class="create-edit-document">
+
+  <h1 ng-init="id = ${image_id if image_id else 0}" class="text-center">
+    <label class="label label-primary" ng-bind="id ? ('Editing an image' | translate) : ('Creating an image' | translate)"></label>
+  </h1>
+
+  <form app-loading app-document-editing="images" app-document-editing-model="image"
+    app-document-editing-controller-name="appDocumentEditingController"
+  % if updating_doc:
+    app-document-editing-id="${image_id}" app-document-editing-lang="${image_lang}"
+  % endif
+    class="document-editing" name="editForm" novalidate ng-submit="editCtrl.submitForm(editForm.$valid)">
+
+    ## PROGRESS BAR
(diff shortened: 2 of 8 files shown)


asaunier:
> image.author = should it be a text area or a username, searched with app-search tool?

This is a text attribute: https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/image.py#L37
=> it should be a text input in the editing form.

This attribute is used to quote the actual author(s) of an image when the user that uploads the image is not the actual author.

In v5, when this attribute is not set, the user that has initially uploaded the image is marked as "author". See for instance http://www.camptocamp.org/images/783717/fr/l2-tonton-coinceurs

When the "author" attribute is set, its value is used as the "author", whereas the user that has initially uploaded the image appears in "image uploaded by". For instance http://www.camptocamp.org/images/208579/fr/face-n-des-grandes-jorasses-quelques-itineraires

asaunier:
> should it be possible to associate routes and wps or just to list them?

One should be able to associate documents (any type - not only WP or routes) from an image detail view page. As in v5. See http://www.camptocamp.org/images/208579/fr/face-n-des-grandes-jorasses-quelques-itineraires#associated_docs

ginold:
>  One should be able to associate documents (any type - not only WP or routes) from an image detail view 

what about the association on an editing page, as other document do ?

asaunier:
> what about the association on an editing page, as other document do ?

I guess so. That would be consistent.

ginold:
So the PR is pretty close to ready, the only things to fix: 

- when saving the date and coming back to editing page, the save button is grayed out, but if you clear or select a new date, it's ok (probably because the date is not  YYYY-MM-DD formated).

- Didn't manage yet to associate outings (wps and routes :+1: )

c2corg_ui/static/js/constants.js
```diff
@@ -31,7 +31,8 @@ app.constants = {
   REQUIRED_FIELDS : {
     waypoints: ['title' , 'lang', 'waypoint_type', 'elevation', 'longitude', 'latitude'],
     routes : ['title' , 'lang', 'activities', 'waypoints'],
-    outings : ['title' , 'lang', 'date_start', 'routes', 'activities']
+    outings : ['title' , 'lang', 'date_start', 'routes', 'activities'],
+    images: ['title', 'image_type']
```
asaunier: I think that would be more relevant that ``title`` is not a mandatory attribute for images (as on Facebook for instance). In v5 many people uses dummy strings as image titles => that would be more relevant to keep the title blank...

c2corg_ui/templates/image/edit.html
```diff
@@ -0,0 +1,292 @@
...
+
+          ## FILENAME
+          <div id="title-group" class="form-group">
+            <label translate>filename</label>
+            <input disabled  type="text" name="filename" ng-model="image.filename" class="form-control" />
+          </div>
+
+
+            ## ACTIVITIES
+            <div class="form-group full" ng-init="activities = ${activities}">
```
asaunier: Incorrect indent

c2corg_ui/templates/image/edit.html
```diff
@@ -0,0 +1,262 @@
...
+
+            ## SUMMARY
+            <div id="summary-group" class="form-group full">
+              <label translate>summary</label>
+              <textarea name="summary" ng-model="image.locales[0].summary" class="form-control"></textarea>
+            </div>
+
+            ## QUALITY
+            <div class="form-group half" ng-style="{'has-success': image.quality}" ng-init="qualityTypes = ${quality_types}">
+              ## I tried with ng-init, ng-selected..so hacking
```
asaunier: What are you trying to do?

asaunier:
Why limiting the PR to the editing?
Shouldn't it also manage the detail view and at least a simple search/listing page?

Please note that changes in the python views will be required when the caching is integrated https://github.com/c2corg/v6_ui/pull/552

ginold:
> Shouldn't it also manage the detail view and at least a simple search/listing page?

I'm surprised you say that, because you always insisted for the PRs to be the most concise possible, thus  I would not think of including Editing + Views  + search page in a single PR ! :P

c2corg_ui/templates/image/edit.html
ginold: apparently "if there is `image.quality`, show ` image.quality`, if not -> show `qualityTypes[1]` by default".

asaunier:
> I'm surprised you say that, because you always insisted on the PRs to be the most concise possible, 

Sure. But the PR must be consistent. What's the use of adding an editing interface if it is not possible to view the edited document?
In my opinion it would be more relevant to start working on the detail view first. And then add the possibility to edit the document.

tsauerwein:
PR https://github.com/c2corg/v6_ui/pull/575 and this PR both add new `Image` view but in different commits. This will be a problem when merging. Could you please turn the two PRs into a single one or base one of the PRs onto the other one?

ginold:
Question: as we can only add images to an existing document (am i right?), there should be no URL `images/add` ?

Or should we be able to create an image that is not associated to anything?

asaunier:
> Or should we be able to create an image that is not associated to anything?

No, new images should always be uploaded from an existing document and be associated to it.

Issue #456, opened 2016-06-25
Loading speed of pages could be increased


asaunier:
The loading speed of the various pages is not excellent.

Things that could be improved:
- [x] Avoid loading all JS lib separately? See https://github.com/c2corg/v6_ui/issues/85 (PR in https://github.com/c2corg/v6_ui/pull/574)
- [x] Add pyramid caching. See https://github.com/c2corg/v6_ui/issues/144
- [ ] Create sprites for icons
- [x] Better caching headers for static files (not sure the JS/CSS files are correctly cached by the browser)  (PR in https://github.com/c2corg/v6_ui/pull/574)
- [x] Make sure some API requests are really needed (eg. #540 - PR in https://github.com/c2corg/v6_ui/pull/579 )

stef74:
Cache browser are configured at 1 hour. Is probably too less.

Exploitez la mise en cache du navigateur pour les ressources suivantes pouvant être mises en cache :
http://c2corgv6.demo-camptocamp.com/node_modules//angular-cookies/angular-cookies.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//angular-gettext/dist/angular-gettext.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//angular-messages/angular-messages.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//angular-moment/angular-moment.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//angular-recaptcha/release/angular-recaptcha.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//angular/angular.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//blueimp-load-image/js/load-image.all.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//bootstrap-slider/dist/bootstrap-slider.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//bootstrap-slider/dist/css/bootstrap-slider.css (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//bootstrap/dist/css/bootstrap.min.css (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//bootstrap/dist/js/bootstrap.min.js (1 hour)
http://c2corgv6.demo-camptocamp.com/node_modules//jquery/dist/jquery.min.js (1 hour)
[...]

PR #552, opened 2016-08-05
Cache pages

tsauerwein:
This PR enables page caching. For example when requesting a document `routes/12345/fr`, the ui server side will check if there is an entry in the cache. If so, a request to the api is made with the previous ETag value (which is stored with the cached page). If the document on the api side has not changed, the cached page is returned. Otherwise the page is rendered, and the cached entry is updated.
An ETag is header is set, which is used by the browser cache (and also Varnish). The ETag header consists of the ETag from the api and the current Git revision (+ a timestamp when running the dev server).

Todo:
* [x] Set up caching/ETag for
  * [x] Detail page
  * [x] Archive page
  * [x] History page
  * [x] Diff page
  * [x] Index page
  * [x] Add/~~edit~~ page
  * [x] Index/auth/account page

Depends on https://github.com/c2corg/v6_api/pull/336

### DIFF
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -1,11 +1,13 @@
 import requests
 from pyramid.config import Configurator
-from pyramid_mako import add_mako_renderer
-from c2corg_ui.lib.cacheversion import version_cache_buster, CACHE_PATH
 from pyramid.httpexceptions import (
     HTTPBadRequest, HTTPInternalServerError)
-from pyramid.view import view_config
 from pyramid.view import notfound_view_config
+from pyramid.view import view_config
+from pyramid_mako import add_mako_renderer
+
+from c2corg_ui.caching import configure_caches
+from c2corg_ui.caching.cacheversion import version_cache_buster, CACHE_PATH
 
 
 def main(global_config, **settings):
@@ -14,6 +16,9 @@ def main(global_config, **settings):
     """
     config = Configurator(settings=settings)
     add_mako_renderer(config, '.html')
+
+    # set up redis cache
+    configure_caches(settings)
 
     # configure connection pool for http requests
     max_connections = int(settings.get('http_request_connection_pool_size'))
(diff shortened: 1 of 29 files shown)


tsauerwein:
This PR would be ready for review.

I didn't set up caching for editing pages (e.g. `/waypoints/edit/733087/fr`). Edit pages for different documents differ in two lines (the lines where the document id and lang are set). So, a separate page for each document and lang would have to be cached. Because there are less edits than views, and to avoid that the cache is filled up too much, I decided no to cache these pages.

c2corg_ui/__init__.py
```diff
@@ -1,11 +1,13 @@
 import requests
 from pyramid.config import Configurator
-from pyramid_mako import add_mako_renderer
-from c2corg_ui.lib.cacheversion import version_cache_buster, CACHE_PATH
 from pyramid.httpexceptions import (
     HTTPBadRequest, HTTPInternalServerError)
-from pyramid.view import view_config
 from pyramid.view import notfound_view_config
+from pyramid.view import view_config
+from pyramid_mako import add_mako_renderer
```
asaunier: Is this still needed?

asaunier:
> I didn't set up caching for editing pages
> Because there are less edits than views, and to avoid that the cache is filled up too much, I decided not to cache these pages.

OK that makes sense. The most important is that detail view pages and very common pages (homepage, advanced search/listing pages if possible) are cached. History and archived versions caching is a plus.

If I understand well, pages are cached a several levels:
* on the API side
* in dogpile for the UI server side
* in Varnish and in the user's browser
Right?

Etags are propagated from the browser to the API to figure out if a new version of the page is available. If yes, the various cache levels are refreshed. Correct?

c2corg_ui/__init__.py
tsauerwein: Yes, the pages are still rendered with Mako.

tsauerwein:
> If I understand well, pages are cached a several levels:

Yes, there will also be a Varnish for the API.

> Etags are propagated from the browser to the API to figure out if a new version of the page is available. If yes, the various cache levels are refreshed. Correct?

Yes. The UI server side also stores the ETag together with the rendered page. For example, if a browser requests a page without providing an ETag, but there is a cache entry for the page, the request to the API will include the ETag from the cache (to check if the entry in the cache is still up-to-date).

tsauerwein:
@asaunier Are you ok with merging this one?

asaunier:
Please do! :+1:

tsauerwein:
Thanks!

Issue #204, opened 2016-02-26
Handle document URLs with no lang param


asaunier:
In v5 documents [1] may contain other document URLs with no lang parameter (only the doc type and the doc id) [2].

Such URLs redirect to the "most appropriate" lang version (depending on the lang prefs and the available lang versions). A similar behaviour is needed in v6.

By the way document URLs should contain slugs (but the current version with no slug should also work). See https://github.com/c2corg/v6_ui/issues/109

[1] for instance http://www.camptocamp.org/summits/37355/fr/mont-blanc
[2] for instance http://www.camptocamp.org/routes/55289

tsauerwein:
From the API this should already be ok: If a language is given as parameter, associated documents include the "best" language.

asaunier:
When such a URL is called, we would need to make a request to the API to get the list of available lang for the given document and make a 301 "Moved permanently" redirection to the URL with the most appropriate lang (+ slug). If there's only one lang available, the choice is easy. If several, we would take the priority listed here https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L2 
Makes sense? Feasible?

tsauerwein:
This works already. For example when you request `/waypoints/38575?l=fr`, associations will include the French translation or the 2nd "best" translation.

    {
      "waypoint_type": "summit",
      "associations": {
        "waypoints": [...],
        "recent_outings": {
          "total": 27,
          "outings": [{
            "locales": [{
              "lang": "fr",                         <-- outing in french
              "version": 1,
              "title": "Aiguilles de Chamonix (et envers des aiguilles)\u00a0: Travers\u00e9e des Aiguilles de Chamonix S-N"
            }],
            "activities": ["mountain_climbing"],
            "version": 1,
            "date_start": "2014-07-16",
            "author": {
              "user_id": 4960,
              "username": "pgatta"
            },
            "document_id": 543208,
            "date_end": "2014-07-16",
            "available_langs": null
          }, {
            "locales": [{
              "lang": "en",                    <-- outing in the 2nd best locale
              "version": 1,
              "title": "Aiguilles de Chamonix: Travers\u00e9e des Aiguilles de Chamonix S-N"
            }],
            "activities": ["mountain_climbing"],
            "version": 1,
            "date_start": "2010-07-26",
            "author": {
              "user_id": 230990,
              "username": "nielsvanveen"
            },
            "document_id": 231002,
            "date_end": "2010-07-26",
            "available_langs": null
[...]

Issue #109, opened 2016-01-19
Document URLs should contain a slug


asaunier:
See #50

PR #558, opened 2016-08-11
Add slug to document URLs

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/109
Fixes https://github.com/c2corg/v6_ui/issues/204

Document URL can be of 3 forms:
* {type}/{id}
* {type}/{id}/{lang}
* {type}/{id}/{lang}/{slug}

The standard one is the third one. When loaded, the 2 others should redirect to it. If the lang is missing, the "best" lang should be found out by the API.

As for now the PR use the standard ``Document.get_document()`` function to get the lang and title of the document (in order to figure the slug out). This is not optimal since this triggers a full document retrieval. We need to create a special service for that in the API. See https://github.com/c2corg/v6_api/issues/364

The serve-side generated slugs uses https://github.com/un33k/python-slugify (there's also https://github.com/dimka665/awesome-slugify both seem quite equivalent).

We also need to slugify the document URLs in the cards (client-side). I have found those 2 angular libs:
* https://github.com/paulsmith/angular-slugify
* https://github.com/lucasconstantino/angular-slug

TODO
* [x] server-side slugs
* [x] client-side slugs
* [x] fix conflicts with the caching system
* [x] update tests

### DIFF
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -3,6 +3,7 @@ goog.provide('app.mapDirective');
 
 goog.require('app');
 goog.require('app.utils');
+goog.require('app.Url');
 goog.require('ngeo.Debounce');
 goog.require('ngeo.Location');
 /** @suppress {extraRequire} */
@@ -63,13 +64,14 @@ app.module.directive('appMap', app.mapDirective);
  *    features to show on the map.
  * @param {ngeo.Location} ngeoLocation ngeo Location service.
  * @param {ngeo.Debounce} ngeoDebounce ngeo Debounce service.
+ * @param {app.Url} appUrl URL service.
  * @constructor
  * @struct
  * @export
  * @ngInject
  */
 app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
-  ngeoDebounce) {
+  ngeoDebounce, appUrl) {
 
   /**
    * @type {number}
@@ -182,6 +184,12 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
    * @private
    */
   this.ignoreExtentChange_ = false;
+
+  /**
+   * @type {app.Url}
+   * @private
+   */
+  this.url_ = appUrl;
 
   /**
    * @type {ol.Map}
@@ -659,13 +667,17 @@ app.MapController.prototype.handleMapFeatureClick_ = function(event) {
     return layer === this.getVectorLayer_();
   }, this);
   if (feature) {
-    var module = /** @type {string} */(feature.get('module'));
-    var id = feature.get('documentId').toString();
-    var lang = /** @type {string} */(feature.get('lang'));
-    var url = app.utils.buildDocumentUrl(module, id, lang);
-    if (url) {
-      window.location.href = url;
+    var module = feature.get('module');
+    var id = feature.get('documentId');
+    var locale = {
+      'lang': feature.get('lang'),
+      'title': feature.get('title')
+    };
+    if (module === 'routes' && feature.get('title_prefix')) {
+      locale['title_prefix'] = feature.get('title_prefix');
     }
+    window.location.href = this.url_.buildDocumentUrl(
+      module, id, /** @type {appx.DocumentLocale} */ (locale));
   }
 };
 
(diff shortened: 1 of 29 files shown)


tsauerwein:
> We also need to slugify the document URLs in the cards (client-side). I have found those 2 angular libs:

What about generating the slugs in the API and including them for each document? This way we are sure the same slugs are used. If we use a different library on the UI server- and client-side, we might get different results.

asaunier:
So the API would return a "slug" attribute in addition to the standard ones?
That sounds interesting.

asaunier:
On the other hand, this slug thing is a pure UI thing. This attribute is probably useless for other consuming apps.

Actually there's probably not a big risk that the JS slugs are different from the python ones and those JS-generated slugs will probably be not available to the search engines, so that's not that important if they are a bit different. The thing will be to avoid that the slug is missing (which will trigger a redirection).

tsauerwein:
> Actually there's probably not a big risk that the JS slugs are different from the python ones and those JS-generated slugs will probably be not available to the search engines

Can you remind me where slugs are created and used on the client-side?

asaunier:
In cards

asaunier:
For the client-side, https://github.com/lucasconstantino/angular-slug seems a good option because it supports non-latin characters as well (yes, in c2corg there are waypoint or route names with russian, greek, chineese, japanese or arabic characters! eg. http://www.camptocamp.org/routes/632364/en/nagisodake--s-route-to-araragi--onsen-place).

This tool is an angular wrapper around the slug.js library which seems to use a unicode library etc.
But that's not very clear what files must be included in our mako template and if there's a minified version.

asaunier:

What changes:
* document URLs now contain a slug based upon the document title (title + title_prefix for routes). If the slug is missing, a 301 redirection is performed towards the URL with the slug.
* slugs are added using a python library when URLs are generated by the UI server-side (SEO links, 301 redirections, archive links).
* for cards (advanced search results and associations) and simple search suggestions, the slug has to be generated client-side. A javascript library (+ an angular wrapper) is used to do so. The results for both pyton and JS tools are consistent for latin-letters titles, a bit less for titles with exotic characters (eg. japanese): those characters are converted by the python tool, ignored by the JS tool (even though the latter seems to support some non-latin alphabet such as russian).
* to ease the use of the slug service, I have created an ``app.Url`` service that is used to generate in JS the document URLs with their slugs.
* tests have been adapted (to support the new URL format and the redirection).

Please note that the 301 redirection now uses a standard document request to the API to get the document title and generate its slug. It would be better if it could request a simple API service that returns the minimal info. See https://github.com/c2corg/v6_api/issues/364

c2corg_ui/static/js/map/map.js
```diff
@@ -659,13 +667,17 @@ app.MapController.prototype.handleMapFeatureClick_ = function(event) {
...
   }, this);
   if (feature) {
-    var module = /** @type {string} */(feature.get('module'));
-    var id = feature.get('documentId').toString();
-    var lang = /** @type {string} */(feature.get('lang'));
-    var url = app.utils.buildDocumentUrl(module, id, lang);
-    if (url) {
-      window.location.href = url;
+    var module = feature.get('module');
+    var id = parseInt(feature.get('documentId'), 10);
```
tsauerwein: Do you need to use `parseInt`? Isn't it already a number?

c2corg_ui/static/js/urlservice.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * @param {function(string):string} slug angular-slug service.
+ * @constructor
+ * @ngInject
+ * @struct
+ */
+app.Url = function(slug) {
+
+  slug.defaults.mode = 'custom';
```
tsauerwein: Is it correct that you are setting the mode to custom, before defining it?

tsauerwein:
That looks good. Do you want to wait for https://github.com/c2corg/v6_api/issues/364 before merging?

asaunier:
@tsauerwein Thanks for your review. I have made the 2 changes you have pointed.

I would prefer we merge the PR without waiting for https://github.com/c2corg/v6_api/issues/364 because it's not a blocking issue (especially with the caching system) and it's better to go forward, no having this PR stalling until we have time to add this new API service. After merging the PR I will create a UI issue to remind us of this needed optimization.

Issue #554, opened 2016-08-05
Use # instead of query params for filters in URL


tsauerwein:
The listing pages use query parameters to encode filters in the URL (e.g. `waypoints?bbox=...&...`). The problem is that the page can not be cached properly, because if the query params are different, the browser cache (and also Varnish) will treat the requests as two different pages (while the rendered page served to the client-side is actually the same no matter what filter is set).

Instead we should encode the filter parameters with `#bbox=...&...`.

asaunier:
> Instead we should encode the filter parameters with #?bbox...&....

Advanced search parameters are added (and read) using ``ngeo.Location`` https://camptocamp.github.io/ngeo/master/apidoc/ngeo.Location.html
So using another syntax depends on its support by ``ngeo.Location``.

PR #559, opened 2016-08-12
Enable ETag header for listing pages

tsauerwein:
Now that https://github.com/c2corg/v6_ui/issues/554 is fixed. This makes it possible that the page can be cached by the browser and Varnish.


asaunier:
Looks good to me.

PR #560, opened 2016-08-12
Add a clear cache script

tsauerwein:
Equivalent to https://github.com/c2corg/v6_api/pull/365


asaunier:
LGTM
Same question about the command names.

PR #562, opened 2016-08-12
Fix minified lib inclusions

asaunier:


tsauerwein:
:+1:

asaunier:
Thanks for the review.

PR #563, opened 2016-08-17
Use public url for images backend

arnaud-morvan:


asaunier:
Looks good to me, thanks.
The travis error is weird. Perhaps it should be relaunched?

PR #565, opened 2016-08-17
Remove debug param stripping in advanced search

asaunier:
Since the advanced search params are passed in the URL using ``#`` instead of ``?``, ``?debug`` in the advanced search URLs is no longer automatically propagated to the API request (by the way would it really be a problem?).

As a result this PR removes the ``debug`` removal.


tsauerwein:
:+1: 

The API would ignore it anyway.

Issue #168, opened 2016-02-16
Preview action


asaunier:
A "preview" action before actually saving an editing form is required (as in v5). The "save" button should be clickable only when the preview has been displayed (to avoid that people create many versions because they have not checked their changes).

asaunier:
The preview being needed when editing a document, we need to use it as a UI AJAX service that would:
* get the attributes from the client UI (without saving them yet)
* generate some HTML using Mako (same renderer and syntax parsers than for the standard detail view)

The generated HTML would then be displayed in the editing flow before allowing to cklick the save button.

asaunier:
Having a preview of the edited document before it is actually saved (at the end of the editing form) is needed so that contributors check their changes (in order to avoid subsequent changes fixing for example formatting errors). In v5 an AJX request is sent to the server that returns some HTML built using  some templates.

Doing so in v6 would be relevant.

The current approach implemented by @ginold is nice. The main downside is that it is not possible to check the final rendering of text attributes with bbcode/markdown formatting.

I was considering we could create a UI webservice that would use the existing ``view.html`` templates, feeding them with the data sent from the editing form instead of the usual API request. The generated HTML/JS code would then be integrated inside the editing form screen.

Several issues though:
* we cannot use the view templates out of the box since some part are not relevant (everything inherited from ``base.html`` + most of the tools like the edit/translate/history buttons etc.). Not sure we should show the associated documents (might be too much stuff in the preview?) even though it would be consistent since the editing form allows to change the associations...
* would the layout behave as expected since the page width will not be the same than in the standard view page => responsiveness issues? Or maybe the layout will adapt out of the box?
* wouldn't this be a bit confusing for contributors to have a preview too similar (!?) to the end view page? I mean that people might think the document is saved whereas it is not?
* would we have conflict between the JS code in the editing form and the preview part?

Should we then:
* display the preview as a separated page when pushing the "preview" button at the end of the editing form? Some ``waypoints/preview`` that would receive the data from the documentediting directive. It would be possible to go back to the editing page, especially if data are changed in the localstorage as in https://github.com/c2corg/v6_ui/issues/250
* stay in the editing page but make the preview webservice use a simplified Mako template with a simplified layout based upon the Mako helpers already used by the view templates?

asaunier:
@tsauerwein Is there a way to define a UI web service / view that would not be cached at all? For instance ``POST waypoints/preview``, that would return some HTML containing the data passed as POST.
[...]

PR #569, opened 2016-08-19
Edited document preview

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/168

TODO:
* [x] basic proto preview UI service
* [x] post document data to the UI service
* [x] build UI service response based on mako templates and the posted data
* [x] integrate the UI service response in the last editing step (client side) + styling
* adapt the editing steps and buttons so that
 * [x] the "preview" step is displayed when clicking the "preview" button
 * [ ] the "save" button is not shown unless the "preview" button has been clicked. It should only be shown at the final step
 * [ ] what happens when browsing to the last (preview) step using the arrows? should this trigger a new preview?
 * [ ] show map?
 * [ ] show associated docs?


c2corg_ui/templates/outing/edit.html
```diff
@@ -563,14 +567,14 @@ <h5 translate ng-show="outing.associations.users.length == 0">You can look up fo
 
       </div>
 
+      ## FIXME: really useful?
       <div class="step step-4"></div>
```
asaunier: @ginold Is this blank step4 really needed?

asaunier:

It is a little simplified version of document editing previewing in order to have at least a base tool for release alpha3. Simplified because:
* associations and map are not shown in the preview (but do we really need to since those parts are already quite explicit?)
* the preview step is not mandatory before saving the document (at first I thought we could show the Save button only after the preview button has been clicked)

What it does:
* add a preview button to the list of buttons on the right hand of the editing interface
* when the preview button is pressed, a wide modal is shown, showing the preview of the edited document
* the preview is generated using a UI webservice that receives the data from the editing form and returns some HTML code based on a preview Mako template. This template uses the same helper functions than the standard view template.

sans titre

Improvements and fixes that could be done (maybe not in this PR):
* [ ] factorise a bit more the view/preview mako templates? (but since helper functions are used, the common code is not that big)
* [ ] add a loading indicator in the modal box until the preview is actually shown?
* [ ] show map and associations?
* [ ] not sure angular code is correctly executed within the modal preview (for instance it seems that the outing dates are not correctly shown with angular moment and that the interface strings translations does not work)
* [ ] some tests?

c2corg_ui/templates/image/preview.html
```diff
@@ -0,0 +1,70 @@
+<%namespace file="helpers/detailed_images_attributes.html"
```
ginold: Can't we use the exact file view.html instead of having preview.html and view.html ? It would be simpler to maintain, we could manage some special cases with `% if preview ....` as it is done for `version` for example.

c2corg_ui/templates/outing/edit.html
ginold: well, if it is empty then hell no! just delete it and rename `step-5` to `step-4`

c2corg_ui/templates/image/preview.html
asaunier: That's not that simple. In preview, you have many parts from view that you don't want to have in preview. For instance:
* the ``base.html`` template inheriting https://github.com/c2corg/v6_ui/blob/preview/c2corg_ui/templates/waypoint/view.html#L7 (not sure you can have a ``if`` condition to include it.
* ditto for the various ``block``

And having many ``if preview`` will make the template even more complicated and hard to read.

Please note that the ``preview.html`` templates are not that big thanks to the use of the various helper functions that you have created :)
https://github.com/c2corg/v6_ui/blob/preview/c2corg_ui/templates/waypoint/preview.html
Perhaps we could improve a bit by using an helper for the description/summary block as well if possible https://github.com/c2corg/v6_ui/blob/preview/c2corg_ui/templates/waypoint/preview.html#L47-L65

c2corg_ui/templates/outing/edit.html
asaunier: Do I have to update the maxSteps here https://github.com/c2corg/v6_ui/blob/ba802ec7f9112b241489d6d2837e8720fc6a7c11/c2corg_ui/templates/outing/edit.html#L36 ?

asaunier:
I have just pushed a commit that removes the useless step in outing editing

I won't do any change related to the improvements and fixes mentioned at the end of
https://github.com/c2corg/v6_ui/pull/569#issuecomment-245708851
They'll be done a further version of the preview (separated PR).

Can someone review this PR please? It's quite urgent.
Thanks :)

ginold:
> Do I have to update the maxSteps

yes, there is one step less because there will be no "review" anymore. Also delete the "review" tab from the menus on every editing page
https://github.com/c2corg/v6_ui/blob/ba802ec7f9112b241489d6d2837e8720fc6a7c11/c2corg_ui/templates/outing/edit.html#L43

tsauerwein:
That looks quite good! Feel free to merge if you want to get it in for the release.

tsauerwein:
> not sure angular code is correctly executed within the modal preview (for instance it seems that the outing dates are not correctly shown with angular moment and that the interface strings translations does not work)

If this is still a problem, we could consider showing the preview in an i-frame to properly separate the pages.

asaunier:
> yes, there is one step less because there will be no "review" anymore. Also delete the "review" tab from the menus on every editing page

I have left the "review" tab (renamed to "summary") because it still contains the quality and "summary of the changes" fields. At least in this version of the preview.

So do I really have changes to do left? The outing ``maxSteps = 4`` seems consistent with the one of the routes: https://github.com/c2corg/v6_ui/blob/preview/c2corg_ui/templates/route/edit.html#L35-L42


> That looks quite good! Feel free to merge if you want to get it in for the release.

OK thanks. I will merge when this step things is clarified.

> If this is still a problem, we could consider showing the preview in an i-frame to properly separate the pages.

That's an option. I'll see that in next version :)

ginold:
> it still contains the quality and "summary of the changes" fields

i think it should be moved to the preview modal in an upcoming PR.

>  So do I really have changes to do left? 
If you want to leave both fields at the "preview" step for now, then no. Good to go :)

asaunier:
OK merging.

Thanks @ginold and @tsauerwein for reviewing!

PR #572, opened 2016-08-19
Redirections

### LINKED ISSUE #190: Manage documents marked as merged
Example in v5: http://www.camptocamp.org/summits/300435/fr/dolomites-de-brenta
In v6, as for now one get an error 500: http://c2corgv6-demo.gis.internal:6555/waypoints/300435/fr => the error seems the same than for https://github.com/c2corg/v6_ui/issues/137

See also https://github.com/c2corg/v6_api/issues/73

### LINKED ISSUE #568: Optimize slugification and use lighter API service
As for now, the slugification service uses a standard "get document" API request that returns the full attributes of the document whereas only the title is needed.
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L423

When https://github.com/c2corg/v6_api/issues/364 is addressed, the UI side should use this new light service instead.

asaunier:
This PR:
* uses the new API info service to get the slug of partial document URLs (without lang or slug) => fixes #568
* handles merged/redirecting documents => fixes #190

### DIFF
--- a/c2corg_ui/views/document.py
+++ b/c2corg_ui/views/document.py
@@ -20,7 +20,7 @@ from c2corg_common.attributes import default_langs, langs_priority
 
 from pyramid.httpexceptions import (
     HTTPBadRequest, HTTPNotFound, HTTPInternalServerError,
-    HTTPMovedPermanently)
+    HTTPMovedPermanently, HTTPFound)
 
 from c2corg_ui.views import etag_cache, get_response, get_or_create_page
 
@@ -195,9 +195,11 @@ class Document(object):
 
         # Manage merged documents (redirecting to another document)
         if 'redirects_to' in document:
-            if lang is None or lang not in document['available_langs']:
+            is_lang_set = lang is not None and \
+                          lang in document['available_langs']
+            if not is_lang_set:
                 lang = self._get_best_lang(document['available_langs'])
-            self._redirect(document['redirects_to'], lang)
+            self._redirect(document['redirects_to'], lang, None, is_lang_set)
 
         # When requesting a lang that does not exist yet, the API sends
         # back an empty list as 'locales'
@@ -422,21 +424,27 @@ class Document(object):
     def _get_response(self, page_html):
         return get_response(self.request, page_html)
 
-    def _redirect(self, id, lang, slug=None):
-        location = ''
+    def _redirect(self, id, lang, slug=None, is_lang_set=False):
         if slug is None:
             location = self.request.route_url(
                 self._API_ROUTE + '_view_id_lang', id=id, lang=lang)
         else:
             location = self.request.route_url(
                 self._API_ROUTE + '_view', id=id, lang=lang, slug=slug)
-        raise HTTPMovedPermanently(location=location)
+        if is_lang_set:
+            raise HTTPMovedPermanently(location=location)
+        else:
+            # The original URL had no lang param, which means it had to be
+            # figured out according to the user's interface and available
+            # langs => the redirection cannot be permanent since it may differ
+            # from one user to another.
+            raise HTTPFound(location=location)
 
     def _redirect_to_full_url(self):
         id = self._validate_int('id')
-        lang = self._validate_lang() \
-            if 'lang' in self.request.matchdict else \
-            self._get_interface_lang()
+        is_lang_set = 'lang' in self.request.matchdict
+        lang = self._validate_lang() if is_lang_set \
+            else self._get_interface_lang()
 
         url = '%s/%d/%s/info' % (self._API_ROUTE, id, lang)
         if log.isEnabledFor(logging.DEBUG):
@@ -454,14 +462,15 @@ class Document(object):
         if 'redirects_to' in data:
             if lang not in data['available_langs']:
                 lang = self._get_best_lang(data['available_langs'])
-            self._redirect(data['redirects_to'], lang)
+            self._redirect(data['redirects_to'], lang, None, is_lang_set)
         else:
             locale = data['locales'][0]
             title = ''
             if self._API_ROUTE == 'routes' and locale['title_prefix']:
                 title += locale['title_prefix'] + ' '
             title += locale['title']
-            self._redirect(data['document_id'], locale['lang'], slugify(title))
+            self._redirect(data['document_id'], locale['lang'],
+                           slugify(title), is_lang_set)
 
     def _get_interface_lang(self):
         return self.request.cookies.get('interface_lang', self._DEFAULT_LANG)
--- a/c2corg_ui/views/document.py
+++ b/c2corg_ui/views/document.py
@@ -39,6 +39,8 @@ class Document(object):
         'limit': 30
     }
 
+    _DEFAULT_LANG = 'fr'
+
     def __init__(self, request):
         self.request = request
         self.settings = request.registry.settings
@@ -244,7 +246,7 @@ class Document(object):
 
     def _get_documents(self):
         params = []
-        lang = self.request.cookies.get('interface_lang', 'fr')
+        lang = self.request.cookies.get('interface_lang', self._DEFAULT_LANG)
         params.append(('pl', lang))
         # Inject default list filters params:
         filters = dict(self._DEFAULT_FILTERS, **{k: v for k, v in params})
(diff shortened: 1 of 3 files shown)


asaunier:
Anyone available for reviewing this PR? :)

tsauerwein:
I will take a look.

asaunier:
Thanks! Feel free to merge the PR if it looks good to you.

c2corg_ui/views/document.py
```diff
@@ -432,12 +441,17 @@ def _redirect_to_full_url(self):
...
-        if self._API_ROUTE == 'routes' and locale['title_prefix']:
-            title += locale['title_prefix'] + ' '
-        title += locale['title']
-        location = self.request.route_url(
-            self._API_ROUTE + '_view', id=data['document_id'],
-            lang=locale['lang'], slug=slugify(title))
-        raise HTTPMovedPermanently(location=location)
+        if 'redirects_to' in data:
+            if lang not in data['available_langs']:
+                lang = self._get_preferred_lang()
```
tsauerwein: Instead of using `prefered_lang`, you would have to select the *best* lang in `data['available_langs']`. Otherwise you might do a redirect to a locale that does not exist (and at this point you already know that it does not exist).

c2corg_ui/views/document.py
```diff
@@ -192,6 +192,12 @@ def _get_document(self, id, lang=None, old_api_cache_key=None):
         if not_modified:
             return not_modified, api_cache_key, None
 
+        # Manage merged documents (redirecting to another document)
+        if 'redirects_to' in document:
+            if lang is None or lang not in document['available_langs']:
+                lang = self._get_preferred_lang()
```
tsauerwein: Same comment as above.

c2corg_ui/views/document.py
```diff
@@ -413,20 +421,43 @@ def _get_api_cache_key_from_etag(self, etag):
...
 
+    def _redirect(self, id, lang, slug=None):
+        location = ''
+        if slug is None:
+            location = self.request.route_url(
+                self._API_ROUTE + '_view_id_lang', id=id, lang=lang)
+        else:
+            location = self.request.route_url(
+                self._API_ROUTE + '_view', id=id, lang=lang, slug=slug)
+        raise HTTPMovedPermanently(location=location)
```
tsauerwein: It's better if we use `HTTPFound` (302) instead of `HTTPMovedPermanently` (301). The browser is very eager in caching 301 redirects, it will *never* do a request again and directly use the redirected URL. I will also change it in the API.

c2corg_ui/views/document.py
asaunier: What is "the best lang in data['available_langs']"?
On the UI side we don't know what lang to choose among ``available_langs`` if ``available_langs`` does not contain the requested ``lang``. In that case, I suggest to fall back on the user's perferred lang and let the API deal with it (return the "best lang" if the preferred lang is not available).
Doesn't it make sense?

c2corg_ui/views/document.py
asaunier: > It's better if we use HTTPFound (302) instead of HTTPMovedPermanently (301).

The goal of using 301 is to make it clear to search engines that the redirected URL should be used instead of the original URL ("permanent redirection"). Not sure 302 would be understood the same way by search engines.

https://moz.com/learn/seo/http-status-codes says that
> Use 301 redirects rather than 302 redirects when redirecting URLs on a site to ensure that link juice (ranking power) is passed between the redirecting web pages.

and https://moz.com/learn/seo/redirection
> 301, "Moved Permanently"—recommended for SEO

tsauerwein: The main problem with using `HTTPMovedPermanently (301)` is that the redirect is actually not permanent. Two examples:

When requesting `/waypoints/1234` (without lang), the redirect depends on the current interface language. Let's say the interface language is `en`, so it will redirect to `/waypoints/1234/en/...`. If the interface language is now changed to `fr`, the browser will directly redirect to `/waypoints/1234/en/...`.

Second problem: If `redirects_to` is set for a document, a permanent redirect to the linked document will be made. But `redirects_to` can be changed! So even if you change `redirects_to`, browsers will still redirect to the old linked document, because they think the redirect is permanent (while it is only temporary!).

c2corg_ui/views/document.py
tsauerwein: If you call `/waypoints/1234/fr` and there is no locale for `fr` for waypoint `1234`, you will get a `404`.

> What is "the best lang in data['available_langs']"?

https://github.com/c2corg/v6_api/blob/b32316f7ee1a9db28107e20c9678c87a7c235468/c2corg_api/views/__init__.py#L168-L175

c2corg_ui/views/document.py
asaunier: > When requesting /waypoints/1234 (without lang), the redirect depends on the current interface language.

OK, makes sense. So we need a 302 when the lang is missing and 301 if it is provided :)

> Second problem: If redirects_to is set for a document, a permanent redirect to the linked document will be made. But redirects_to can be changed!

Why would it change? This case has never happened in v5 since 2007.

My concern is that we need to make sure that redirecting documents are not referenced by search engines and that the target document is always taken into accound instead. It seems we need the 301 for this, doesn't it?

c2corg_ui/views/document.py
asaunier: OK, that makes sense to do as you suggest.

c2corg_ui/views/document.py
tsauerwein: > Why would it change? This case has never happened in v5 since 2007.

How did you check this? Our model does not prevent that `redirects_to` can not be changed anymore if it has been set once.

> My concern is that we need to make sure that redirecting documents are not referenced by search engines and that the target document is always taken into accound instead. It seems we need the 301 for this, doesn't it?

The redirected page will also be indexed with 302. This graphic explains it quite well: https://seohackercdn-seohacker.netdna-ssl.com/wp-content/uploads/2011/10/301-and-302-Redirect-small2.png

asaunier: > How did you check this? Our model does not prevent that redirects_to can not be changed anymore if it has been set once.

Because merging documents is a moderator-only feature, undoing a merge requires an hand-change in the DB in v5 and I'm pretty sure such a operation has never been requested (there are very few people that have ssh access to the server).

Of course documents might be merged by error and rollback would then be necessary. But this will happen very rarely (this is not a public feature). And the page ranking issue is much more important.

> The redirected page will also be indexed with 302. This graphic explains it quite well: https://seohackercdn-seohacker.netdna-ssl.com/wp-content/uploads/2011/10/301-and-302-Redirect-small2.png

Well, the bottom of the graphic explains that 
capture d ecran 2016-08-31 a 17 18 14

In order to actually transfer the indexing of the old page to the new one, a 301 is recommended.

c2corg_ui/views/document.py
asaunier: When the API returns ``redirects_to`` it also provides the available langs. So it's easy to figure the best lang out as in your link.

But what should we do when the ``lang`` param is not provided (URL such as ``waypoints/123456``)? We don't know what langs are available for this document. Is there a way to ask the API for the available langs?

tsauerwein: > But what should we do when the lang param is not provided (URL such as waypoints/123456)? We don't know what langs are available for this document. Is there a way to ask the API for the available langs?

Isn't this PR all about handling this? You are doing a request to the "document info" service with the interface language. If a locale in that language does not exist, you get the *best* locale.

c2corg_ui/views/document.py
tsauerwein: > In order to actually transfer the indexing of the old page to the new one, a 301 is recommended.

If it really is a permanent redirect, yes. I am worried that accidentally a redirect is set to a wrong document. And then you have a hard undoing the (not so) permanent redirect.

asaunier:
I have updated the PR so that:
* missing or unavailable langs are better managed
* a 302 (temporary) redirection is made for URL not providing the lang.

When redirecting from a URL providing both the document id and lang, I have preferred to stick to a 301 (permanent) redirection because:
* it's critical that the documents have a good page ranking in google and other search engines. Which means that only one document is indexed for a given WP/route/...
* I agree that the 301 makes it more difficult to fix when 2 documents have been merged by error but I don't think it has ever happened with the v5 (this is a moderator only feature). Same problem actually than if a moderator had dropped the document by mistake.
For me the first point is more important than the second one.

asaunier:
By the way, merging documents is not only about flagging one as redirecting to the other, it should also transfer all the associations of the redirecting doc to the target one (perhaps it's not done yet, as in v5). Which means that merging documents is (almost) irreversible anyway.

tsauerwein:
> By the way, merging documents is not only about flagging one as redirecting to the other, it should also transfer all the associations of the redirecting doc to the target one (perhaps it's not done yet, as in v5).

The API endpoint to mark a document as *redirected* is not implemented yet. This is part of the moderator tools.

> Which means that merging documents is (almost?) irreversible anyway.

We should make sure that we communicate this clearly in the UI (something like "Are you really sure...").


It would be cool to have some tests for the logic that you added in this PR. But feel free to merge.

asaunier:
> The API endpoint to mark a document as redirected is not implemented yet. This is part of the moderator tools.

OK, I have added an API issue to make sure we keep track of the things to do:
https://github.com/c2corg/v6_api/issues/386

> We should make sure that we communicate this clearly in the UI (something like "Are you really sure...").

Sure!

> It would be cool to have some tests for the logic that you added in this PR.

There was already some testing about the slug redirection. I have added some more with 89f89e9

tsauerwein:
:+1:

asaunier:
Thanks @tsauerwein for the review and tips!

Issue #570, opened 2016-08-19
Problem with document detail view cache?


asaunier:
Now and then I have observed 500 errors when loading a document detail view page, with the following stack

```
ValueError: Expecting value: line 1 column 1 (char 0)

Traceback (most recent call last)
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid_debugtoolbar/panels/performance.py", line 57, in resource_timer_handler
￼￼result = handler(request)
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py", line 62, in excview_tween
￼￼reraise(*attrs['exc_info'])
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/compat.py", line 148, in reraise
￼￼raise value
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py", line 22, in excview_tween
￼￼response = handler(request)
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/router.py", line 158, in handle_request
￼￼view_name
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/view.py", line 547, in _call_view
￼￼response = view_callable(context, request)
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 413, in viewresult_to_response
￼￼result = view(context, request)
File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 123, in _class_requestonly_view
￼￼response = getattr(inst, attr)()
File "/home/alex/Desktop/v6_ui/c2corg_ui/views/waypoint.py", line 59, in detail
￼￼return self._get_or_create_detail(id, lang, render_page)
File "/home/alex/Desktop/v6_ui/c2corg_ui/views/document.py", line 83, in _get_or_create_detail
￼￼self._get_cache_key)
File "/home/alex/Desktop/v6_ui/c2corg_ui/views/document.py", line 119, in _get_or_create
￼￼not_modified, api_cache_key, loaded_data = load_data(old_api_cache_key)
File "/home/alex/Desktop/v6_ui/c2corg_ui/views/document.py", line 78, in load_data
[...]

PR #573, opened 2016-08-19
Increase cache lock timeout

tsauerwein:
Same as https://github.com/c2corg/v6_api/pull/374
Closes https://github.com/c2corg/v6_ui/issues/570


asaunier:
👍

PR #574, opened 2016-08-19
Improve handling of assets

tsauerwein:
- JavaScript libraries are concatenated into a single file
- CSS files of dependencies are concatenated into a single file
- The expire time for all assets is set to 1 year. When a new version is deployed, the cache version in the URL changes, and all assets have to be requested again.

Closes https://github.com/c2corg/v6_ui/issues/85

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "angular-gettext": "2.3.4",
     "angular-gettext-tools": "2.2.3",
     "angular-messages": "1.5.8",
+    "slug": "0.9.1",
     "angular-slug": "git://github.com/lucasconstantino/angular-slug#v0.0.4",
     "eslint": "2.13.1",
     "eslint-config-openlayers": "5.0.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "angular-gettext": "2.3.4",
     "angular-gettext-tools": "2.2.3",
     "angular-messages": "1.5.8",
+    "slug": "0.9.1",
```
asaunier: It's in the dependencies of ``angular-slug`` https://github.com/lucasconstantino/angular-slug/blob/master/package.json#L15
Shall we also add it to our own dependencies?

asaunier:
👍

package.json
tsauerwein: Depending on the npm version `slug` might be installed in `node_modules/slug` or in `node_modules/angular-slug/node_modules/slug`. When using an explicit dependency, it will always be in `node_modules/slug`.

tsauerwein:
Thanks for the review

Issue #551, opened 2016-08-03
Avoid that page elements "jump" when loading


tsauerwein:
On detail pages (e.g. http://c2corgv6.demo-camptocamp.com/routes/737035/fr), when loading/reloading the page, the elements below the map jump down when the map is shown.

The problem is that the height of the map is set on the `viewport` element (in [viewdetails.less](https://github.com/c2corg/v6_ui/blob/e1724c8c11fa5483bd78954a9c042f54a1132bc1/less/viewdetails.less#L497-L517)). When the page is loaded, the viewport element does not exist yet, so that the browser does not know how much space the map element will take. Only when the map is initialized, the height defined in the CSS is applied.

It would be better to set the height on the map element itself.

fjacon:
@ginold : it seems fixed ? can you confirmed ?

asaunier:
Not fixed at all, at least in the demo.

PR #575, opened 2016-08-26
Images detail view + list page

ginold:
screenshot from 2016-08-30 12 22 34
screenshot from 2016-08-30 12 38 29
- if auhor is not set, how to get the user that posted the image?
- i'm not sure of the available query options for images (filters), so the query may vary
- in the view, the map is super stretched if moved anywhere below oO.

:P


c2corg_ui/static/js/auth/authservice.js
```diff
@@ -75,32 +75,42 @@ app.Authentication.prototype.isAuthenticated = function() {
...
     for (var i = 0; i < users.length; i++) {
       if (userid === users[i].id) {
         return true;
       }
     }
     return false;
+  } else if (imageType) {
+    if (imageType === 'collaborative') {
+      return true;
+    // TODO: check if user == author
```
arnaud-morvan: Remaining TODO here

tsauerwein:
> in the view, the map is super stretched if moved anywhere below oO.

You shouldn't set the height directly on `ol-viewport`. See also https://github.com/c2corg/v6_ui/issues/551

c2corg_ui/views/image.py
```diff
@@ -0,0 +1,99 @@
+from pyramid.renderers import render
+from pyramid.view import view_config
+
+from c2corg_ui.views.document import Document
+
+
+class Outing(Document):
```
arnaud-morvan: should be named ``Image``

c2corg_ui/static/js/auth/authservice.js
tsauerwein: Currently the API does not include the author for images. When would you need it? When requesting a single image or also when requesting a list of images?

c2corg_ui/static/js/card.js
```diff
@@ -119,6 +125,15 @@ app.CardController.prototype.createURL = function() {
 
 
 /**
+ * Don't create any url on edit and add pages.
+ * @export
+ * @return {string | undefined}
```
tsauerwein: Why can this be `undefined`?

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -75,32 +75,42 @@ app.Authentication.prototype.isAuthenticated = function() {
 /**
  * Checks if the current user has rights to access/edit the document.
  * TODO: rights to personal documents only for current user.
+ * @param {Array<string>} users
+ * @param {?string} imageType
  * @return {boolean}
  * @export
  */
-app.Authentication.prototype.hasEditRights = function(users) {
+app.Authentication.prototype.hasEditRights = function(users, imageType) {
```
tsauerwein: You are using this method to check the rights for images and outings. Would it simplify the logic to have two separate methods?

c2corg_ui/static/js/card.js
ginold: Probably copy-paste from the function just below, gone wrong :P thanks for spotting it

c2corg_ui/static/partials/cards/images.html
```diff
@@ -0,0 +1,26 @@
+<a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <img ng-src="{{cardCtrl.createImg()}}" alt="{{locale['title']}}">
+  <div class="list-item-info">
+    <div class="outing-date">{{ doc['date_time'] | date: 'dd/MM/yyyy'}}</div>
```
tsauerwein: Class `outing-date`?

c2corg_ui/static/js/auth/authservice.js
ginold: I think it's necessary only for view pages - we would like to show who is the author of the image (if `author` is set) or to write "image posted by: xxxx". Also, if the `image_type` is `personal`, you could only edit the image if it is yours.

tsauerwein: Ok, I will add it: https://github.com/c2corg/v6_api/issues/376

c2corg_ui/templates/image/index.html
```diff
@@ -0,0 +1,40 @@
...
+
+<%block name="pagetitle"><title ng-bind="mainCtrl.page_title('Images')">${show_title('Images')}</title></%block>
+
+<%block name="moduleConstantsValues">
+  module.value('mapFeatureCollection', null);
+  module.value('advancedSearchFilters', {
+    'date': {'type': 'date'},
+    'act': {'type': 'list'},
+    'oalt': {'type': 'range'}
+  });
```
tsauerwein: I will have to index these fields in ElasticSearch. Which fields to you need? `activities, categories, elevation, date_time`? Maybe also `image_type`?

c2corg_ui/static/js/auth/authservice.js
tsauerwein: Actually it is no the `author`, it is the user who uploaded the image.

c2corg_ui/templates/image/index.html
ginold: Yes, I think those are enough.

c2corg_ui/static/js/auth/authservice.js
ginold: Probably not a bad idea. Would it be ok to write something like :

```
hasEditRIght(users, imageType) {
    isAuthenticated...
    ....
    isModerator...
   ....
    if(!imageType && users) {  checkRightsOuting() }
    else if (imageType) {  checkRightsImage() } 
}

```
because the first part of the function is always common.

c2corg_ui/static/js/auth/authservice.js
ginold: Yes, the `author` isn't unbedingt the one who uploaded, but if it is not set it should show "posted by: xxx". Can we get this info from somewhere (who posted the image)?

tsauerwein: I will have to add it in the API, see https://github.com/c2corg/v6_api/issues/376

c2corg_ui/static/js/auth/authservice.js
tsauerwein: Yes, that would be good.

ginold:
exposure_time question: normally, the extracted value from exif should look something like `0.002` seconds. Do we leave it as is or should we convert it to something more readable like 1/250? 
If we leave it, it means that this value should be translated on the client to `1/250s`. 

The same problem is in the editing where user can enter the `exposure_time` value : how do we determine what the user can type? it works easily when the input is like <input>1/ {value} s</input> but what if the user wants to enter "2s" ?

tsauerwein:
> Do we leave it as is or should we convert it to something more readable like 1/250?
> If we leave it, it means that this value should be translated on the client to 1/250s.

I guess it would be nice to format it like this. The value is stored as float, so it should be done when showing the value.

> The same problem is in the editing where user can enter the exposure_time value : how do we determine what the user can type? it works easily when the input is like <input>1/ {value} s</input> but what if the user wants to enter "2s" ?

I think changing the exposure time is not a very common operation. So in my opinion it's ok to show `0.002` for the editing.

ginold:
> So in my opinion it's ok to show 0.002 for the editing.
It's a shame that the user will have to convert 1/320 to 0.003125 by himself if he wants to enter a new value, I guess :P

ginold:
i think i will put a converter that shows the actual value - user type '0.000325' and it transaltes to "1/320"

PR #576, opened 2016-08-26
disable deleting the last associated wp

ginold:
related to #511
 
when creating a route with associated waypoints, one of them NEEDS to be the main waypoint, therefore you will not be able to save a route with associated waypoint where none of them is the main one. 

- when associating the 1st wp to route, it will appear automatically as the main wp
- when deleting the main wp, while you still have some other wps, the first one will become the main wp

Like this, when you are on a view page, you will always find some associated waypoints where least one is the main one -> this one cannot be deleted and thus you will not be able to delete all the associated waypoints.

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -180,7 +180,11 @@
            <app-card app-card-doc="doc"></app-card>
         % if showDelete:
         <app-delete-association child-doctype="${type}" parent-id="${document['document_id']}" child-id="doc.document_id"
-                                ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+                                ng-if="userCtrl.auth.isAuthenticated()
+                                % if document['doctype'] == 'routes':
+                                    && ${associations}.length > 1
+                                % endif
+                                "></app-delete-association>
         % endif
       </div>
     </div>
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -304,7 +304,7 @@ app.DocumentEditingController.prototype.submitForm = function(isValid) {
     };
     this.api_.updateDocument(this.module_, this.id_, data).then(function() {
       window.location.href = this.url_.buildDocumentUrl(
-        this.module_, this.id_, data['locales']['0']);
+        this.module_, this.id_, this.documentService.document['locales'][0]);
     }.bind(this)
     );
   } else {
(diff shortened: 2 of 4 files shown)


asaunier:
> when creating a route with associated waypoints, one of them NEEDS to be the main waypoint, 

Are you sure of this?
I know that a route must at least have one associated WP (as described in issue #511) but I am not sure a main waypoint is really mandatory. For instance for raids (several-day-long trips), a main waypoint is not relevant I think. I don't remember if it is described in an issue or somewhere else. @stef74 @loicperrin ?

ginold:
Ok, I haven't thought of it that way. 
If stef74 and ioicperrin confirm your opinion, I will have to find another way :P ...

loicperrin:
I confirm that a route needs at least one associated WP, but not necessary a main WP.
See http://www.camptocamp.org/forums/viewtopic.php?id=290601

ginold:
Ok, so i suggest not letting to delete the last associated wp from route - check when there is only 1 left.

c2corg_ui/templates/helpers/view.html
```diff
@@ -180,7 +180,11 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
            <app-card app-card-doc="doc"></app-card>
         % if showDelete:
         <app-delete-association child-doctype="${type}" parent-id="${document['document_id']}" child-id="doc.document_id"
-                                ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+                                ng-if="userCtrl.auth.isAuthenticated()
+                                % if document['doctype'] == 'routes':
+                                    && ${associations}.length > 1
+                                % endif
```
asaunier: ``${associations}.length > 1`` doesn't look like a nice way to test how many documents are associated because ``${associations}`` will print the whole associations list in the HTML. Which could be huge.
Rather user a full python code: ``len(associations)``.

A more concise syntax would be
```
ng-if="userCtrl.auth.isAuthenticated() ${'&& ' + len(associations) + ' > 1' if document['doctype'] == 'routes' else ''}"
```

Mmmm, but that test will not be updated when the associations list is updated. But neither would the ``${associations}.length`` way I think!?

ginold: > ${associations} will print the whole associations list in the HTML. Which could be huge.

`${associations}` is a string, in this case `detailsCtrl.documentService.document.associations.`

asaunier: Then OK for me :)
Feel free to merge and mark the related issue as "ready to be tested".

Issue #540, opened 2016-07-27
Prevent an "updatePreferredLanguage" request on every single page loading


asaunier:
At page load, the interface language is detected and saved at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/langservice.js#L97

This function also sends a request to the API in order to update the user's lang setting (used for instance to translate the emails sent by the API): https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/langservice.js#L110-L112

Problem: the API request is done whenever a page is loaded, even though the user has not changed his/her interface language and this most of the time uselessly loads the API.

A possibility would be to trigger this API request only when the user actually choose another language in the interface language dropdown and when the user logs in.

PR #579, opened 2016-08-30
Handle interface language

tsauerwein:
* Set the interface language on login
* Only send the interface language to the API if a new one is selected

Closes https://github.com/c2corg/v6_ui/issues/540
Requires https://github.com/c2corg/v6_api/pull/375


asaunier:
👍

tsauerwein:
Thanks for the review!

Issue #296, opened 2016-04-13
Redirect after login might fail


tsauerwein:
1. Go to http://c2corgv6-demo.gis.internal/
2. Click on "Sign in".
3. Click on "No account yet?"
4. Remember that you actually have an account and click again on "Sign in" in the top menu.
5. The URL now looks like this: http://c2corgv6-demo.gis.internal/auth?from=http%3A%2F%2Fc2corgv6-demo.gis.internal%2Fauth%3Ffrom%3Dhttp%253A%252F%252Fc2corgv6-demo.gis.internal%252F
6. If you now successfully login, you will be redirected to the login page. Which is confusing because you are not sure if you are logged-in or not.

asaunier:
Indeed :)

To combine with https://github.com/c2corg/v6_ui/issues/270 maybe?

Issue #270, opened 2016-04-06
Security issue with auth URLs?


asaunier:
As for now we use a GET parameter to pass the URL the user should be redirected to after authenticating. For instance ``http://.../auth?from=http://.../waypoints/add``

The client suggests that it might be a security breach, for instance if someone forge a URL with a malicious "from" parameter that would redirect to some other site and that we should use POST instead of GET.

Perhaps also a problem with relative URL:

> Je me rends compte que de ne pas interdire les url relatives est une faille de sécutité. Qqun peut mettre des liens de login dans la nature (avec un truc bien foutu qui incite à se connecter, du style "urtiisez votre compte c2c pour vous connecter sur mon site"), avec redirection vers une page qui reseemble à la home c2c, puis au bout de 10s un popupp s'ouvre demandant de s'identifier à nouveau : le hacker récupère les mots de passe.

(perhaps it was "ne pas interdire les url absolues"?)

fbunoz:
No : the absolute url must be forbidden, to be sure to be redirected to c2c.
In the login url, we show a relative url as parameter (it is more sexy).
In the server, to prevent the pishing, we cut he url to remove the domain name and transform it to relative url.
If the parameter is empty, we redirect to the referer (after cutting the referer url to remove the domain name).

gberaudo:
Maybe we do not need absolute URLs. In this case we can block redirecting to absolute URLs
in angular (with a special case for redirection URLs received in API responses).

fbunoz:
I forgot a particular case.
I don't know how the login will work when the UI will not be c2c, but it will be an external site (partner site). Maybe the full url would be needed ?

asaunier:
> I don't know how the login will work when the UI will not be c2c, but it will be an external site (partner site). Maybe the full url would be needed ?

I don't think it's a problem. The ``auth?from=`` URL is only used to display an auth form in c2corg UI site and then to redirect to the refering page. It is will not be used anyway to authenticate users in a partner website interface (direct access to the API).

PR #580, opened 2016-08-31
Only accept relative URLs for login redirect

tsauerwein:
Closes https://github.com/c2corg/v6_ui/issues/296
Closes https://github.com/c2corg/v6_ui/issues/270


asaunier:
Looks good thanks!

By the way a minor comment: I wonder if should not rename the ``from`` parameter to ``to`` since it actually contains the URL that should be loaded afterwards (and this URL might not be exactly the one we come from).

tsauerwein:
> I wonder if should not rename the from parameter to to

Ok for me. Thanks for the review.

PR #581, opened 2016-08-31
Improve map icon selection

tsauerwein:
The icons contain transparent areas that are not selectable. To make it easier to click on an icon, a transparent circle is rendered behind the actual icon.


asaunier:
👍

PR #582, opened 2016-08-31
Image detail-view + edit + list pages

ginold:
a combination of #575 and #550 with further improvements and some other fixups.

Don't know why the commits appear in disorder, they are ordered in my git log. 

Closes https://github.com/c2corg/v6_ui/issues/586


tsauerwein:
FYI https://github.com/c2corg/v6_api/pull/380 and https://github.com/c2corg/v6_api/pull/379 are merged.

https://github.com/c2corg/v6_api/pull/380 adds a `creator` attribute for the image detail service:
```
"creator": {
    "username": "...",
    "user_id": 123,
    "name": "..."
},
```

And https://github.com/c2corg/v6_api/pull/379 indexes more fields for image in ES: https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/image_mapping.py#L11-L20

You will have to run the `fill_index` script again to update ES:
https://github.com/c2corg/v6_api/tree/master/c2corg_api/scripts/migration#initialize-elasticsearch

ginold:
@tsauerwein, @asaunier this pr can start being reviewed.

c2corg_ui/static/js/card.js
```diff
@@ -119,6 +125,15 @@ app.CardController.prototype.createURL = function() {
 
 
 /**
+ * Don't create any url on edit and add pages.
```
asaunier: I don't understand what this comment means. Do you really need it here?

c2corg_ui/static/partials/cards/images.html
```diff
@@ -0,0 +1,26 @@
...
+
+    <br>
+    <span class="quality" ng-if="doc['quality']" uib-tooltip="{{'quality'| translate}}" tooltip-placement="top">
+      <span class="glyphicon glyphicon-file"></span>
+      <span x-translate>{{doc['quality']}}</span>
+    </span>
+
+    <span class="elevation-max" ng-if="doc['elevation']" uib-tooltip="{{'elevation'| translate}}">
+      <span translate class="value-title glyphicon glyphicon-sort-by-attributes"></span>
+      <span class="value">{{doc['elevation']}}&nbsp;m</span>
```
asaunier: Are you sure one need the ``elevation`` attribute for images? 
I know it is defined in the image model (https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/image.py#L41 => why?) but anyway I don't think it is a relevant information in an image card.

c2corg_ui/static/partials/cards/images.html
```diff
@@ -0,0 +1,26 @@
...
+    <div class="outing-date">{{ doc['date_time'] | date: 'dd/MM/yyyy'}}</div>
+    <div ng-if="doc['author']"><span class="value-title" translate>author</span>: <span class="value">{{doc['author']}}</span></div>
+
+    <span class="list-item-title">{{locale.title}}</span>
+    <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{locale.lang}})</span>
+
+    <br>
+    <span class="quality" ng-if="doc['quality']" uib-tooltip="{{'quality'| translate}}" tooltip-placement="top">
+      <span class="glyphicon glyphicon-file"></span>
+      <span x-translate>{{doc['quality']}}</span>
```
asaunier: Not sure "quality" is really relevant for images either.

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,32 +89,67 @@ app.Authentication.prototype.isAuthenticated = function() {
...
-      if (userid === users[i].id) {
-        return true;
-      }
-    }
-    return false;
-    // by default, user has rights to every doc
+
+  if (users.length) {
+    return this.hasEditRightsOuting_(users);
+  // 'default' is set when imageType is Null in the mako template
```
asaunier: This comment should be in the next block.

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,32 +89,67 @@ app.Authentication.prototype.isAuthenticated = function() {
...
-  if (users) {
-    for (var i = 0; i < users.length; i++) {
-      if (userid === users[i].id) {
-        return true;
-      }
-    }
-    return false;
-    // by default, user has rights to every doc
+
+  if (users.length) {
```
asaunier: This condition is a bit confusing. Why would a list of users mean that we are checking the rights for an outing?

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,32 +89,67 @@ app.Authentication.prototype.isAuthenticated = function() {
...
+    return this.hasEditRightsOuting_(users);
+  // 'default' is set when imageType is Null in the mako template
+  } else if (imageType !== 'default') {
```
asaunier: What happens if ``imageType`` is not defined?

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,32 +89,67 @@ app.Authentication.prototype.isAuthenticated = function() {
...
+    return this.hasEditRightsImage_(imageType);
+
   } else {
+    // by default, user has rights to every doc
     return true;
```
asaunier: You don't need to have an ``else`` block for a default ``return`` statement.
You can simply write:
```
if (cond A) {
  return 'foo';
}
return 'bar';
```

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,32 +89,67 @@ app.Authentication.prototype.isAuthenticated = function() {
...
+ * @return {boolean}
+ * @private
+ */
+app.Authentication.prototype.hasEditRightsImage_ = function(imageType) {
+  if (imageType === 'collaborative') {
+    return true;
+    // TODO: check if user == author
+  } else {
+    return false;
+  }
```
asaunier:
```
app.Authentication.prototype.hasEditRightsImage_ = function(imageType) {
  return imageType === 'collaborative';
};
```

c2corg_ui/static/js/documentservice.js
```diff
@@ -181,7 +181,7 @@ app.Document.prototype.isCollaborative = function(type) {
     return false;
   } else if (type === 'outings' || type === 'articles') {
     // check if user == owner or participant of the outing/article
-    return this.auth_.hasEditRights(this.document.associations['users']);
+    return this.auth_.hasEditRights(this.document.associations['users'], null);
```
asaunier: Do you really need to pass a ``null`` value?
``hasEditRights`` should manage itself the case where no ``imageType`` is passed.

c2corg_ui/static/js/user.js
```diff
@@ -108,8 +108,8 @@ app.UserController.prototype.logout = function() {
  * @return {boolean}
```
asaunier: missing ``@param`` comments

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,32 +89,67 @@ app.Authentication.prototype.isAuthenticated = function() {
 /**
  * Checks if the current user has rights to access/edit the document.
  * TODO: rights to personal documents only for current user.
+ * @param {Array<string>} users
+ * @param {?string} imageType
  * @return {boolean}
  * @export
  */
-app.Authentication.prototype.hasEditRights = function(users) {
+app.Authentication.prototype.hasEditRights = function(users, imageType) {
```
asaunier: I don't find very elegant/convenient to pass an argument that is only relevant for a given type of document.
What about for instance passing a generic ``options`` argument where you could pass the image type when needed?

c2corg_ui/templates/image/view.html
```diff
@@ -66,59 +66,63 @@ <h1 class="routes">
...
+      <section id="document-informations" class="collapse in">
+        ${get_image_general(image)}
+        ${get_image_camera_settings(image)}
+      </section>
+    </div>
+
+      ## DESCRIPTION
+      % if locale['summary'] or locale['description']:
+      <div class="view-details-description col-xs-12  description">
+        <h3 class="heading show-phone"><span translate>personal comments</span></h3>
```
asaunier: Rather "description" instead of "personal comments"

c2corg_ui/static/js/addassociation.js
```diff
@@ -83,11 +83,12 @@ app.AddAssociationController = function(appApi, appDocument) {
  */
 app.AddAssociationController.prototype.associate = function(doc) {
   var parentId, childId;
+  var type = this.parentDoctype;
```
asaunier: ``type`` is a bit confusing compared to ``doc['type']``. It's better to keep "parent" somewhere. At least ``parentType``.

c2corg_ui/templates/helpers/view.html
```diff
@@ -397,11 +397,11 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
...
 ## Float buttons
-<%def name="show_float_buttons(outingUsers, doctype, doc_id, doc_lang, other_langs, missing_langs)">\
+<%def name="show_float_buttons(associatedUsers, doctype, doc_id, doc_lang, other_langs, missing_langs, image_type = 'default')">\
   <div class="float-buttons">
-      <div ng-if="::userCtrl.hasEditRights(${dumps(outingUsers)})" class="float-button float-edit"
-           tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}"
-           protected-url-btn url="${request.route_path(doctype + '_edit', id=doc_id, lang=doc_lang)}">
+      <div ng-if="'${image_type}' == 'collaborative' || userCtrl.hasEditRights(${dumps(associatedUsers)})" class="float-button float-edit"
+       tooltip-placement="left" uib-tooltip="{{'Edit' | translate}}"
+       protected-url-btn url="${request.route_url(doctype + '_edit', id=doc_id, lang=doc_lang)}">
```
asaunier: Are you sure of this ``request.route_url``? It looks like a regression after @tsauerwein has changed it to ``request.route_path``.

c2corg_ui/templates/helpers/view.html
```diff
@@ -397,11 +397,11 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
 
 
 ## Float buttons
-<%def name="show_float_buttons(outingUsers, doctype, doc_id, doc_lang, other_langs, missing_langs)">\
+<%def name="show_float_buttons(associatedUsers, doctype, doc_id, doc_lang, other_langs, missing_langs, image_type = 'default')">\
```
asaunier: I feel uncomfortable with using this ``default`` image type since it actually does not exist.
I would rather make it ``collaborative`` by default. Then your test below
```
ng-if="'${image_type}' == 'collaborative' || userCtrl.hasEditRights(${dumps(associatedUsers)})"
```
would become
```
% if image_type != 'collaborative' :
    ng-if="userCtrl.hasEditRights(${dumps(associatedUsers)})"
% endif
```

c2corg_ui/static/js/addassociation.js
ginold: `type === 'images'` is needed, but if you ask about `var type = this.parentDoctype;` it's not, I just found it shorter to write 3 times 'type' rather than 'this.parentDoctype'.

c2corg_ui/templates/image/filters.html
```diff
@@ -0,0 +1,68 @@
...
+from c2corg_common.attributes import default_langs, activities, image_categories, image_types
+%>
+<%namespace file="../helpers/filters.html" import="add_multiselect"/>
+
+<form app-search-filters app-search-filters-controller-name="appSearchFiltersController">
+  <div class="filters">
+    <div class="row simple-filters">
+
+      <div class="filter">
+        <label translate>date_taken</label><br>
```
asaunier: Rather ``date_time`` as in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/image.py#L55 so that the same string is used as in the diff page?
Or a human readable string such as ``Date``?

asaunier: By the way I don't think the date is a "level 1" criteria for images. Not sure if the list of advanced search criteria is available anywhere.

Here is my suggestions for level 1 criterias (always displayed): 
* title
* categories
* activities

c2corg_ui/templates/image/filters.html
```diff
@@ -0,0 +1,68 @@
...
+
+    <div class="row collapse" id="moreFilters">
+
+      <div class="col-xs-12 col-sm-6 filter">
+        <label translate>title</label>
+        <input type="text" class="form-control" ng-model="filtersCtrl.filters.q">
+      </div>
+
+      <div class="col-xs-12 col-sm-6 filter">
+        <label translate>elevation</label><br>
```
asaunier: Searching on elevation for images is very rare. I think we can remove this filter.

c2corg_ui/templates/image/filters.html
```diff
@@ -0,0 +1,68 @@
...
+        <app-slider filter="oalt" filters-list="filtersCtrl.filters" max="8850"></app-slider>
+      </div>
+
+      <div class="col-xs-12 col-sm-6 filter">
+        <label translate>type</label><br>
+        ${add_multiselect('im_typ', image_types)}
+      </div>
+
+      <div class="filters-bottom-buttons">
```
asaunier: Perhaps we could create helper functions for those buttons ("show results", "clear filters", "all filters") since they will be used "as is" for all doc types? Since it would impact other documents as well, we could do it in a dedicated issue/PR.

c2corg_ui/templates/image/helpers/detailed_images_attributes.html
```diff
@@ -0,0 +1,108 @@
...
+    <span class="glyphicon glyphicon-info-sign"></span>
+    <p translate>General</p>
+  </div>
+  <span class="detail-text accordion">
+
+    % if image.get('activities') :
+      <article class="value activities">
+        <span translate class="value-title">activities</span><br>
+        % for type in image.get('activities') :
+        <span x-translate  class="route-activity icon-${type}" uib-tooltip="{{mainCtrl.translate('${type}')}}" ></span>
```
asaunier: Is ``x-translate`` really needed here? I see no string to be translated.

c2corg_ui/static/js/addassociation.js
asaunier: Please use ``parentType`` at least. Else it is difficult to understand the type of what document you are testing.

c2corg_ui/static/js/card.js
ginold: Maybe better to put it just before the if statement.

c2corg_ui/static/js/auth/authservice.js
ginold: because for now only outings require the list of users to know if you can edit them or not.

c2corg_ui/templates/image/view.html
```diff
@@ -0,0 +1,161 @@
...
+    % if not version:
+       , "associations": ${dumps(image.get('associations')) | n}
+    % endif
+  });
+</%block>
+
+<header class="view-details-title">
+  <h1 class="routes">
+    <div class="icons-header">${get_route_activities(image, 'top')}</div>
+    <span class="title">${show_route_title(locale)}</span><br>
```
asaunier: Are you sure of this "route" helpers?
By the way shouldn't we renamed ``get_route_activities`` to ``get_activities`` since it may be used for routes but also outings or images?

c2corg_ui/templates/image/view.html
```diff
@@ -0,0 +1,161 @@
...
+    <label class="badge" onClick="window.location='${request.route_url('images_index')}' ">
```
asaunier: Why not using a simple HTML link here instead of an angular click handler?

c2corg_ui/templates/image/view.html
```diff
@@ -0,0 +1,161 @@
...
+    <section id="document-informations" class="collapse in">
+      ${get_image_general(image)}
+      ${get_image_camera_settings(image)}
+    </section>
+  </div>
+
+  ## DESCRIPTION
+  % if locale['summary'] or locale['description']:
+  <div class="view-details-description col-xs-12  description">
+    <h3 class="heading show-phone"><span translate>personal comments</span></h3>
```
asaunier: ``personal comments`` => ``description``?

c2corg_ui/templates/image/view.html
```diff
@@ -0,0 +1,161 @@
...
+        % endif
+
+      </div>
+    </span>
+  </div>
+  % endif
+
+  <div class="view-details-description col-xs-12  description">
+    <span class="lead">
+      <img class="image" src="${image_url}/${image['filename']}" alt="${locale['title']}">
```
asaunier: Do you use the full size file here?
If yes, it's probably better to use a resized version with a link towards the full size file. See for instance in v5: http://www.camptocamp.org/images/796074/fr/bietschhorn

c2corg_ui/static/js/documentediting.js
```diff
@@ -151,6 +151,12 @@ app.DocumentEditingController = function($scope, $element, $attrs, appLang,
   this.submit = false;
 
   /**
+   * @type {Date}
+   * @export
+   */
+  this.today = new Date();
```
asaunier: Can't you directly set "today" in the template (expression) instead of here?

c2corg_ui/templates/image/edit.html
```diff
@@ -0,0 +1,316 @@
...
+              <div class="help-block" ng-messages="editForm.iso_speed.$error">
+                <p ng-message="number" class="label label-danger" style="{{ (editForm.iso_speed.$invalid && editForm.iso_speed.$touched) ? 'display:inline' : ''}}" translate>This field must be a number.</p>
+              </div>
+            </div>
+
+            ## WIDTH
+            <div class="form-group half" ng-if="image.width">
+              <label translate>width</label>
+              <div class="input-group">
+                <input type="number" disabled min="0" minlength="1" name="width" ng-model="image.width" class="form-control" />
```
asaunier: This info should be automatically detected, not user-provided. Ditto for height.

c2corg_ui/static/js/documentediting.js
```diff
@@ -504,6 +504,22 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
   app.utils.pushToArray(object, property, value, event);
 };
 
+/**
+ * @param {number} value
+ * @export
+ */
+app.DocumentEditingController.prototype.convertExposureTime = function(value) {
+  if (value === 0) {
+    this['scope']['image']['exposure_time_converted'] = this.alerts.gettext('cannot be 0');
```
asaunier: Not sure to understand this line! How can an error message be saved a document attribute?

c2corg_ui/static/js/documentediting.js
```diff
@@ -504,6 +504,22 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
   app.utils.pushToArray(object, property, value, event);
 };
 
+/**
+ * @param {number} value
+ * @export
+ */
+app.DocumentEditingController.prototype.convertExposureTime = function(value) {
```
asaunier: Please consider moving this function to a dedicated ``imageediting`` controller, as already done for outings.
You could also move the ``date_time`` conversion introduced at https://github.com/c2corg/v6_ui/pull/582/commits/ff1b7f5997ba9fc2aa583d4815f43010d3f78d60

c2corg_ui/static/js/documentediting.js
```diff
@@ -504,6 +504,22 @@ app.DocumentEditingController.prototype.pushToArray = function(object, property,
...
+  if (value === 0) {
+    this['scope']['image']['exposure_time_converted'] = this.alerts.gettext('cannot be 0');
+  } else if (value < 1 && value) {
+    this['scope']['image']['exposure_time_converted'] = '1/' + (1 / value) + ' s';
+  } else if (value >= 1) {
+    this['scope']['image']['exposure_time_converted'] = value + ' s';
+  } else if (!value) {
+    this['scope']['image']['exposure_time_converted'] = '';
```
asaunier: Perhaps you could factorize a bit in order to write ``this['scope']['image']['exposure_time_converted']`` only once.

c2corg_ui/templates/image/edit.html
```diff
@@ -193,17 +195,20 @@ <h2 class="heading show-phone" translate>camera settings</h2>
...
+            <div class="form-group half" ng-class="{ 'has-error': editForm.exposure_time.$invalid, 'has-success': editForm.exposure_time.$valid && image.exposure_time}">
               <label translate>exposure_time</label>
-              <div class="input-group">
-                <span class="input-group-addon">1/</span>
-                <input type="number" min="1" name="exposure_time" ng-model="image.exposure_time" class="form-control">
-                <span class="input-group-addon">s</span>
+              <div class="input-group flex">
+                {{editCtrl.convertExposureTime(image.exposure_time)}}
+                <input type="number" min="0" max="30" name="exposure_time" ng-model="image.exposure_time" ng-change="editCtrl.convertExposureTime(image.exposure_time)" class="form-control"
+                       uib-tooltip="{{'You can type 0.003125 and it will be converted to 1/320s' | translate}}">
```
asaunier: This help message is a bit complicated. Do we really need it?

c2corg_ui/static/js/auth/authservice.js
asaunier: That does not look robust (and explicit for code readers).
For instance "personal articles" should also be editable by associated users only I think.

It would be clearer to check for a type for instance.

c2corg_ui/static/js/documentediting.js
ginold: writing for example `ng-init="today = new Date"(`) will not work.

Maybe we could use a mako variable and render it with pyton `today = ${today}` ? 
why is it more appropriate to set it in the template rather than here?

c2corg_ui/templates/image/edit.html
ginold: they are shown only if the property already exists within the object.  If yes, it will appear in a `disabled` input, so the user cannot edit it in any case.

c2corg_ui/templates/image/edit.html
ginold: i would have no idea that you should type the value in this format if there were no help message :P

c2corg_ui/static/js/documentediting.js
asaunier: > Maybe we could use a mako variable and render it with pyton today = ${today} ? 

That's an option.

> why is it more appropriate to set it in the template rather than here?

Because it is needed only for images, not the other doc types.

I think you should create an ``ImageEditing`` controller. You could move this property here as well.

c2corg_ui/templates/image/edit.html
asaunier: If you cannot edit them, why do you need to display them anyway?
It adds complexity to the editing form, and I don't think it is justified.

c2corg_ui/templates/image/edit.html
asaunier: Perhaps, but who will type "0.003125" anyway?
Even if you know that it will be converted, no one will calculate the value of "1/320s".

c2corg_ui/static/js/documentediting.js
ginold: `Today` is also used for outings, but it has been moved to the `outingEditingController`. So maybe I should remove it from there and leave only `today` in `documentEditingController`?

c2corg_ui/templates/image/edit.html
ginold: in other words this field is useless, at least if we want to stick to it within this format.

c2corg_ui/static/js/documentediting.js
asaunier: Mmmm
I would say it's simpler to create a new controller based upon the outing editing controller and to have ``today`` in the new image editing controller as well (because this property is not needed for other document types).

c2corg_ui/static/js/documentediting.js
ginold: i used `aletrs.gettext` only to get a translated version of the message string t:P  but i'm not sure if that service does it like the `mainController's` `translate()`

asaunier: The alert service should be used to display alerts, not to translate strings.
Anyway I don't think it the place for adding an error message, it should be done in the form itself using the usual validation & message tools, as for the other attributes.

c2corg_ui/templates/image/view.html
ginold: I don't remember the choice for it, probably that an `a` tag is a pain in the eye to style because it has a lot of defaults styles :P

c2corg_ui/templates/image/view.html
ginold: that's probably what's happening, we created the functions when routes arrived but then it turned out to be useful somewhere else than routes. I would prefer to  do it in a separate PR because it requires to change names in lots of pages and files.

asaunier: Could you create an issue for this? Else we will forget it.

ginold:
I updated the PR, you can re-review again.

c2corg_ui/static/partials/cards/images.html
```diff
@@ -0,0 +1,16 @@
+<a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <img ng-src="{{cardCtrl.createImg()}}" alt="{{locale['title']}}">
```
asaunier: Isn't it still the big image file here?

c2corg_ui/templates/helpers/view.html
```diff
@@ -401,20 +401,28 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
 
 
 ## Float buttons
-<%def name="show_float_buttons(outingUsers, doctype, doc_id, doc_lang, other_langs, missing_langs)">\
+<%def name="show_float_buttons(associatedUsers, doctype, doc_id, doc_lang, other_langs, missing_langs, image = 'null')">\
```
asaunier: Rather use ``None`` as default value than ``'null'`` (string).

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -89,31 +89,68 @@ app.Authentication.prototype.isAuthenticated = function() {
...
   if (roles.indexOf('moderator') > -1) {
     return true;
   }
-  // we are checking for editing rights of an outing
-  if (users) {
-    for (var i = 0; i < users.length; i++) {
-      if (userid === users[i].id) {
+  if (doctype === 'outings') {
+    return this.hasEditRightsOuting_(options['users']);
+
```
asaunier: Unneeded blank line

c2corg_ui/templates/helpers/view.html
```diff
@@ -401,20 +401,28 @@ <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($even
...
+  <%
+      if image != "null":
```
asaunier: ``if image:``

c2corg_ui/templates/image/helpers/detailed_images_attributes.html
```diff
@@ -0,0 +1,110 @@
+<%namespace file="../../helpers/view.html" import="show_areas, show_attr, get_document_up_down, get_document_min_max"/>
+<%! import math %>
+
+## GENERAL
+<%def name="get_image_general(image)">\
+<div class="name-icon-value general" ng-click="detailsCtrl.openTab($event)">
```
asaunier: Indenting the content of the function would be nice.

c2corg_ui/templates/image/helpers/detailed_images_attributes.html
```diff
@@ -0,0 +1,110 @@
...
+    <p><span class="value-title" translate>image_type</span>: <span class="value">${image['image_type']}</span></p>
+    % endif
+  </span>
+</div>
+</%def>
+
+## CAMERA SETTINGS
+<%def name="get_image_camera_settings(image)">\
+
+% if image.get('camera_name') or image.get('fnumber') or image.get('exposure_time') or image.get('iso_speed') or \
```
asaunier: Indenting this "if" block would be nice.

asaunier:
Thanks for addressing most of my previous comments. I have added some (minor) others.

c2corg_ui/static/partials/cards/images.html
ginold: No, it uses a new function `app.utils.createImageUrl(filename, suffix)` so that it will download a Medium Image (MI) in this case.

asaunier: > SM was too small and distorted

I guess you mean SI.
Isn't MI a bit too big for cards? Why SI images are distorted? You mean that's because they are squared (are they?)? What's the problem exactly?

c2corg_ui/templates/helpers/view.html
ginold: Ahh, i was returning `Null` and it gave an error `"Null is not defined`" and i was like....DUUH it is not defined xd

c2corg_ui/static/partials/cards/images.html
ginold: I'd love to show you but the images loaded only once and i cannot find them anymore :P SI are distorted because they are simply too small, you can see, even on the cards, that they are pixelated (and they are squared as well).

asaunier: Are the cards that big?! :p 

OK let's try with MI but honestly I'm afraid those files are a bit too heavy for a list page.

asaunier: OK for testing with MI, I was confused thinking you wanted to use the BI images. MI are not that bigger than SI and preserve the whole image, on the contrary of SI that are cropped squared images.

ginold:
Updated, should be ready for merging anytime :)  have a look.

asaunier:
OK I will take a look shortly.

I have created a separated issue concerning the fact that currently images with no geometry are not returned by the advanced search. See https://github.com/c2corg/v6_ui/issues/605

c2corg_ui/templates/image/edit.html
```diff
@@ -0,0 +1,315 @@
+<%!
+from c2corg_common.attributes import default_langs, activities, quality_types, image_types, image_categories \
```
asaunier: The trailing \ is not needed.

asaunier:
I have seen a minor problem else looks good.
Please address this remaining problem and merge, thanks!

Issue #248, opened 2016-03-24
More understandable message when email nonce is invalid


gberaudo:
This could be:
- a bad copy paste by the user;
- a outdated nonce (only the last one is valid, for a short period of time).

tsauerwein:
And the error message should not be shown with this error popup that disappears after 2 seconds. Because after 2 seconds you are left with an empty page.

PR #583, opened 2016-08-31
Show clearer message in case of authentication errors

tsauerwein:
Closes https://github.com/c2corg/v6_ui/issues/248


asaunier:
LGTM

I don't remember: are strings within ``this.alerts_.gettext()`` automatically extracted and added to the PO files / Transifex?

tsauerwein:
> I don't remember: are strings within this.alerts_.gettext() automatically extracted and added to the PO files / Transifex?

Yes, these strings show up in the POT file. I think the extractor checks for method calls with `gettext`.

PR #588, opened 2016-09-01
Increase rate limit for simple search

tsauerwein:
300 ms is still fast enough and this makes 6-times less requests to the API.


asaunier:
:+1:

PR #589, opened 2016-09-05
Col fix

ginold:
small CSS fix for columns on list pages.

it is related to the tooltip that are sometimes cut in the columns of cards of the advanced search pages !


asaunier:
Don't hesitate to explain more why you suggest such changes.
I guess it is related to the tooltip that are sometimes cut in the columns of cards of the advanced search pages.
LGTM.

PR #590, opened 2016-09-05
update association context

ginold:
small update for associations context on editing pages (you should not be able to find already associated docs in the search results).


asaunier:
Not sure to understand what this PR does.

ginold:
for example on the route editing page under 'associations', in the simple-search results you could still find a document that has already been associated.

I added a fix that will check that `app-simple-search` has a `div.section.associations` somewhere in parents and therefore should strip the docs that are contained in `doc.associations`

asaunier:
Please merge

PR #607, opened 2016-09-07
Fix some translate directives

asaunier:
This PR:
* fix some incorrect "translate" directives that made variables being extracted and pushed to Transifex
* replaced some ``<span>{{type | translate}}</span>`` by ``<span x-translate>{{type}}</span>`` to be consistent with what we have done most of the time (for instance in the waypoint edit template)

By the way, does ``<span x-translate>{{type}}</span>`` really work? I was wondering if it could explain why some (all?) dropdown values were not translated (see https://github.com/c2corg/v6_ui/issues/602). Or maybe it's simply because those strings are not explicitely available in the templates and thus not extracted/pushed to Transifex.

### DIFF
--- a/c2corg_ui/templates/waypoint/edit.html
+++ b/c2corg_ui/templates/waypoint/edit.html
@@ -90,7 +90,7 @@ from c2corg_common.attributes import default_langs, climbing_outdoor_types, publ
                  ng-class="{ 'has-error': editForm.lang.$touched && editForm.lang.$invalid, 'has-success': editForm.lang.$valid}">
               <label><span translate>lang</span> *</label>
               % if waypoint_lang:
-              <input class="form-control" disabled x-translate value="${waypoint_lang}">
+              <input class="form-control" disabled value="{{mainCtrl.translate('${waypoint_lang}')}}">
               % else:
               <select name="lang" ng-options="lang as mainCtrl.translate(lang) for lang in ['${"','".join(default_langs) |n}'] | translate"
                       ng-model="waypoint.locales[0].lang" class="form-control" required><option value=""></option></select>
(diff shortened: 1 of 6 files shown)


asaunier:
> By the way, does <span x-translate>{{type}}</span> really work?

It seems not and is most likely one explanation for https://github.com/c2corg/v6_ui/issues/602

On the contrary ``<span>{{type | translate}}</span>`` seems to work and "type" is not added to the POT file by the extractor.

asaunier:
PR updated and ready to be reviewed!

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -90,7 +90,7 @@ <h2 class="heading show-phone"><span translate>title</span></h2>
                  ng-class="{ 'has-error': editForm.lang.$touched && editForm.lang.$invalid, 'has-success': editForm.lang.$valid}">
               <label><span translate>lang</span> *</label>
               % if waypoint_lang:
-              <input class="form-control" disabled x-translate value="${waypoint_lang}">
+              <input class="form-control" disabled value="{{mainCtrl.translate('${waypoint_lang}')}}">
```
tsauerwein: Why are you not using the `translate` filter here?

tsauerwein:
> On the contrary <span>{{type | translate}}</span> seems to work and "type" is not added to the POT file by the extractor.

If that is the case, good! Otherwise `x-translate="{{type}}"` should work.

c2corg_ui/templates/waypoint/edit.html
asaunier: You mean as a filter like ``value="{{'${waypoint_lang} | translate}}'"``?
In that case ``${waypoint_lang}`` is pushed to the POT files by the extractor.
Using ``x-translate`` as a filter does not work (no effect).

If as an attribute (as the original ``x-translate``), the translation does not work because the directive does not apply to the ``value`` content.

asaunier:
OK thanks!

c2corg_ui/templates/waypoint/edit.html
tsauerwein: > You mean as a filter like value="{{'${waypoint_lang}' | translate}}'"?
> In that case ${waypoint_lang} is pushed to the POT files by the extractor.

Didn't you say that using `translate` as a filter does not extract the key into the POT files? I am confused? :)

asaunier: It seems the extractor is smart enough to extract only strings, not variables.
For instance ``{{'foo' | translate}}`` will make the string "foo" added to the POT file.
On the other hand ``{{foo | translate}}`` won't.

Issue #613, opened 2016-09-10
Route creation : WP unavailable after creation (error 500)


mgerbaux:
When creating a route, WP that are associated during the creation process are unavailable after (500 Internal Server Error). Example : http://c2corgv6.demo-camptocamp.com/waypoints/109614/fr/habert-de-chamechaude

Edit : It seems no to be always the case...

tsauerwein:
@asaunier The API returns the document without error (http://c2corgv6-api.demo-camptocamp.com/waypoints/109614). Something in the UI logs?

asaunier:
There's probably an error in the mako templates/helpers. The UI apache logs say:

> [Tue Sep 13 09:23:37.936215 2016] [wsgi:error] [pid 25832] [remote 10.27.71.4:63314]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/waypoint/view.html", line 67, in render_moduleConstantsValues
[Tue Sep 13 09:23:37.936226 2016] [wsgi:error] [pid 25832] [remote 10.27.71.4:63314]     ${associated_routes_featurelist(waypoint)}
[Tue Sep 13 09:23:37.936234 2016] [wsgi:error] [pid 25832] [remote 10.27.71.4:63314]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/helpers/view.html", line 258, in render_associated_routes_featurelist
[Tue Sep 13 09:23:37.936240 2016] [wsgi:error] [pid 25832] [remote 10.27.71.4:63314]     % if 'geometry' in route and route.get('geometry').get('geom_detail'):
[Tue Sep 13 09:23:37.936251 2016] [wsgi:error] [pid 25832] [remote 10.27.71.4:63314] AttributeError: 'NoneType' object has no attribute 'get

asaunier:
The API no longer returns the associated routes geometries for a WP document. So this problem should no longer be observed.

stef74:
@mgerbaux please test again and close if it's ok.

PR #618, opened 2016-09-13
Waypoint view fixes

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/613

I have also added the elevation attribute to the WP general section (by the way it is useful for the preview since the preview has no map at the moment). Not sure it is really necessary to display the areas+coords (wrong projection by the way)+elevation twice in the doc, but if OK then it's more consistent to show the elevation as well.
sans titre

### DIFF
--- a/c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
+++ b/c2corg_ui/templates/waypoint/helpers/detailed_waypoint_attributes.html
@@ -30,6 +30,10 @@
     <p translate>General</p>
   </div>
   <span class="detail-text accordion">
+    % if waypoint.get('elevation'):
+      <p><span translate class="value-title">elevation</span>: <span class="value">${waypoint['elevation']} m</span></p>
+    % endif
+
     % if waypoint.get('waypoint_type'):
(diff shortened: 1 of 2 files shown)


ginold:
i thought it was fixed. In fact, the left coordinates + location should be only shown on mobile because the location-post-it is only visible on tablet/desktop.

asaunier:
OK that would make sense.

As for now, both blocks are shown (as in my screenshot that I have just made with my instance using the master branch). Should we create an issue for this?

Anyway I have added the elevation attribute to the "general" block (not the "location" block). So it's not really related to this duplicated block issue.

asaunier:
I have removed the commit fixing the associated routes geometries since the API no longer returns associated routes geometries.

Only remains a small change displaying the WP elevation in the document itself (not only on the map).

@ginold please review

ginold:
add `display: none;` and remove `display: flex;` in `viewdetails.less` at line 434 so that the location tab is not displayed on tablet/desktop, only on phone.

ginold:
```
      &.location {
        display: none;
        @media @phone {
          .flex();
        }
        .areas {
          @media (min-width: @phone-max) {
            display: none;
          }
        }
      }
```

asaunier:
OK thanks, I have updated the PR as you suggest it.

PR #620, opened 2016-09-13
make docker image publishing work again

mfournier:
See error at https://travis-ci.org/c2corg/v6_ui/jobs/159540799#L3954 for example.

cf https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A

This seems to have been broken (for non-master branches) since fe10809.


tsauerwein:
Is it correct that you want to create a docker image for each branch, not only master?

If so feel free to merge.

Issue #619, opened 2016-09-13
Changes required to match API new simple/advanced search attributes


asaunier:
This PR in the API makes the simple/advanced search + association attributes consistent: https://github.com/c2corg/v6_api/pull/391

Those changes have impacts on the UI side. Merging PR https://github.com/c2corg/v6_api/pull/391 will most likely break the UI.

The following changes are required:
* [x] Users in outings now have their id as ``document_id`` instead of ``id``
* [x] Route associations for waypoints no longer include ``geometry_detail`` => remove the associated routes geometry in WP maps (unless we can get the geometry with some other way?)
* [x] The structure for recent_outings has been changed from ``'recent_outings': {'outings': ...}}`` to ``'recent_outings': {'documents': ...}}`` (same for ``all_routes``).

asaunier:
> Users in outings now have their id as document_id instead of id

@tsauerwein do you have examples of actions that are impacted by this change?

tsauerwein:
> @tsauerwein do you have examples of actions that are impacted by this change?

For example here:
https://github.com/c2corg/v6_ui/blob/f72c289c22b4d85c7cbff0e82062b69b70e35a12/c2corg_ui/static/js/outingediting.js#L81
https://github.com/c2corg/v6_ui/blob/f72c289c22b4d85c7cbff0e82062b69b70e35a12/c2corg_ui/static/js/outingediting.js#L152-L155
https://github.com/c2corg/v6_ui/blob/ab598ce7865714a083df20fe1e79652c3680f4e1/c2corg_ui/tests/views/data/outing.json#L69-L76

asaunier:
Concerning https://github.com/c2corg/v6_ui/blob/f72c289c22b4d85c7cbff0e82062b69b70e35a12/c2corg_ui/static/js/outingediting.js#L152-L155
I thought we could remove this piece of code. But if I do, ``id`` is not set in the ``associations.users`` item and the API fails saving the outing saying that
> description: "Required"
location: "body"
name: "associations.users.0.id"

Shoulldn't the API accept ``document_id`` instead of ``id`` in the outings recording requests?

tsauerwein:
> Shouldn't the API accept document_id instead of id in the outings recording requests?

Yes, that would now be more consistent. It can be changed here:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/schema_utils.py#L69-L70

asaunier:
You mean changing to
```
[...]

PR #621, opened 2016-09-13
API consistency changes

asaunier:
Closes https://github.com/c2corg/v6_ui/issues/619

### DIFF
--- a/c2corg_ui/externs/appx.js
+++ b/c2corg_ui/externs/appx.js
@@ -276,7 +276,7 @@ appx.Image;
 /**
  * @typedef {{
  *   total: number,
- *   outings: Array.<appx.Outing>
+ *   documents: Array.<appx.Outing>
  * }}
  */
 appx.RecentOutings;
@@ -285,7 +285,7 @@ appx.RecentOutings;
 /**
  * @typedef {{
  *   total: number,
- *   routes: Array.<appx.Route>
+ *   documents: Array.<appx.Route>
  * }}
  */
 appx.AllRoutes;
--- a/c2corg_ui/static/js/auth/authservice.js
+++ b/c2corg_ui/static/js/auth/authservice.js
@@ -131,7 +131,7 @@ app.Authentication.prototype.hasEditRightsOuting_ = function(users) {
   }
   if (u) {
     for (var i = 0; i < u.length; i++) {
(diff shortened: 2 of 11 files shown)


asaunier:
This PR addresses all points mentioned at https://github.com/c2corg/v6_ui/issues/619

Commit fdb5ae7 might not be needed, it depends if there's a consistency issue in the API. See https://github.com/c2corg/v6_ui/issues/619#issuecomment-247307134

I have noticed 2 problems in the outing editing page:
* the user name is missing in the card of the current user (by default in the associated users). Most likely not related to this PR since the same behaviour is observed in the demo. I have created a new issue for this: https://github.com/c2corg/v6_ui/issues/623
* when using the route association tool in the outing editing form, the following error is returned many times after selecting a result. Then again when moving to the next step or when making a new search. Ditto with the user association tool. No error is shown in debug mode! No problem in the detail view page. No problem on the demo (yet?).
> deps.js:157 TypeError: Cannot read property 'length' of undefined
    at so.Z.ig (build.js:426)
    at fn (eval at compile (deps.js:272), <anonymous>:4:509)
    at m.$digest (deps.js:182)
    at m.$apply (deps.js:185)
    at HTMLInputElement.<anonymous> (build.js:353)
    at HTMLInputElement.dispatch (deps.js:5)
    at HTMLInputElement.r.handle (deps.js:5)
    at Object.trigger (deps.js:6)
    at HTMLInputElement.<anonymous> (deps.js:6)
    at Function.each (deps.js:4)

Don't know what we could do for the latter error...

asaunier:
A quick review of this PR would be cool because in the meantime, several pages are broken since the changes in the API.

tsauerwein:
> Don't know what we could do for the latter error...

In `build.json` you could add the option `"formatting": "pretty_print"` to get more context.

asaunier:
> In build.json you could add the option "formatting": "pretty_print" to get more context.

Thanks for the tip. Well, even with that option I'm still a bit confused with this error.

asaunier:
@ginold Is the PR OK for you? Do you see other places where the associated users ``id`` should be converted to ``document_id``?

asaunier:
Merging because it's a pain in the neck to have all those API-changes-related errors when working on the UI.

PR #624, opened 2016-09-15
Add config for the demo@c2corg instance

asaunier:


mfournier:
Cherry-picked to alpha3 as 9c8b859. Thanks for taking care of this @asaunier !

PR #630, opened 2016-09-19
Add basic print stylesheet

asaunier:
Very basic version of the print stylesheet. Many improvements are still needed to have a nice printing layout but won't be done in this PR (done by the association later?).

I have used a standard CSS file, not "less" files as for the actual screen styles, because I thought those styles would not be that numerous. Should we use less instead? Perhaps in the future if we figure out we need to add many styles for printing?


asaunier:
Not sure it is really needed in the print layout, but I have found a way to print the URL of links:
> @media print{
  a:after {
    content:" (" attr(href) ") ";
  }
}

See http://www.partage-it.com/faites-bonne-impression-avec-twitter-bootstrap/

asaunier:
@ginold could you review this PR? Thanks :)

ginold:
I don't think there is any advantage of using css over less - this css file will surely grow bigger and less could save a few lines. 

I'm not a fan of writing the link href on the printed form, but it's just my opinion :P 

```
else { 
 looksGood = true;
} 
```

asaunier:
I have updated the PR in order to use less for the print styles as well.

> I'm not a fan of writing the link href on the printed form, but it's just my opinion :P

OK. Not sure we really need it either, I just wanted to save the tip somewhere, just in case it would be useful.

PR #632, opened 2016-09-19
Add basic support of Google Analytics

asaunier:
Simple integration based upon the code of v5:
https://github.com/c2corg/camptocamp.org/blob/1e60b94803765ca09fba533755dad0ecd4cad262/apps/frontend/modules/common/templates/_tracker.php

Please note though that:
* dimensions from v5 are not supported yet. They were used to track form factor and user status (connected or not)
* forum pages are not impacted by this PR. See https://github.com/c2corg/v6_forum/issues/46

I have also noticed there were AngularJS libs for integrating Google Analytics. For instance:
* https://www.npmjs.com/package/angular-google-analytics / https://github.com/revolunet/angular-google-analytics
* http://angulartics.github.io/
Should we use such tools instead?

### DIFF
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -143,6 +143,7 @@ class NotFound():
           'error_msg': self.context.detail if self.context.detail else '',
           'ign_api_key': self.settings['ign_api_key'],
           'bing_api_key': self.settings['bing_api_key'],
+          'ganalytics_key': self.settings['ganalytics_key'],
           'image_backend_url': self.settings['image_backend_url'],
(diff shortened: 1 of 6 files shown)


ginold:
just `ganalytics_key` missing, but it will surely be generated. 
Looks good.

asaunier:
Yes, I have used a dummy "FIXME" key instead of the one used currently in v5, to avoid mixing the prod pages and the v6 ones still in dev/demo.

I have added an issue listing things to do before going live: https://github.com/c2corg/v6_ui/issues/639

Thanks for reviewing.

PR #636, opened 2016-09-20
Associated user documents id change

asaunier:
The API now use 'document_id' instead of 'id' for associated users as well.
See https://github.com/c2corg/v6_api/commit/e0f61cde92e45830b856b3f930f4b3d5e1f7d271

See also https://github.com/c2corg/v6_ui/issues/619#issuecomment-247307134

### DIFF
--- a/c2corg_ui/static/js/outingediting.js
+++ b/c2corg_ui/static/js/outingediting.js
@@ -147,10 +147,6 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
       outing.date_end = window.moment(outing.date_end).format('YYYY-MM-DD');
     }
 
-    var associations = outing.associations;
-    for (var i = 0; i < associations['users'].length; i++) {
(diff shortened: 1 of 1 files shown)


c2corg_ui/static/js/outingediting.js
```diff
@@ -147,10 +147,6 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
       outing.date_end = window.moment(outing.date_end).format('YYYY-MM-DD');
     }
 
-    var associations = outing.associations;
-    for (var i = 0; i < associations['users'].length; i++) {
-      associations['users'][i]['id'] = associations['users'][i]['document_id'];
```
ginold: You removed a part of the code but added nothing back - is this one already done elsewhere?

asaunier: The API now uses ``document_id`` instead of ``id``for users as well. See https://github.com/c2corg/v6_api/commit/e0f61cde92e45830b856b3f930f4b3d5e1f7d271

Issue #641, opened 2016-09-21
Hard-coded forum URL in sidemenu


asaunier:
The forum URL is hard-coded in the navigation menu:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/sidemenu.html#L25
It should rather be passed from the config. Probably not a big change since Mako variable ``discourse_url`` is already available:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L56
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/index.py#L20
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/__init__.py#L149

Issue #625, opened 2016-09-16
Home Page / Feed V6.0


fjacon:
The activity feed will be a combination of:

- **Activities and regions filters**. Filters remove feed elements that do no match. If no filter is set, all recently created documents would be shown => default feed (for anonymous visitors as well).
 * no lang fitler
 * The feed does not advertise creations of new users.
 * For images: a feed item for grouped images upload (not one per uploaded image).
- Followed documents. They add feed elements, even not matching filters. Actually if the user has set no filters, following documents has no effect since the feed elements are shown anyway.

It should be made obvious somewhere that filters are activated: filter icon or on/off button + a button opening the filter settings interface.

**Only users may be followed**

The home page feed only shows:
* documents created by the followed users or outings they have been associated to. _Note: outings should be shown only once even though several followed users are associated._
* documents (including outings) matching the filters.

Relies on https://github.com/c2corg/v6_api/issues/403

asaunier:
For the records, the "old" issue is here https://github.com/c2corg/v6_ui/issues/226

By the way the comment above only describes the activity feed. But the homepage also have a few other components. For instance an introduction text (for unregistered visitors) + image (see the wireframes).

asaunier:
This issue is based upon the ``feed`` services recently added to the API:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/feed.py
* **homepage activity feed for anonymous users**: ``/feed``
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/feed.py#L23-L58
```
        Request:
            `GET` `/feed[?pl=...][&limit=...][&token=...]`
        Parameters:
            `pl=...` (optional)
            When set only the given locale will be included (if available).
            Otherwise all locales will be returned.
            `limit=...` (optional)
            How many entries should be returned (default: 10).
            The maximum is 50.
            `token=...` (optional)
            The pagination token. When requesting a feed, the response includes
            a `pagination_token`. This token is to be used to request the next
            page.
```
* **homepage activity feed for authenticated users**: ``/personal-feed``
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/feed.py#L61-L82
```
        Request:
            `GET` `/personal-feed[?pl=...][&limit=...][&token=...]`
        Parameters: See above for '/feed'.
```

The activity feed UI will also be used in the user profile page (see https://github.com/c2corg/v6_ui/issues/335):
[...]

PR #642, opened 2016-09-21
Homepage first version

asaunier:
Here is a very first version of the homepage with:
* some intro text + image, shown only for not authenticated users (as in the wireframes)
* the outing advanced search feed (unfiltered) as a preview of the still-to-come activity feed.

A screenshot:
sans titre
I'm sure we could have something much nicer (for instance showing the whole image instead of a small part of the sky :P ). I need the help of a real artist! @ginold ?

By the way concerning the image, I have noticed that HTML5 provides a special ``srcset`` attribute that can be used to specify several image files whose sizes fit the screen size/resolution (responsive). See 
https://developer.mozilla.org/fr/docs/Web/HTML/Element/Img
http://www.alsacreations.com/article/lire/1621-responsive-images-srcset.html
My code doesn't use it yet but we should probably do something like that to avoid loading a big image in a phone screen.

Related to https://github.com/c2corg/v6_ui/issues/625
By the way, the PR fixes https://github.com/c2corg/v6_ui/issues/641


ginold:
screenshot from 2016-09-23 17 41 05
screenshot from 2016-09-23 17 40 50

asaunier:
I knew you were the real artist of the team :)
I will have a look at your changes but at least the screenshots are awesome.
Showing the activity icons is cool (as in the wireframes) but it's not very consistent to show only 3. Would it be possible to show all of them (with tooltips)?

asaunier:
By the way, we'll probably need a "read more" link at the end of the small paragraph, pointing to an article with more details.

c2corg_ui/templates/index.html
```diff
@@ -1,4 +1,48 @@
...
+        facilitate information sharing between
+        mountain addicts and contribute to the safety of mountain activities.
+      </p>
+
+      <div class="intro-shortcuts">
+        <a href="${request.route_url('routes_index')}">
+          <button class="btn green-btn">
+            <span class="glyphicon glyphicon-arrow-right"></span>
+            <span translate>Topoguide</span><br>
+            <span translate class="small">Cartes, itinéraires, sommets...</span>
```
asaunier: It would be better to have the strings in english: "Maps, routes, summits..."

c2corg_ui/templates/index.html
```diff
@@ -1,4 +1,48 @@
...
+          </button>
+        </a>
+        <a href="${discourse_url}">
+          <button class="btn blue-btn" translate>
+            <span class="glyphicon glyphicon-th-list"></span>
+            <span translate>Forum</span><br>
+            <span translate class="small">Partagez avec nous!</span>
```
asaunier: In english as well

c2corg_ui/templates/index.html
```diff
@@ -1,4 +1,48 @@
...
+          </button>
+        </a>
+      </div>
+
+    </div>
+  </section>
+  <section ng-hide="!userCtrl.auth.isAuthenticated()" class="site-intro">
+    <div class="intro-banner">
+      <h1 class="welcome-title text-center" translate>Feed</h1>
```
asaunier: Perhaps "Activity feed"?
What about showing this title for all users, whether authenticated or not?

asaunier:
What about using different versions of the image depending on the screen?

ginold:
Sure. I thought that it would be nice to have the possibility to choose from existing images or upload your own like on facebook, but that maybe in the faaar future :P

ginold:
> Would it be possible to show all of them (with tooltips)?

i find it too much and there will be problems with different sizes of the screen, where 1 or 2 activities can show just below, which is odd.

screenshot from 2016-09-26 09 32 12

less/home.less
```diff
@@ -0,0 +1,94 @@
+.home-page {
```
ginold: rename the file name. It's not the best we have thought of :P

asaunier:
> Sure. I thought that it would be nice to have the possibility to choose from existing images or upload your own like on facebook, but that maybe in the faaar future :P

In the future the image will probably be picked among a pool of nice images but as for now that's out of the scope.

> i find it too much and there will be problems with different sizes of the screen, where 1 or 2 activities can show just below, which is odd. And you imagine that this icons take half of the screen on mobile.

That's not consistent to show only some activities. All of them are important. We should either show all of them or none.
What about listing the icons on 2 lines on wide screens and hide then on mobile screens?

ginold:
updated the pr with all icons and redirection to outings?activity. I think we should leave the question of whether to show or not the menu for another time.

asaunier:
Could you provide a screenshot?

> I think we should leave the question of whether to show or not the menu for another time.

+1

asaunier:
You have to rebase your branch in order to get the c2corg_common fix you made yesterday. Else Travis fails.

c2corg_ui/templates/index.html
```diff
@@ -1,4 +1,53 @@
...
-<app-map></app-map>
+<div class="home-page">
+  <section ng-hide="userCtrl.auth.isAuthenticated()" class="site-intro not-logged">
+    <div class="intro-banner">
+      <h1 class="welcome-title text-center" translate>The mountain sports community</h1>
+
+      <div class="route-activities text-center">
+        % for activity in activities:
+        <a href="${request.route_url('outings_index')}#&act=${activity}">
+          <span class="activity icon-${activity}" uib-tooltip="{{'${activity}' | translate}}"></span>
```
asaunier: I think ``{{'${activity}' | translate}}`` will add ``${activity}`` to Transifex because of the quotes => the extractor will think it's a standard string, not a variable.
You probably have to use ``mainCtrl.translate()`` here instead.

c2corg_ui/templates/index.html
```diff
@@ -1,4 +1,53 @@
...
+            <span class="glyphicon glyphicon-arrow-right"></span>
+            <span translate>Topoguide</span><br>
+            <span translate class="small">Maps, routes, summits...</span>
+          </button>
+        </a>
+        <a href="${discourse_url}">
+          <button class="btn blue-btn" translate>
+            <span class="glyphicon glyphicon-th-list"></span>
+            <span translate>Forum</span><br>
+            <span translate class="small">Share with !</span>
```
asaunier: Share with who?

c2corg_ui/templates/index.html
```diff
@@ -1,4 +1,53 @@
...
+        facilitate information sharing between
+        mountain addicts and contribute to the safety of mountain activities.
+      </p>
+
+      <div class="intro-shortcuts">
+        <a href="${request.route_url('routes_index')}">
+          <button class="btn green-btn">
+            <span class="glyphicon glyphicon-arrow-right"></span>
+            <span translate>Topoguide</span><br>
+            <span translate class="small">Maps, routes, summits...</span>
```
asaunier: Rather write ``Conditions, summits, routes, huts, climbing sites, images...`` if it's not too long.

ginold: too long!

asaunier: OK, then "Conditions, summits, routes..." (maps are not that important I think)

asaunier: One for the whole sentence is good.

ginold:
screenshot from 2016-09-28 12-17-09
screenshot from 2016-09-28 12-17-20

ok for merging?

c2corg_ui/templates/index.html
asaunier: What about using a clearer title here?

asaunier:
Have the changes in the activity icon an impact in the other pages that use such icons?

asaunier:
Ahem, what about my question?

ginold:
ops, i clicked on the wong PR :P 
I will bring a fixup soon.
As for icons - it's even better and centered, no impacts.

ginold:
you can just quickly check if the icons correspond to the names, I though there was a confusion between mountain_climbing and rock_climbing.

c2corg_ui/templates/index.html
ginold: showing the title or the feed? i think it's a good thing to show a default activity feed for not logged users and a personalized one for logged ones. 

Clearer title = activity feed? can be.

asaunier: Yes, I mean "Feed" => "Activity feed".
It's more less generic and i'ts better to change it before a translation is added in Transifex (because I think that the translations are lost in Transifex when updating the source string!?). Since the PR has been merged, I think you have to make a new one.

asaunier:
> you can just quickly check if the icons correspond to the names, I though there was a confusion between mountain_climbing and rock_climbing.

I have just updated the demo: http://c2corgv6.demo-camptocamp.com/
Yes "mountain_climbing" and "rock_climbing" icons are swapped (again :P)

asaunier:
I notice that when authenticated, the background is briefly shown. Would it be possible to skip it (ng-cloak?) and avoid that the image is loaded since we don't need it?

ginold:
would you have a second image for logged users? it would be consistent to fill the empty space :P

asaunier:
No image for authenticated users. This section is not displayed for them, see the wireframes.

PR #643, opened 2016-09-21
Image slide

ginold:
fixes partly #478

- uses small/big images for thumbnails/gallery
- edit-button is working and redirects to the editing page
- more fields are saved and sent to the API
- user can geolocate the image on the editing page

To-do : API side - return all the relevant fields of a document's associated images.

### DIFF
--- a/c2corg_ui/static/js/utils.js
+++ b/c2corg_ui/static/js/utils.js
@@ -119,23 +119,13 @@ app.utils.getImageFileBase64Source = function(file) {
  * @export
  */
 app.utils.createImageSlide = function(file, imageUrl) {
-  var img;
-  var ahref;
-  // if !src => it's an already existing image, else it's an image that has just been uploaded
-  if (!file['src']) {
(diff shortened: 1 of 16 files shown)


c2corg_ui/static/js/utils.js
```diff
@@ -121,21 +121,18 @@ app.utils.getImageFileBase64Source = function(file) {
...
-};
+  var smallImage = app.utils.createImageUrl(file.filename, 'SI');
+  var bigImage = app.utils.createImageUrl(file.filename, 'BI');
 
+  ahref = '<a href="' + imageUrl + bigImage + '" data-info-id="' + file['image_id'] + '-slide">';
+  img = '<img src="' + imageUrl + smallImage + '">';
+
+  return '<figure id="' + file['image_id'] + '">' +
+              ahref + img +
+              '</a>' +
```
asaunier: That would make more sense to directly embed the ``<a>`` and ```` code in the HTML above, that would be more readable. And also more consistent. For instance ``ahref`` contains the opening ``<a>`` tag whereas the closing tag is in the HTML above.

asaunier:
A minor comment else looks good to me!

Issue #605, opened 2016-09-07
Images advanced search issues


asaunier:
* [x] Since the image advanced search uses a geometry filtering criteria (map), images with no geometry are not returned by the advanced search. It should be possible to disable the map filtering (by default?). Related to https://github.com/c2corg/v6_ui/issues/504

PR #648, opened 2016-09-23
Add a map filter switcher in search maps

asaunier:
Add a checkbox on the advanced search maps to enable (default)/disable the map extent filtering.
Fixes https://github.com/c2corg/v6_ui/issues/504
Fixes https://github.com/c2corg/v6_ui/issues/605

Still TODO:
* [x] improve styling and position on the map
* [x] make sure duplicate requests to the API are not triggered
* [x] handle case where no search results have a geometry (eg. images) => currently the map recentering is wrong


asaunier:
This PR is ready to be reviewed!

tsauerwein:
That looks good!

Is it always enabled? Would it make sense to disable it by default for images?

asaunier:
Thanks.
Yes it's always enabled by default.
Disabling it by default for images would make sense indeed. I will have a look and see if it is reasonably possible.

asaunier:
I have updated the PR to add the possibility to specify if the map filter is ON by default (OFF for images, ON for other types).

c2corg_ui/static/js/map/map.js
```diff
@@ -145,6 +146,18 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
   this.showRecenterTools;
 
   /**
+   * @type {boolean}
+   * @export
```
tsauerwein: Do you need the `@export`?

tsauerwein:
:+1:

c2corg_ui/static/js/map/map.js
asaunier: I think we need it, else the directive attribute value (in the HTML) is not taken into account. Most likely because Closure renames the property where its name is fixed in the directive scope list above.

asaunier:

PR #649, opened 2016-09-26
Call transifex-send in before_deploy

arnaud-morvan:
Fix deploy script not launched after trasifex push.

Note that ``all_branches: true``  is incompatible with ``tags: true``.
Seems we have to make a choice between tags and branches here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,14 +53,10 @@ script:
 after_failure:
   cat v6_api/nohup.out
 
+before_deploy:
+- make transifex-send
+
 deploy:
-- provider: script
-  script: make transifex-send
-  skip_cleanup: true
-  on:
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
@marc: docker build has passed on 4th try. But we should make a choice between tags and branches.

mfournier:
Thanks for tracking this down @arnaud-morvan !

I vote for branches. I feel this will give us more flexibility on the long run.

asaunier:
I can't say if your changes are the good ones but thanks for fixing the problem. Have you tested that new translations are actually pushed to transifex? perhaps we will be able to test only when the PR is merged?

arnaud-morvan:
Superseeded by #675

Issue #272, opened 2016-04-06
Support of the "quality" field


asaunier:
All documents (except maybe user profiles) have a "quality" attribute that is used to figure out they should be improved or even filtered (for instance to hide outings almost empty).

TODO:
* [x] Use "quality_types" values from https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L610-L616 in the editing forms
* [x] Send the quality value to the API when editing a document (required attribute)
* [x] Display the quality attribute in the document detail views and list views (more instructions to come)
* [x] Set a default quality value for documents migrated from v5: ``correct``(3rd level)
* [x] Set a default quality value when creating a document (default value of the dropdown): ``medium``(2nd level), whatever the document type

Related to https://github.com/c2corg/v6_common/issues/18 and https://github.com/c2corg/v6_api/issues/220

stef74:
For quality display use a colored disc : http://www.camptocamp.org/images/755110/fr/picto-completude-couleurs-seules-20px

An bulble infos show the quality name

asaunier:
You can post images directly in the issues by drag-dropping them in a comment:
![1462784786_839360498](https://cloud.githubusercontent.com/assets/1192331/16228529/dd78dc8c-37b7-11e6-9e93-622de8bc07a8.png)

asaunier:
Currently the quality is displayed as text in the documents (not with the disk icons shown above).
And not displayed in the cards I think.

ginold:
the quality field is now displayed in the view as well as on the cards.
Can be closed?

ginold:
@asaunier  ?

stef74:
@shortoland please test
For me quality icon is not good in outing ... quality not show in route ? (reminder the specification for card content : https://docs.google.com/spreadsheets/d/1W2kEAyGPc7cw9yKe3hyYIMTNgUDLPsFUD0ccTS8jI5Y/edit?usp=sharing)

shortoland:
@ginold Are you sure it is ready for testing ? I still see text and no disk icons.

shortoland:
![image100](https://cloud.githubusercontent.com/assets/20488365/18544524/4c186d7c-7b35-11e6-8ba1-ee68c4fc8c80.png)

stef74:
@ginold it is ready for test ?

ginold:
yep it's already online.

shortoland:
[...]

PR #652, opened 2016-09-26
add 'excellent' icon & remove quality icon from associations

ginold:
further improvements for #272

### DIFF
--- a/less/helpers.less
+++ b/less/helpers.less
@@ -331,7 +331,11 @@ p {
 .center-items-align {
   align-items: center;
 }
-
+.associated-documents {
+  .quality {
+    display: none;
+  }
+}
 .quality {
   width: 25px;
   height: 23px;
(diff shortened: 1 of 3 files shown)


less/helpers.less
```diff
@@ -331,7 +331,11 @@ p {
 .center-items-align {
   align-items: center;
 }
-
+.associated-documents {
+  .quality {
+    display: none;
```
asaunier: Is it to avoid the position conflict with the "remove association" icon in the bottom right corner of cards?
Does it mean the quality is only shown in the advanced search pages?

ginold:
as the client stated 

>  Associated WP: quality shouldn't be displayed (Desktop & Mobile)

and 

> Associated route to WP: quality shouldn't be displayed (Desktop & Mobile)

that mainly means that the quality is only displayed on the list and view  pages.

asaunier:
OK, looks good to me then.

PR #653, opened 2016-09-27
Install JDK for Ubuntu 16.04

asaunier:
My workstation has just been reinstalled with Ubuntu 16.04. It seems no longer possible to install ``openjdk-7-jdk`` but it works as expected with ``openjdk-8-jdk`` (``openjdk-9-jdk`` not tested).


tsauerwein:
Thanks, please merge.

Issue #658, opened 2016-09-28
Can not create outing with minimal required fields


tsauerwein:
On http://c2corgv6.demo-camptocamp.com/outings/add I am trying to create an outing with the minimal required fields (marked with a star): Activity (selected: hiking), taken routes (selected: Mont Granier : Pilier des Grottes), date_start (select: 27.09.2016), outing title (using default).

But I can't click on "Save", the title message is saying "What is missing?" (I'd really like to know.. ;).

tsauerwein:
Probably related to https://github.com/c2corg/v6_ui/issues/631

Issue #631, opened 2016-09-19
JS errors when using the association tools in the outing editing page


asaunier:
When creating (updating?) an outing, when using the route association tool (or the user association tool), routes are actually added but the console shows many JS errors:

> deps.js:157 TypeError: Cannot read property 'length' of undefined
    at so.Z.ig (build.js:426)
    at fn (eval at compile (deps.js:272), <anonymous>:4:482)
    at m.$digest (deps.js:182)
    at m.$apply (deps.js:185)
    at HTMLInputElement.<anonymous> (build.js:353)
    at HTMLInputElement.dispatch (deps.js:5)
    at HTMLInputElement.r.handle (deps.js:5)
    at Object.trigger (deps.js:6)
    at HTMLInputElement.<anonymous> (deps.js:6)
    at Function.each (deps.js:4)

The association tools seem to work. Perhaps it prevents the "Save" button from behind activated?

These errors seem to have been introduced by the changes related to the API changes https://github.com/c2corg/v6_ui/pull/621
See also https://github.com/c2corg/v6_ui/pull/621#issuecomment-247314448 and following comments. Especially @tsauerwein 's:
> In build.json you could add the option "formatting": "pretty_print" to get more context.

PR #660, opened 2016-09-28
Make sure constant name is not renamed

tsauerwein:
With the Closure compiler you have to be consistent on how you access object properties. E.g. if you have a property `obj.name = '...'`, you will have to access the property with `obj.name` (and not `obj['name']`). Otherwise the compiler can not properly rename the properties. If you want to access the property with `obj['name']` (as it was the case in this issue), you will have to set the property also with ` obj['name'] = '...'`.

See also: https://developers.google.com/closure/compiler/docs/limitations#implications-of-global-variable-function-and-property-renaming

Closes https://github.com/c2corg/v6_ui/issues/631
Closes https://github.com/c2corg/v6_ui/issues/658

### DIFF
--- a/c2corg_ui/static/js/constants.js
+++ b/c2corg_ui/static/js/constants.js
@@ -10,12 +10,12 @@ goog.provide('app.constants');
 app.module.constant('constants', app.constants);
 
 app.constants = {
-  SCREEN : {
+  SCREEN: {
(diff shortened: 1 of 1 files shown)


asaunier:
:+1: 

Really cool that you have found the fix for https://github.com/c2corg/v6_ui/issues/631 :)

Issue #627, opened 2016-09-17
Translations missing in Transifex


shortoland:
The following strings are now translated in last french version, and also not present in Transifex:

- [x] After disconnecting, there is the following confirmation message: "You have been disconnected"

![image1](https://cloud.githubusercontent.com/assets/20488365/18607271/7f9fa748-7cc8-11e6-9d37-887d5aab6f82.png)

- [ ] maps: + and - icons, up-left corner: tooltips "zoom in" and "zoom out"

- [x] http://c2corgv6.demo-camptocamp.com/routes#bbox=-857237,3946514,1657236,7253485:

Choices for "Configuration" and "Qualité du document" are not translated:

![image2](https://cloud.githubusercontent.com/assets/20488365/18607292/3330ff6e-7cc9-11e6-9fa6-c1cd14b7cf57.png)

- [x] Routes search results (http://c2corgv6.demo-camptocamp.com/routes#bbox=-857237,3946514,1657236,7253485): tooltips rock-rating and labande-rating not translated:

![image3](https://cloud.githubusercontent.com/assets/20488365/18607313/e3b62c92-7cc9-11e6-8b83-2b63e545299a.png)

- [x] Route creation (http://c2corgv6.demo-camptocamp.com/routes/add)
Button "Upload a GPS track".

![image4](https://cloud.githubusercontent.com/assets/20488365/18607357/76700242-7cca-11e6-8c76-1c772b52e471.png)

- [ ] Possibles associations categories not translated:

![image5](https://cloud.githubusercontent.com/assets/20488365/18607366/aa081220-7cca-11e6-99d0-c409f6efafb2.png)

- [x] Summary (http://c2corgv6.demo-camptocamp.com/routes/add):
Qualities of document not translated:

![image6](https://cloud.githubusercontent.com/assets/20488365/18607415/f310c984-7ccb-11e6-99f4-b93656c8f71f.png)

- [x] WP creation, "Références cartographiques" help not translated:

![image7](https://cloud.githubusercontent.com/assets/20488365/18607442/0de81bda-7ccd-11e6-98c4-b2f293cb24b9.png)

- [x] Sorties search: Filters "Evaluation des conditions", "Quality" and  "Fréquentation" not translated:

![image8](https://cloud.githubusercontent.com/assets/20488365/18607486/1913d34a-7cce-11e6-8f5b-677788ad26cf.png)

[...]

PR #664, opened 2016-09-28
I18n fixes

asaunier:
More i18n fixes, related to https://github.com/c2corg/v6_ui/issues/627#issuecomment-250198344

Not sure it is enough though, I have noticed that some dropdowns are not correctly translated in the editing forms despite I have added the missing ``mainCtrl.translate()`` (as for other dropdowns that are now translated as expected!).


tsauerwein:
:+1:

Issue #655, opened 2016-09-27
Incorrect behaviours of advanced search pages


shortoland:
- [ ] slider handles are not fully reset when clearing filters (see https://github.com/c2corg/v6_ui/pull/460#issuecomment-229400429)
- [x] Multi-select boxes sometimes become no longer updatable after interacting with some other filters (other multi-select boxes?). For instance:
  - Click on "Topoguide"
  - Select "escalade" activity
  - Click on "Tous les filtres" (please note that choices for "Quality du document" are not all clickable)
  - Click on "Voir les résultats"
  - Issue: it's impossible to modify "Activité"
- [x] Filtering any document (route, outing, WP) on quality returns wrong results

Alex said: 
- [ ] a problem with double-slider filters in the primary filters section:
  - move an handle of a slider to some position
  - scroll the list until the "sticky" buttons appear at the top of the list
  - click on "see all filters"
  - the slider filter that is then shown is not in the expected status (handle not at the right position)
  - if clicking on "view results", the original slider is shown in the expected status
  - no problem with other filter types (text input or checkbox panel)
  - no problem if clicking on "see all filters" when the primary section is shown
- [ ] sticky filter problem:
  ![screenshot from 2016-10-28 14-43-09](https://cloud.githubusercontent.com/assets/7814311/19806807/67999238-9d1d-11e6-8a81-b9c13c364a2a.png)

- [x] Route search (activity : Neige galce et mixte): the field associated to 'matériel glacier' are not dispalyed correctly (Chrome).
![bug matos glacier](https://cloud.githubusercontent.com/assets/19892089/21420302/3e277a74-c82d-11e6-8074-fe3ee7f96480.jpg)

#1252

shortoland:
@asaunier No worries. I have had a new item on quality's filter.

asaunier:
> Filtering any document (route, outing, WP) on quality returns wrong results

Seems related to the API or ElasticSearch (index problem?).

I think it once worked and that the ``qa`` parameter used to add a quality filter in the API request is correct: https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mapping.py#L54

I am currently rebuilding the ElasticSearch index in my own instance to see if the results are consistent with the quality filter.

tsauerwein:
[...]

PR #666, opened 2016-09-29
Use a slider for quality filter

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/655#issuecomment-250410959


asaunier:
It seems Travis is in troubles and fails because of unrelated problems:
```
The following packages have unmet dependencies:
 postgresql-9.4-postgis-scripts : Depends: postgresql-9.4-postgis-2.3-scripts but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
apt-get.diagnostics
apt-get install failed
```
Can we do something?

asaunier:
Updated to get the Travis fix

PR #669, opened 2016-09-29
add 'view details' btn on image-slide

ginold:


c2corg_ui/static/js/imageuploader.js
```diff
@@ -299,6 +299,7 @@ app.ImageUploaderController.prototype.save = function() {
       scope['photo'] = images[i];
       scope['photo']['image_id'] = id;
       scope['photo']['edit_url'] = '/images/edit/' + imageIds[i]['document_id'] + '/' + images[i]['locales'][0]['lang'];
+      scope['photo']['view_url'] = '/images/' + imageIds[i]['document_id'] + '/' + images[i]['locales'][0]['lang'];
```
asaunier: Do you have access to the image title?
If yes, you could use the ``appUrl`` service and its ``buildDocumentUrl`` function:
https://github.com/c2corg/v6_ui/blob/e5ca771ddf5a95e66352a53c2064af921719cd63/c2corg_ui/static/js/urlservice.js#L39

This way the slug will automaticallly be added (avoiding to make an additional request when the image doc is loaded with no slug).

ginold:
updated!

asaunier:
Please merge.

PR #675, opened 2016-09-30
[travis] Call publish in after_success

arnaud-morvan:
This is an alternative for #649.
As in all other c2corgv6 projects, publish is called in after_success.


arnaud-morvan:
@mfournier : can you take a look.
For now build don't pass because of postgis depedency.
Here I suppose we get tags and branches.

Issue #598, opened 2016-09-06
User Preference Specifications


fjacon:
* Filtering preferences interface only (email, password etc. kept separated in the existing account page)
* Create new account menu entry + new “filters settings” page with activities and regions switches (icons - all ON by default) + region switches:
 * Activities: all ON by default ? or all OFF by default to show there is no activity filter activated yet (would be more consistent with regions)
 * Regions: a map with a vector layer showing the regions, hovering a feature would make the region name appear, clicking a feature would toggle on/off the filtering on this region.
 * Toggling on/off an activity or a region would make a request to the API feed filters service.
 * When loading the page a request to the API feed filters service would retrieve the list of filters of the current user.

Todo:
* [x] Add a route/view/template/menu entry for the filters settings page
* [ ] Create a directive for the activity switches (with read/write access the API service)
* [ ] Update the map directive to enable and show a region vector layer with hovering and click interactions + read/write access to the API service.

Relies on https://github.com/c2corg/v6_api/issues/402

asaunier:
@tsauerwein made me realize that v5 region filters also include countries and administrative areas (FR departements or CH cantons) in addition to ranges. I think ranges are most of the time enough, but yes there are cases where the other region types are useful. For instance if you leave in a place where no range has been defined yet (but then it probably means there are not much activity in this area). Or if you only want to see the outings in Valais.

It's not that easy to display all 3 types of regions at the same time in a map (or maybe by displaying each type in a distinct layer? but then you might not see that a region is selected if the layer is not activated).
On the other hand it's more intuitive (you figure out easily what regions are interesting for you even though you don't know the name on c2c and see the exact limits of the regions).

A problem of a region text selector is that it might take quite a lot of space in the page since there are many regions: 678 according to http://www.camptocamp.org/areas (among them 90 ranges http://www.camptocamp.org/areas/list/atyp/1).
We could use tabs and scrollbared sections but then again you might not see that some region is selected?

If using a map selector, it's probably a good idea to only retrieved simplified geometries of the areas!

Since there is quite a lot of regions, I wonder if this preferences page should not be redis-cached on the UI-side with the list of areas retrieved by a UI-server-side request to the API (one for all). Then the client-side would only make API requests to get/update the user-related list of prefed regions.

tsauerwein:
> It's not that easy to display all 3 types of regions at the same time in a map (or maybe by displaying each type in a distinct layer? but then you might not see that a region is selected if the layer is not activated).

[...]

PR #672, opened 2016-09-29
Add Area documents support to the UI

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/668

PR #676, opened 2016-09-30
User preferences interface

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/598
Requires https://github.com/c2corg/v6_api/pull/421

* [x] checkbox "follow only"
* activities
 * [x] read and show current pref from API
 * [x] update pref and send to API
 * [x] check when the API service is ready (see https://github.com/c2corg/v6_api/issues/402) 
* areas
 * [x] set a simple search tool filtered on areas (related to https://github.com/c2corg/v6_api/issues/419)
 * [x] show area cards (requires https://github.com/c2corg/v6_ui/pull/672)

### DIFF
--- a/c2corg_ui/externs/appx.js
+++ b/c2corg_ui/externs/appx.js
@@ -275,6 +275,18 @@ appx.Image;
 
 /**
  * @typedef {{
+ *   locales: Array.<Object>,
+ *   type : string,
+ *   area_type: string,
+ *   quality: string,
+ *   document_id:number
+ * }}
+ */
+appx.Area;
+
+
+/**
+ * @typedef {{
  *   total: number,
  *   documents: Array.<appx.Outing>
  * }}
@@ -305,3 +317,12 @@ appx.mapApiKeys;
  * }}
  */
 appx.DiscourseEmbedded;
+
+/**
+ * @typedef {{
+ *   'followed_only': boolean,
+ *   'activities': Array.<string>,
+ *   'areas': Array.<appx.Area>
+ * }}
+ */
+appx.UserPreferences;
--- a/c2corg_ui/static/js/preferences.js
+++ b/c2corg_ui/static/js/preferences.js
@@ -1,0 +1,154 @@
+goog.provide('app.PreferencesController');
+goog.provide('app.preferencesDirective');
+
+goog.require('app');
+goog.require('app.Alerts');
+goog.require('app.Api');
+goog.require('app.Authentication');
+goog.require('app.utils');
+
+
+/**
+ * Directive managing the user preferences.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.preferencesDirective = function() {
+  return {
+    restrict: 'A',
+    controller: 'appPreferencesController',
+    controllerAs: 'prefCtrl'
+  };
+};
+
+
+app.module.directive('appPreferences', app.preferencesDirective);
+
+/**
+ * @param {angular.Scope} $scope Scope.
+ * @param {app.Authentication} appAuthentication
+ * @param {app.Alerts} appAlerts
+ * @param {app.Api} appApi Api service.
+ * @param {string} authUrl Base URL of the authentication page.
+ * @constructor
+ * @ngInject
+ * @export
+ */
+app.PreferencesController = function($scope, appAuthentication, appAlerts,
+    appApi, authUrl) {
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
+
+  /**
+   * @type {app.Alerts}
+   * @private
+   */
+  this.alerts_ = appAlerts;
+
+  /**
+   * @type {app.Api}
+   * @private
+   */
+  this.api_ = appApi;
+
+  /**
+   * @type {Array.<string>}
+   * @export
+   */
+  this.activities = [];
+
+  /**
+   * @type {Array.<appx.Area>}
+   * @export
+   */
+  this.areas = [];
+
+  /**
+   * @type {boolean}
+   * @export
+   */
+  this.followed_only = false;
+
+  if (appAuthentication.isAuthenticated()) {
+    this.api_.readPreferences().then(function(response) {
+      var data = /** @type {appx.UserPreferences} */ (response['data']);
+      this.activities = data.activities;
+      this.areas = data.areas;
+      this.followed_only = data.followed_only;
+
+      this.scope_.$watch(function() {
+        return this.followed_only;
+      }.bind(this), function(newValue, oldValue) {
+        if (newValue !== oldValue) {
+          this.save_();
+        }
+      }.bind(this));
+    }.bind(this));
+  } else {
+    app.utils.redirectToLogin(authUrl);
+  }
+};
+
+
+/**
+ * @param {string} activity
+ * @export
+ */
+app.PreferencesController.prototype.updateActivities = function(activity) {
+  if (this.activities.indexOf(activity) > -1) {
+    this.activities = this.activities.filter(function(item) {
+      return item !== activity;
+    });
+  } else {
+    this.activities.push(activity);
+  }
+  this.save_();
+};
+
+
+/**
+ * @param {appx.Area} area
+ * @export
+ */
+app.PreferencesController.prototype.addArea = function(area) {
+  var alreadyInList = this.areas.some(function(a) {
+    return a.document_id === area.document_id;
+  });
+  if (alreadyInList) {
+    return;
+  }
+  this.areas.push(area);
+  this.save_();
+};
+
+
+/**
+ * @param {number} id Area document_id.
+ * @export
+ */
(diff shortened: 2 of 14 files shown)


asaunier:
I'm using the API branch ``feed-preferences`` in my API instance and use it as a backend for this user pref interface. I have a weird 403 error when loading the user prefs from the API:
> Getting preferences failed:
debug_authorization of url http://c2corgv6-demo.gis.internal:6545/users/feed-preferences (view name '' against context <c2corg_api.RootFactory object at 0x7f350706ca20>): ACLDenied permission 'authenticated' via ACE '<default deny>' in ACL [('Allow', 'system.Everyone', 'public'), ('Allow', 'system.Authenticated', 'authenticated'), ('Allow', 'group:moderators', 'moderator')] on context <c2corg_api.RootFactory object at 0x7f350706ca20> for principals ['system.Everyone'] : Forbidden

The http://c2corgv6-demo.gis.internal:6545/users/feed-preferences request indeed provide no ``Authorization`` header (with a ``JWT token``) and no ``OPTIONS`` is made previously, on the contrary of the ``users/account`` request (usernames, password, email, etc.).

I don't understand this difference since both tools use almost the same ``ApiService`` methods:
https://github.com/c2corg/v6_ui/blob/1fa5a994f35c37eba33c66fb51edf4fc7c09ae34/c2corg_ui/static/js/apiservice.js#L381-L432

Is there something special to do in order to pass the JWT token along the GET request?
By the way the POST "feed-preferences" request works fine (passes the JWT token as expected).

tsauerwein:
> Is there something special to do in order to pass the JWT token along the GET request?

This is a more a UI issue. You have to adapt this function:
https://github.com/c2corg/v6_ui/blob/671f86f74f20c94577c77a8ebac7d5e988ac9093/c2corg_ui/static/js/auth/authservice.js#L332

asaunier:
This PR is ready for reviewing. It requires c2corg/v6_api#421 as API backend.

Here is an example overview:
sans titre

I have chosen not to display any pref change confirmation as I initially intended because the "success" alerts were breaking the flow of the changes (you had to close them before making further changes) and I found it not user-friendly. Only errors are "alerted".

tsauerwein:
FYI https://github.com/c2corg/v6_api/pull/421 is merged.

c2corg_ui/static/js/preferences.js
```diff
@@ -0,0 +1,157 @@
...
+};
+
+
+/**
+ * @param {appx.Area} area
+ * @export
+ */
+app.PreferencesController.prototype.addArea = function(area) {
+  var id = area.document_id;
+  for (var i = 0, n = this.areas.length; i < n; i++) {
```
tsauerwein: You could use [Array.some](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/some). It is supported by all modern browsers: http://kangax.github.io/compat-table/es5/#test-Array_methods_Array.prototype.some

c2corg_ui/templates/preferences.html
```diff
@@ -0,0 +1,46 @@
...
+<div class="preferences">
+  <section app-preferences class="section">
+    <h1 translate>Filter preferences</h1>
+    <p>
+      <label>
+        <input type="checkbox" ng-model="prefCtrl.followed_only">
+        <span translate>In the homepage feed see only statuses from people you follow</span>
+      </label>
+    </p>
+    <p translate>Here you may set activity and region preferences, used for instance in the activity feed.</p>
```
tsauerwein: Why "...for instance in the activity feed"? Is it used else where?

c2corg_ui/templates/preferences.html
```diff
@@ -0,0 +1,46 @@
...
+<%block name="pagetitle"><title ng-bind="mainCtrl.page_title('Filter preferences')">${show_title('Filter preferences')}</title></%block>
+<%block name="metarobots"><meta name="robots" content="noindex,nofollow"></%block>
+
+<div class="preferences">
+  <section app-preferences class="section">
+    <h1 translate>Filter preferences</h1>
+    <p>
+      <label>
+        <input type="checkbox" ng-model="prefCtrl.followed_only">
+        <span translate>In the homepage feed see only statuses from people you follow</span>
```
tsauerwein: Maybe "Show only updates from followed users in the homepage feed"?

c2corg_ui/templates/preferences.html
```diff
@@ -0,0 +1,46 @@
...
+      </label>
+    </p>
+    <p translate>Here you may set activity and region preferences, used for instance in the activity feed.</p>
+    <div ng-hide="prefCtrl.followed_only">
+      <h2 translate>activities</h2>
```
tsauerwein: Capital letter

tsauerwein: I would also add a "explanation/help section" saying something like "Only status updates with the selected activities and in the selected areas are shown in your homepage feed. Status updates from followed users will always be shown."

c2corg_ui/externs/appx.js
```diff
@@ -275,6 +275,18 @@ appx.Image;
 
 /**
  * @typedef {{
+ *   locales: Array.<Object>,
```
ginold: don't we already have the type `appx.DocumentLocale;` ?

asaunier: Perhaps. I think I have copied-pasted ``appx.Image``:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js#L268
I will make the change in  ``appx.Image`` as well.

c2corg_ui/templates/preferences.html
tsauerwein: I don't really like the idea of making the user interface more complicated/less clear only for a technical reason.

If later in the future we add a new feature where these preferences are used elsewhere, we can also update this translation.

c2corg_ui/static/js/preferences.js
```diff
@@ -0,0 +1,157 @@
...
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.preferencesDirective = function() {
+  return {
+    restrict: 'A',
+    scope: true,
+    controller: 'appPreferencesController',
+    controllerAs: 'prefCtrl',
+    bindToController: true
```
ginold: if you write `scope: true` and `bindToController: true`, what actually is bound to the controller when you don't set any value for example with `{someObject: '='}` ? is it of any use in this case?

c2corg_ui/templates/preferences.html
asaunier: ``activities`` already exists in the translation strings (in our code and in Transifex). ``Activities`` doesn't.
I admit ``Activities`` would be better but it it would create a duplicate in translations! And it's not that easy to change the translation keys without losting translations in Transifex (OK, not that a big deal since this is a one-word translation).
We have to be VERY careful when creating translations!

c2corg_ui/static/js/preferences.js
asaunier: No idea. This is a standard setting in our directives. I will try to omit them.

c2corg_ui/static/js/preferences.js
asaunier: Thanks for the tip! But I am not really convinced there's a gain using it because it's as long and a little less readable (because "some" is not very well known to most developers?).

* A:
```
  for (var i = 0, n = this.areas.length; i < n; i++) {
    if (this.areas[i].document_id === id) {
      // selected area is already in the list
      return;
    }
  }
```
* B:
```
  if (this.areas.some(function(a) {
    return a.document_id === id;
  })) {
    // selected area is already in the list
    return;
  }
```
I think I prefer A.

c2corg_ui/externs/appx.js
asaunier: Actually all document typedef use ``locales: Array.<Object>`` (not only Image or Area).
And the definition of ``appx.DocumentLocale`` seems a bit too restrictive to be used instead.
```
/**
 * @typedef {{
 *   title: string,
 *   title_prefix: ?string,
 *   lang: string
 * }}
 */
appx.DocumentLocale;
```
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js#L60-L67

asaunier:
PR updated according to your suggested changes @tsauerwein @ginold (except the externs).
sans titre

I will merge when Travis is OK and if you don't object in the meantime.

c2corg_ui/templates/preferences.html
ginold: I'm always asking myself what's the best way of doing - translating strings as capital or low string words? in both cases we will use sometimes a Capitalized word and sometimes lowercased words. We should have an angular filter that caps/lowers the letters when needed.

asaunier: Most of the time, strings should be capitalized. Except for attribute names (we have no control on them since coming from the database) and words used inside a more complicated sentence (but there are not many such cases).

PR #680, opened 2016-10-03
Column fix + pagination style

ginold:
due to some glitches, we had to go back to colum-based CSS instead of flexbox.

in addition, i suggest a new style for pagination buttons and layout for mobile - they were poorly positioned. 

+ some map elements improvements

screenshot from 2016-10-03 10-50-25

### DIFF
--- a/c2corg_ui/static/partials/map/map.html
+++ b/c2corg_ui/static/partials/map/map.html
@@ -8,7 +8,11 @@
     class="btn btn-primary map-reset-feature" translate>Reset feature</button>
   <div class="map-filter-switch" ng-show="::mapCtrl.advancedSearch">
     <label>
-      <input type="checkbox" ng-model="mapCtrl.enableMapFilter"><span translate>Filter on map extent</span>
+      <input type="checkbox" ng-model="mapCtrl.enableMapFilter">
+      <div>
+        <span translate class="text">Filter on map extent</span>
+        <span class="glyphicon glyphicon-globe"></span>
+      </div>
     </label>
   </div>
 </div>
--- a/c2corg_ui/templates/outing/index.html
+++ b/c2corg_ui/templates/outing/index.html
@@ -27,7 +27,8 @@
   <div class="documents-nav-buttons text-center">
     <button class="btn btn-default show-documents-filters-phone" translate>Filters</button>
     ${add_doctype_selector()}
-    <button class="btn btn-default" protected-url-btn url="${request.route_path('outings_add')}">
+    <button class="btn btn-default orange-btn" protected-url-btn url="${request.route_path('outings_add')}"
+            uib-tooltip="{{'Add a new outing' | translate }}" tooltip-placement="right">
       <span class="glyphicon glyphicon-plus-sign"></span>
     </button>
     <app-pagination></app-pagination>
--- a/c2corg_ui/static/partials/pagination.html
+++ b/c2corg_ui/static/partials/pagination.html
@@ -1,15 +1,14 @@
 <div ng-if="pageCtrl.total > pageCtrl.limit" id="pagination">
-  <span>{{pageCtrl.offset + 1}} -
-     {{(pageCtrl.offset + pageCtrl.limit < pageCtrl.total) ? pageCtrl.offset + pageCtrl.limit : pageCtrl.total}} /
-     {{pageCtrl.total}}</span>
+  <span class="page-numbers">{{pageCtrl.offset + 1}} -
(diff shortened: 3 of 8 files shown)


c2corg_ui/static/js/imageuploader.js
```diff
@@ -306,7 +306,7 @@ app.ImageUploaderController.prototype.save = function() {
       scope['photo'] = images[i];
       scope['photo']['image_id'] = id;
       scope['photo']['edit_url'] = '/images/edit/' + imageIds[i]['document_id'] + '/' + images[i]['locales'][0]['lang'];
-      scope['photo']['view_url'] = this.url_.buildDocumentUrl('images', imageIds[i]['document_id'], images[i]['locales'][0], images[i]['locales'][0]['lang'] );
+      scope['photo']['view_url'] = this.url_.buildDocumentUrl('images', imageIds[i]['document_id'], images[i]['locales'][0], images[i]['locales'][0]['lang']);
```
asaunier: This change has already been done by @tsauerwein in a previous PR. See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/imageuploader.js#L309

It's weird that github doesn't report a conflict!? Have you rebased your branch?

c2corg_ui/static/partials/map/map.html
```diff
@@ -8,7 +8,11 @@
     class="btn btn-primary map-reset-feature" translate>Reset feature</button>
   <div class="map-filter-switch" ng-show="::mapCtrl.advancedSearch">
     <label>
-      <input type="checkbox" ng-model="mapCtrl.enableMapFilter"><span translate>Filter on map extent</span>
+      <input type="checkbox" ng-model="mapCtrl.enableMapFilter">
+      <div>
+        <span translate class="text">Filter on map extent</span>
+        <span class="glyphicon glyphicon-globe"></span>
```
asaunier: Are you sure this really works? I am testing this branch in my instance: no "glyphicon-globe" appears!? By the way I am not really convinced that we should display such an icon, since we don't have that much space when the map is small.

c2corg_ui/templates/outing/index.html
```diff
@@ -27,7 +27,7 @@
   <div class="documents-nav-buttons text-center">
     <button class="btn btn-default show-documents-filters-phone" translate>Filters</button>
     ${add_doctype_selector()}
-    <button class="btn btn-default" protected-url-btn url="${request.route_path('outings_add')}">
+    <button class="btn btn-default orange-btn" protected-url-btn url="${request.route_path('outings_add')}">
```
asaunier: What about adding a tooltip on such buttons as well? Or even a real title?
I'm afraid many people won't understand what this button is for.

c2corg_ui/static/partials/map/map.html
asaunier: I get it: the globe is shown only on mobile screen, instead of the text. That looks like a good idea indeed, even though showing only the icon is not very explicit.

asaunier:
Your branch seems not up-to-date. Please rebase, it helps testing for real and checking that recent changes are compatible with yours.

c2corg_ui/static/js/imageuploader.js
ginold: just rebased - no conflicts.

c2corg_ui/templates/outing/index.html
ginold: tooltip ok, but it won't work well on smartphone and adding text takes space, and we are already short on space. I don't really know what would be the solution :P

c2corg_ui/static/js/imageuploader.js
asaunier: The change no longer appears in https://github.com/c2corg/v6_ui/pull/680/files

c2corg_ui/templates/outing/index.html
asaunier: Could you at least add a tooltip?

c2corg_ui/static/js/imageuploader.js
ginold: :confused:  maybe it was ignored

asaunier:
The changes look good to me. Would be cool to add the tooltip for "add" buttons as well.
Please merge when Travis is OK.

c2corg_ui/static/js/imageuploader.js
asaunier: That's just that this change was already done in master and thus there is no longer a difference between what you did and master.

PR #681, opened 2016-10-04
use BI in view pages & add link to full image

### LINKED ISSUE #670: Incorrect image size displayed in image detail view pages
It seems that the original file is displayed in the image detail view pages: that's a problem, especially if the original image is quite big (>1 MB)
* [x] the "BI" image should be displayed instead
* [x] the original image is displayed (direct link - not in the current page) when clicking on the BI image.

As in v5: http://www.camptocamp.org/images/807159/fr/cabane-de-graetli-chocolat-cali-sieste-et-le-papa-qui-s-impatiente-sachant-la-journee-est-encore-lon

ginold:
fixes #670

### DIFF
--- a/c2corg_ui/templates/image/helpers/detailed_images_attributes.html
+++ b/c2corg_ui/templates/image/helpers/detailed_images_attributes.html
@@ -1,6 +1,17 @@
 <%namespace file="../../helpers/view.html" import="show_areas, show_attr, get_document_up_down, get_document_min_max"/>
 <%! import math %>
 
+## CREATE IMAGE LINK
+<%def name="create_image_link(url, suffix, title)">\
+  <%
+    i = url.rfind('.')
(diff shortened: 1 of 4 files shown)


c2corg_ui/templates/image/helpers/detailed_images_attributes.html
```diff
@@ -1,6 +1,17 @@
 <%namespace file="../../helpers/view.html" import="show_areas, show_attr, get_document_up_down, get_document_min_max"/>
 <%! import math %>
 
+## CREATE IMAGE URL
+<%def name="create_image_url(url, suffix, title)">\
```
arnaud-morvan: this do not return an url, could be renamed ``image_with_link`` for exemple.

asaunier:
:+1:

PR #684, opened 2016-10-05
avoid element jump on load

### LINKED ISSUE #551: Avoid that page elements "jump" when loading

### LINKED ISSUE #612: Incorrect recentering in detail view maps
There's something weird with the maps embedded in the detail view pages.
See for instance: http://c2corgv6-demo.gis.internal:6555/routes/708857/fr/mont-brion-tour-du-mont-brion
=> map is no longer recentered on the vector features but a bit to the west.

When zooming in/out using the +/- buttons of the map, the features remain stuck in the bottom right corner!?

This must be quite recent because I have only noticed it this morning.

ginold:
fixes #551 
fixes #612

### DIFF
--- a/less/viewdetails.less
+++ b/less/viewdetails.less
@@ -149,6 +149,7 @@
     width: 90% !important;
     margin-right: 5%;
     padding: 10px 0 10px 0;
+    height: 170px;
     @media @phone {
       width: 88% !important;
       margin-right: 6%;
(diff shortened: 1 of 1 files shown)


less/viewdetails.less
```diff
@@ -524,22 +525,21 @@
...
-    @media @phone {
-      height: 170px !important;
-    }
-    @media @phone {
-      margin-bottom: 0;
-    }
+    height: 100%;
   }
   /*is 100% by defaul - it stretches the map!*/
   .ol-unselectable {
```
tsauerwein: If you could get rid of this style on ` .ol-unselectable`, https://github.com/c2corg/v6_ui/issues/612 could be fixed as well.

asaunier:
Looks good to me for #551 
Concerning #612 I cannot test because I never had this problem with my workstation, only with my laptop at home. Feel free to merge, I will test again at home then,

tsauerwein:
> Concerning #612 I cannot test because I never had this problem with my workstation, only with my laptop at home.

You can simulate this with Chrome: https://developers.google.com/web/tools/chrome-devtools/device-mode/emulate-mobile-viewports#device_pixel_ratio_dpr

Issue #131, opened 2016-02-02
Parsing C2C custom tags in documents text attributes


asaunier:
In v5 a log of formatting is possible and based upon several syntaxes, mainly BBcode and Markdown.
For the latter syntaxes a PR is available at #127 

Many custom tags are also available:
- new tags such as wikilinks `[[summits/123456|link title]]`, `[toc 2 right]` or `L#` (climbing routes pitches)
- existing tags but with special attributes, not supported out of the box by the parsers used in #127. For instance `[img=188119 right]Lever de soleil sur le Mont Blanc[/img]`

See the documentations about those syntaxe:
- http://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte
- http://www.camptocamp.org/articles/305462/fr/aide-topoguide-l-balise-de-description-des-longueurs

asaunier:
wikilinks => see #156

asaunier:
### Images

First thing is to figure out weither to an extension for the bbcode parser or for the markdown parser.

Extending the bbcode parser looks fairly easy, see the doc/example at http://bbcode.readthedocs.org/en/latest/formatters.html

On the other hand the markdown parser seems a bit more flexible and powerful (and in the future we might want to get rid of the bbocode!?) => then it's probably smarter to only write markdown extensions?

There's a quite annoying issue with images: the ``img`` tags contains the document_id, not the filename itself. For instance ``[img=188119 right]Lever de soleil[/img]``.
=> we will have to make a query to the database (ie. to the API) to get the actual filename.

* Maybe we could convert the bbcode tag to an angular directive that would request the API to get the filename to use?
* Maybe we could change the filename to use the document_id instead? A problem is that if a new file has been uploaded to a given image document, the document_id is the same but the filename is obviously different...

stef74:
Liste of non supported syntax 
- [ ] L#
- [ ] [Image]
- [ ] [toc]
- [ ] [q]
- [ ] [c]
- [ ] [important],[warning]
- [ ] column / table

fjacon:
Highest priority : images

asaunier:
The tricky thing with images is that the tag contains the image id but not the real filename (eg. ``[img=188119 right]Lever de soleil[/img]``).

Perhaps a solution could be to convert the tag to some angular directive that makes an AJAX request to the API to get the image filename based upon the document id => embedded images would then be loaded asynchronously.

tsauerwein:
[...]

PR #695, opened 2016-10-05
Enable markdown extensions for TOC and linebreaks

asaunier:
This PR activates 2 Python-Markdown extensions to support:
* line breaks, see https://pythonhosted.org/Markdown/extensions/nl2br.html
* table of contents, see https://pythonhosted.org/Markdown/extensions/toc.html

## Line breaks

handled pretty fine.

## Tables of contents

are supported quite well except the following limitations:

* the default TOC marker is ``[TOC]``. I have changed it to ``[toc]`` but it only works with ``[toc]`` in the text to convert,  variants such as ``[toc 2 right]`` are not supported (remains unreplaced). I have tried with ``[toc(.*)]`` but with no more success. Of course using ``[toc 2 right]`` as the marker is possible but that's obviously not relevant. Perhaps we could convert all ``[toc(.*)]`` variants to ``[toc]`` using the migration script?
* the extension has a ``baselevel`` option that can be used to specify the "biggest" level of header that is taken into account in the TOC. For instance if ``baselevel=3``, only ``<h3>``,  ``<h4>``, etc. will be taken into account, not ``<h1>`` or  ``<h2>``. On the other hand it's not possible to limit the number of sublevels (in v5 ``[toc 2]`` means that only "h2" and "h3" are referenced, not "h4" or "h5". Maybe not a big deal.
* with the current setting, the TOC is actually created and ids are added to the section headers:
```
<div class="toc">
<ul>
<li><a href="#presentation">Présentation</a><ul>
<li><a href="#historique">Historique</a></li>
...
```
```
<h3 id="presentation">Présentation</h3>
<h4 id="historique">Historique</h4>
```
The weird thing is that when I click on the TOC links, the URL in the address bar becomes ``waypoints/37355/fr/mont-blanc#/presentation`` instead of ``waypoints/37355/fr/mont-blanc#presentation`` (note the ``#/``) => the first URL does not redirect to the section whereas the second one does. Same problem in Chrome or FF!

Here is what the generated TOC looks like for http://c2corgv6.demo-camptocamp.com/waypoints/37355/fr/mont-blanc

capture d ecran 2016-10-05 a 23 18 22

## Tables

I have also tried to enable the "tables" extension https://pythonhosted.org/Markdown/extensions/tables.html

hoping it would help manage the v5 ``L#`` tag (see https://github.com/c2corg/v6_ui/issues/131#issuecomment-237195445)

but this was quite a disaster because:

* there are conflicts with some of the C2C-custom wikilinks (``[[routes/12345|foo bar]]``), perhaps because of the ``|`` sign (used also as a cell separator for tables combined with some combinations of blank spaces and ``-``). See for instance:
capture d ecran 2016-10-05 a 23 00 55
* the syntax of this parser is not really flexible and intuitive, it's a bit painful to figure out what syntaxes work and what don't.
* anyway it seems it has no effect on ``L#`` cells. And if by chance it does, the first line is emphasized (bold) as the table headers.

So for me this extension does not help at all.

### DIFF
--- a/c2corg_ui/templates/utils/format.py
+++ b/c2corg_ui/templates/utils/format.py
@@ -3,6 +3,8 @@ import markdown
 import html
 
 from c2corg_ui.format.wikilinks import C2CWikiLinkExtension
+from markdown.extensions.nl2br import Nl2BrExtension
+from markdown.extensions.toc import TocExtension
 
 
 _markdown_parser = None
@@ -14,6 +16,8 @@ def _get_markdown_parser():
     if not _markdown_parser:
         extensions = [
             C2CWikiLinkExtension(),
+            Nl2BrExtension(),
+            TocExtension(marker='[toc]', baselevel=2),
         ]
         _markdown_parser = markdown.Markdown(output_format='xhtml5',
                                              extensions=extensions)


asaunier:
Anyone for this easy review? :)

tsauerwein:
I was overwhelmed by your long description and didn't notice that your changes are rather small.. ;)

But feel free to merge. 

> Perhaps we could convert all [toc(.*)] variants to [toc] using the migration script?

We could, yes.

> ... hoping it would help manage the v5 L# tag (see #131 (comment))

Didn't we want to simplify the `L#` syntax anyway?

> The weird thing is that when I click on the TOC links, the URL in the address bar becomes

This could be related to `ngeoLocation`. To check.

asaunier:
Thanks, merged!

>> Perhaps we could convert all [toc(.*)] variants to [toc] using the migration script?
> We could, yes.

OK, see https://github.com/c2corg/v6_ui/issues/702

> Didn't we want to simplify the L# syntax anyway?

Yes, but at least a basic support will be needed (simple incrementation of the pitch numbers L1 > L2 > ..., so that the most common case is handled + table formating)

> This could be related to ngeoLocation.

Thanks for the tip!

asaunier:
Parsing of images and L# has been moved to:
* img => https://github.com/c2corg/v6_ui/issues/712
* L# => https://github.com/c2corg/v6_ui/issues/713

fbunoz:
>>  Perhaps we could convert all [toc(.*)] variants to [toc] using the migration script?

> We could, yes.

No !
Don't change any unsupported tags using migration script !
Adding the tags in the docs needed thousands hours of works from contributors and moderators for ten years, We don't want to lose this work in few hours of migrating !
It's better to show the unsupported tags (or to mask it). When these tags will be supported, all will be ok.

fbunoz:
> The weird thing is that when I click on the TOC links, the URL in the address bar becomes ``waypoints/37355/fr/mont-blanc#/presentation`` instead of ``waypoints/37355/fr/mont-blanc#presentation`` (note the ``#/``) => the first URL does not redirect to the section whereas the second one does. Same problem in Chrome or FF!

The url is OK for me (FF) on the preview. I can't try when reading the doc because the saving fails.

asaunier:
> Don't change any unsupported tags using migration script !

No problem. Actually it saves us some extra work.

PR #699, opened 2016-10-06
[WIP] Image transition

ginold:
When we recreate the small images while keeping their original aspect ratio, I will be able to get a transition from thumbnail to the big image. The plugin we're using for this requires the height and width of the big image that has not yet been loaded when opening the gallery. It will be enough to multiply the small image dimensions by the coefficient it has been scaled down to, like : 

```
        // to change : multiplying coefficient
        var dimensions = {
          'height': $(linkEl.children[0]).prop('naturalHeight') * 6,
          'width': $(linkEl.children[0]).prop('naturalWidth') * 6
        };
```

### DIFF
--- a/c2corg_ui/static/js/viewdetails.js
+++ b/c2corg_ui/static/js/viewdetails.js
@@ -154,34 +154,27 @@ app.ViewDetailsController.prototype.initPhotoswipe_ = function() {
     var parseThumbnailElements = function(el) {
       var thumbElements = el.childNodes;
       var items = [];
-      var figureEl;
-      var linkEl;
-      var item;
-      var id;
-      var info;
 
       for (var i = 0; i < thumbElements.length; i++) {
-        figureEl = thumbElements[i]; // <figure> element
-        linkEl = figureEl.children[0]; // <a> element
+        var figureEl = thumbElements[i]; // <figure> element small image
+        var linkEl = figureEl.children[0]; // <a> element big image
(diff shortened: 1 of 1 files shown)


arnaud-morvan:
SI will be 150*150.
Does this suppose that the BI will be 900*900 ? Note that the BI will not be a square.

ginold:
you mean they always will be squared? won't we resize them down proportionally?

arnaud-morvan:
I've a pending pull request where I fixed the SI resizing using cropping.
I can keep the aspect ratio for SI images, but no one ask me to do so.
See : https://github.com/c2corg/v6_images/issues/5#issuecomment-251681943

ginold:
as Alex stated: 

> Discussed a few minutes ago with @ginold
> 
> thumbnails (SI images) don't have to square. They currently are because of the old v5 thumbnails that do are square.
> when regenerating the thumbnails, it would be an option to keep the original w/h ratio.

but sure, maybe it was too late. If you still can resize the SI keeping the proportions - that would be great.

asaunier:
@arnaud-morvan @ginold 

> If you still can resize the SI keeping the proportions - that would be great.

For me both formats (square or unchanged-ratio) have pros and cons, and are OK.
So as you want if both work as expected in the UI.
@stef74 What do you think?

asaunier:
Anyone knows if we should still do something with this PR? Is it still relevant?

Issue #594, opened 2016-09-05
rename non-route-specific functions


ginold:
some functions that seem to be route-specific are in fact used also for other documents, so their names should be more generic, for example `get_route_activities` shoud become `get_document_activities` because it's used within outings, routes and images.

asaunier:
There's also ``show_route_title`` which seems used for all document types: 
https://github.com/c2corg/v6_ui/search?utf8=%E2%9C%93&q=show_route_title
https://github.com/c2corg/v6_ui/blob/d5b38e1ab35851437826cf39c2887a52275f3c4d/c2corg_ui/templates/helpers/view.html#L61

fjacon:
Should we do this rapidly ?

asaunier:
That's not critical but the more we wait, the more those tools will be used (when adding new document types such as articles, books, areas, etc.) => more changes to do later.

PR #703, opened 2016-10-06
Adds articles to UI

dkocich:


c2corg_ui/static/partials/cards/articles.html
```diff
@@ -0,0 +1,20 @@
+<a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <div class="article-author text-center">
```
ginold: in this case, maybe not the most correct and consistent way of doing, but you would have to keep the classes with `outing-` because they are defined and styled in the CSS whereas `article-` are not and will not display correctly.

c2corg_ui/templates/article/edit.html
```diff
@@ -0,0 +1,601 @@
...
+
+<section class="create-edit-document">
+
+  <h1 ng-cloak ng-init="id=${article_id if article_id else 0}"
+      ng-bind="id ? ('Editing an article' | translate) : ('Creating an article' | translate)"
+      class="text-center"></h1>
+
+  <form app-loading app-document-editing="articles" app-document-editing-model="article"
+    app-document-editing-controller-name="apparticleEditingController"
+  % if updating_doc:
```
ginold: I would indent the `%if `block with the other lines

c2corg_ui/templates/article/view.html
```diff
@@ -0,0 +1,127 @@
...
+  % else:
+    <meta name="robots" content="index,follow">
+  % endif
+</%block>
+
+<%block name="moduleConstantsValues">
+  module.value('documentData', {
+    "document_id": ${article.get('document_id')},
+    "lang": "${lang}",
+    "type": "o",
```
ginold: type a

c2corg_ui/templates/base.html
```diff
@@ -55,7 +55,7 @@
     <button class="btn btn-default menu-open-close header"><span class="glyphicon glyphicon-align-justify"></span></button>
     <div class="page-header">
       <div class="quick-search">
-        <app-simple-search dataset="wr"></app-simple-search>
+        <app-simple-search dataset="wrc"></app-simple-search>
```
ginold: wr c ? what is c? 
Usually the letters are the first letter of a document type - Waypoint, Route, Outing, C...?

asaunier: "c" is "articles". See https://github.com/c2corg/v6_common/blob/master/c2corg_common/document_types.py#L9

c2corg_ui/templates/article/view.html
asaunier: rather "c"

c2corg_ui/templates/article/edit.html
asaunier: I wouldn't. Since we have not done it for other document types editing forms.
In addition it's related to an attribute within an HTML element, it's not 2 elements embedded one in another.

dkocich:
Please review and comment someone ( @asaunier @ginold ?). Many thanks.

c2corg_ui/static/js/maincontroller.js
```diff
@@ -78,7 +78,7 @@ app.MainController.prototype.isPath = function(path) {
     return 'home';
   // if topoguide, it can be all kinds of documents
   } else if (path === 'topoguide') {
-    return location.indexOf('waypoints') > -1 || location.indexOf('routes') > -1 || location.indexOf('outings') > -1;
+    return location.indexOf('waypoints') > -1 || location.indexOf('routes') > -1 || location.indexOf('outings') > -1 || location.indexOf('articles') > -1;
```
asaunier: I realize I have not added ``areas`` here as well.
@ginold is it needed?

c2corg_ui/static/js/utils.js
```diff
@@ -196,7 +198,7 @@ app.utils.getTemplate = function(path, $templateCache) {
  */
 app.utils.detectDocumentIdFilter = function(ngeoLocation) {
   // see app.utils.getDoctype() for types definitions
-  var associatedDocTypes = ['w', 'r', 'a'];
+  var associatedDocTypes = ['w', 'r', 'a', 'c'];
```
asaunier: This change is relevant only if we need to pass some article id to an advanced search page (eg. ``waypoints#c=12345``).
I cannot think of a case where we would need this!?

c2corg_ui/static/partials/cards/articles.html
```diff
@@ -0,0 +1,20 @@
+<a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <div class="outing-author text-center">
```
asaunier: Using an "outing" class for an article is weird :P

c2corg_ui/static/partials/cards/articles.html
```diff
@@ -0,0 +1,20 @@
...
+    <div class="outing-date">{{ doc['date_end'] | date: 'dd/MM/yyyy'}}</div>
+    <span class="icon-user"></span><br>
+    <b class="author-name">{{doc['author']['name']}}</b>
```
asaunier: I don't think those attributes are useful for article cards: date, author (especially for collaborative articles, that may have several authors).

The attributes you have added below seem OK to me (title, lang, quality, activities). Perhaps "article_type"?
Is there a document or an issue where the client would have listed the attributes for article cards?

c2corg_ui/templates/article/edit.html
```diff
@@ -0,0 +1,143 @@
...
+                </div>
+              </div>
+
+            </section>
+
+            <div class="form-group half half-group">
+
+              <div ng-class="{ 'has-success': article.article_type}">
+                <label><span translate>article_type</span> *</label>
+                <select class="form-control" ng-model="article.article_type" ng-options="rat as mainCtrl.translate(rat) for rat in ${article_types}"><option value=""></option></select>
```
asaunier: OK when creating a new article.
But when editing a collaborative article, only moderators should be allowed to change the ``article_type`` to ``personal``.
On the other hand, no problem to change a ``personal`` article to ``collaborative``.

c2corg_ui/templates/article/edit.html
```diff
@@ -0,0 +1,143 @@
...
+                <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="article.article_type"></span>
+              </div>
+
+              ## ARTICLE CATEGORIES
+              <div ng-class="{'has-success': article.article_categories}" ng-init="articleCategories = ${article_categories}">
+                <div class="ng-hide">{{article.articleC = article.articleCategories || articleC[1]}}</div>
```
asaunier: Why do you need this?

c2corg_ui/templates/article/edit.html
```diff
@@ -0,0 +1,143 @@
...
+                <select ng-options="type as mainCtrl.translate(type) for type in articleCategories" ng-model="article.article_categories" class="form-control"><option value=""></option></select>
+                <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="article.article_categories"></span>
+              </div>
+
+
+            </div>
+          </div>
+        </section>
+
+        ## MAP
```
asaunier: Articles have no geometry! => no need for a map

c2corg_ui/templates/article/edit.html
```diff
@@ -0,0 +1,143 @@
...
+
+            ## SUMMARY
+            <div id="summary-group" class="form-group full">
+              <label translate>summary</label>
+              <textarea name="summary" ng-model="article.locales[0].summary" class="form-control"></textarea>
+            </div>
+
+            ## DESCRIPTION
+            <div id="description-group" class="form-group full">
+              <label translate>description</label>
```
asaunier: ``description`` is probably not a relevant field name for an article.
In v5 it's ``article core`` (really better??) See http://www.camptocamp.org/articles/edit

c2corg_ui/templates/article/filters.html
```diff
@@ -0,0 +1,63 @@
...
+        <input type="text" class="form-control" ng-model="filtersCtrl.filters.q">
+      </div>
+
+      <div class="filter">
+        <label translate>Activities</label><br>
+        ${add_multiselect('act', activities)}
+      </div>
+
+      <div class="filter">
+        <label translate>Article categories</label><br>
```
asaunier: Perhaps simply ``categories``? (already exist in the translated strings, in Transifex).
Or ``article_categories`` if this attribute name will be added to the strings to translate.

c2corg_ui/templates/article/filters.html
```diff
@@ -0,0 +1,63 @@
...
+        ${add_multiselect('acat', article_categories)}
+      </div>
+
+    </div>
+
+    <div class="row collapse" id="moreFilters">
+
+      <div class="col-xs-12 col-sm-6 filter">
+        <label translate>Type</label><br>
```
asaunier: Maybe ``article_type``?

c2corg_ui/templates/article/index.html
```diff
@@ -0,0 +1,42 @@
...
+    <button class="btn btn-default orange-btn" protected-url-btn url="${request.route_path('articles_add')}"
+            uib-tooltip="{{'Create a new article' | translate }}" tooltip-placement="right">
+      <span class="glyphicon glyphicon-plus-sign"></span>
+    </button>
+    <app-pagination></app-pagination>
+  </div>
+  <app-advanced-search document-type="articles" use-map="false"></app-advanced-search>
+</section>
+<div class="map-right">
+  <app-map app-map-advanced-search="false" app-map-show-recenter-tools="false" app-map-default-map-filter="false"></app-map>
```
asaunier: No map for the advanced search page because articles have no geometries.

As for now, the map filtering is also (temporarly) disabled for "areas" advanced search as well. See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/area/index.html

c2corg_ui/templates/helpers/list.html
```diff
@@ -3,6 +3,7 @@
     <option translate value="${request.route_url('routes_index')}" ng-selected="{{mainCtrl.isPath('routes')}}">Routes</option>
     <option translate value="${request.route_url('waypoints_index')}" ng-selected="{{mainCtrl.isPath('waypoints')}}">Waypoints</option>
     <option translate value="${request.route_url('outings_index')}" ng-selected="{{mainCtrl.isPath('outings')}}">Outings</option>
+    <option translate value="${request.route_url('articles_index')}" ng-selected="{{mainCtrl.isPath('articles')}}">Articles</option>
```
asaunier: I don't think we'll add articles to this list: they have their own entry in the sidemenu. https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/sidemenu.html#L37-L40

c2corg_ui/templates/helpers/view.html
```diff
@@ -74,6 +74,15 @@
   ${get_attr(version, 'comment', md=False)  if version['comment'] else '' | n}
 </%def>
 
+<%def name="show_article_title(locale, is_tab_title=False)">\
```
asaunier: Are you sure we need this?
There's not ``title_prefix`` for articles. You should be able to use the standard ``show_title`` function.

c2corg_ui/static/partials/cards/articles.html
ginold: it is, as is strange using routes-outings for preferences :P we should have some generic `document-author`, `document-outings`, but it won't forcibly be the same. Maybe we can open an issue along the lines of the issue where we want to rename "route_" functions that have nothing to do with route.

asaunier: Not sure to get what you mean.
But anyway we don't need such data in the article cards so this ``<div class="outing-author text-center">`` and its content can be removed => problem fixed, isn't it?

c2corg_ui/static/js/maincontroller.js
ginold: ah sorry, i confused with other stuff. Yeah so you should add it here too. It will highlight the "topoguide" in the sidemenu if you are on /areas.

asaunier: OK thanks.

So it means:
* ``areas`` should be added to the list for ``topoguide`` => @dkocich could you do it in this PR?
* ``articles`` should not be added to the list for ``topoguide`` since they have their own menu entry => should another path condition be added?
Something like
```
} else if (path === 'articles') {
```
?

c2corg_ui/templates/helpers/view.html
dkocich: I am not sure. Show_title cause errors (TypeError: unsupported operand type(s) for +: 'dict' and 'str')... In areas you used show_route_title. Maybe it is better to use it here too and it will be easier to find and change it later with other parts named by "route"...

asaunier: You might indeed have to use ``show_route_title``.
It's probably related to this issue: https://github.com/c2corg/v6_ui/issues/594

c2corg_ui/static/js/maincontroller.js
dkocich: I added areas to the list.

dkocich:
I think I changed everything already. Please tell me when I should merge.

c2corg_ui/views/document.py
```diff
@@ -366,9 +366,9 @@ def render_page(content1, content2):
                 'version1': version1,
                 'version2': version2,
                 'geometry1': doc_v1['geometry']['geom']
-                if doc_v1['geometry'] else None,
+                if 'geometry' in doc_v1 else None,
```
asaunier: I wonder if this change is correct. OK for testing first whether the document has a geometry attribute or not (eg. articles). But are we sure ``doc_v1['geometry']`` is never ``None`` when there's a ``geometry`` attribute?
@tsauerwein what do you think?

In addition I think the indentation could be improve. It would be more readable to write 
```
'geometry1': doc_v1['geometry']['geom']
    if 'geometry' in doc_v1 else None,
```
wouldn't it?

asaunier:
@dkocich Thanks for the changes!

Next time, could you push your branch to the main repository, not to your fork? It's simpler when people want to get your branch and test it in their instance (no need to add a remote etc.).

The code looks OK to me, except perhaps concerning https://github.com/c2corg/v6_ui/pull/703#pullrequestreview-4221372

I have installed your branch in my instance and noticed a few remaining problems:

## general

* [ ] it seems the article categories https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L612-L623  and article types https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L625-L628 are not available in the "translation" template https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/i18n.html

By the way is the article type "collab" really relevantly named whereas for images it's "collaborative"?
https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L625-L628
https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L576-L580

## advanced search page

* the map filter is actually disabled but the container remains. We have the same problem for areas. See https://github.com/c2corg/v6_ui/issues/733 OK to fix this in a separated PR.
* [ ] the article titles are not displayed (indexing problem in my instance?)
* [ ] are the ``article_type`` and ``article_categories`` correctly justified? should be left aligned?
sans titre

## detail view pages

* [ ] there's a "f" in the top left corner of the page!?
* [ ] the activity icons are not displayed correctly in the title => black and too big (probably a style problem). By the way, I'm not sure it is relevant to display those activity icons in the title because for articles relevant to all activities it could make a bunch of icons to show. On the other hand, sometimes you have no activity set. Activity is not an attribute as important as for routes or outings. Perhaps we could simply remove those icons from the title section (we already have them in the "info" section).
* [ ] "articles" badge is cut to the right 
sans titre

ginold:
as for the big icons and badge : you probably forgot to put a class somewhere and the icons got super big, therefore the badge is cut because it is pushed.

asaunier:
Thanks for the tip @ginold 

I would recommend to simply remove the activity icons from the article title section.

dkocich:
Yes, that was again a problem with "article" instead of  "route" classes which are used for everything. I removed icons and title with badge is ok.

c2corg_ui/views/document.py
dkocich: Today I sometimes get an error when I want to view different document types ( f.e. here /outings/736946/fr/la-jumelle-3eme-pucelle-aisha ).
"json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)". 

I am not sure if this is caused by this change or because I did some other changes elsewhere / I rebased / worked on books API, etc... (error was still showing when I used original code changed here).

asaunier: Are you sure your API is up?

I think I got this error when trying to load a document while my API was down. Or perhaps your API does not return a JSON response. Try to load in a browser the API request your UI is doing.

dkocich:
branches to main repo - OK

### general
TODO - I will add translations; we changed it to collab with Tobias just to have it shorter. Maybe we can change it back later.

### advanced seach. page
I see title. I am not sure if show and how to align other parts
.
selection_001

### detail view page
I fixed it. 

selection_003

c2corg_ui/views/document.py
dkocich: Thanks. API was ok, it works now. I was just missing new col. (is_profile_public) in Users table, so some documents couldn't be selected.

asaunier:
> TODO - I will add translations; we changed it to collab with Tobias just to have it shorter. Maybe we can change it back later.

Sometimes shorter is more confusing (especially for translators), and it's better to reuse existing keywords (and existing translations). But it might be a problem here because for instance in French "article" and "image" have different gender (?), thus "collaborative" is not translated the same way for both. Perhaps we should have use clearer keywords such "collaborative image" / "collaborative article" etc.

OK leave as is for now but we might have to fix it in the future.

> I see title - it is your DB problem.

OK

> I am not sure if show and how to align other parts

No clue. Perhaps @ginold can help?

> Later we can choose color for article badge (class is not defined yet, but route badge is blue I think...)

OK. I think there are other document types with the gray badge (areas, images?). Anyway it's not critical. It can be improved later.

I don't think we need the "description" header for the article core:
* it's not a description
* it's not relevant to hide the article core => it's the main content
I would prefer the article core is shown as now but with no header.

dkocich:
> I don't think we need the "description"

Ok, changed. I will also add "article core" text in new commit with other things to translations.

asaunier:
Is the PR complete now or do you still have things to commit?

c2corg_ui/templates/article/edit.html
```diff
@@ -0,0 +1,156 @@
...
+
+            ## SUMMARY
+            <div id="summary-group" class="form-group full">
+              <label translate>summary</label>
+              <textarea name="summary" ng-model="article.locales[0].summary" class="form-control"></textarea>
+            </div>
+
+            ## DESCRIPTION
+            <div id="description-group" class="form-group full">
+              <label translate>article core</label>
```
tsauerwein: @asaunier Did you really mean to use "article core"?

asaunier: Why not? It's the fieldname used in v5. Can you think of a better name?

tsauerwein: I understood that you didn't want a label at all.

But "article core" sounds strange. "Article body" maybe?

asaunier: "Article body" sounds good, thanks.
The fieldname is useful in the editing form, not in the detail view, I think.

dkocich: Ok. let it be "article body" on editing page. On view page, it should be empty now without any title...

c2corg_ui/templates/article/index.html
```diff
@@ -0,0 +1,39 @@
...
+<header class="page-main-title">
+  <h3><span translate>Articles</span>
+    <span class="badge" ng-cloak>{{resCounter}}</span>
+  </h3>
+</header>
+<section class="documents-list-section">
+  <%include file="filters.html" />
+  <div class="documents-nav-buttons text-center">
+    <button class="btn btn-default show-documents-filters-phone" translate>Filters</button>
+    ${add_doctype_selector()}
```
asaunier: We don't need the doctype selector in the article advanced search, since it's not part of the guidebook.

c2corg_ui/templates/article/view.html
```diff
@@ -0,0 +1,143 @@
...
+      % endif
+    </section>
+  </div>
+
+  <div class="description">
+    ${get_document_description(locale, 'article')}
+##     ${get_document_locale_text(locale, 'description')}
+  </div>
+
+  ## ASSOCIATIONS
```
asaunier: This part is not correct.
You have used what was done for areas (by the way the links of the buttons use the ``a`` parameter, which relates to areas, not articles).

We should rather use the associations rendering of waypoints or routes.
I think you should also be able to associate documents from the current article (as with waypoints, routes etc.)

asaunier: > We should rather use the associations rendering of waypoints or routes.

Mmmm perhaps it's already the case, since there's another association block a few lines below.

Anyway this "areas-associations"-like block is not relevant here. Please remove it.

c2corg_ui/views/document.py
tsauerwein: > > Thanks. API was ok, it works now.

Is the problem resolved, or not?

c2corg_ui/templates/article/view.html
dkocich: I did some little changes, but still I have an error and no new cards with associatited documents are showing after my click. I am still missing something - probably here? https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/addassociation.js

selection_004

c2corg_ui/templates/article/edit.html
asaunier: > On view page, it should be empty now without any title...

The orange band still remains (empty). Is it possible to remove it completely?

c2corg_ui/views/document.py
dkocich: I did not have any problem during my work. I am not sure if other documents contain geometry if they do not have any. For the articles, I think we did not put it to schema so the attribute should never be there.

asaunier: For me it's not correct because it's possible to have ``doc_v1['geometry']: null``. In that case ``'geometry' in doc_v1`` is ``True`` but ``doc_v1['geometry']['geom']`` will cause an error.

In addition the current indentation of 
```
                'geometry1': doc_v1['geometry']['geom']
                if 'geometry' in doc_v1 else None,
```
is confusing and very readable.

I recommend
```
                'geometry1': doc_v1['geometry']['geom']
                    if 'geometry' in doc_v1 and doc_v1['geometry'] else None,
```

c2corg_ui/static/partials/cards/articles.html
```diff
@@ -0,0 +1,21 @@
+<a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
+  <div class="list-item-info">
+
+    <span class="list-item-title">{{locale.title}}</span>
+      <div class="text-center value">{{doc['article_type'] | translate }}</div>
```
asaunier: Indent seems not correct here.

c2corg_ui/static/partials/cards/articles.html
```diff
@@ -0,0 +1,21 @@
...
+
+    <div class="article-categories">
+      <span ng-repeat="category in doc.categories" class="value">{{category | translate }}</span>
+    </div>
+
+    <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{locale.lang}})</span>
```
asaunier: Shouldn't this be moved right after the title?

c2corg_ui/static/partials/cards/articles.html
```diff
@@ -0,0 +1,21 @@
...
+    <br>
+    <div class="quality" ng-if="doc['quality']" ng-class="doc['quality']"
+         uib-tooltip="{{'quality'| translate}} : {{doc['quality'] | translate}}" tooltip-placement="left"></div>
+
+    <div class="article-activities">
```
asaunier: To be moved right after "article-categories"?

asaunier:
The PR contains an empty ``c2corg_ui/tests/views/data/article_archive.json`` that seems used nowhere.
Is it really needed? If not, please remove it, thanks.

dkocich:
Do You have any other comments about what to change? Please tell me when to merge.

asaunier:
What about https://github.com/c2corg/v6_ui/pull/703/files#r83592645 ?

dkocich:
Ok. I missed that comment and now it is repaired, but without suggested indentation - it creates errors... 

make -f config/travis check
c2corg_ui/views/document.py:369:21: E131 continuation line unaligned for hanging indent
c2corg_ui/views/document.py:371:21: E131 continuation line unaligned for hanging indent
make: *** [flake8] Error 1

c2corg_ui/templates/article/preview.html
```diff
@@ -0,0 +1,36 @@
...
+  <h1 class="routes">
+    <div class="icons-header routes">${get_route_activities(article, 'top')}</div>
+    <span class="title">${locale['title']}</span><br>
+    <label class="badge article">
+      <span class="glyphicon glyphicon-arrow-right"></span>&nbsp;<span translate>article</span>
+    </label>
+  </h1>
+</header>
+
+<section class="view-details-section" app-view-details>
```
asaunier: Could you remove ``app-view-details`` here, as in this pending PR: https://github.com/c2corg/v6_ui/pull/740/files ?

asaunier:
> Ok. I missed that comment and now it is repaired, but without suggested indentation - it creates errors...

You should avoid the errors adding a ``\`` at the end of the first line:
```
'geometry1': doc_v1['geometry']['geom'] \
￼    if 'geometry' in doc_v1 and doc_v1['geometry'] else None,
```

asaunier:
Except the 2 last requested changes, it looks good to me and ready for merging, thanks!

dkocich:
That did not help.

c2corg_ui/views/document.py:368:57: E502 the backslash is redundant between brackets
c2corg_ui/views/document.py:370:57: E502 the backslash is redundant between brackets

asaunier:
OK, I don't know how we could fix this. Then forget it.

Concerning the preview change, forget it too if not already done, I will do it in my PR when your PR is merged.

By the way:
> make -f config/travis check
c2corg_ui/views/document.py:369:21: E131 continuation line unaligned for hanging indent
c2corg_ui/views/document.py:371:21: E131 continuation line unaligned for hanging indent
make: *** [flake8] Error 1

Are you sure this is the correct command. I would have rather expected ``make -f config/dkocich check``.

dkocich:
Yes, that was what I took from Travis log here when I saw error. Locally I use make -f config/dkocich check...

I did not fixup the other change so I am ready to merge and after you can merge the other PR.

Issue #711, opened 2016-10-08
Outing creation: minor problems


loicperrin:
- ~~On preview page, "{{ '2016-09-24T22:00:00.000Z' | amDateFormat:"dddd Do MMMM YYYY" }}" is displayed instead of the date~~ -> duplicates #168 
- [x] When creating an outing an routes/54997/fr/aiguille-occidentale-du-soreiller-arete-s, the default title is of the outing is "Aiguille Occidentale du Soreiller : Cresta S" whereas the language is french. 
- [ ] When adding a participant to an outing, the search field is then filled with "undefined (en)" instead of "Rechercher" (it is OK before adding a participant)

asaunier:
> On preview page, "{{ '2016-09-24T22:00:00.000Z' | amDateFormat:"dddd Do MMMM YYYY" }}" is displayed instead of the date

This is already reported at https://github.com/c2corg/v6_ui/issues/168#issuecomment-245893257
> not sure angular code is correctly executed within the modal preview (for instance it seems that the outing dates are not correctly shown with angular moment and that the interface strings translations does not work)

=> striking it as duplicate

ginold:
>  When creating an outing an routes/54997/fr/aiguille-occidentale-du-soreiller-arete-s, the default title is of the outing is "Aiguille Occidentale du Soreiller : Cresta S" whereas the language is french.

what is the problem with that? If you're creating an outing from this route, it will take the route's name by default, and if you are creating a new outing and then add a new associated route, it will take its name as the outing name.

loicperrin:
> what is the problem with that?
The name of the route is : "Aiguille Occidentale du Soreiller : Arête S", and the name of the outing is by default "Aiguille Occidentale du Soreiller : **Cresta** S" instead of  "Aiguille Occidentale du Soreiller : **Arête** S"

stef74:
@loicperrin please test and close ?

loicperrin:
Ok for the default title is of the outing
But the participant association works less than before: now it does not find anyone and it proposes to create a WP/route/outing:
![issue711](https://cloud.githubusercontent.com/assets/19718055/19404087/a478df0e-926b-11e6-8df0-75a609eeccd5.jpg)
It 's a pitty that after a "bug fix", the fonctionnality works less than before...

loicperrin:
It works (but I had a problem before deconnection).

PR #720, opened 2016-10-11
Duplicate and default outing title

### LINKED ISSUE #719: duplicate naming in outing-editing
screenshot from 2016-10-11 12-27-13

we have 2 times "itineraire emprunte" which I find superfluous. Maybe i can move the input field just above the first title and remove the 2nd one? Or change the naming of the field.

ginold:
related to (fixes) #711  "When creating an outing an routes/54997/fr/aiguille-occidentale-du-soreiller-arete-s, the default title is of the outing is "Aiguille Occidentale du Soreiller : Cresta S" whereas the language is french."

fixes #719


asaunier:
PO files should not be part of the repo.

Issue #694, opened 2016-10-05
Delete association between an outing and a route


shortoland:
I go on the outing page http://c2corgv6.demo-camptocamp.com/outings/723667/fr/pic-de-l-affron-depuis-le-refuge-wallon, entered by BenB.

I click on the trash included in the corresponding route card:

![image4](https://cloud.githubusercontent.com/assets/20488365/19118859/12fd5fa4-8b1e-11e6-9555-01e8585f3b55.png)

After confirmation, I can delete the association between the route and the outing:

![image5](https://cloud.githubusercontent.com/assets/20488365/19118895/37f6d4ca-8b1e-11e6-8b31-27529ef0c3e2.png)

I don't think it's normal I am allowed to do that.

asaunier:
Indeed, only moderators (and associated users) should be able to edit or remove an association in an outing document.

You have no moderator permissions since the "edit" button does not appear on the right side of the page.

tsauerwein:
> Indeed, only moderators (and associated users) should be able to edit or remove an association in an outing document.

I don't think the API enforces this constraint at the moment.

asaunier:
Then we should create an API issue too.
It's not consistent to prevent people to edit the outing but to let them change the associations.

* [ ] By the way the "add association" tool should be shown in personal documents (as outings) only for moderators and associated users.

loicperrin:
2 questions:
- If I am the author of the outing, it is authorized to desassociate the last associated route? I think it should not
- Are the moderators permissions available in demo? In fact I have modified an outing created by Tintin, but maybe it is due to moderators rights (and so it would be correct)

asaunier:
> If I am the author of the outing, it is authorized to desassociate the last associated route? I think it should not

Why?

> Are the moderators permissions available in demo?

Yes

tsauerwein:
> Then we should create an API issue too.

See https://github.com/c2corg/v6_api/issues/414

loicperrin:
> Why?

Because the outing has no more associated route. It is the same kind of problem as for routes when we delete the last associated WP.

asaunier:
> Because then the outing has no more associated route. It is the same kind of problem as for routes when we delete the last associated WP.

Right!

ginold:
what we can do for now is to not let the user to delete the last associated route, because an outing requires at least one.

stef74:
this issue is tagged pull request ready, normal ?

asaunier:
[...]

PR #726, opened 2016-10-12
Don't delete the last associated route to outing + deleting only for appropriate user

ginold:
related to #694

but i'm not 100% sure it works as there have been issues with `userCtrl.hasEditRights`

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -242,7 +242,12 @@
           <app-card app-card-doc="doc"></app-card>
           % if type == 'routes':
             <app-delete-association child-doctype="routes" parent-id="${document['document_id']}" child-id="doc.document_id"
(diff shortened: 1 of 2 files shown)


ginold:
@asaunier the PR seems to work now - I can delete the routes only from my own outing but not the last remaining one.

asaunier:
If the PR is ready for reviewing, please update the title (and label the PR as ready?).

ginold:
As you have moderator rights, you should check if you can delete a route from and edit any outing.

asaunier:
> As you have moderator rights, you should check if you can delete a route from and edit any outing.

You can change your own permissions by setting the ``moderator`` attribute to ``True`` in your ``users.user`` table :P

ginold:
it works as expected for moderators. May i merge?

asaunier:
Please do, thanks!

PR #730, opened 2016-10-13
preferences style update

ginold:


tsauerwein:
Could you share a screenshot?

ginold:
@tsauerwein 
screenshot from 2016-10-13 15-26-21

asaunier:
that would make sense to do similar changes for the "my account" page as well, eg. the title color etc.

PR #731, opened 2016-10-13
Feed

ginold:
first version of the feed :)
screenshot from 2016-10-13 15-22-17


asaunier:
I'm not very found of the orange band. I would prefer something simpler/softer (a bit like Twitter?).

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -339,7 +339,9 @@ app.Authentication.prototype.needAuthorization = function(method, url) {
       url.indexOf('/users/register') !== -1) {
     return false;
   }
-
+  if (url.indexOf('/personal-feed') !== -1) {
```
asaunier: Why not adding this condition to the next test block?

c2corg_ui/static/js/apiservice.js
```diff
@@ -380,6 +380,20 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
 
 
 /**
+ * @param {string} url
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.readFeed = function(url) {
```
asaunier: Wouldn't it be more consistent to have the URL defined in this service (as for the other functions) and only pass a "type" parameter (standard, personal, profile) as argument?

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,121 @@
...
+   */
+  this.api_ = appApi;
+
+  /**
+   * @type {app.Authentication}
+   * @private
+   */
+  this.auth_ = appAuthentication;
+
+
```
asaunier: extra blank line :)

stef74:
Too many 'blank' on facebook we have one colonum but it's more "dense".

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,121 @@
...
+app.FeedController.prototype.getHomeFeed_ = function() {
+  if (this.auth_.isAuthenticated()) {
+    this.api_.readFeed('/personal-feed').then(function(response) {
+      var data = response['data']['feed'];
+      this.documents = data;
+    }.bind(this));
+  } else {
+    this.api_.readFeed('/feed').then(function(response) {
+      var data = response['data']['feed'];
+      this.documents = data;
```
asaunier: This code seems the same than for "personal"? If yes, what about defining a named function?

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,121 @@
...
+    }.bind(this));
+  }
+};
+
+
+/**
+ * @private
+ */
+app.FeedController.prototype.getUserProfileFeed_ = function() {
```
asaunier: Do we really need a separated function for the profile?

c2corg_ui/templates/index.html
```diff
@@ -37,14 +37,29 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
+
+        <article class="feed-list-item" protected-url-btn url="${request.route_path('outings_add')}">
+          <div class="list-item vertical-align">
+            <button class="orange-btn btn">add an outing</button>
+          </div>
+        </article>
+
+        <article class="feed-list-item" ng-repeat="doc in feedCtrl.documents"
+                 ng-init="type = feedCtrl.documentType(doc.document.type);">
+          <p class="action" ng-cloak><b><a ng-href="todo">{{doc['user']['username']}}</a></b>
```
asaunier: Is is really relevant to have a link inside a card?

c2corg_ui/templates/index.html
```diff
@@ -37,14 +37,29 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
+            <span x-translate>has {{doc.change_type}}</span> a(n)
```
asaunier: I am not very found of those splitted translations, that makes them difficult to translate (context etc.).

c2corg_ui/templates/index.html
asaunier: Maybe it's OK. For instance there are links in Twitter cards.

c2corg_ui/templates/index.html
asaunier: And it doesn't work for german for instance, because the document type is not at the end of the sentence.

asaunier:
Don't spend too much time on the styling now. It's more important that the feed actually works (auth or not, profile) + infinite scrolling + actual content (eg. images).

We can improve the layout (styling + number of columns?) as a second step.

c2corg_ui/templates/index.html
tsauerwein: For outings it would also be cool to show something like "xyz went hiking with ...".

tsauerwein: And to avoid the "a(n)", we could write "... created/added a new waypoint/article/...".

tsauerwein: The date of the change should also be shown somewhere.

c2corg_ui/templates/index.html
ginold: of course it is, when you are watching the news on your homepage you can directly go to the mentioned document.

ginold:
@asaunier updated the PR and ready for review.
Featuring:

+ infinite scroll
+ different feed for logged and not-logged users
+ bling bling timline (sorry Alex, I had to add some of my contribution!)
+ feed does not load if it's already loading (no multiple requests)
+ clickable cards
+ responsive 

If you are trying it for yourself - be patient - the feed takes some time to load.

screenshot from 2016-10-18 13-45-59

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,153 @@
...
+   * @type {angular.$sce} $sce Angular Strict Contextual Escaping
+   * @private
+   */
+  this.sce_ = $sce;
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.location_ = $attrs['location'];
```
asaunier: That would be more consistent to:
* name the attribute according to the directive, eg. ``appFeedLocation``
* retrieve this param by making the constructor a ``@struct`` and pass the argument in the directive's scope or in ``bindToController`` (as in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/imageuploader.js)

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,153 @@
...
+   */
+  this.busy = true;
+
+  /**
+   * @type {string}
+   * @private
+   */
+  this.feedType_;
+
+  switch (this.location_) {
```
asaunier: I wonder if ``this.location_`` and ``this.feedType_`` are not a bit duplicates.

What about dropping ``location`` and use only ``app-feed-type`` in the HTML.
In the homepage, this attribute would be omitted and
* if is auth => personal
* if not => standard
In a user profile we would add ``app-feed-type="profile"`` to the directive.

asaunier: By the way do you really need to distinguish here the personal and standard feeds? The check could be done only in the apiservice depending on the user authentication.

tsauerwein:
Really cool! :+1: 

Maybe add some margin around the document box inside the bubble?

For outings you could also list the other participants.

> If you are trying it for yourself - be patient - the feed takes some time to load.

I gave it a try on my system. The API request to fetch the feed takes 200-400ms. Could it be related to https://github.com/c2corg/v6_api/issues/353? You can also start the API server with `.build/venv/bin/pserve --reload development.ini`.

c2corg_ui/static/js/feed.js
asaunier: So instead of ``app-feed-type``, you could only have an optional ``app-feed-profile="true"`` attribute that would be passed to the apiservice.

c2corg_ui/static/js/apiservice.js
```diff
@@ -380,6 +387,39 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
 /**
+ * @param {string} type
+ * @param {undefined | string} token
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.readFeed = function(type, token) {
+  var url = '';
+  var userId;
+
+  switch (type) {
```
asaunier: This switch is somehow a duplicate of the switch in the directive controller. As explained below, I would recommend to only have a "isProfile" boolean parameter. If ``true`` this function would use ``/profile-feed``, else it would test if the user is authenticated to choose among the standard or personal feeds.

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,153 @@
...
+app.FeedController.prototype.getDocumentsFromFeed = function() {
+  this.busy = true;
+  this.api_.readFeed(this.feedType_, this.nextToken_).then(function(response) {
+    var data = response['data']['feed'];
+    for (var i = 0; i < data.length; i++) {
+      this.documents.push(data[i]);
+    }
+    this.nextToken_ = response['data']['pagination_token'];
+    this.busy = false;
+    data = [];
```
asaunier: Really needed?

c2corg_ui/templates/index.html
```diff
@@ -37,14 +37,52 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
+        <article class="feed-list-item" ng-repeat="doc in feedCtrl.documents"
+                 ng-init="type = feedCtrl.documentType(doc.document.type);">
+
+          <div class="timeline-bullet icon-{{type}}s"></div>
+
+          <p class="action-line">
+            <span class="action" ng-cloak>
+              <b><a ng-href="todo">{{doc['user']['username']}}</a></b>
+              <span ng-bind-html="feedCtrl.createActionLine(doc)"></span>
+              <b translate>{{type}}</b>
```
asaunier: That wouldn't work for german since the past participle should be shown at the end of the sentence?!

asaunier:
> I gave it a try on my system. The API request to fetch the feed takes 200-400ms. Could it be related to c2corg/v6_api#353? You can also start the API server with .build/venv/bin/pserve --reload development.ini.

In my instance, with ``build/venv/bin/pserve --reload development.ini`` the performances of the feed are indeed way more acceptable.

By the way, it's not very fluid to wait that the user has scrolled until the end of the current feed items before triggering a new feed request. Would it be possible to trigger it sooner?

tsauerwein:
Sometimes when scrolling down, the "feed line" does not go down to the bottom. Both in FF and Chrome.

c2corg-feed-missing-line

c2corg_ui/templates/index.html
```diff
@@ -37,14 +37,52 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
+          <div class="timeline-bullet icon-outings"></div>
+          <div class="list-item vertical-align">
+            <button class="orange-btn btn">add an outing</button>
+          </div>
+        </article>
+
+        <article class="feed-list-item" ng-repeat="doc in feedCtrl.documents"
+                 ng-init="type = feedCtrl.documentType(doc.document.type);">
+
+          <div class="timeline-bullet icon-{{type}}s"></div>
```
tsauerwein: Could you add a tooltip for these icons? Like for the other icons.

ginold:
@tsauerwein 

> Maybe add some margin around the document box inside the bubble? 

you mean padding around the card, so that it does not stick to the borders?

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,153 @@
...
+ * @export
+ */
+app.FeedController.prototype.getDocumentsFromFeed = function() {
+  this.busy = true;
+  this.api_.readFeed(this.feedType_, this.nextToken_).then(function(response) {
+    var data = response['data']['feed'];
+    for (var i = 0; i < data.length; i++) {
+      this.documents.push(data[i]);
+    }
+    this.nextToken_ = response['data']['pagination_token'];
```
ginold: do you think this will work? 

```
var token = response['data']['pagination_token'];
token ? (this.nextToken_ = token, this.busy = false) : this.busy = true; // disable scroll
```

tsauerwein: From the idea yes, but I would prefer a proper if-else construct instead of doing side-effects in the ternary operator:

```
if (token) {
   this.nextToken_ = token;
   this.busy = false;
} else {
   this.busy = true; // disable scroll
}

**Comment by asaunier** [comment]
When viewing the homepage feed, it would be useful for authenticated users:
* to be reminded that they have filters and "followship" enabled
* to provide them a shortcut (button?) towards their preferences interface
* to provide them a button (or checkbox?) to temporarly disable the filters (in order to see the standard feed without having to make changes in their preferences)

Perhaps it could be done later (in a next PR) but at least it's worth taking this future evolutions into account for writing the current code. Especially the apiservice related part.
For instance temporarly disabling the filters could be done by:
* adding a ``filter=0`` parameter to the ``/personal-feed`` request
* or switching to ``/feed``
until the "disable filters" tool is turned of. When switching on/off the "disable filters" tool, it's probably more relevant to reset the feed (not only to apply the change to the subsequent infinite scrolling calls).

**Comment by tsauerwein** [comment]
> adding a filter=0 parameter to the /personal-feed request

I just removed that parameter! So, you will have to query the public feed.

**Comment by ginold** [comment]
+ tooltips
+ error msg if fetching fails
+ stop scrolling if no more feed (token)
+ fix timeline bar
+ preferred language
+ change the feed type logic

File: c2corg_ui/static/js/apiservice.js [hunk 15]
```diff
@@ -380,6 +387,37 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
 /**
+ * @param {undefined | string} token
+ * @param {string} lang
+ * @param {boolean} isProfile
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.readFeed = function(token, lang, isProfile) {
+  var url;
+  var params = {'pl': lang};
+  if (token) params['token'] = token;
```
asaunier: Is it really allowed by the linter to have one-line "if" blocks?

c2corg_ui/static/js/apiservice.js
```diff
@@ -380,6 +387,37 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
+
+  if (isProfile || this.auth_.isAuthenticated()) {
+    params['u'] = this.auth_.userData.id;
```
asaunier: ``u`` is needed only for the user profile feed. Not for the auth homepage feed:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/feed.py#L73

c2corg_ui/static/js/apiservice.js
```diff
@@ -380,6 +387,37 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
+  }
+  if (isProfile) {
```
asaunier: You could move ``params['u'] = this.auth_.userData.id;`` here.

c2corg_ui/templates/index.html
```diff
@@ -37,14 +37,53 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
-    <button class="btn orange-btn" id="add-my-outing" protected-url-btn url="${request.route_path('outings_add')}"><span class="glyphicon glyphicon-plus-sign"></span><span translate>add an outing</span></button>
-    <app-advanced-search document-type="outings" use-map="false"></app-advanced-search>
+    <div class="feed-title flex flexend-align" ng-show="!feedCtrl.busy && !feedCtrl.error">
+      <span class="icon-news"></span>
+      <h3 translate>Activity feed</h3>
+    </div>
+    <h3 translate class="text-center text-danger" ng-show="feedCtrl.error">Sorry, there was an error while getting the feed.</h3>
+
+    <div app-feed app-feed-profile="false" app-loading>
+      <h4 class="text-success text-center" translate ng-show="feedCtrl.busy">Loading feed, please wait...</h4>
```
asaunier: What about a small spinning wheel as well?

c2corg_ui/templates/index.html
```diff
@@ -37,14 +37,53 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
+      <div class="feed-list" ng-show="feedCtrl.documents.length > 0"
+           infinite-scroll="feedCtrl.getDocumentsFromFeed()"
+           infinite-scroll-immediate-check="false"
+           infinite-scroll-distance="0.7"
```
asaunier: That triggers the new feed request before the user has reached the end of the feed, right?
Cool :)

c2corg_ui/static/js/feed.js
```diff
@@ -0,0 +1,157 @@
...
+    case 'created':
+      line += 'has created a new ';
+      break;
+    case 'updated':
+      line += 'has updated the ';
+      break;
+    default:
+      break;
+  }
+  return line + this.documentType(doc['document']['type']);
```
asaunier: This means the strings to translate will be "has created a new outing", "has updated the waypoint", etc. right?
Then you should add those strings to the ``i18n.html`` template.

c2corg_ui/static/js/apiservice.js
ginold: no erros with make check !

c2corg_ui/templates/index.html
ginold: what do you mean? i'm asking myself if i should leave 'app-loading' that will show the spinning wheel as usual for each feed fetch, or just leave the 'loading, please wait...'

c2corg_ui/templates/index.html
ginold: yep, that's the % of the height where the directive is set.
You can read more on https://sroze.github.io/ngInfiniteScroll/documentation.html

c2corg_ui/static/js/apiservice.js
asaunier: Indeed. But I think we prefer always having the curly braces for readability reasons (OK that's minor).

asaunier:
Oh by the way (no need to address this in this PR): what happens if users scroll the feed, clicks on a card (open a new page) and then want to go back to the feed by pressing the "back" button: will the feed be reset or will it be restored at the same place?

If reset, I think it might be frustrating for the users to go back to the beginning...

ginold:
yep it will be reset. The same happens on facebook! the only thing  i see is happening, is that facebook somehow remembers the position 'in height' where you last left the page and then scrolls all the way down, loading all the posts again. (or maybe it's a browser behavior, i don't know)

asaunier:
On facebook, clicking on a card opens a new tab, so the feed is still available.

c2corg_ui/templates/index.html
asaunier: No, only a small spinning wheel cursor right next to the "loading please wait" message (would be cool but not mandatory)

ginold:
..not on my browser :P so those things are managed by the browser rather than fb. I've read somewhere that you cannot programmaticaly tell to open a page in a new tab or in a new window - it's up to the browser.

asaunier:
Don't know. But reset the feed every time one clicks on a card is a serious usability issue...
Perhaps we will have to pass some kind of permalink to the address bar?

asaunier:
> Don't know. But resetting the feed every time one clicks on a card is a serious usability issue...
Perhaps we will have to pass some kind of permalink to the address bar?

But no need to find a solution now. We could also wait for the feedbacks of the testers.
Please simply address the last suggestions I did and I think the PR will be OK for merging.

c2corg_ui/static/js/apiservice.js
ginold: I will leave it, it this case it's really short and we save 2 extra lines :P

asaunier: Readability and code consistency is more important than saving lines. And closure will make the lines optimisation when building the JS.
Whatever I won't fight for this :)

asaunier:
Please merge when Travis passes, thanks :)

PR #734, opened 2016-10-14
Change typo in English word

dkocich:
"Information" is not plural in English (but it can be maybe used like that in French?). I found it 12x at "preview" and "view" pages. It would be nice for the user to see a web page without any grammar mistakes. What do you think about this change?

More info f.e. here http://dictionary.cambridge.org/dictionary/english/information OR http://ell.stackexchange.com/questions/17748/information-or-informations

### DIFF
--- a/c2corg_ui/templates/area/preview.html
+++ b/c2corg_ui/templates/area/preview.html
@@ -23,7 +23,7 @@ ${show_preview_warning()}
 
   <div class="view-details-informations col-xs-12 informations">
     <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#document-informations">
-      <span translate>Informations</span><span class="glyphicon glyphicon-menu-down"></span>
+      <span translate>Information</span><span class="glyphicon glyphicon-menu-down"></span>
     </h3>
     <section id="document-informations" class="collapse in">
       ${get_area_general(area)}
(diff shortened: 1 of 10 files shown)


tsauerwein:
Thanks! Please merge when Travis has passed.

asaunier:
This change makes sense, clearly.
But changing a translation in the code makes Transifex (the online translation service that we use to make people translate the UI strings) think it's a new string. The old string (with all translations already done) is then lost!

So please don't merge.

asaunier:
@shortoland what do you think? OK if this (small) translation string is updated (the available translations in Transifex will be lost)?

tsauerwein:
This sounds interesting:

> ... Translation Memory is a database of previously completed translations. Each time a translation is submitted, it's saved to this database. In the future, if a phrase similar to the original source string appears, the translation is presented to the translator as a suggestion.

http://docs.transifex.com/setup/introduction-to-translation-memory/

shortoland:
"Information" can be plurial in french ;-)
I found it twice in Transifex translations. I corrected that.
According to me it's not a problem to have mistakes in us english, as it won't be displayed (only en english is).

dkocich:
@shortoland I believe this is a problem also of UK-English (Cambridge dictionary is from UK).

So if translation memory is enabled I do not see any problem with merging. http://docs.transifex.com/setup/enabling-translation-memory-autofill/

Please decide because I am not sure now.

asaunier:
OK go for it, since it's a one-word translation.

> So if translation memory is enabled I do not see any problem with merging.

That's not that simple. Even though this feature is enabled, changing the translation string requires a translator action (for each of the 7 languages of the project) to recreate the translation. In the meantime, the string is simply not translated.

We have to be very careful when creating or updating the translated strings.

PR #739, opened 2016-10-17
Get geolocation from image EXIF

### LINKED ISSUE #737: Retrieve image coordinates in EXIF data if any
And automatically fill in the geometry when uploading an image.

Example of image with coordinates in EXIF data: http://www.camptocamp.org/images/466723/fr/tete-tsernou-et-tete-a-pierre-grept-depuis-la-pointe-n

ginold:
fixes #737

### DIFF
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -383,7 +383,7 @@ app.ImageUploaderController.prototype.getImageMetadata_ = function(file) {
     if (exif) {
       angular.extend(file['metadata'], exif.getAll());
       if (file['metadata']['GPSLatitude']) {
-        this.setGeolocation_(file);
+        this.getGeolocation_(file);
       }
       return;
     }
@@ -395,7 +395,7 @@ app.ImageUploaderController.prototype.getImageMetadata_ = function(file) {
  * @param {File} file
  * @private
  */
-app.ImageUploaderController.prototype.setGeolocation_ = function(file) {
+app.ImageUploaderController.prototype.getGeolocation_ = function(file) {
   var lat = file['metadata']['GPSLatitude'].split(',');
   var lon = file['metadata']['GPSLongitude'].split(',');
   lat = app.utils.convertDMSToDecimal(lat[0], lat[1], lat[2], file['metadata']['GPSLatitudeRef']);
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -234,7 +234,9 @@ app.ImageUploaderController.prototype.uploadFile_ = function(file) {
       'id': file['name'] + '-' + new Date().toISOString(),
       'activities': [],
       'categories': [],
-      'image_type': this.image_type_
+      'image_type': this.image_type_,
+      'elevation': null,
+      'geometry': {}
     }
   });
   this.getImageMetadata_(file);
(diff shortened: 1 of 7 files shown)


c2corg_ui/templates/image/view.html
```diff
@@ -33,6 +32,10 @@
 </%block>
 
 <%block name="moduleConstantsValues">
+  <%
+    geometry = loads(image['geometry']['geom'])['coordinates'] if image['geometry'] else Null
```
asaunier: 4-space indent in python

asaunier: ``Null`` => ``None``

c2corg_ui/static/js/imageuploader.js
```diff
@@ -380,17 +382,26 @@ app.ImageUploaderController.prototype.getImageMetadata_ = function(file) {
...
 
 
 /**
- * @export
+ * @private
  */
-app.ImageUploaderController.prototype.addGeoinfo = function(doc, file) {
-  file['metadata']['geo'] = doc;
+app.ImageUploaderController.prototype.setGeolocation_ = function(file) {
+  var lat = file['metadata']['GPSLatitude'].split(',');
```
ginold: yeah, forgot to not to go to the function if there are no gps data !

c2corg_ui/static/partials/imageuploader.html
```diff
@@ -35,17 +35,6 @@
...
         <div class="image-attributes btn-group" ng-show="file['processed']">
-          <!-- GEOLOCATION-->
-          <div class="btn-group">
-            <button class="btn btn-default dropdown-toggle" ng-class="{'btn-success': file.metadata.geo}" type="button" data-toggle="dropdown">
-              <span class="glyphicon glyphicon-globe"></span>
-            </button>
-            <div class="dropdown-menu geo">
-              <p translate ng-if="!file.metadata.geo">You can add geolocation info by typing in the input field.</p>
-              <label ng-if="file.metadata.geo">{{file.metadata.geo.label}}</label>
-              <app-simple-search dataset="wr" app-select="uplCtrl.addGeoinfo(doc, file)"></app-simple-search>
```
asaunier: Perhaps you can also remove ``addGeoinfo`` from the controller if no longer used?

ginold: already done

asaunier: OK, I had not paid attention that the old function was now ``setGeolocation_`` in the diff.

c2corg_ui/templates/image/view.html
```diff
@@ -33,6 +32,10 @@
 </%block>
 
 <%block name="moduleConstantsValues">
+  <%
+      geometry = loads(image['geometry']['geom'])['coordinates'] if image['geometry'] and image['geometry']['geom'] else Null
```
asaunier: What about Null/None?

``if image['geometry'] and image['geometry']['geom']`` => OK for me, but perhaps more consistent to use the same syntax for testing as below: ``if image.get('geometry') and image.get('geometry').get('geom')``?

Anyway I don't find it very consistent to get the coordinates this way. It would be better to use the same way than for waypoint for instance:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L52
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/waypoint.py#L48

So you probably have to add ``'geometry': self._get_geometry(waypoint['geometry']['geom']),`` to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/image.py#L42
Ditto for the archive view, a few lines below.

ginold:
updated

c2corg_ui/templates/image/view.html
```diff
@@ -1,7 +1,7 @@
 <%!
 from c2corg_common.attributes import activities, image_categories, image_types
 from c2corg_ui.templates.utils import get_lang_lists
-from json import dumps
+from json import dumps, loads
```
asaunier: This change is no longer relevant.

PR #740, opened 2016-10-17
Fix compilation of the preview in editing forms

asaunier:
Fix
> not sure angular code is correctly executed within the modal preview (for instance it seems that the outing dates are not correctly shown with angular moment and that the interface strings translations does not work)

see https://github.com/c2corg/v6_ui/issues/168#issuecomment-245893257

=> translations and date are now displayed as expected

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -538,17 +538,6 @@ app.DocumentEditingController.prototype.toggleOrientation = function(orientation
  * @export
  */
 app.DocumentEditingController.prototype.preview = function() {
-  var template = $('#preview-container').clone();
-  this.modal_.open({
-    animation: true,
-    template: this.compile_(template)(this.scope),
-    controller: 'appPreviewModalController',
-    controllerAs: 'previewModalCtrl',
-    size: 'xl'
-  });
-
-  // TODO: loading indicator?
-
   var config = {
     headers: {
       'Content-Type': 'application/json; charset=UTF-8',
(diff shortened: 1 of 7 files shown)


c2corg_ui/static/js/documentediting.js
```diff
@@ -559,12 +548,24 @@ app.DocumentEditingController.prototype.preview = function() {
...
   };
+  // TODO: loading indicator?
   this.http_.post(url, payload, config).
     catch(function(response) {
       this.alerts.addError('A server error prevented creating the preview');
     }.bind(this)).
     then(function(response) {
-      $('#preview-container-content').html(response['data']);
+      var template = angular.element('#preview-container').clone();
+      template.contents().filter(function() {
```
tsauerwein: You want to get the element with id `preview-container-content` inside the element `#preview-container`? Maybe you could use [find](https://api.jquery.com/find/) instead?

asaunier: I know :)
But the angular doc https://docs.angularjs.org/api/ng/function/angular.element says that 
> find() - Limited to lookups by tag name

asaunier:
Thanks for reviewing @tsauerwein I will merge when the articles are merged, this way I will apply the preview change to them as well.

> Great that you found a fix.

Yes, now I have no excuse for not implementing the original design in the user profiles https://github.com/c2corg/v6_ui/issues/335 :P

c2corg_ui/static/js/documentediting.js
asaunier: OK, then we can use ``find``. I have updated the PR (also with the change in articles).
I will merge when Travis passes then.

Issue #741, opened 2016-10-17
Associations problems in articles


asaunier:
I have just upgraded the demo (ditto in my dev instance). If I look at an article, say http://c2corgv6.demo-camptocamp.com/articles/713242/fr/procedure-documentation-technique-petzl I notice the following problems:

* [x] when searching for a document to associate (I have typed "grange"), the console reports that
```
Uncaught TypeError: Cannot read property 'length' of undefined
app.Document.hasAssociation	documentservice.js:102
(anonymous function)		simplesearch.js:319
```
* [x] I have not been able to associate a WP or a route (not tested other doc types) because of ``invalid association type : association`` (the associations request status is "400 Bad Request")
* [x] we don't need the following section:
![sans titre](https://cloud.githubusercontent.com/assets/1192331/19445056/28e95c5e-9493-11e6-8598-c66f39265e10.png)

asaunier:
* [ ] Perhaps it's possible to remove the commented out code at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/article/view.html#L62-L72 ?

dkocich:
Ok. I am working on that right now.

PR #745, opened 2016-10-18
Fix issue 741 - associations problems in article

dkocich:
Close https://github.com/c2corg/v6_ui/issues/741

### DIFF
--- a/c2corg_ui/static/partials/suggestions/articles.html
+++ b/c2corg_ui/static/partials/suggestions/articles.html
@@ -1,0 +1,6 @@
+  <span class="suggestion-text"> 
+    <span class="suggestion-title">
+      <span ng-bind-html="highlight(label, _query)"></span> 
+    </span><br>
+  </span>
+  <span class="icon-container"><span ng-bind-html="::activitiesHtml"></span></span>
--- a/c2corg_ui/static/js/addassociation.js
+++ b/c2corg_ui/static/js/addassociation.js
@@ -88,7 +88,8 @@ app.AddAssociationController.prototype.associate = function(doc) {
   // if the parent doc is a route and the child doc is a waypoint OR
   // if the parent doc is an outing, inverse the IDs.
   if ((parentType === 'routes' && doc['type'] === 'w') ||
-      parentType === 'outings' || parentType === 'images') {
+      parentType === 'outings' || parentType === 'images' ||
+      (parentType === 'articles' && doc['type'] === 'w')) {
     childId = this.parentId;
     parentId = doc['document_id'];
   } else {
--- a/c2corg_ui/static/js/documentservice.js
+++ b/c2corg_ui/static/js/documentservice.js
@@ -54,12 +54,14 @@ app.Document = function(appAuthentication, $rootScope) {
    *  routes: Array<number>,
    *  waypoints: Array<number>,
    *  users: Array<number>,
-   *  images: Array<number>
+   *  images: Array<number>,
+   *  areas: Array<number>,
+   *  articles: Array<number>
    * }}
    * @private
    */
   this.associationsIds_ = {
-    'routes': [], 'waypoints': [], 'images': [], 'users': [], 'areas': []
+    'routes': [], 'waypoints': [], 'images': [], 'users': [], 'areas': [], 'articles': []
   };
 };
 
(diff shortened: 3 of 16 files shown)


dkocich:
TODO - I still have a problem with deleting of articles associations. I will fix it.

I am not sure if cards visible under "Add associations"-search should be available also created separately for the articles in every document view.html page?

image

Should be article visible also in the upper right corner simple search?
image

Other things I am not sure about:
- Here we have HTML also for other types (areas, outings, users) which do not show in the simple search.  Is it ok to not show them? https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials/suggestions

asaunier:
> I am not sure if cards visible under "Add associations"-search should be available also created separately for the articles in every document view.html page?

Not sure to understand your question.
Do you mean "article" cards that would be shown in the detail view pages ("view.html") of associated documents?

> Here we have HTML also for other types (areas, outings, users) which do not show in the simple search. Is it ok to not show them?

It's no use searching for outings in the simple search tool (there would be too many results since many have the same name).

On the other hand "areas" should be supported by the simple search widget.
"users" would make sense too but I don't remember if the client has asked for it. If yes, that could be done in the user profile PR I am working on currently.

dkocich:
> Do you mean "article" cards that would be shown in the detail view pages ("view.html") of associated documents?

@asaunier Yes. Now everywhere is only association to routes/outling/waypoints... But it would make sense to see association to some articles when user looks to some waypoint (or route/image) that he wants to visit 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L168

Code should be ok now, but travis is broken.

asaunier:
> Now everywhere is only association to routes/outling/waypoints... But it would make sense to see association to some articles when user looks to some waypoint (or route/image) that he wants to visit

Right!
We probably need a ``show_associated_articles`` helper as well.

dkocich:
> We probably need a show_associated_articles helper as well.
OK. I added it to few places (article, image, outing, route, waypoint).

@asaunier 
But still I am not sure why we support some many possible associations between types ( https://github.com/c2corg/v6_common/blob/master/c2corg_common/associations.py ), but we are missing them on the UI. Is it a mistake or on purpose somewhere (planned for the other versions, available only on first creation, etc.)?
F. e. we here search and want to associate only waypoints and routes and not other types (...dataset="wr"...) https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/route/view.html#L140

asaunier:
> Is it a mistake or on purpose somewhere

It's most likely because the association tool settings (and associated docs section) have not been updated when new association types have been added.

Could you please create a new UI issue listing the missing associations for each document type, according to https://github.com/c2corg/v6_common/blob/master/c2corg_common/associations.py ?

c2corg_ui/static/partials/suggestions/articles.html
```diff
@@ -0,0 +1,6 @@
+  <span class="suggestion-text"> 
+    <span class="suggestion-title">
+      <span ng-bind-html="highlight(label, _query)"></span> 
+    </span><br>
+  </span>
+  <span class="icon-container"><span ng-bind-html="::activitiesHtml"></span></span>
```
asaunier: Does this add the activity icons to the right of the article suggestion?
That sounds relevant indeed but won't we have the same problem than with the article detail view titles when there are a lot of activities?
We can try as is but I'm curious if this will be OK.

asaunier: > I saw some icons in the other cases so I tried to add this also. And in case of "routes" there are also same icons. This is the case how it looks with many activities...

OK. Let's do it like this. Testers will tell us if this is a problem.

> Problem is sometimes that long title - then the icons are pushed under each other instead of being next to each other.

I think we have the same problem with other document types (for instance you may have routes with very long title). In that case, the title should be wrap to a second line.
Could you create an issue for this?

dkocich:
Thanks. I will merge probably when @ginold will say that it is ok. Other things will be solved in special issues+PR.

ginold:
it does look ok. If needed, solve the problems in the special pr.

asaunier:
Please have a look at Travis errors before merging the PR.

PR #748, opened 2016-10-18
Use pserve instead Gunicorn for development server

tsauerwein:
Same as https://github.com/c2corg/v6_api/pull/430


asaunier:
:+1:

Issue #754, opened 2016-10-19
Feed improvements


ginold:
as from #731 

- [ ] what happens if users scroll the feed, clicks on a card (open a new page) and then want to go back to the feed by pressing the "back" button: will the feed be reset or will it be restored at the same place?
Perhaps we will have to pass some kind of permalink to the address bar?

- [x] show a message when no more feed (error while fetching, loading feed, no more feed)

**When viewing the homepage feed, it would be useful for authenticated users:**

- [x] to be reminded that they have filters and "followship" enabled
- [x] to provide them a shortcut (button?) towards their preferences interface
- [x] to provide them a button (or checkbox?) to temporarly disable the filters (in order to see the standard feed without having to make changes in their preferences)

Perhaps it could be done later (in a next PR) but at least it's worth taking this future evolutions into account for writing the current code. Especially the apiservice related part.
For instance temporarly disabling the filters could be done by:

- [x]  switching to /feed until the "disable filters" tool is turned of. When switching on/off the "disable filters" tool, it's probably more relevant to reset the feed (not only to apply the change to the subsequent infinite scrolling calls).

- [x] create a partial template with the feed
- [x] remove app-loading directive

asaunier:
* [ ] it seems that the user name used in the action descriptions is the username (or login?). It would be more relevant to use the ``name`` attribute ("guidebook name") as in the outing cards:
![sans titre](https://cloud.githubusercontent.com/assets/1192331/19533514/606ff256-9640-11e6-9ff4-daf2ef88d737.png)

ginold:
> ... as in the outing cards

the ones in the feed ARE cards, so they are no different and they use `doc['author']['name']`, not 'username'

asaunier:
The problem is not in the card but in the comment above: "foobar has created a new outing" => the latter should use ``name`` as well.

Actually I think that the login should never be passed to (and used by) the UI, see https://github.com/c2corg/v6_api/issues/436
Only ``name`` and ``forum_username`` are relevant in the UI.

asaunier:
* [ ] Another point with feed improvement: are images (with filenames) supported? Please note that the images feed cards are a bit different then in the advanced search since they may contain several image filenames.

ginold:
if a user changes the feed type, disables the filters and so on - should the complete feed be reloaded ? or just add the other, new documents at the end of the feed?

asaunier:
> should the complete feed be reloaded ?

Yes

[...]

PR #755, opened 2016-10-19
add feed partial template + less file

ginold:
related to #754

@asaunier, when you integrate this into the preferences/user profile, you can use 
`<app-feed app-feed-profile="false" class="feed" ></app-feed>` 
anywhere you want. 

Then, don't forget to add the parent class into the `feed.less` file on the top. You may still want to change the bg color or the width of the feed in the less file. 

For example, just add `.preferences` next to `.home-page`, but then if you want to have a different color of the feed bg, type at the bottom of the file a new line `.preferences .feed { bg-color }` (!important will maybe not be needed)

### DIFF
--- a/c2corg_ui/templates/index.html
+++ b/c2corg_ui/templates/index.html
@@ -37,53 +37,5 @@ from c2corg_common.attributes import activities
 
     </div>
   </section>
-
-  <div class="feed">
-    <div class="feed-title flex flexend-align" ng-show="!feedCtrl.busy && !feedCtrl.error">
-      <span class="icon-news"></span>
-      <h3 translate>Activity feed</h3>
-    </div>
-    <h3 translate class="text-center text-danger" ng-show="feedCtrl.error">Sorry, there was an error while getting the feed.</h3>
-
-    <div app-feed app-feed-profile="false" app-loading>
-      <h4 class="text-success text-center" translate ng-show="feedCtrl.busy">Loading feed, please wait...</h4>
-      <div class="feed-list" ng-show="feedCtrl.documents.length > 0"
-           infinite-scroll="feedCtrl.getDocumentsFromFeed()"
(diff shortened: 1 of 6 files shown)


asaunier:
Do you really need ``class="feed"`` in ``<app-feed app-feed-profile="false" class="feed" ></app-feed>``?
Can't you use ``app-feed`` as the CSS selector instead?

> Then, don't forget to add the parent class into the feed.less file on the top.

Is it really necessary to add the parent class? Why can't I simply plug the directive in the user profile page and have it displayed out of the box?

ginold:
> Do you really need class="feed"

not really, i could write in css app-feed, but as it turns out, the good practices of css tell to always have a class instead of a general element call. (better to style .feed than app-feed)

>  Why can't I simply plug the directive in the user profile page and have it displayed out of the box?

i can simplify and it will be 'plugable' out ouf the box, but you will still surely have some adaptations to do.

ginold:
updated and ready to be plugged-in outta box!

c2corg_ui/templates/index.html
```diff
@@ -37,53 +37,5 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
...
-          <div class="list-item">
-            <app-card app-card-doc="doc['document']" class="{{type}}"></app-card>
-          </div>
-        </article>
-        <h4 class="text-success text-center" translate ng-show="feedCtrl.busy">Loading feed, please wait...</h4>
-      </div>
-
-    </div>
-  </div>
+  <app-feed app-feed-profile="false" id="feed"></app-feed>
```
asaunier: I don't think you need to explicitly set ``app-feed-profile="false"``. ``false`` is the default value I think.

ginold: seems to work :+1:

Issue #746, opened 2016-10-18
Wrong color for summary block


tsauerwein:
The bar of the summary block uses a different green value than for example for the activity icons.

![c2corg-color-summary](https://cloud.githubusercontent.com/assets/266474/19479784/aea1b2b4-9548-11e6-9292-58f1e73d0adf.png)

PR #758, opened 2016-10-20
Change summary block color in styles

dkocich:
Now the summary block color should be same as icons, left sidebar, etc. green color...

Close https://github.com/c2corg/v6_ui/issues/746

### DIFF
--- a/less/viewdetails.less
+++ b/less/viewdetails.less
@@ -558,7 +558,7 @@
 .document-summary{
   color: #626262;
   border: 1px solid #D8D8D8;
-  border-left: 0.5em solid #5FE45F;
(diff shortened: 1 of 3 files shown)


less/viewdetails.less
```diff
@@ -558,7 +558,7 @@
 .document-summary{
   color: #626262;
   border: 1px solid #D8D8D8;
-  border-left: 0.5em solid #5FE45F;
+  border-left: 0.5em solid #A9D361;
```
tsauerwein: You can use the color variable here:
https://github.com/c2corg/v6_ui/blob/master/less/variables.less#L12

tsauerwein:
Thanks, please merge once Travis has passed.

PR #759, opened 2016-10-20
add feed partial template + less file 

ginold:
copy of the precedent PR (i had some strange conflict errors).
@asaunier I think it's ok to merge?

it uses 'name' instead of 'username' in the cards.


asaunier:
Looks good to me thanks. OK for merging when Travis passes.

Issue #751, opened 2016-10-19
Problems with titles and icons in simple search suggestions


dkocich:
In this comment is an example picture of the problematic situation.

https://github.com/c2corg/v6_ui/pull/745#discussion_r84028506

Activities icons are under each other and that makes row height too big. They should stay next to each other.

dkocich:
It looks like two problems.

**1st - long titles pushing all icons under each other**

This one is solved with setting of:
`whitespace: nowrap` to `normal` for suggestion title

--- Before
![image](https://cloud.githubusercontent.com/assets/6165660/19558554/7bed7edc-96cb-11e6-9363-3fdb9b51803e.png)
--- After
![image](https://cloud.githubusercontent.com/assets/6165660/19558805/cc5793d4-96cc-11e6-9c83-1bb4f5fec338.png)


**2nd - when there are only 2 icons and normal title, they are also under each other**

If the icons have `float: none` and `float: right` is set only to their container, it is ok only if there are few icons. Otherwise, with many icons it do not work and they fill too much space. With a setting of `max-width` for them they start to overflow... I am not sure how to solve this now.

--- Before
![image](https://cloud.githubusercontent.com/assets/6165660/19558940/5ffa159e-96cd-11e6-9257-dc55c5d9341b.png)
--- After
![image](https://cloud.githubusercontent.com/assets/6165660/19559989/c55d7f2a-96d2-11e6-985b-42fe682554b3.png)

Any tips? @ginold 
Maybe merge only first problem and have a separate issue later for other problem because it is not so important?

PR #760, opened 2016-10-20
Repair simplesearch icons to show correctly with long title (not unde…

dkocich:
…r each other)

Related to https://github.com/c2corg/v6_ui/issues/751

### DIFF
--- a/less/simplesearch.less
+++ b/less/simplesearch.less
@@ -90,6 +90,7 @@ app-simple-search span.twitter-typeahead {
       }
     }
     .suggestion-title {
+      white-space: normal !important;
       font-weight: bold;
       color: graytext;
     }


asaunier:
Looks good to me. I don't know who made the styling of the simple search tool? @ginold ? @tsauerwein ? OK for you guys?

But does it really solve both problem you have reported at https://github.com/c2corg/v6_ui/issues/751 ?

If it solves only the first problem, I think it's better to keep the issue open, perhaps renaming it to make it less centered on the first problem. Thanks!

dkocich:
Yes, it solves only the first problem now. I am not sure how to fix the second issue with CSS/less... I will have to add something to the search function to add/remove styles based just on the number of icons of each searched item (if there are two icons for an item, apply new styles...).

PR #761, opened 2016-10-20
Turn on/off personal feed on homepage

ginold:
related to #754

Now you can turn on and off personal feed if you want to see the standard, non filtered feed, and access directly to your preferences as well.

screenshot from 2016-10-20 15-14-01

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -390,9 +390,10 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
  * @param {undefined | string} token
  * @param {string} lang
  * @param {boolean} isProfile
+ * @param {boolean} isPersonal
  * @return {!angular.$q.Promise<!angular.$http.Response>}
  */
-app.Api.prototype.readFeed = function(token, lang, isProfile) {
+app.Api.prototype.readFeed = function(token, lang, isProfile, isPersonal) {
   var url;
   var params = {'pl': lang};
   if (token) params['token'] = token;
@@ -400,7 +401,7 @@ app.Api.prototype.readFeed = function(token, lang, isProfile) {
   if (isProfile) {
     url = '/user-profile';
     params['u'] = this.auth_.userData.id;
-  } else if (this.auth_.isAuthenticated()) {
+  } else if (this.auth_.isAuthenticated() && isPersonal) {
     url = '/personal-feed';
   } else {
     url = '/feed';
--- a/c2corg_ui/static/js/feed.js
+++ b/c2corg_ui/static/js/feed.js
@@ -91,6 +91,13 @@ app.FeedController = function($scope, $attrs, appAuthentication, appApi, appLang
    */
   this.isProfile;
 
+  /**
+   * Used also in the template
+   * @type {boolean}
+   * @export
+   */
+  this.isPersonal = !this.isProfile;
+
   this.getDocumentsFromFeed();
 };
 
@@ -102,21 +109,19 @@ app.FeedController = function($scope, $attrs, appAuthentication, appApi, appLang
  */
 app.FeedController.prototype.getDocumentsFromFeed = function() {
   this.busy = true;
-  this.api_.readFeed(this.nextToken_, this.lang_.getLang(), this.isProfile).then(function(response) {
+  this.api_.readFeed(this.nextToken_, this.lang_.getLang(), this.isProfile, this.isPersonal).then(function(response) {
+    this.error = false;
     var data = response['data']['feed'];
     var token = response['data']['pagination_token'];
     for (var i = 0; i < data.length; i++) {
       this.documents.push(data[i]);
     }
-    this.error = false;
-
     if (token) {
       this.nextToken_ = token;
       this.busy = false;
     } else {  // if no token = reached the end of the feed - disable scroll
       this.busy = true;
     }
-
   }.bind(this), function() { // Error msg is shown in the api service
     this.busy = false;
     this.error = true;
@@ -127,7 +132,7 @@ app.FeedController.prototype.getDocumentsFromFeed = function() {
 /**
  * Creates a HTML with action that user used on the document in the feed.
  * Will be useful for verbs like 'created', 'updated', 'associated xx', 'went hiking with xx'.
- * @return {string | *} line
+ * @return {string} line
  * @export
  */
 app.FeedController.prototype.createActionLine = function(doc) {
@@ -144,6 +149,17 @@ app.FeedController.prototype.createActionLine = function(doc) {
       break;
   }
   return line + this.documentType(doc['document']['type']);
+};
+
+/**
+ * Switches between /personal-feed and /feed
+ */
+app.FeedController.prototype.toggleFilters = function() {
+  this.isPersonal = !this.isPersonal;
+  this.nextToken_ = undefined;
+  this.documents = [];
+  this.getDocumentsFromFeed();
+  window.scrollTo(0, 0);
 };
 
 /**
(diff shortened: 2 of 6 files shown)


c2corg_ui/static/js/apiservice.js
```diff
@@ -390,17 +390,18 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
+app.Api.prototype.readFeed = function(token, lang, isProfile, personalFeed) {
   var url;
   var params = {'pl': lang};
   if (token) params['token'] = token;
 
   if (isProfile) {
     url = '/user-profile';
     params['u'] = this.auth_.userData.id;
-  } else if (this.auth_.isAuthenticated()) {
+  } else if (this.auth_.isAuthenticated() && personalFeed) {
```
asaunier: I guess you no longer need to test if user is authenticated.
If ``personalFeed`` and not authenticated should not occur and would be contradictory. It would be rejected by the API anyway.

c2corg_ui/static/js/apiservice.js
```diff
@@ -390,17 +390,18 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
  * @param {undefined | string} token
  * @param {string} lang
  * @param {boolean} isProfile
+ * @param {boolean} personalFeed
  * @return {!angular.$q.Promise<!angular.$http.Response>}
  */
-app.Api.prototype.readFeed = function(token, lang, isProfile) {
+app.Api.prototype.readFeed = function(token, lang, isProfile, personalFeed) {
```
asaunier: What about a more explicit var name for a boolean such as ``isPersonal``?
More consistent with ``isProfile``.

c2corg_ui/static/js/feed.js
```diff
@@ -91,6 +91,13 @@ app.FeedController = function($scope, $attrs, appAuthentication, appApi, appLang
    */
   this.isProfile;
 
+  /**
+   * Used also in the template
+   * @type {boolean}
+   * @export
+   */
+  this.personalFeed = !this.isProfile;
```
asaunier: Same comment than in the APIService: ``isPersonal``?

c2corg_ui/static/js/feed.js
```diff
@@ -147,6 +152,17 @@ app.FeedController.prototype.createActionLine = function(doc) {
 };
 
 /**
+ * Switches between /personal-feed and /feed
+ */
+app.FeedController.prototype.turnPersonalFeedOnOff = function() {
```
asaunier: ``toggleFilters``?

c2corg_ui/templates/index.html
```diff
@@ -38,4 +38,20 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
     </div>
   </section>
   <app-feed id="feed"></app-feed>
+
+  <div class="float-buttons" ng-if="userCtrl.auth.isAuthenticated()">
+    <div class="float-button-container">
+      <div class="float-button" ng-click="feedCtrl.turnPersonalFeedOnOff()" ng-class="{'personal-feed-on': feedCtrl.personalFeed}">
+        <span class="glyphicon glyphicon-user"></span>
+      </div>
+      <p class="float-button-text" x-translate ng-bind="feedCtrl.personalFeed ? 'Personal feed on' : 'Personal feed off'">Personal feed</p>
```
asaunier: We need to find a way to make ``Personal feed on`` and ``Personal feed off`` be extracted by the i18n tool.
Translate filters maybe?
At least add them to the ``i18n.html`` template but that's a poor solution.

c2corg_ui/static/js/apiservice.js
ginold: if you are logged and scroll through your personal-feed and then log out -> it will throw a big error on the screen. Checking for both allows to continue to scroll, but loading the normal /feed.

c2corg_ui/templates/index.html
```diff
@@ -38,4 +38,20 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
     </div>
   </section>
   <app-feed id="feed"></app-feed>
+
+  <div class="float-buttons" ng-if="userCtrl.auth.isAuthenticated()">
+    <div class="float-button-container">
+      <div class="float-button" ng-click="feedCtrl.toggleFilters()" ng-class="{'personal-feed-on': feedCtrl.isPersonal}">
+        <span class="glyphicon glyphicon-user"></span>
+      </div>
+      <p class="float-button-text" x-translate ng-bind="feedCtrl.isPersonal ? 'Personal feed on' : 'Personal feed off'">Personal feed</p>
```
asaunier: Is it really not possible to mark those strings as to be translated instead of adding them to ``i18n.html``?
For instance
```
<p class="float-button-text" x-translate ng-bind="feedCtrl.isPersonal ? ('Personal feed on' | translate) : ('Personal feed off' | translate)">Personal feed</p>
```

What is this ``Personal feed`` string? Shouldn't it be added to the translated strings as well (simply by using ``translate`` instead of ``x-translate``?

asaunier:
Not sure why Travis fails. Perhaps rebase your PR?

c2corg_ui/templates/index.html
tsauerwein: This could be rewritten as:

```js
<p class="float-button-text" x-translate ng-show="feedCtrl.isPersonal">Personal feed on</p>
<p class="float-button-text" x-translate ng-show="!feedCtrl.isPersonal">Personal feed off</p>
```

tsauerwein:
> Not sure why Travis fails. Perhaps rebase your PR?

I restarted the build, it was related to https://github.com/c2corg/v6_api/pull/442.

c2corg_ui/templates/index.html
ginold: because it's clearer or because the translations won't work properly?

ginold: the thing is that like this, even with ng cloak, both lines will shortly appear on load, and when they disappear the buttons are stack one upon another - 1-2 seconds of hässlich :(

asaunier: What about my suggestion with the translate filters? 
Such a solution has already been used for instance here:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/edit.html#L29

Adding translations to ``i18n.html`` should be avoided as much as possible because if you change or remove the strings in the code, you have to make sure they are also updated in this template.

c2corg_ui/templates/index.html
```diff
@@ -38,4 +38,20 @@ <h1 class="welcome-title text-center" translate>The mountain sports community</h
     </div>
   </section>
   <app-feed id="feed"></app-feed>
+
+  <div class="float-buttons" ng-if="userCtrl.auth.isAuthenticated()">
+    <div class="float-button-container">
+      <div class="float-button" ng-click="feedCtrl.toggleFilters()" ng-class="{'personal-feed-on': feedCtrl.isPersonal}">
+        <span class="glyphicon glyphicon-user"></span>
+      </div>
+      <p class="float-button-text" ng-bind="feedCtrl.isPersonal ? ('Personal feed on' | translate) :  ('Personal feed off' | translate)">Personal feed</p>
```
asaunier: Isn't a ``translate`` directive missing (to translate the default ``Personal feed``)?
Or perhaps we don't need this default ``Personal feed``?

ginold: it's a "dummy" text so that there is anything while the pages loads. I think I should to set it to "personal feed off" by default.

asaunier:
OK for merging this PR when Travis passes.

Issue #732, opened 2016-10-14
Sidemenu for beta and v6.0


asaunier:
For the beta and v6.0 there is no accident report section (and no "community"/association section?).

It would make sense to remove those to entries in the sidemenu. Or perhaps to make them point to the v5  URLs? Since the articles will be available in v6 beta, we could make the "association" sidemenu entry point to some v6 article. 
@stef74 @HSERGERAERT what do you think?

In addition I wonder if we really need the "Homepage" entry since clicking on the logo already gives access to the homepage (and it's a common way to do so on most websites). Should we remove it?
It would allow to save some space vertically and allow to display the ad banner for smaller screens (when the screen is not high enough, the ad banner is hidden).

stef74:
First I said yes ... but it can be a teasing. No link with v5 (indexation
problem) ... just infobox that say " bientôt ;-)"

You can remove "home" entry

2016-10-14 10:08 GMT+02:00 Alex Saunier <notifications@github.com>:

asaunier:
> No link with v5 (indexation
problem) ... just infobox that say " bientôt ;-)"

What do you mean by "infobox"?

Honestly I would prefer we stay simple, we have too many things to do.
Possible solutions:
* remove the entry
* hard-coded URL to v5
* disable link (entry is here but it's not clickable)

stef74:
disable link ...

2016-10-14 10:57 GMT+02:00 Alex Saunier <notifications@github.com>:

mgerbaux:
> we could make the "association" sidemenu entry point to some v6 article. 

_Camptocamp association_ can point to the corresponding V6 article of http://www.camptocamp.org/articles/106726/fr/camptocamp-association
 _Base de données accidents_ can point to the corresponding V6 article of http://www.camptocamp.org/articles/697210/fr/base-serac-de-recits-d-incidents-et-accidents

asaunier:
> Camptocamp association can point to the corresponding V6 article of http://www.camptocamp.org/articles/106726/fr/camptocamp-association

OK

> Base de données accidents can point to the corresponding V6 article of http://www.camptocamp.org/articles/697210/fr/base-serac-de-recits-d-incidents-et-accidents

What about pointing to the v5 URL http://www.camptocamp.org/xreports ?

mgerbaux:
> What about pointing to the v5 URL http://www.camptocamp.org/xreports ?
[...]

PR #764, opened 2016-10-21
Fix sidemenu links

asaunier:
Fix https://github.com/c2corg/v6_ui/issues/732

@ginold recommends to keep the "Home" entry in the menu.
By the way he suggested that when the sidemenu is not high enough to show the ad banner, the latter could be resized (shrunk) instead of simply be hidden. Anyway I won't do it in this PR.


tsauerwein:
LGTM

No sure why Travis is not triggered... :S

asaunier:
Perhaps because github seems to be a bit too loaded?

asaunier:
Hmmm I don't know if the Travis tests will be run ever?
I will merge because those changes are quite trivial and the tests pass in my instance.
Thanks for reviewing!

PR #766, opened 2016-10-21
Show name instead of username

tsauerwein:
Closes https://github.com/c2corg/v6_ui/issues/765

Requires https://github.com/c2corg/v6_api/pull/447

PR #772, opened 2016-10-21
Rename detailed attributes helper files to view.html

asaunier:
As for now, the helper templates dedicated to a given document type are named ``<doctype>/helpers/detailed_<doctype>_attributes.html``

This PR suggests to rename them all to ``<doctype>/helpers/view.html`` because:
* the ``<doctype>`` in the actual filename is redundant
* it's more consistent with the general ``helpers/view.html`` helper file
* functions in those files are used in the doctype-related view and preview pages


asaunier:
Thanks for the review.
Please note that it will probably create conflicts in other PR, eg. https://github.com/c2corg/v6_ui/pull/766 (which is already conflicting by the way).

Issue #712, opened 2016-10-08
Parsing of [img] tags


asaunier:
c2corg v5 uses a special [img] tag:
* a filename is never provided
* always based upon the image document_id => this way when the image is modified (title, new file, etc.), it's always up-to-date when embedded in documents.
* it's possible to specify what size of the image to use as well as alignment params

See the tag usage at http://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte#image

The tricky part of this tag is to convert the image id into a real image URL. This is discussed in this comment+following ones: https://github.com/c2corg/v6_ui/issues/131#issuecomment-245639695

Part of issue https://github.com/c2corg/v6_ui/issues/131

asaunier:
My opinion on the approach to use (to be discussed!):

> Advantages of using some kind of image proxy on the API side could be:
* the code would be simpler => probably some sqlalchemy on the API side (+some caching!?) and a quite simple tag conversion in the UI, a bit like https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/format/wikilinks.py
* if the image is updated (new filename but same document id), the UI part is still valid and its cache does not have to be invalidated to take into account the new filename => only the output of the proxy service on the API side has to be refreshed.

arnaud-morvan:
I think proxy service in API is the better way too. I will start working on it ASAP.

arnaud-morvan:
Done for the API proxy service, could use [api_url]/images/proxy/{document_id} to get a redirection to image.
No option for now concerning the resized images, proxy redirect to original image.

asaunier:
> No option for now concerning the resized images, proxy redirect to original image.

That's obviously a problem. We can't ask for the original image when only a thumbnail is embedded in the document :P 

https://github.com/c2corg/v6_api/pull/422#issuecomment-252650951
> no cache for now, not sure this is necessary.

Why?
Document detail views are actually cached, as well the API request to build them. Why not caching the API requests that are used to display an image? If not, whenever someone displays an article with embedded images, API/DB requests will be done to get the image filenames. A bit a pity, don't you think?

tsauerwein:
Views that use the application cache make a query to the database to get the current cache version of a document, for every request.

In this case, only one query is made to fetch a single column. When using the cache, you would first make a query to get the cache version. And if there is no entry in the cache, you would make a second query to get the filename. So, using the application cache is not really beneficial.

But we should check that Varnish caches requests to this view for a small amount of time.

[...]

PR #775, opened 2016-10-22
Add custom img tag

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/712

Specs of the img tag: http://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte#image

TODO
* [x] base parser
* [x] support optional positioning
* [x] support optional size
* [x] support optional caption
* [ ] open image slideshow
* [x] link towards image detail view page? (or maybe not needed since it is available in the image slideshow)
* [x] CSS styling for the generated HTML elements


asaunier:
> link towards image detail view page? (or maybe not needed since it is available in the image slideshow)

@ginold would it be possible that when clicking on an image embedded in a document description (see for instance http://www.camptocamp.org/articles/244529/fr/piolets-cascade-de-glace) an action opens the big image in the usual slideshow?

asaunier:
There a difficult part with converting those ``img`` tags in v6: the proxy URL created by @arnaud-morvan with https://github.com/c2corg/v6_api/pull/424 uses an API URL, so we cannot use a relative path here. But the parser is not aware of the API base URL.

The solution I can think of would be to create an ``app-img`` directive and make the parser transform the ``[img]`` tags to ``<app-img app-img-id="12345" app-img-position="right" app-img-size="big">Image caption</app-img>``.
Then the directive would generate the final HTML code https://github.com/c2corg/v6_ui/blob/84d472512134bbc1a154dbab47bc7b656b542c8c/c2corg_ui/format/img.py#L8-L14 and inject the API URL (angular directives can access the API base URL).

Any better solution?

Perhaps a problem of this solution would be that search engines would most likely don't understand the directive as in image and would not reference the image correctly.

tsauerwein:
> But the parser is not aware of the API base URL.

Initialize the [parser](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/utils/format.py#L10) in [main](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/__init__.py#L13).

asaunier:
Sounds good indeed. Thanks!

ginold:
> ...  opens the big image in the usual slideshow?

surely yes.

asaunier:
> Initialize the parser in main. In the main function you have access to the settings, so you can get the API URL. Then initialize the parser instance with this URL.

@tsauerwein could you precise your suggestion?
How would you initialize the parser instance with the API URL found in settings in ``main()``?
Do you mean I should call ``_get_markdown_parser(settings)`` in ``main()`` and change ``_get_markdown_parser`` so that it pass ``settings.get('api_url')`` to the ``C2CImageExtension`` constructor?

tsauerwein:
Something similar to `configure_caches`:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/__init__.py#L20
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/caching/__init__.py#L38

asaunier:
This PR is not ready yet. Still required:
* [ ] CSS styling for the generated HTML elements
* [ ] action to open the image in the slideshow

In addition there is an unexpected behaviour when using ``xml.etree`` to create the HTML elements (https://docs.python.org/3.4/library/xml.etree.elementtree.html#element-objects): when the ````'s ``src`` (or ``<a>``'s ``href``) is relative => the image tag (or link) is generated as expected, with the correct ``src`` (or ``href``).
But if it is absolute (eg. an API URL), a link HTML tag is generated inside the  ``src`` (or ``href``) attribute!
For instance:
```
<img alt="321640" src="<a href="http://c2corgv6-demo.gis.internal:6545/images/proxy/321640">http://c2corgv6-demo.gis.internal:6545/images/proxy/321640</a>" />
```

I have not found out where this behaviour comes from and how to prevent it...

tsauerwein:
> I have not found out where this behaviour comes from and how to prevent it...

Does the parser replace URLs with links?

asaunier:
> Does the parser replace URLs with links?

I think so. It's probably the reason of this behaviour, thanks for the tip.
So perhaps I should simply make sure this extension is run after the absolute URL "linkification".

asaunier:
> Does the parser replace URLs with links?

The problem was that the BBcode parser was run after the Markdown parser, and it replaces the absolute URLs it finds by ``<a href="">`` tags. I have switched the order and it works as expected: 61fe5cc

Not sure though that the switching as not broken other tags.
See also https://github.com/c2corg/v6_ui/issues/131#issuecomment-256396802

asaunier:

It's quite simpler than the v5 parser (250-line long):
https://github.com/c2corg/camptocamp.org/blob/72e777075f2e6260b1d2ef563ab0f90c383038ba/plugins/sfPunBBCodeParserPlugin/lib/sfPunBBCodeParser.class.php#L521-L777

but at least it handles the most common cases (sizes, left/center/right/inline, caption) and could be improved later (actually I don't even know if we need everything from the v5 parser since most options are not described in http://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte#image).

I have created some simple styles inspired from the ones from v5:
https://github.com/c2corg/camptocamp.org/blob/72e777075f2e6260b1d2ef563ab0f90c383038ba/web/static/sass/_base.scss#L494-L532

As for now the parser does not add the code needed to open the image in a slideshow. I would need help for this and it makes sense to do it in a dedicated PR.

sans titre

c2corg_ui/format/img.py
```diff
@@ -0,0 +1,89 @@
...
+        self.config = {
+            "api_url": ['', 'Base URL of the API. Defaults to ""']
+        }
+
+        super(C2CImageExtension, self).__init__(*args, **kwargs)
+
+    def extendMarkdown(self, md, md_globals):  # noqa
+        self.md = md
+
+        IMG_RE = r'\[img=(\d+)([a-z_ ]*)(/\]|\](.*?)\[/img\])'  # noqa
```
tsauerwein: The regex could be defined on the class so that it is only compiled once.

c2corg_ui/format/img.py
```diff
@@ -0,0 +1,89 @@
...
+        pattern = C2CImage(IMG_RE, self.getConfigs())
+        pattern.md = md
+        # append to end of inline patterns
+        md.inlinePatterns.add('c2cimg', pattern, "<not_strong")
```
tsauerwein: What does this `not_strong` do?

c2corg_ui/format/img.py
asaunier: I have done as in https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/wikilinks.py#L44-L48

but in some other extensions it's done as you suggested: https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/nl2br.py#L24

I will make the change for the current extension and also for the other extension we have added (C2CWikilink).

c2corg_ui/format/img.py
asaunier: ``not_strong`` is a standard python-markdown inline patterns https://github.com/waylan/Python-Markdown/blob/master/markdown/inlinepatterns.py#L79

I have no clue what it does :P 
But this syntax ``"<not_strong"`` says that the current extension must be included/run before the ``not_strong`` pattern.
Actually I have done as in the standard wikilinks extension: https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/wikilinks.py#L51
because this argument is required and this value seems to work fine here as well.

asaunier:
I have made the change about the regexp: e128912

I will merge when Travis passes if no one has objected in the meantine.

Issue #644, opened 2016-09-21
v5 > v6 URLs migration


asaunier:
Redirections from old v5 URLs to new v6 URLs are needed:
- [ ] detail view pages of summits/huts/parkings/sites/products => waypoints + users => profiles (?)
- until v6.x is fully ready, the v5 website will stay available read-only (maybe only the guidebook), for instance "incident reports"
  - [ ] use a special subdomain for v5 pages
  - [ ] switch v5 to read only
  - [ ] make a 301 (permanent) redirection for migrated documents (waypoints, routes, outings etc.). Such v5 pages should no longer be accessible => always redirect to v6.
  - [ ] make a 302 (temporary) redirection (?) to v5-subdomain URLs for not migrated document types
- [ ]  manage static image URLs + forum URLs

asaunier:
This cannot be a "beta" ticket. It involves changes on the actual prod apache config.

asaunier:
Moved to Trello.

PR #776, opened 2016-10-23
Fix profile link in wiki tags

### LINKED ISSUE #774: Wrong URL is used for [users] tags
The `[users]` tags are replaced with a link pointing to `/users/377` instead of `/profiles/377`. For example on http://c2corgv6.demo-camptocamp.com/articles/121965/en/credits

tsauerwein:
Closes #774

### DIFF
--- a/c2corg_ui/format/wikilinks.py
+++ b/c2corg_ui/format/wikilinks.py
@@ -29,9 +29,14 @@ class C2CWikiLinkExtension(Extension):
 class C2CWikiLinks(Pattern):
 
     def handleMatch(self, m):  # noqa
-        if m.group(2).strip():
-            url = m.group(2).strip()
+        url = m.group(2).strip()
+        if url:
             url = '/' + url if url[0] != '/' else url
+
+            if url.startswith('/users/'):
+                # '/users' was renamed to '/profiles' (v5 -> v6)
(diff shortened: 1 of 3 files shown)


asaunier:
You have answered my question at https://github.com/c2corg/v6_ui/issues/774#issuecomment-255578475

To be honest I am not very fond of this approach.
Because it means that we introduce some kind of aliases, like "users" -> "profiles".
(In addition we would need to do the same for old v5 summits, huts, parkings, ..., wikilinks.)

It means more complexity added to the v6.

In addition we have decided to restrict the number of formatting tags (for instance no longer support shortcuts such as ``[c]`` instead of ``[code]`, see https://github.com/c2corg/v6_ui/issues/131 and https://github.com/c2corg/v6_api/issues/417 )

I think only the 2 following options are OK:
* make a (temporary) apache redirection https://github.com/c2corg/v6_ui/issues/644 and have people updating the wikilinks in the document by hand
* convert the wikilinks once for all during the migration, see https://github.com/c2corg/v6_api/issues/417#issuecomment-255579022

tsauerwein:
> convert the wikilinks once for all during the migration, see c2corg/v6_api#417 (comment)

Yes, that would be even better.

Issue #778, opened 2016-10-23
Outing editing


asaunier:
* [x] As for now, the date(s) of the outing and the duration text field are placed in the "weather" section:
<img width="968" alt="capture d ecran 2016-10-23 a 13 01 26" src="https://cloud.githubusercontent.com/assets/1192331/19625733/f2531b02-9920-11e6-8c96-dbe3e6eff36c.png">

That doesn't make sense.

A possible solution would be to move the date(s) and duration text field to the first step and move the "weather" field to the "conditions" section right below (which could be renamed "Weather & conditions" as the step itself).

PR #779, opened 2016-10-23
Remove duration and duration_difficulties fields

asaunier:
Requires https://github.com/c2corg/v6_common/pull/35 and https://github.com/c2corg/v6_api/pull/448

Those fields are somehow duplicates of the ``timing`` text field. They aim to give more precision but:
* complicate the UI
* are in minutes!

PR #781, opened 2016-10-23
Move dates+weather fields within outing editing form

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/778

This PR is based upon PR https://github.com/c2corg/v6_ui/pull/779 (both are a bit independent but the latter removes the "duration" fields, as a result impact quite the same section of the outing editing form). Only the second commit is to be reviewed.

* dates are moved at the top of step 1, before the activities (by the way it makes it more obvious it's an outing report)
* timing is moved right after the description of the trip
* weather is moved in the conditions section, the later being renamed to "weather & conditions"

sans titre
sans titre2

### DIFF
--- a/c2corg_ui/templates/outing/edit.html
+++ b/c2corg_ui/templates/outing/edit.html
@@ -57,8 +57,38 @@ updating_doc = outing_id and outing_lang
           <h2 class="heading show-phone" translate>general informations</h2>
           <div class="content collapse in" id="title-edit">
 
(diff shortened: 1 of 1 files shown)


asaunier:
PR rebased to fix reported conflict. OK for merging?

ginold:
:+1:

asaunier:
Thanks @ginold for the review!

PR #784, opened 2016-10-23
Add a robots.txt file

asaunier:
The proposed ``robots.txt`` file disables all indexing of the site (useful for the demos - of course it should be updated at golive). See http://robots-txt.com/ressources/robots-txt-disallow-all/

According to this page it might be not enough to block the indexing. Using meta robots might be needed: http://robots-txt.com/meta-robots/
Actually it's already the case:
* generic template: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L29
* overridden in specific templates: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L35-L41
If we have to change those values, it means we'll have to make sure to roll the changes back before golive as well.

The ``robots.txt`` file must be made available at the route of the website. Which is not that easy with Pyramid: 
http://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/static.html#static-view
> It can't serve top-level file URLs such as "/robots.txt" and "/favicon.ico".

A solution is to use a server alias:
http://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/pylons/static.html#top-level-file-urls

Note sure the change suggested in this PR works. It seems not in my dev instance => "page not found" (but not using Apache!?)

### DIFF
--- a/c2corg_ui/templates/area/view.html
+++ b/c2corg_ui/templates/area/view.html
@@ -26,7 +26,8 @@ area['doctype'] = 'areas'
   % if version:
     <meta name="robots" content="noindex,follow">
   % else:
-    <meta name="robots" content="index,follow">
+    ## FIXME: set back to "index,follow" before going to prod
+    <meta name="robots" content="noindex,nofollow">
   % endif
 </%block>
 
--- a/apache/wsgi.conf.in
+++ b/apache/wsgi.conf.in
@@ -3,6 +3,8 @@ WSGIDaemonProcess c2corg_ui-{instanceid} python-path={site_packages}
 
 WSGIScriptAlias {base_url} {base_dir}/apache/app-c2corg_ui.wsgi
 
+Alias /robots.txt {base_dir}/c2corg_ui/static/robots.txt
+
 <Location {base_url}>
   WSGIProcessGroup c2corg_ui-{instanceid}
   WSGIApplicationGroup %{{GLOBAL}}
--- a/c2corg_ui/static/robots.txt
+++ b/c2corg_ui/static/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /
--- a/c2corg_ui/templates/article/view.html
+++ b/c2corg_ui/templates/article/view.html
@@ -25,7 +25,8 @@ other_langs, missing_langs = get_lang_lists(article, lang)
   % if version:
     <meta name="robots" content="noindex,follow">
   % else:
-    <meta name="robots" content="index,follow">
+    ## FIXME: set back to "index,follow" before going to prod
+    <meta name="robots" content="noindex,nofollow">
   % endif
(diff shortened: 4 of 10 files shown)


tsauerwein:
> Using meta robots might be needed:  .... Actually it's already the case:

Are you this is configured correctly? On http://c2corgv6.demo-camptocamp.com/ I see:

     <meta name="robots" content="index,follow">

asaunier:
> Are you this is configured correctly?

What is the problem?
As for now the meta is configured so that the website IS indexed by search engines.
Perhaps we could set the homepage to "noindex,follow" by overridding the "metarobots" mako block because there are not much indexable content on the homepage (but that doesn't hur).

tsauerwein:
> What is the problem?
> As for now the meta is configured so that the website IS indexed by search engines.

I thought the point of this PR was to make sure that nothing is indexed?

asaunier:
Indeed.

But in that PR I have only added a ``robots.txt``, hoping it would be enough, because changing all the "robots" meta adds some extra work we'll have to revert at golive (and that's no an on/off switch, the meta is not the same depending on the page):
https://github.com/c2corg/v6_ui/search?utf8=%E2%9C%93&q=metarobots

Perhaps we don't have to change all those "metarobots", only the "index" ones, the "noindex" ones may remain as is. It seems only the ``base.html`` template has an "index" meta:
https://github.com/c2corg/v6_ui/blob/58a61545bb296bde756b3c30e8fd4023ce584edd/c2corg_ui/templates/base.html#L29
So it's much easier to change.

asaunier:
> Perhaps we don't have to change all those "metarobots", only the "index" ones, the "noindex" ones may remain as is. It seems only the base.html template has an "index" meta:

The ``view.html`` templates also have an "index" meta.
Changed at 7f91d79

c2corg_ui/templates/area/view.html
```diff
@@ -26,7 +26,8 @@
   % if version:
     <meta name="robots" content="noindex,follow">
   % else:
-    <meta name="robots" content="index,follow">
+    ## FIXME: set back to "index" before going to prod
+    <meta name="robots" content="noindex,follow">
```
tsauerwein: Shouldn't it be `noindex,nofollow`?

asaunier: Indeed

asaunier:
Updated the "robots" meta to "noindex,nofollow".

asaunier:
I have just updated the PR because there was a path error in the alias definition at https://github.com/c2corg/v6_ui/blob/8e222eb351e294ba0284276c909824c7aa877c57/apache/wsgi.conf.in#L6

Now the path is OK but the alias still doesn't work with pserve, probably only works with apache?

OK for merging?

PR #786, opened 2016-10-24
add feed action type to UI 'added_photos'

ginold:


c2corg_ui/templates/i18n.html
```diff
@@ -376,5 +376,8 @@
 <span translate>has updated the article</span>
 <span translate>has updated the route</span>
 <span translate>has updated the image</span>
+<span translate>has added photos to the route</span>
+<span translate>has added photos to the outing</span>
+<span translate>has added photos to the waypoint</span>
```
tsauerwein: What about other document types images can be added to?

ginold: + regions and articles ? books too ?

tsauerwein: Can't you split up the translation string? E.g. only translate "has created a new", "has updated the" and "has added photos to the"?

ginold: no because it will not work (or we don't know how to) for example with german translations : 

`<span translate>has added photos to the</span><span translate>outing</span>`
...hat Fotos zum Tourenbericht hinzugefügt.

tsauerwein: Ah yes, I remember. But did you try this? https://angular-gettext.rocketeer.be/dev-guide/annotate/#interpolation

asaunier:
Are there still changes to add to this PR?

What about

> regions and articles

and

> But did you try this? https://angular-gettext.rocketeer.be/dev-guide/annotate/#interpolation

?

ginold:
Still need to add the strings for the other documents.

>  interpolation?

I don't get how to use it in this case. 

<span translate>has added photos to the {{outing}}</span> ?
didn't we say that we should not put {{}} into a translate tag because the extractor will extract '{{}}' as a string?

asaunier:
> didn't we say that we should not put {{}} into a translate tag because the extractor will extract '{{}}' as a string?

I don't think so.

The problem was that when writing
```
<span translate>{{some expression}}</span>
```
or
```
<span translate>${'some mako variable'}</span>
```
``{{some expression}}`` or ``${'some mako variable'}`` were extracted and added to the POT files.

But having something like ``has created a {{document_type}}`` added to the translation is OK. Don't know if it works (I think so), but anyway you can try by running by hand the extracting command (https://github.com/c2corg/v6_ui/blob/master/Makefile#L176).

Please consider though that we might have other problems. For instance in French, "has created a ..." would be translated differently depending on the gender of the document type: "a créé UNE (sortie)" vs "a créé UN (itinéraire)".

tsauerwein:
> Please consider though that we might have other problems. For instance in French, "has created a ..." would be translated differently depending on the gender of the document type: "a créé UNE (sortie)" vs "a créé UN (itinéraire)".

Ok, maybe let's go with the [initial idea](https://github.com/c2corg/v6_ui/blob/a8f3e0b16ee89d0b8f0d60f2860241b01fe91a44/c2corg_ui/templates/i18n.html#L368). Thanks for trying though.

It's probably not even that bad considering that we have a limited number of document types and event types.

asaunier:
OK
There is still something that looks strange to me with this PR.

It's supposed to handle the case of image uploads to a document, right?
But then the card should show a subset of the uploaded images. Is it planned? Done in a coming PR?

Don't know yet what happen when clicking on the images/card?! Should we open the document where the images have been uploaded? Or the image itself (its detail view page? in a slideshow?)?
What do you think?

ginold:
To simplify - I would still redirect to the image-view if clicking on one of the images, or to the document it has been uploaded to when clicking on the card.

asaunier:
Both approaches make sense in some ways.
What will be the document type icon on the left of the card?

ginold:
I updated the pr with new strings and images within the feed cards: 
screenshot from 2016-10-26 15-01-00

c2corg_ui/templates/i18n.html
```diff
@@ -369,12 +369,25 @@
 <span translate>has created a new outing</span>
 <span translate>has created a new waypoint</span>
 <span translate>has created a new article</span>
+<span translate>has created a new boom</span>
```
asaunier: boom => book

c2corg_ui/templates/i18n.html
```diff
@@ -369,12 +369,25 @@
...
+
 <span translate>has updated the outing</span>
 <span translate>has updated the waypoint</span>
 <span translate>has updated the article</span>
 <span translate>has updated the route</span>
 <span translate>has updated the image</span>
+<span translate>has updated the book</span>
+<span translate>has updated the area</span>
+
+<span translate>has added photos to the route</span>
```
asaunier: photos => images (btw people may upload drawings and schemas, not only photos)

In addition shouldn't you write "has added photos to route" instead? (no "the")

c2corg_ui/static/js/feed.js
```diff
@@ -176,6 +184,18 @@ app.FeedController.prototype.toggleFilters = function() {
   window.scrollTo(0, 0);
 };
 
+
+/**
+ * @param {string} filename
+ * @param {string} suffix
+ * @return {string | undefined}
```
asaunier: Why "undefined"?
The function at least returns a string starting with ``this.imageUrl_``.

c2corg_ui/templates/i18n.html
ginold: 'no the' ? no idea, but that's what I would say.

tsauerwein: Same for the other the translations: "has updated book XY" is fine.

asaunier:
> I updated the pr with new strings and images within the feed cards: 

Really cool!

So you add the images on top of a wp/route/outing/... card. That seems a smart idea.
What happens if a user has created an outing and then added images to the outing? Will there be 2 cards in the feed (1 for the outing creation + 1 for the image uploading on this outing using the same outing card) or only one?

tsauerwein:
> What happens if a user has created an outing and then added images to the outing? Will there be 2 cards in the feed (1 for the outing creation + 1 for the image uploading on this outing using the same outing card) or only one?

If the same user who created the outing adds images, there will be only one entry. If another user added images, there will be a separate entry.

asaunier:
> If the same user who created the outing adds images, there will be only one entry. If another user added images, there will be a separate entry.

OK, that sounds good to me

c2corg_ui/templates/i18n.html
ginold: "has added photos to route XY" sounds ok for me but "has updated book XY" not anymore :confounded:  haha, hopefully it's not up to me to judge. Do we also have English correctors?

ginold: Maybe not the best way, but when comparing to french it would be '...a ajouté des images à LA sortie", "...a mis à jour LA sortie", "...a créé UNE sortie", most of the time the articles work the same way in English.
TBC.

ginold: already updated. You can re-review.

asaunier:
Minor comment, else it looks good to me, thanks!

What do you think about the size of MI images? Your input is needed here: https://github.com/c2corg/v6_images/issues/6#issuecomment-256309185

ginold:
I think it's good to go. I will merge if no one objects.

asaunier:
I have rebased the branch and fixed the conflict in ``i18n.html``. Will merge the PR when Travis passes, unless you want to do it @ginold

Issue #657, opened 2016-09-27
Advanced search pages in mobile screens


asaunier:
As for now, the advanced search pages are hard to use with a smartphone because there are too many elements for such a small screen and the elements are too small making them barely usable:
- [x] it would be better to show the sections separately:
  - search results (cards) by default with buttons to open the filters and the map
  - map (full screen? see https://github.com/c2corg/v6_ui/issues/428) with a button to close the map and go back to the results
  - filters: all filters at once with a button to show the results or clear the filters (not primary filters only until the user has clicked "all filters")
- [x] horizontal scrolling should be neutralized, especially in the filters section because it conflicts with the use of the horizontal double-sliders (when trying to move the slider's handles, the full section is scrolled horizontally)
- [x] All filters should be displayed on a dedicated line (one filter per line). As for now the primary filters (generally 3) are displayed on the same line which makes it harder to use them (whereas they are the main filters!).
- [x] Activities cannot be selected anymore once you scroll down the list.

stef74:
@asaunier what is the remaining job for beta ?

asaunier:
Nothing has been done yet.

stef74:
In landscape mode advanced search is not usable 
![img_0763](https://cloud.githubusercontent.com/assets/3962997/19640035/3cbf6a0a-99db-11e6-8a1e-b7eebc464c1e.PNG)

stef74:
@ginold as you mentionned in #428 a simple way is on mobile 

> Maybe we could add a switch to hide/show the map and display only the cards by default?

ginold:
@asaunier what do you think of this issue as #799 has been merged?

asaunier:
I think it's better to keep it open until further testing is done by the association.

stef74:
When is avalaible on demo server ?

2016-11-02 9:35 GMT+01:00 Alex Saunier <notifications@github.com>:

asaunier:
No idea. @fjacon @arnaud-morvan ?

stef74:
Great Job !

PR #799, opened 2016-10-26
Advanced search filters improvements

### LINKED ISSUE #692: Problem with scroll down in advanced search pages
I am using Windows 10 and Firefox 49.0.1.

I am on the route search page, I select "VTT" activity, then click on "Tous les filtres":
http://c2corgv6.demo-camptocamp.com/routes#bbox=-857237,3946514,1657236,7253485&act=mountain_biking

The criteria are displayed, but I am unable to go down and click "Voir les résultats": when scrolling down, the page automatically goes up again.

ginold:
probably fixes #692 
related to #655, #657

Some filters improvements:

+ stickyFilters directive and controller
+ when you first open 'more filters' and scroll down the list - it will not jump and become a 'sticky filter'.
+ only when you scroll down the list without first opening the dropdown - it will become a 'sticky filter'
+ solves the problem where 'activities' sometimes became unusable 
+ solves the problem with reaching the list's bottom 'show results'
+ one filter per line on mobile
+ hide the map on mobile by default, let the user toggle it with a button
+ dropdown menus FINALLY don't overflow to the right under he map - they are repositioned (as it was done for image uploader).
+ dropdown menus are not position: fixed - they move with the list as you scroll (more natural)

screenshot from 2016-10-26 11-12-49

This pr only works for routes, but before I add the changes for other documents, I would like to know it what i'm doing is ok.

### DIFF
--- a/c2corg_ui/static/partials/advancedsearch.html
+++ b/c2corg_ui/static/partials/advancedsearch.html
@@ -4,3 +4,5 @@
               ng-mouseleave="searchCtrl.onMouseLeave(doc.document_id)"></app-card>
   </div>
 </div>
+<p translate class="text-warning text-center no-results" ng-class="{'show': searchCtrl.noResults}">No results</p>
+
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -21,9 +21,8 @@ app.imageUploaderDirective = function() {
     },
     templateUrl: '/static/partials/imageuploader.html',
     link: function(scope, el, attrs, ctrl) {
-      // dropdown-menus can overlap the modal and it's impossible to fix it with overflow-y: srcoll + overflow-x: visible.
       el.on('click', '.dropdown-toggle', function() {
-        ctrl.repositionMenu_(this);
+        app.utils.repositionMenu({'menu': this, 'boxBoundEl': '.images-container', 'checkBottom': true});
       });
     }
   };
@@ -419,24 +418,6 @@ app.ImageUploaderController.prototype.openModal = function() {
     controller: 'AppImageUploaderModalController as imageModalCtrl',
     size: 'xl'
   });
-};
-
-
-/**
- * @param {HTMLElement} el
- * @private
- */
-app.ImageUploaderController.prototype.repositionMenu_ = function(el) {
-  var menu = $(el).next();
-  menu.css('opacity', 0); // prevent from jumping on the screen
-  setTimeout(function() { // because at the moment of the click the menu is hidden and result would give 0.
-    var boxBounds = $('#image-uploader')[0].getBoundingClientRect();
-    var menuBounds = menu[0].getBoundingClientRect();
-    if (menuBounds.right > boxBounds.right) {
-      menu.css('left', -Math.abs(menuBounds.right - boxBounds.right + 20));
-    }
-    menu.css('opacity', 1);
-  }, 50);
 };
 
 
--- a/c2corg_ui/static/js/main.js
+++ b/c2corg_ui/static/js/main.js
@@ -40,7 +40,7 @@ goog.require('app.sidemenu');
 goog.require('app.simpleSearchDirective');
 goog.require('app.slideInfoDirective');
 goog.require('app.sliderDirective');
-goog.require('app.stickyFilters');
+goog.require('app.stickyFiltersDirective');
 goog.require('app.suggestionDirective');
 goog.require('app.userDirective');
 goog.require('app.userProfileDirective');
--- a/c2corg_ui/static/js/search/advancedsearch.js
+++ b/c2corg_ui/static/js/search/advancedsearch.js
@@ -95,6 +95,12 @@ app.AdvancedSearchController = function($scope, appApi, ngeoLocation,
   this.documents = [];
 
   /**
+   * @type {boolean}
+   * @export
+   */
+  this.noResults = false;
+
+  /**
    * @type {?number}
    * @export
    */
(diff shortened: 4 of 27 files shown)


asaunier:
Please rebase your PR to upgrade your dependencies ("moment"). It seems the reason why snyk complains.

asaunier:
Is the PR ready for reviewing? Still need a rebase anyway :P

ginold:
rebased. You can start reviewing while i'm working on a few things in this pr.

asaunier:
I see you have made changes to the route filters. Don't you need to apply such changes to other document types?

ginold:
maybe yes. I just moved 2 filters above because the dropdown overflows the bottom of the screen. As i said in the description - i did changes only for the routes for now. I'm waiting for your confirmation to know if what i'm doing is ok.

asaunier:
Don't your changes break the other doc types anyway?

As far as I can see, your PR looks OK to me. I will test your branch in my instance.

ginold:
yes they do, I mean i created a directive that I haven't put in the other templates yet. But if you told me this is completely wrong etc, I would have done some additional work for nothing :P

asaunier:
My concern is about the map filtering. What happens when the map is hidden? I have to test it live.

ginold:
Don't worry - Your doubts would be adequate if the map had `display: none;` - it would in fact not work and the bounding box would never be generated. But i use `opacity: 0;` which means that the map is still virtually there, but transparent, invisible.

asaunier:
@ginold I have tested your branch in my instance and it works really great, including in mobile mode. You rulezzzzzzz :)

Since the map is initially "transparent" on a mobile screen, what about making it use the whole screen when finally displayed? As for now it uses only half or 2 thirds of the screen height. Anyway it's hard to see both the cards and the map at the same time.

screencapture-localhost-6555-routes-1477569719472

ginold:
updated pr with the so long-awaited feature: 

+ dropdown menus FINALLY don't overflow to the right under he map - they are repositioned (as it was done for image uploader).
+ dropdown menus are not position: fixed - they move with the list as you scroll (more natural)

there are some bugs, but they don't appear every time. A deep testing is welcome!

asaunier:
Perhaps I'm getting old but I have found nothing special to complain about in the code. Perhaps you're etting better :)

Anyway I think it's an excellent work!

Concerning the behaviour, I have found a few minor problems (perhaps already there before your changes), but nothing critical. If you cannot find easy and fast fixes, I think it's OK to merge the PR anyway and to list the problems in a new issue:
* in mobile screens the map could be bigger (as discussed above) - not critical anyway
* in mobile screens when there is less than < 30 results, the pagination bar is not shown, if the map is open one can see the cards in the background below the map - not critical anyway
* in mobile screens sometimes I have observed that that "see results" "clear filters" buttons were shifted to the left and not fully visible. But perhaps I had resized the screen in the meantime? I cannot reproduce the problem. See the screenshot below:
screencapture-localhost-6555-waypoints-1477588526266
* when scrolling down the list, the "sticky" buttons appear at the top of the list. If I click on the "all filters" button and if there are too many filters available, it's not possible to scroll down the filters and click on the "show results" or "clear filters" buttons. I think this problem is not related to this PR (see https://github.com/c2corg/v6_ui/issues/692)
* when searching doctypes with no map filter (eg. articles), in mobile screens since the "map" button in the bottom right corner is not displayed, there is a white gap at the right hand of the pagination bar (OK, really minor :P)

ginold:
Updated! Hopefully I fixed all the bugs you mentioned. You can retest.

asaunier:
* [ ] a problem with double-slider filters in the primary filters section:
 * move an handle of a slider to some position
 * scroll the list until the "sticky" buttons appear at the top of the list
 * click on "see all filters"
 * the slider filter that is then shown is not in the expected status (handle not at the right position)
 * if clicking on "view results", the original slider is shown in the expected status
 * no problem with other filter types (text input or checkbox panel)
 * no problem if clicking on "see all filters" when the primary section is shown

c2corg_ui/static/partials/advancedsearch.html
```diff
@@ -4,3 +4,5 @@
               ng-mouseleave="searchCtrl.onMouseLeave(doc.document_id)"></app-card>
   </div>
 </div>
+<p translate class="text-warning text-center no-results" ng-class="{'show': searchCtrl.noResults}">No results</p>
```
asaunier: A dream come true! Thanks :)

asaunier:
I have only tested with Chromium. Looks to work fine except the comment above.
Is it OK with other browsers?
For instance this ``.calc()`` things in the styles? I guess so: https://developer.mozilla.org/fr/docs/Web/CSS/calc#Compatibilité_des_navigateurs

Thanks for enlarging the map. Not a big deal but its size is still a bit weird, especially at the top:
screencapture-localhost-6555-routes-1477652483834

ginold:
>  For instance this .calc() things in the styles?

I hope so. This is just a function that returns 

```
.calc(@prop; @val) {
  @{prop}: calc(~'@{val}');
  @{prop}: -moz-calc(~'@{val}');
  @{prop}: -webkit-calc(~'@{val}');
  @{prop}: -o-calc(~'@{val}');
}
```

So it is adapted to all browsers.

ginold:
>  a problem with double-slider filters in the primary filters section

this is annoying, true. I think the two sliders found in .simple-search and all filters don't sync. There is probably some work to do with the slider-directive, when the slide occurs it should sync across the same type of slider. Maybe for another PR?

asaunier:
> Maybe for another PR? This is not critical if one slider doesn't sync (it's ok for both on page load thanks to the data from url).

OK for me. Please merge and add this problem to issue https://github.com/c2corg/v6_ui/issues/655
Thanks!

Issue #705, opened 2016-10-07
New missing translations in Transifex


shortoland:
The following translations are missing in Transifex:
- [x] Left-side menu: "Community" is not translatable:

![image1](https://cloud.githubusercontent.com/assets/20488365/19180918/b3a4fff6-8c69-11e6-9301-a942d174421f.png)
- [x] WP creation: http://c2corgv6.demo-camptocamp.com/waypoints/add, I select "access" type.

The criteria "Accessibilité en transports en commun" and "Stationnement payant" have common parameters, whereas I think they should both have own and independant ones. For example "seasonal" should be translated by "service saisonnier" in "Accessibilité" and only "saisonnier" in "Stationnement":

![image2](https://cloud.githubusercontent.com/assets/20488365/19181159/687cfab8-8c6b-11e6-9630-f76cafc3b3f9.png)

![image3](https://cloud.githubusercontent.com/assets/20488365/19181161/6aee0418-8c6b-11e6-9ac9-50a39a4bda5a.png)

Same for answer "no".
- [x] WP creation: in case of type "site d'escalade", "sae", "produits locaux" or "station meteo":

Title "Types & styles" is not translated in Transifex. There is another field "Types & styles", but this one is missing.

![image4](https://cloud.githubusercontent.com/assets/20488365/19181334/b49aebac-8c6c-11e6-9cac-b61243b82aa1.png)
- [x] WP, route, outing display (all types), "Commentaires" area:

Field "No thread yet ?" is missing in Transifex

![image6](https://cloud.githubusercontent.com/assets/20488365/19181548/216659be-8c6e-11e6-9f92-4d91deb5486c.png)
- [x] WP display: http://c2corgv6.demo-camptocamp.com/waypoints/726442/fr/pont-de-giers

Field "Associated waypoint_children" is not translatable:

![image7](https://cloud.githubusercontent.com/assets/20488365/19181820/d3eada6e-8c6f-11e6-8e11-ad5fc022da11.png)
- [x] Outing creation (any activity), "Accès & refuge/cabane": parameters for "Fréquentation", "Remontées mécaniques", "Refuge" and "Conditions de l'accès" have been translated in Transifex, but we cannot see those translations in french:

![image10](https://cloud.githubusercontent.com/assets/20488365/19184379/160b28ce-8c7d-11e6-9bbd-e2994ceac5ad.png)
[...]

PR #806, opened 2016-10-27
Fix translations

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/705

Requires https://github.com/c2corg/v6_api/pull/457

### DIFF
--- a/c2corg_ui/static/partials/cards/areas.html
+++ b/c2corg_ui/static/partials/cards/areas.html
@@ -4,7 +4,7 @@
   <br>
   <div class="list-item-info flex baseline-align">
     <div class="quality" ng-if="doc['quality']" ng-class="doc['quality']"
-         uib-tooltip="{{'quality'| translate}} : {{doc['quality'] | translate}}" tooltip-placement="left"></div>
-    <span x-translate>{{doc['area_type']}}</span>
+         uib-tooltip="{{'quality' | translate}} : {{doc['quality'] | translate}}" tooltip-placement="left"></div>
+    {{doc['area_type'] | translate}}
   </div>
 </a>
--- a/c2corg_ui/static/partials/feed.html
+++ b/c2corg_ui/static/partials/feed.html
@@ -3,7 +3,7 @@
   <h3 translate>Activity feed</h3>
 </div>
 
-<h4 translate class="text-center text-danger" ng-show="feedCtrl.error">Sorry, there was an error while getting the feed.</h4>
+<h4 class="text-center text-danger" translate ng-show="feedCtrl.error">Sorry, there was an error while getting the feed.</h4>
 <h4 class="text-success text-center" translate ng-show="feedCtrl.busy">Loading feed, please wait...</h4>
 <h4 class="text-success text-center" translate ng-show="feedCtrl.noFeed">This user has no recent activity.</h4>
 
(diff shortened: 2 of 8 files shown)


c2corg_ui/static/partials/cards/areas.html
```diff
@@ -4,7 +4,7 @@
   <br>
   <div class="list-item-info flex baseline-align">
     <div class="quality" ng-if="doc['quality']" ng-class="doc['quality']"
-         uib-tooltip="{{'quality'| translate}} : {{doc['quality'] | translate}}" tooltip-placement="left"></div>
-    <span x-translate>{{doc['area_type']}}</span>
+         uib-tooltip="{{'quality' | translate}} : {{doc['quality'] | translate}}" tooltip-placement="left"></div>
+    {{doc['area_type'] | translate}}
```
tsauerwein: Doesn't this extract the translation string `doc['area_type']`?

c2corg_ui/static/partials/feed.html
```diff
@@ -16,14 +16,14 @@ <h4 class="text-success text-center" translate ng-show="feedCtrl.noFeed">This us
   <article class="feed-list-item" protected-url-btn url="/outings/add" ng-if="!feedCtrl.userId">
     <div class="timeline-bullet icon-outings" uib-tooltip="{{'outing'| translate}}" tooltip-placement="left"></div>
     <div class="list-item vertical-align">
-      <button class="orange-btn btn">add an outing</button>
+      <button class="orange-btn btn" translate>add an outing</button>
```
tsauerwein: Upper-case?

c2corg_ui/templates/i18n.html
```diff
@@ -378,3 +378,7 @@
 <span translate>has updated the image</span>
 <span translate>Personal feed off</span>
 <span translate>Personal feed on</span>
+
+## misc
+<span translate>see more results</span>
```
tsauerwein: Upper-case?

c2corg_ui/static/partials/feed.html
asaunier: That would make sense.
But this string is already used somewhere else.

asaunier:
I have updated the PR so that ``requirements.txt`` use the correct version of c2corg_common. See https://github.com/c2corg/v6_common/pull/39

Concerning upper-casing some translated strings:
* I agree there are many strings in the code with unconsistant case. It would definitely make sense to review the code and change the case of those strings. I think it should be done in a dedicated issue/PR.
* same problem as usual: the existing translations for those change strings are lost (not completely thanks to Transifex's translation memory but it requires manual actions for all those strings in all languages) => than again a dedicated issue/PR would be relevant to do the changes only once.

c2corg_ui/static/partials/cards/areas.html
asaunier: No, because the extractor is smart enough to understand ``doc['area_type']`` is a variable, not a fixed string.

I have tried to run the extractor on this file like this:
```
node tools/extract-messages.js c2corg_ui/static/partials/cards/areas.html > foo.txt
```
and only ``quality`` is extracted. Which is correct.

asaunier:
OK for merging?

tsauerwein:
:+1:

Issue #804, opened 2016-10-26
Associated images are missing in articles


asaunier:
The API actually returns images in the ``associations`` attribute.
But no associated image is displayed (carousel) in the article detail view pages.

See for instance http://c2corgv6.demo-camptocamp.com/articles/224203/fr/noeud-de-pecheur
and the API response: http://c2corgv6-api.demo-camptocamp.com/articles/224203?l=fr

I think the images are simply not included in the detail view template https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/article/view.html

For instance the following code is missing:
```
  ## IMAGES
  % if not version:
    ${get_image_gallery()}
  % endif
```
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L112-L115

Don't know if it's enough.

PR #807, opened 2016-10-27
Add image HTML code to article page

dkocich:
Close https://github.com/c2corg/v6_ui/issues/804

It looks that HTML was the only part. Associated images are missing because they cannot be retrieved from the server.
"...
`1281991596_1519631005SI.jpg:1 GET http://c2corgv6-images.demo-camptocamp.com/active/1281991596_1519631005SI.jpg 404 (Not Found)`
... etc."

I tried to add my own and they show in the header.

image

I noticed that when I go to "topoguide - images" and I want to get to the detail page of an image I added on an `article/route/waypoint` page, then it does not work. It does not look as an article-related mistake because I tried to do the same with routes/waypoint also on the demo server ( http://c2corgv6.demo-camptocamp.com/ ). I will probably create new issue for it...

### DIFF
--- a/c2corg_ui/templates/article/view.html
+++ b/c2corg_ui/templates/article/view.html
@@ -55,6 +55,12 @@ other_langs, missing_langs = get_lang_lists(article, lang)
     ${show_archive_data('articles', article, locale, version)}
   % endif
 
+
+  ## IMAGES
+  % if not version:
+    ${get_image_gallery()}
+  % endif
+
   <div class="view-details-informations col-xs-12  informations">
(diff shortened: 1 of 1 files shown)


asaunier:
> I noticed that when I go to "topoguide - images" and I want to get to the detail page of an image I added on an article/route/waypoint page, then it does not work. It does not look as an article-related mistake because I tried to do the same with routes/waypoint also on the demo server ( http://c2corgv6.demo-camptocamp.com/ ). I will probably create new issue for it...

Gasp! I have created the issue: https://github.com/c2corg/v6_ui/issues/809
Thanks for noticing.

Issue #803, opened 2016-10-26
Categories in article editing


asaunier:
I have noticed several problems with the ``categories`` in the editing form:
* [x] attribute name is ``categories``, not ``article_categories`` => the field is never set in the editing form and changes are never saved (since not pushed with the right name to the API)
* [x] the field i18n name should be ``categories``, not ``article_categories`` as well (the actual attribute name should generally be used as i18n string, especially because the diff page automatically use the attribute name to legend the diffs). Same thing in helper used for view and preview.
* [x] as for now the field is a single-choice select whereas it should be a multi-choice select (see https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/article.py#L21)
* [x] By the way I notice there is an unexpected ``{{article.article_type}}`` displayed outside of the ``article_type`` select, (to be removed?) See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/article/edit.html#L84

PR #811, opened 2016-10-27
Article categories in the edit form, fix of issue 803

dkocich:
Close https://github.com/c2corg/v6_ui/issues/803

Now I can view an article, edit and save it correctly.

1
image

2
image

3
image

### DIFF
--- a/c2corg_ui/static/js/constants.js
+++ b/c2corg_ui/static/js/constants.js
@@ -30,7 +30,7 @@ app.constants = {
   },
   REQUIRED_FIELDS: {
     'waypoints': ['title' , 'lang', 'waypoint_type', 'elevation', 'longitude', 'latitude'],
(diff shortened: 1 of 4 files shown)


dkocich:
I did not do anything with languages. When I switch to FR it looks ok (but style can be maybe better later in case of many categories)

image

asaunier:
I meant "in the editing form". That's because yesterday I have spent time fixing some fields that were not correctly translated in other editing forms and I wanted to make sure it was already OK for articles :P See https://github.com/c2corg/v6_ui/pull/806

asaunier:
If the translations are OK in the editing form, feel free to merge, thank you very much :+1:

dkocich:
FR, SPA, DE are ok... IT, BA, CA, EU are missing.

image

PR #815, opened 2016-10-28
Fix inconsistencies in config

arnaud-morvan:
Same as https://github.com/c2corg/v6_api/pull/461 but for UI side

### DIFF
--- a/config/default
+++ b/config/default
@@ -17,7 +17,7 @@ export ign_api_key = 9qxhimhvm24i40p8j86n7de3
 export bing_api_key = AudizIfCd3NAdt91ubJMGkMI-swfHxe1R-_U7KiLxCHqepDy70txQ-_-89_eevxc
 
 export ganalytics_key = FIXME
-export recaptcha_site_key = 6LdWkR4TAAAAAJEW4VaASKHU55wP42yYNy7Mbgdb
+export recaptcha_site_key = 6LfWUwoUAAAAAAxud1qqok6wOJJlCUsYXxHizRhc
 
 export logging_level = WARNING
 export skip_captcha_validation = False
--- a/config/default
+++ b/config/default
@@ -21,10 +21,10 @@ export ign_api_key = 9qxhimhvm24i40p8j86n7de3
 export bing_api_key = AudizIfCd3NAdt91ubJMGkMI-swfHxe1R-_U7KiLxCHqepDy70txQ-_-89_eevxc
 
 export ganalytics_key = FIXME
+export recaptcha_site_key = 6LdWkR4TAAAAAJEW4VaASKHU55wP42yYNy7Mbgdb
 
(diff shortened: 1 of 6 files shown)


config/default
```diff
@@ -1,15 +1,11 @@
...
+export debug_port ?= 6553
 
 # the size of the connection pool for http requests (to the API)
 export http_request_connection_pool_size = 200
 
-host = c2corgv6.demo-camptocamp.com
-url = http://$(host)
-
-export api_url = http://c2corgv6-api.demo-camptocamp.com
+export api_url = http://localhost:6543
```
asaunier: Are you sure of this?
``api_url`` is the "external" URL, it cannot be "localhost" by default.
The former value looks more relevant and consistent with the params below.

arnaud-morvan: what default is for you ? dev / main or exoscale

arnaud-morvan: Seem OK to use the main dev api instance as a default

config/default
```diff
@@ -21,11 +17,11 @@ export ign_api_key = 9qxhimhvm24i40p8j86n7de3
 export bing_api_key = AudizIfCd3NAdt91ubJMGkMI-swfHxe1R-_U7KiLxCHqepDy70txQ-_-89_eevxc
 
 export ganalytics_key = FIXME
+export recaptcha_site_key = 6LdWkR4TAAAAAJEW4VaASKHU55wP42yYNy7Mbgdb
```
asaunier: Is it normal that this key is not the same than for the API: https://github.com/c2corg/v6_api/blob/2ce9c6425e6bd85a7d294a9e4776be5e19ad4f5e/config/default#L75

config/dev
```diff
@@ -1,5 +1,11 @@
+instanceid = dev
+
 include Makefile
 
+export base_url = /${instanceid}
```
asaunier: Do we really use such base URLs?

I think we all use ``/`` in our dev instances, ditto for demos. By the way I don't think it would work (there are often relative URLs in the UI such as ``/images/12345``).

What about fixing ``export base_url = /`` once for all in ``config/default``?

arnaud-morvan: This is used by wsgi, it's a common way of all c2c projects to set instance name in path to allow many devs on the same server.

config/default
arnaud-morvan: This is the public key, API hold the private one.

config/dev
arnaud-morvan: In ui dev config, I've set api_url to the make serve one : http://localhost:6543
if you want to use dev wsgi instance in ui dev config set api_url to http://vhost/instanceid/
for the main instance the base url is / so it reachable with http://vhost/

asaunier:
Thanks for the explanation, the PR looks good to me.

Issue #769, opened 2016-10-21
User profile improvements and fixes


asaunier:
- [x] Fix user/profile icon on the map. See https://github.com/c2corg/v6_ui/issues/390#issuecomment-255041415
- [x] Show `forum_username` in user profile when  https://github.com/c2corg/v6_api/pull/447 is merged.

asaunier:
* [ ] Show (guidebook) moderator rights (what about forum moderator rights?)

asaunier:
@ginold do you think you could have a look at 
> Fix user/profile icon on the map

ginold:
Yes I'm working on it right now

stef74:
@ginold what is the status ?

ginold:
not merged yet, but you can see the expected result under https://github.com/c2corg/v6_ui/pull/822

asaunier:
> ￼ Show (guidebook) moderator rights (what about forum moderator rights?)

This part is not done yet because changes in the API are still required (low prio): https://github.com/c2corg/v6_api/issues/463

"forum moderator rights" info is not available (part of Discourse, not in the API).

stef74:
@asaunier it is relevant to close and open dedicted issue for modereator right (not critical for v6.0 ... probably v6.1 or 2). ?

asaunier:
> it is relevant to close and open dedicted issue for modereator right (not critical for v6.0 ... probably v6.1 or 2). ?

👍

PR #822, opened 2016-10-31
User profile icon

ginold:
related to #769
related to #390

As @asaunier  suggested, it's a good idea to have a separate profile.svg icon, following what has been done for other document types.

I extracted the bootstrap's glyphicon user icon: 
screenshot from 2016-10-31 17-43-51

It's better than the other we have because it has no background and can be colored entirely with no empty spaces inside itself. 

The question is: are we going to use this icon for the user profile only? should we replace the other user-icon completely (used in outing-cards for exemple) ?


asaunier:
> I extracted the bootstrap's glyphicon user icon:

The background circle seems a bit too small compared to the user icon (or the latter is too big?), doesn't it? We have to be consistent with the sizes of other map icons.

> The question is: are we going to use this icon for the user profile only? should we replace the other user-icon completely (used in outing-cards for exemple) ?

That would be consistent to have the same icon everywhere.
On the other hand, the current icon is not that different and perhaps it helps figure out the card with such icons are not "profile" cards but "outing" cards?

ginold:
Yes, I can make it fit into the circle.

The advantage of this change is that it is very simple: just added a new svg and that's it, no need for regenerating the fonts and replacing all the icons.

asaunier:
> The advantage of this change is that it is very simple:

What change?

ginold:
> What change?

 "just added a new svg and that's it, no need for regenerating the fonts and replacing all the icons"

asaunier:
OK, go for it.

ginold:
updated - the user-profile icon is smaller and fit's into the circle. And I don't know what i was talking about regenerating the icons - this one is already available with glyphicon-user :p 

I will merge if no objections.

asaunier:
I had a quick look, seems cool, thanks. Please merge.

PR #823, opened 2016-10-31
Map btns alignment

### LINKED ISSUE #801: Style for map tools looks broken (FF)
c2corg-map-tools

It looks better on Chrome.

ginold:
fixes #801

and a new suggestion: i put the map-filter-switch on the bottom and there is always space for all the map's buttons.

screenshot from 2016-10-31 18-52-54

### DIFF
--- a/c2corg_ui/static/partials/map/map.html
+++ b/c2corg_ui/static/partials/map/map.html
@@ -1,18 +1,19 @@
 <div class="map" ngeo-map="::mapCtrl.map">
-  <app-baselayer-selector app-baselayer-selector-map="::mapCtrl.map"></app-baselayer-selector>
-  <div class="map-recenter-tools" ng-show="::mapCtrl.showRecenterTools">
-    <app-geolocation app-geolocation-map="::mapCtrl.map"></app-geolocation>
-    <app-map-search app-map-search-map="::mapCtrl.map"></app-map-search>
+
+  <div class="map-upper-controls">
+    <div class="map-recenter-tools" ng-show="::mapCtrl.showRecenterTools">
+      <app-geolocation app-geolocation-map="::mapCtrl.map"></app-geolocation>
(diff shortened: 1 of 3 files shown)


asaunier:
Interesting!
A small drawback of this change is that the "filter on map extent" option is a bit less visible.
Please also note that the layer names in the base layer selector are longer when actually translated (test when browsing in French for instance).

ginold:
ah those french translations... :P you are right, in french for example it's too long and on smaller screens it slightly overflows the borders, but it's already surely better. I think "limiter la recherche à la carte visible" is a bit too literal. Maybe simply "filter par carte" or something? 

Or i can simply move this filter just above the base-layer selector and the turn is played (le tour est joué).

asaunier:
>  in french for example it's too long

What about making the text wrap on 2 lines?
I agree the translations is a bit too long but it's important that it's clear enough, else users won't understand what it's used for.

In airbnb it's "Rechercher quand je déplace la carte"
https://www.airbnb.fr/s/Lausanne--Suisse?page=1&s_tag=kfMa3CwJ&allow_override%5B%5D=
@shortoland What do you think?

ginold:
wrapping on 2 lines:
screenshot from 2016-11-02 11-19-53

moved up:
screenshot from 2016-11-02 11-21-43


which one do you like more?

asaunier:
I don't have a strong opinion.
The first option is more "discrete" (but is it really what we want?), because more integrated.

My concern is that this tool is not obvious enough.

Would it be possible/complicated to leave it in the top right corner (with a max-width and a text wrap if needed) and if the map is too narrow (conflict with the search input), to make it jump on the next line?

ginold:
will try!

ginold:
PR updated: filter switch on top - goes onto the next line when no space - and wraps even more if really little space.

asaunier:
That's great!! Thanks a lot!
Please merge.

PR #824, opened 2016-11-01
list page with no map fix

### LINKED ISSUE #733: Remove/hide the map container in advanced search page for no-geom document types
When the map filter is disabled in an advanced search page (for instance for articles, areas or books), the map directive is actually disabled but the map container remains => it should be removed (the badge section with the number of results should probably be moved somewhere else).

sans titre

ginold:
screenshot from 2016-11-01 13-26-09

fixes #733

just put the `<div class="documents-list-section-container no-map"></div>` around the content of an index.html file. 

I have also made a small change: the title ("Outings" or "Waypoints" for example) has been moved inside .map-right because it was shown on the map and therefore it was like a part of the map. It is not shown on mobile anymore as it was previously, on the top. You still know what document type you are seeing because of the select box where the doc type is selected.

This was a problem because if there is no map - where should this big title go? It's true that now it's less evident what we're seeing. Maybe a title just above the list would do the trick, but it would be redoundant with the select box.

To be approved.

### DIFF
--- a/c2corg_ui/templates/area/index.html
+++ b/c2corg_ui/templates/area/index.html
@@ -18,15 +18,16 @@
     <span class="badge" ng-cloak>{{resCounter}}</span>
   </h3>
 </header>
+<div class="documents-list-section-container no-map">
+  <section class="documents-list-section">
+    <%include file="filters.html" />
 
-<section class="documents-list-section">
-  <%include file="filters.html" />
-  <div class="documents-nav-buttons text-center">
-    <button class="btn btn-default orange-btn show-documents-filters-phone" translate>Filters</button>
-    ${add_doctype_selector()}
+    <div class="documents-nav-buttons text-center">
+      <button class="btn btn-default orange-btn show-documents-filters-phone" translate>Filters</button>
+      ${add_doctype_selector()}
+      <app-pagination class="areas"></app-pagination>
+    </div>
 
-    <app-pagination class="areas"></app-pagination>
-  </div>
-
-  <app-advanced-search document-type="areas" use-map="false"></app-advanced-search>
-</section>
+    <app-advanced-search document-type="areas" use-map="false"></app-advanced-search>
+  </section>
+</div>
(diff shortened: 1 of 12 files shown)


asaunier:
Could you provide screenshots for both cases (with/without map)?

> This was a problem because if there is no map - where should this big title go? It's true that now it's less evident what we're seeing. Maybe a title just above the list would do the trick, but it would be redoundant with the select box.

We need to be consistent: have the "big title" in both cases. At the same place. It's not very relevant that this "big title" is shown or not whether there is a map or not. The map is a filter as another.

The select box is not obvious enough to figure out what type are the document listed in the page. I think we need the "big title" somewhere anyway.

ginold:
> The select box is not obvious enough

we can still stye the select element, but yeah...

I'm having a hard time deciding where to put the bit main title - as of now, there is no place. 
Maybe we could add a title above the filters, similar to the detail-view pages?

ginold:
with map, as usual: 
screenshot from 2016-11-02 11-42-34

no map, title above the filters:
screenshot from 2016-11-02 11-44-07

asaunier:
Option 2 is OK for me, but should be used also when a map filter is provided.

ginold:
what I meant is that if there is a map - show the title on the map, when not - show it above the filters.

But you want to say that the title should be permanently moved above the filters, whenever there is a map or not?

asaunier:
> But you want to say that the title should be permanently moved above the filters, whenever there is a map or not?

Above filters AND map, yes.
This way the layout is the same, whether there is a map or not.

ginold:
ok, I stick to the version with the main title on the left, always above the filters, with or without map.
The title is fixed so it will be visible at all times.

I updated the pr, you can check it out.

asaunier:
> ok, I stick to the version with the main title on the left,

That's not what I have recommended :P 
I suggested to have the main title above both the filters+results column AND the map column. This way it's more obvious that the map is part of the filters and results system.

ginold:
like this? 
screenshot from 2016-11-02 14-36-02

asaunier:
Genau!

ginold:
i preferred when it was only on the left or right, when it spans all across the width it's too long. 
But you can test it out - pr updated.

asaunier:
That looks excellent to me! Including the full width title.
Please merge, thanks a lot.

PR #827, opened 2016-11-02
mobile vesion of conditions levels table

### LINKED ISSUE #410: Smartphone : problems when creating an outing

ginold:
related to #410

this is a suggestion for fixing the `conditions_levels` table of an outing, which is very difficult to navigate on mobile.

After browsing https://css-tricks.com/responsive-data-table-roundup/, I decided that the easiest way would be to duplicate the table: show one on desktop and another one on mobile only (that's what one of the listed plugin's just did). 

That said, tables are not very suitable for mobiles...:P 

Awaiting your opinion.

screenshot from 2016-11-01 17-29-22

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -380,48 +380,7 @@
           % endif
 
           % if conditions_levels_exist:
-            <div class="table-responsive">
-              <table class="table table-striped conditions-levels">
-                <thead>
-                  <tr>
-                    <th class="location"><span translate>location</span> |
-                      <span translate>altitude</span> |
-                      <span translate>orientations</span>
-                    </th>
-                    <th class="soft-snow" translate>soft snow</th>
-                    <th class="total-snow" translate>total snow</th>
-                    <th class="comment" translate>comment</th>
-                  </tr>
-                </thead>
-                <tbody>
-                  % for condition in cond:
-                  <tr>
-                    ## print empty td for alignement
-                    % if 'level_place' and condition.get('level_place'):
-                      <td>${condition['level_place']}</td>
-                    % else:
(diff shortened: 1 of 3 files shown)


asaunier:
> I decided that the easiest way would be to duplicate the table: show one on desktop and another one on mobile only

I guess it's an acceptable compromise (there are generally not dozens of condition levels lines).

Your proposal looks good. Please note that fields "soft snow" and "total snow" are only numeric, then short => perhaps they could be on the same line?

Anyway it looks good to me.

ginold:
> Please note that fields "soft snow" and "total snow" are only numeric

I'm afraid i have noticed a few cases where they were not numerical but also text, someone wrote something like "20cm blabla mais 30cm si blabla". The reason you can still enter text is because when I parsed those values as 'numeric only', i remember having some errors.

 So...it's maybe more careful to leave it like this.

asaunier:
> I'm afraid i have noticed a few cases where they were not numerical but also text, someone wrote something like "20cm blabla mais 30cm si blabla". The reason you can still enter text is because when I parsed those values as 'numeric only', i remember having some errors.

No big deal: the form has only small inputs for that and the title says "(cm)". It's OK if people write ">20" or "+ de 20" or even "entre 10 et 20 cm" etc.
What I mean is that they are not supposed to write long sentences. So it's OK to have both fields on the same line. If people write tons of text, too bad but that's not really a problem: the text will wrap on several lines.

I do find it more relevant to have both fields on the same line. Unless it brings too much complexity in the code?

ginold:
It was pretty easy:
screenshot from 2016-11-02 15-25-57

how do you like it? can i merge if it's ok ?

asaunier:
Thanks, please merge!

PR #837, opened 2016-11-03
Add books to the UI

dkocich:


asaunier:
Books should not be added to the side menu. They are part of the guidebook section.

They should be added to the document type drop down, at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/list.html#L1-L9
I would recommend to place this doctype between images and areas.

ginold:
... add added to main controller isPath() function.

c2corg_ui/static/js/constants.js
```diff
@@ -31,6 +31,7 @@ app.constants = {
   REQUIRED_FIELDS: {
     'waypoints': ['title' , 'lang', 'waypoint_type', 'elevation', 'longitude', 'latitude'],
     'articles': ['title', 'activities', 'categories', 'article_type'],
+    'books': ['title', 'activities', 'book_types'],
```
ginold: type or types?

asaunier: "types" indeed https://github.com/c2corg/v6_api/blob/books/c2corg_api/models/book.py#L28
(a book may be of several types)

asaunier:
Some changes (API calls, caching, old sitemap views) to take into account when https://github.com/c2corg/v6_ui/pull/844 is merged.

asaunier:
Is this PR ready to be reviewed?

dkocich:
Yes, anyone can review. Thanks. You will probably need to migrate again after the API part was merged...

asaunier:
Looks pretty good, thanks :)

I have noticed a few weird things though:

* [ ] It is possible to associate a waypoint to a book (makes sense) but the book is then not visible in the WP detail view page (OK in the book page => the WP is listed). OK for book-route associations.
* [ ] In book detail view pages, please use ``quality`` as attribute title, not ``document_quality`` (which is a duplicate that should not be used at all). Please also use the model attribute names in the UI as well (else the attribute title won't be translated in the diff pages): "available_langs" => "langs"
* [ ] In cards, there are too many data. Please remove the publication date. In addition you don't need to write "author" in front of the author name. It's quite self explanatory. Put this info right after the book title.
* [ ] Some fonts are weird (too big or bold) in "association" cards (same layout then in advanced search pages should be used): 
capture d ecran 2016-11-16 a 01 54 27

dkocich:
Thanks.

I just noticed I have a problem in the editing form. I was trying to edit "available_langs" (online document language) instead of "langs" (languages of printed books that are available). I will fix it and then merge.

Issue #753, opened 2016-10-19
license information


stef74:
> the info about the licence of the document is shown nowhere.
#449

asaunier:
Are you sure this is really a "beta" issue?
It's rather a "cosmetic" work depending on the document type or some other attribute.

stef74:
it's a go live issue ...

2016-10-19 13:58 GMT+02:00 Alex Saunier <notifications@github.com>:

asaunier:
So the milestone should rather be "go live"?

stef74:
For me go live = beta + critical bug ... but I changed for go live

asaunier:
Complements to this issue:
* depending on the document type (+ some additional attribute for images or articles), the licence of the document should be made obvious in the detail view pages.
* see the CC logo (depends on the licence of the document) at the bottom of http://www.camptocamp.org/routes/211687/fr/mont-de-grange-traversee-bel-air-coicon
* something similar is needed for v6 (logo + text and links toward the licence page). Note sure where to place this part since now comments are in the way. @ginold what do you think?

Reminder:
* WP, routes, books, regions, collaborative articles, collaborative images => licence is CC BY SA https://creativecommons.org/licenses/by-sa/3.0/deed.fr
* outings, profiles, personal articles, personal images, accident reports => licence is CC NY NC ND https://creativecommons.org/licenses/by-nc-nd/3.0/deed.fr

ginold:
>   Note sure where to place this part 

this can be put just before the comments or just after the description text. Or add one row to the 'information block' with 'licence : CC BY SA' and tooltip with the explanation text.

asaunier:
> add one row to the 'information block' with 'licence : CC BY SA' and tooltip with the explanation text.

:+1: 
Please use the CC logo, not "licence : CC BY SA"

ginold:
BY - attribution
SA - share
ND - not commercial
NC - no changes
NY - ??? i only find new york :P

asaunier:
NY => typo, I meant BY

stef74:
@loicperrin @desnoes @shortoland please test

asaunier:
> @loicperrin @desnoes @shortoland please test

problably not on the official demo yet (OK on the old one)

loicperrin:
There is a little problem with tooltip on licence info in some documents: route, WP/access.
When I put the mouse over the icon "ShareAlike", the tooltip flashes. 
[...]

PR #838, opened 2016-11-03
Creative commons

ginold:
related to #753

Adds licence info to the view pages:
screenshot from 2016-11-03 16-29-58

A tooltip will appear when hovering over an icon. Texts copied from https://creativecommons.org


c2corg_ui/static/img/fonts/README.md
```diff
@@ -9,7 +9,7 @@ You can use an online tool such as Icomoon - https://icomoon.io/app/#/select
 5) click download
 6) the zipped file will contain the style.css file as well as fonts folder
 7) put the contents of fonts folder into static/img/fonts and give them a special name (for example "heavy_metal")
-8) update the css, renaming 
+8) copty style.css to the less folder and update it renaming
```
asaunier: copty => copy

c2corg_ui/templates/helpers/view.html
```diff
@@ -6,6 +6,61 @@
     from c2corg_ui.views.document import get_slug
 %>
 
+<%def name="get_licence(doc)">\
+<%
+  doctype = doc['type']
+  CC_ND_NC = False
+  CC_BY_SA = False
+
+  if doctype == 'o' or doctype == 'profile' or doctype == 'xreport':
```
asaunier: Do we really have doctypes ``profile`` or ``xreport``? That does look consistent with the 1-letter doctypes!

c2corg_ui/templates/helpers/view.html
```diff
@@ -6,6 +6,61 @@
     from c2corg_ui.views.document import get_slug
 %>
 
+<%def name="get_licence(doc)">\
+<%
+  doctype = doc['type']
+  CC_ND_NC = False
+  CC_BY_SA = False
```
asaunier: Actually we have cases where images are under "copyright" licence, not creative commons => book covers for instance. See https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L576-L580

asaunier: Instead of ``CC_ND_NC`` (by the way "BY" is missing here) and ``CC_BY_SA`` you should rather figure out if the document is "personal" or "collaborative" (or "copyright" for images).

And then in the HTML code you decide what icons to use depending on this.

c2corg_ui/static/img/fonts/README.md
```diff
@@ -26,6 +26,7 @@ to
     url('fonts/heavy_metal.woff?56tqll') format('woff'),
     url('fonts/heavy_metal.svg?56tqll#icomoon') format('svg');```
 
+9) @import 'heavymetal.less'; in c2corg_ui.less
```
asaunier: What's this used for?
I see no mention to ``heavymetal.less`` anywhere. And why don't you add this ``@import 'heavymetal.less';`` to ``c2corg_ui.less`` directly anyway?

asaunier:
The approach is interesting!
I really like the idea of the tooltip on each clause of the CC licences, explaining what the clause means.

But:
* we need a link towards the full licence page, see https://github.com/c2corg/v6_ui/issues/753#issuecomment-258121900
* some images are not CC but "copyright". For instance http://www.camptocamp.org/images/811920/fr/valais-randonnees-et-plus-si-affinites (see the logo at the bottom of the page)

c2corg_ui/templates/helpers/view.html
ginold: I havent checked what the abbreviation for profiles and reports was :(

ginold:
The link to the page should be in the language of the interface. Something to keep in mind.

ginold:
what text should go with copyright ? the one from v5? :

"This picture depicts a book cover. It is the property of its editor and/or author. It is presented here only for illustration purposes.
The images associated to this page are available under the license specified in the original document of each image."

there is "this picture" "cover book" "this article" and maybe more. Can we have something more generic for all copyrights?

asaunier:
"copyright" only applies to images. And actually only to book covers. That's an exception (btw only moderators may select such an image type). Else only CC-compatible contents are allowed in the guidebook/wiki.

I would simply use "This picture depicts a book cover. It is the property of its editor and/or author. It is presented here only for illustration purposes".

ginold:
Updated:

+ checking if document is of type personal/collaborative/copyright
+ the icons forward to the creative-commons website in the language of the interface.
+ different icons for creative-commons and copyright 

see for example : 

copyright: /images/722245/fr/
collaborative: /routes/617373/fr/
personal: outings/665064/fr/

the only question I have: what is the 1-letter abbreviation for 'xreports'?

asaunier:
> what is the 1-letter abbreviation for 'xreports'?

Surprisingly it's "x" :P
https://github.com/c2corg/v6_common/blob/reports/c2corg_common/document_types.py#L7

asaunier:
> what is the 1-letter abbreviation for 'xreports'?

Surprisingly it's "x" :P
https://github.com/c2corg/v6_common/blob/reports/c2corg_common/document_types.py#L7

asaunier:
> what is the 1-letter abbreviation for 'xreports'?

Surprisingly it's "x" :P
https://github.com/c2corg/v6_common/blob/reports/c2corg_common/document_types.py#L7

ginold:
don't know why it hid from my eyes :P 

Are really all books collaborative as you mentioned in #753 ? that feels counter intuitive.

asaunier:
> Are really all books collaborative as you mentioned in #753 ? that feels counter intuitive.

That's not the books that are collaborative, that's the pages that describe them.
Ditto for WP and routes: the actual summit is of course not collaborative, but the page on c2corg is :)

Will test your branch shortly.

asaunier:
This looks like an excellent work! I really like the way you have used the icons and the tooltips!
The code looks OK to me.

Just one thing: in my instance the lang suffix of the CC URLs is missing!?

ginold:
welcome to the next episode of "it works only in debug, i don't know why".

asaunier:
If it works in debug and not in build mode, it's probably a closure issue. Renamed property because not exported? Hmm, no, ``lang`` is actually exported.
What about setting ``lang`` with the actual current lang, not with ``appLang``?

ginold:
^ it's hard to test...maybe a cache issue? I mean in debug and non debug mode the current language is correctly console.logged, it just don't appear in the template even after a make build and clear-cache. But even if I add some random things to the template, like a big H1 title in the middle, it won't appear.

ginold:
appLang uses gettextcatalog.currentLanguage, so it's the same.

tsauerwein:
You are calling `getLang`? This method should be exported: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/langservice.js#L95

ginold:
the function in `appLang` itself should be exported? isn't it enough to export the the details-controller `lang` property?

```
  /**
   * @type {app.Lang}
   * @export
   */
  this.lang = appLang;
```

and then call

`ng-href="https://creativecommons.org/licenses/by-nc-nd/3.0/deed.{{detailsCtrl.lang.getLang()}}"`

asaunier:
``lang`` is not renamed but ``getLang`` is, since it's not tagged as exported.

ginold:
OK, I think it was it! thanks. You can recheck and i will merge. It should be working now.

asaunier:
Do you really need a separated commit for exporting ``getLang()``?

asaunier:
I can't test because my API is currently broken (WIP) but if now the links are OK in prod mode, than please merge (perhaps squash some commits before, up to you).

Issue #834, opened 2016-11-03
Add the geonames search tool in editing maps


asaunier:
In order to ease the recentering on a location near where the WP or route is located.

PR #841, opened 2016-11-03
Activate map recentering tools in editing forms

asaunier:
Fixes https://github.com/c2corg/v6_ui/issues/834

This change is quite simple. A small downside is that the geonames tool comes with the "recenter on my position" button, which is probably not really relevant here (?).

capture d ecran 2016-11-03 a 22 28 04

### DIFF
--- a/c2corg_ui/templates/image/edit.html
+++ b/c2corg_ui/templates/image/edit.html
@@ -165,7 +165,7 @@ from c2corg_common.attributes import default_langs, activities, quality_types, i
               </div>
             </div>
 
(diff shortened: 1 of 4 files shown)


ginold:
>   small downside is that the geonames tool comes with the "recenter on my position" button, which is probably not really relevant here 

Could be nice if you are a c2c addict and you and you want to immediately create a new doc from the place you are at :P All in all, not a big deal, better have it than not in my opinion.

asaunier:
Thanks @ginold and @tsauerwein for the feedback. I fully agree with you.
And yes, people might get so addicted they will create WP "in situ" with their smartphones!

Issue #742, opened 2016-10-17
Incorrect criteria when we select 6b+, P1+, A0+, ...


fbunoz:
When we select a criteria which contain a '+', it is not taken account.
For exemple, I select the free climbing rating range between 6a+ and 6b+ : the number of results doesn't change about all grades.
http://c2corgv6.demo-camptocamp.com/routes#bbox=-607746,3535588,1407745,7664411&act=rock_climbing&rrat=6a%20,6b%20&l=fr

If I choose a range between 6a and 6b+, only 6a is taken account.
Ther is the same problem with all ratings whose some grades contin a '+'.

asaunier:
It seems that the "+" symbol is understood as the space " " symbol (which is urlencoded "%20").

asaunier:
I guess we should URL-encode the values of the criterias when adding them to the URL
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/encodeURIComponent

Perhaps here: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/searchfilters.js#L175

@tsauerwein would it make sense to do it in ``ngeo.Location``?
https://github.com/camptocamp/ngeo/blob/master/src/services/location.js

stef74:
@asaunier for me this point is key for beta ...

tsauerwein:
> @tsauerwein would it make sense to do it in ngeo.Location?

PR is here: https://github.com/camptocamp/ngeo/pull/2002

asaunier:
Argh, I was just about to make a PR :P 
But in v6_ui. It's probably better to do it in ngeo (if it's OK to upgrade our ngeo/openlayers).

PR #848, opened 2016-11-04
Urlencode filters values when added into the address bar

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/742

PR created just for the records (it was ready) since a change in ngeo is pending: https://github.com/camptocamp/ngeo/pull/2002

### DIFF
--- a/c2corg_ui/static/js/search/searchfilters.js
+++ b/c2corg_ui/static/js/search/searchfilters.js
@@ -172,7 +172,14 @@ app.SearchFiltersController.prototype.getFilterFromPermalink = function(key) {
 app.SearchFiltersController.prototype.handleFiltersChange_ = function() {
   // ignore the initial $watchCollection triggering (at loading time)
   if (!this.loading_) {
-    this.location.updateFragmentParams(this.filters);
+    var filters = {};
+    for (var key in this.filters) {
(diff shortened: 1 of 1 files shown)


tsauerwein:
If you want, we can use this PR as long as ngeo is not updated.

asaunier:
It depends how long it will take before your PR is merged and if we have problems upgrading ngeo/OL in v6_ui.

That could be a temporary fix, yes, especially since the beta probably needs a fast fix.

asaunier:
@tsauerwein Your PR in ngeo has been approved (question from @gberaudo though?). Do you think it will be merged soon?
That we could upgrade our ngeo/OL soon?

Just trying to figure out it's better whether to merge my current PR now and revert it when your PR is merged + ngeo/OL are upgraded or to wait a bit?
If it can be done today it's worth waiting. Else I would prefer to merge my PR so that the demo has at least a temporary fix.

tsauerwein:
> @tsauerwein Your PR in ngeo has been approved (question from @gberaudo though?). Do you think it will be merged soon?
> That we could upgrade our ngeo/OL soon?

I will merge the ngeo PR once Travis has passed and can then look into updating ol/ngeo.

asaunier:
No longer needed since done in ngeo.

PR #854, opened 2016-11-05
Support of tags "important" and "warning"

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/131

Overview:
capture d ecran 2016-11-05 a 17 09 03

In v5: http://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte#paragraph


stef74:
:smile:

PR #873, opened 2016-11-07
Fix color of the quality circle

dkocich:
I noticed that there is still an issue after renaming of the quality 'good' to 'fine'. Should change black circle to green now.


asaunier:
Arf, thanks for noticing :)
Please merge when Travis is OK.

PR #883, opened 2016-11-08
fix quality translations in modal

### LINKED ISSUE #880: I18n is broken in the quality modal box when editing a document
Perhaps a modal template compiling problem (wrong scope?). See the preview modal for instance.

Please note that the attribute name is "quality", not "document_quality".

@ginold Always test in French, this way you can see when i18n is broken (you can't if browsing in english).

ginold:
fixes  #880

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -50,9 +50,11 @@
       </div>
       % endif
       ## QUALITY
-      <div class="full comment" ng-class="{'has-success': saveCtrl.quality}" ng-init="qualityTypes = ${quality_types}">
-        <label translate>document_quality</label>
-        <select ng-options="type for type in qualityTypes" ng-model="saveCtrl.quality" class="form-control"><option value=""></option></select>
+      <div class="full comment" ng-class="{'has-success': saveCtrl.quality}">
+        <label translate>quality</label>
+        <select ng-model="saveCtrl.quality" ng-options="type as saveCtrl.langService.translate(type) for type in ${quality_types}" class="form-control">
+          <option value=""></option>
+        </select>
       </div>
 
       <div class="text-center">
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -600,10 +600,17 @@ app.module.controller('appDocumentEditingController', app.DocumentEditingControl
  * We have to use a secondary controller for the modal so that we can inject
  * uibModalInstance which is not available from the first level controller.
  * @param {Object} $uibModalInstance modal from angular bootstrap
+ * @param {app.Lang} appLang Lang service.
  * @constructor
  * @ngInject
  */
-app.ConfirmSaveController = function($uibModalInstance, appDocument) {
+app.ConfirmSaveController = function($uibModalInstance, appDocument, appLang) {
+
+  /**
+   * @export
+   * @type {app.Lang}
+   */
+  this.langService = appLang;
 
   /**
    * @type {Object} $uibModalInstance angular bootstrap
@@ -622,6 +629,7 @@ app.ConfirmSaveController = function($uibModalInstance, appDocument) {
    * @export
    */
   this.quality = appDocument.document.quality;
+
 };
 
 
--- a/c2corg_ui/static/js/langservice.js
+++ b/c2corg_ui/static/js/langservice.js
@@ -99,6 +99,16 @@ app.Lang.prototype.getLang = function() {
 
 
 /**
+ * @export
+ * @param {string} str
+ * @return {string}
+ */
+app.Lang.prototype.translate = function(str) {
+  return this.gettextCatalog_.getString(str);
+};
+
+
+/**
  * @param {string} lang
  * @param {boolean=} opt_syncWithApi
  */


c2corg_ui/templates/helpers/edit.html
```diff
@@ -50,9 +50,14 @@ <h3 class="heading show-phone" translate>summary</h3>
...
       % endif
       ## QUALITY
-      <div class="full comment" ng-class="{'has-success': saveCtrl.quality}" ng-init="qualityTypes = ${quality_types}">
-        <label translate>document_quality</label>
-        <select ng-options="type for type in qualityTypes" ng-model="saveCtrl.quality" class="form-control"><option value=""></option></select>
+      <div class="full comment" ng-class="{'has-success': saveCtrl.quality}">
+        <label translate>quality</label>
+        <select ng-model="saveCtrl.quality" class="form-control">
+          <option value=""></option>
+          % for q in quality_types:
```
ginold: it didn't work with `ng-options mainCtrl.translate(type)`

asaunier: Have you tried to change the scope when compiling the modal template? I would prefer we use the usual way to build and translate a dropdown.

asaunier: I had a similar problem (I think!?) with the preview modal where translations where not done as expected (ditto for dates converting with moment). It seems I have finally found another solution but I remember that @tsauerwein gave me a relevant tip, suggesting to use the parent scope or something like that, else ``mainCtrl`` was not available.

ginold: by changing the scope you mean create a `scope.$new(true)` ? yes i tried. This doesn't work either: `ng-options="type as mainCtrl.translate(type) for type in ${quality_types}"` nor with `ng-init`.

I admit I don't know how to fix it. If @tsauerwein  has an idea, then great.

Should I merge as is (because at least it works), awaiting a possible solution ?

asaunier: > by changing the scope you mean create a scope.$new(true)

No clue. Perhaps I am wrong. But I have the feeling it should be possible to make the usual way work in your modal box.

Have you tried ``$scope.$parent`` or ``$rootScope`` or... ?
It seems that ``mainCtrl`` is not available in the scope provided to the template.

> Should I merge as is (because at least it works), awaiting a possible solution ?

There is no hurry.

tsauerwein: Is `saveCtrl` available?  E.g. is the quality field shown correctly when editing an existing document with the field set?

ginold:
so I changed the solution a little and it works - I added `appLang`  to `saveController` and created a `translate(str)` function inside `appLang`.

asaunier:
That's still a workaround!
And why adding the ``addLang`` service? You could directly inject ``gettextCatalog`` and use its ``getString`` method.
I prefered your initial proposal with the mako loop and the translate directive. It was not consistent with the way other translated dropdown were created but at least it was simple and introduced no more complexity in the code.

If we cannot find an easy solution we will do so. But have you tried to replace
```
template: this.compile_(template)(this.scope),
```
by
```
template: this.compile_(template)(this.scope.$parent),
```
in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/documentediting.js#L580-L584 ?

I have tried. It does not work yet: the quality values are all replaced by "undefined":
capture d ecran 2016-11-16 a 00 45 40

But at least ``mainCtrl`` is actually available into ``this.scope.$parent`` whereas it's not in ``this.scope``. ``saveCtrl`` is not available in both of them.

If the dropdown is built using
```
<select ng-model="saveCtrl.quality" ng-options="type for type in ${quality_types}" class="form-control">
```
the quality values are displayed correctly, but remain untranslated.
If using
```
<select ng-model="saveCtrl.quality" ng-options="type as mainCtrl.translate(type) for type in ${quality_types}" class="form-control">
```
"undefined" is displayed. I notice that the ``translate`` method of the main controller seems to be  never called here.

I have the feeling there is a small error somewhere! @tsauerwein do you have an idea?

ginold:
I have tested all what you said, in fact, and have the same results as you. 
I hope that @tsauerwein comes with other answer than '42'.

asaunier:
@ginold We can't spend much more time on this issue. Please use the initial work around you have suggested (the one with the mako loop). At least it was simple.
Thanks!

tsauerwein:
> And why adding the addLang service? You could directly inject gettextCatalog and use its getString method.

Sorry, I didn't have the time to look into it. But injecting the `gettextCatalog` sounds like a valid solution. It would be more consistent with the rest of the code than using the Mako loop IMHO.

ginold:
:confused: please vote for the best solution :P

asaunier:
For me both solutions are not consistent :) I have not a strong preference.

But OK to follow @tsauerwein's recommendation. 
Actually OK for merging the PR as is.

ginold:
>  Actually OK for merging the PR as is.

you sure? ok with `appLang` + `translate` method?

asaunier:
Yes :)
As I said it's a workaround anyway. Using ``gettextCatalog`` in ``app.ConfirmSaveController`` instead of ``appLang`` would make sense but then in the mako template we would write 
```
ng-options="type as saveCtrl.gettextCatalog.getString(type) for type in ${quality_types}"
```
instead of
```
ng-options="type as saveCtrl.langService.translate(type) for type in ${quality_types}"
```
but finally that's not more consistent compared to the expected
```
ng-options="type as mainCtrl.translate(type) for type in ${quality_types}"
```
and having a ``translate`` function in the ``appLang`` service is not silly.

Let's get rid of this issue and merge the PR :)

PR #885, opened 2016-11-08
fix wrong menu-selected

### LINKED ISSUE #876: Incorrect menu highlight in the homepage
Menu entry "topoguide" should not be hilighted in the homepage:
capture d ecran 2016-11-07 a 21 48 58

ginold:
fixes #876

i'm not sure what happened, but i just moved the code around and it fixed it :confused:

### DIFF
--- a/c2corg_ui/static/js/maincontroller.js
+++ b/c2corg_ui/static/js/maincontroller.js
@@ -76,11 +76,10 @@ app.MainController.prototype.isPath = function(path) {
   // path = '/'
   if (path === location) {
     return 'home';
-  // if topoguide, it can be all kinds of documents
+  // if topoguide, it can be all kinds of documents. Articles have their own line in the sidemenu.
   } else if (path === 'topoguide') {
-    return location.indexOf('waypoints') > -1 || location.indexOf('routes') > -1 || location.indexOf('outings') > -1
-            || location.indexOf('areas') > -1 || location.indexOf('images') > -1 || location.indexOf('books')
-            || location.indexOf('articles') > -1;
+    return location.indexOf('outings') > -1 || location.indexOf('waypoints') > -1 || location.indexOf('routes') > -1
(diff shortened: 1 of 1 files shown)


c2corg_ui/static/js/maincontroller.js
```diff
@@ -78,9 +78,9 @@ app.MainController.prototype.isPath = function(path) {
     return 'home';
   // if topoguide, it can be all kinds of documents
   } else if (path === 'topoguide') {
-    return location.indexOf('waypoints') > -1 || location.indexOf('routes') > -1 || location.indexOf('outings') > -1
-            || location.indexOf('areas') > -1 || location.indexOf('images') > -1 || location.indexOf('books')
```
asaunier: Was this ``location.indexOf('books')`` (missing ``> -1``) the reason of the problem?
I am a bit in doubt that changing the order of the tests actually fixes it?!

ginold: oh maybe, I didn't spot the missing -1 !

I really don't find those lines too long. I mean...look at our html templates...:P

asaunier: The template lines ARE too long!

ginold: oh yeah they are! next to them, those are little tiny perfect lines.
Anyway - ok for merging?

asaunier: Changing the tests order is not relevant to me when the only required change was to add the missing ``> -1``.
I admit it is probably peanuts but it's better to test the most usual document first (waypoints, routes, outings). Book pages are more rarely used.

ginold: yep you're right. I will change it and merge afterwards.

ginold: I tried with another, probably clearer version : 

```
    var paths = ['/outings', '/waypoints', '/routes', '/images', '/areas', '/books'];
    return $.inArray(location, paths) > -1;
```

do you find it ok?

PR #888, opened 2016-11-08
correct display of article categories in information box

gottferdom:
In an article, categories are displayed like activities. I don't think that's the expected behaviour, so this PR.
I'm juste getting started with developping with v6, please be kindly with me :sweat_smile:

### DIFF
--- a/c2corg_ui/templates/article/helpers/view.html
+++ b/c2corg_ui/templates/article/helpers/view.html
@@ -24,7 +24,7 @@
         % for type in article.get('categories') :
           <span x-translate class="value">${type}</span>${'' if loop.last else ', '}
         % endfor
-      </p>
+      </article>
     % endif
 
     % if article.get('article_type'):
--- a/c2corg_ui/templates/article/helpers/view.html
+++ b/c2corg_ui/templates/article/helpers/view.html
@@ -19,12 +19,11 @@
     % endif
 
     % if article.get('categories'):
-      <article class="value activities">
-          <span translate class="value-title">categories</span><br>
+      <p><span class="value-title" translate>categories</span>:
         % for type in article.get('categories') :
           <span x-translate class="value">${type}</span>${'' if loop.last else ', '}
         % endfor
-        </article>
+      </p>
     % endif
 
     % if article.get('article_type'):
--- a/c2corg_ui/templates/article/helpers/view.html
+++ b/c2corg_ui/templates/article/helpers/view.html
@@ -19,7 +19,8 @@
     % endif
 
     % if article.get('categories'):
-      <p><span class="value-title" translate>categories</span>:
+      <article>
+        <span class="value-title" translate>categories</span>:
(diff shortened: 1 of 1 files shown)


gottferdom:
@asaunier thinks I need to be more explicit. So let it be: in the articles left information box, article categories were formatted the same way activities are. I changed this to be consistent with other fields such as article type. Although there may be several categories, I don't think it was appropriately formatted.

asaunier:
@ginold @dkocich could you review please?

c2corg_ui/templates/article/helpers/view.html
```diff
@@ -19,12 +19,11 @@
     % endif
 
     % if article.get('categories'):
-      <article class="value activities">
-          <span translate class="value-title">categories</span><br>
+      <p><span class="value-title" translate>categories</span>:
```
ginold: I would write like this, similar to what you can find in routes:

```
    % if article.get('categories'):
      <article>
        <span translate class="value-title">categories</span>:
        % for cat in article.get('categories') :
        <span x-translate class="value">${cat}</span>${'' if loop.last else ', '}
        % endfor
      </article>
    % endif

```

gottferdom:
OK sure. As I'm not used to v6_ui yet, I didn't think it would be different. Let's do it your way :smile:

By the way, why use an `article` tag? Just curious, but the semantics is strange IMHO.

ginold:
very probably it is not *the best* element that could have been chosen, but in order to stay coherent with the existing code we may want to reuse it. Why not creating an issue and rewrite some of the elements that are not in the good place? maybe we will have some time next year :confused: 

article : "element represents a component of a page that consists of a self-contained composition in a document" - If we really want and believe in it - it could pass in this case too :P ....

other than that :ok:

besson-genostar:
Yeah, it's not a top priority right now :innocent: 
Don't spend to much time on this trivial PR (it mas mostly for testing purposes). If all is OK, let's merge!

asaunier:
The diff is a bit weird though. There's an opening ``<article>`` tag but it's "closed" by a ``</p>``?
That seems not OK for merging...

besson-genostar:
Oops, my bad!

gottferdom:
(I shouldn't intervene on c2c with my work account, sorry)

asaunier:
Is it possible to squash the commits (3 commits for that change is a bit too much :P)?

gottferdom:
Sure, you pedantic dev!
The one merging PR can squash them, isn't it? (I assume it is not my role - at least not yet, while you guys are working heavily on the release - to merge PRs)

tsauerwein:
@asaunier When merging you can select "Squash and merge".

asaunier:
Pedantic is my lastname.

> The one merging PR can squash them, isn't it? 
> When merging you can select "Squash and merge".

OK, then @gottferdom feel free to merge.

PR #892, opened 2016-11-09
CSS code cleaning

dkocich:
I sometimes look into the results of the code inspection and there are some smaller changes in the CSS.
- It has no meaning to specify units for zero values
- I deleted some duplicate properties (I was not sure in one case because both of them had different value)

Maybe it could be useful if everyone can also look later if there are some possible issues that can be fixed to save some time later (my inspection shows me around 150 warnings).

### DIFF
--- a/less/c2corg_ui.less
+++ b/less/c2corg_ui.less
@@ -268,7 +268,7 @@ app-pagination.areas, app-pagination.articles {
         background-color: @bright-green;
       }
       .glyphicon {
-        top: 0px;
+        top: 0;
         margin-left: -5px;
       }
       a,  .glyphicon {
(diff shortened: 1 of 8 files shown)


ginold:
>  my inspection shows me around 150 warnings

...and it doesn't show 300 others :P i'm sorry!

asaunier:
Thanks @dkocich 
What tool did you use to perform such an inspection?
I'm curious if we have also a tool to check the HTML (and detect mis-closed tags etc.).

Since the changes are only unit removals for "0" values and fixes of duplicated styles, it looks safe (no visible effect) to merge the PR.
@ginold OK for you?

ginold:
:+1:

PR #902, opened 2016-11-11
Add health web-service

tsauerwein:
See also https://github.com/c2corg/v6_api/pull/499

Example:

```
{
	"version": "458187d",
	"redis": "ok",
	"api": "ok",
	"redis_keys": 33,
	"api_status": {
		"pg": "ok",
		"es_indexed_docs": 0,
		"es": "ok",
		"redis_keys": 13025,
		"last_es_syncer_run": "2016-11-11T12:27:52.023859+01:00",
		"redis": "ok",
		"version": "15fbece"
	}
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,15 @@ Captcha configuration is done through https://www.google.com/recaptcha/admin .
 The public key is used in the UI. The corresponding secret key is used in the API.
 
 
+Health service
+--------------
+
+To get information about a running UI instance, e.g. its version and the
+status of its components, the health service `/health` can be used:
+
+    GET http://localhost:6543/health
+
+
 Developer Tips
 --------------
 
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -140,6 +140,9 @@ def main(global_config, **settings):
     config.add_route('sitemap_index', '/sitemap.xml')
     config.add_route('sitemap', '/sitemaps/{doc_type:[a-z]{1}}/{i:\d+}.xml')
 
+    # health services
+    config.add_route('health', '/health')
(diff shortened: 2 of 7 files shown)


asaunier:
How would you see this ``health`` service used in the UI?
Having the version displayed somewhere in the pages would make sense, but shouldn't this info be integrated in mako templates (eg. in ``base.html``) and cached? Then do we need a UI webservice?

tsauerwein:
The aim of this PR was not to show a version in the pages. This health web-service is for developers to verify the deployed version and to get information about the status. Also it can be used for monitoring purposes.

asaunier:
OK

But when @mfournier suggested to get the current version
> Which returns the current revision (proposed by @mfournier).

I think he considered showing the current version in the pages, so that tested may know the revison of the demo.

So I am just wondering if the ``health`` services could be used for that purpose.

Anyway OK for merging, thanks!

tsauerwein:
Ok, I pushed a commit to show the version in the sidebar. The links points to Github.

c2corg-version

asaunier:
OK, I understand, thanks.

I wonder if this info could not be placed somewhere else. It's useless for most visitors and the current position uses space in the sidemenu and will make the ad banner disappear for even smaller screens (when there is not enough space to show the ad banner it is removed).

tsauerwein:
I'll let you figure this out. Feel free to drop/adapt the last commit.

asaunier:
Please merge as is, I don't really have time/inspiration/skills to propose something else :P

tsauerwein:
I reduced the alpha value of the color a bit, to make it less prominent:

c2corg-version

PR #905, opened 2016-11-14
Add reports/accidents to the UI

dkocich:
There are still going to be some changes...

EDIT:
Current problems I know about

- some problems with dates
- [x] readable format parsing problems (show 15th November instead of 2016-11-15)

- [x] when creating a report, none of the personal data is saved to DB (is it API problem? I am not sure now)
- [x] function to show associated waypoints on the report page has an error during the creation of SEO links?
- [x] Edit form - date field has to be clicked and some date set before it is possible to apply other changes 

- [ ] From-To date calendar in advanced filter does not work
- [x] SVG icon is not visible on header map in report/view page (text is visible and sometimes disappear when I hover over it)
- [ ] I am not sure that access rights really works correctly - I tried to edit my own document and I can't get already existing personal data to be visible in the new edit form... but maybe it is problem with restarting of my localhost, cache keys, etc.

- [ ] fix `XreportEditingController` as mentioned [in comment below](https://github.com/c2corg/v6_ui/pull/905#discussion_r90652076)


dkocich:
Hello @asaunier @ginold ... maybe you can slowly start with the review when you will have some time

asaunier:
What are - in short - the points still to be addressed with this PR?

dkocich:
It should be ok and stable now. Some leftovers can be fixed separately maybe if nobody would notice any obvious fixes during the review.

asaunier:
I will review the PR when I have time if no one else has done it in the meantime, thanks :)
Is the checklist in the initial comment up-to-date?

asaunier:
Is there an API (or database) available with reports data that could be used for testing the UI?

dkocich:
Thanks. The checklist is up to date and DB is on the remote server - Tobias did test PG migration to his instance yesterday and I filled ES index without problems.

asaunier:
What do you call the remote server? The "demov6" API?

tsauerwein:
> What do you call the remote server? The "demov6" API?

It's the database 'c2corg_tsauerwein' on the dev server.

asaunier:
I have noticed that in v6, both in the UI and in the API, the module/view name is "reports", whereas it was "xreports" in v5. It's a bit a pity because:
* "reports" is a bit too generic/confusing (outings are trip reports)
* changing the name will require to make redirections from v5 to v6 URLs since they are no longer the same. We will also need to convert "xreports" wikilinks (see https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/document.py#L234-L238)

Honestly I find this renaming not very relevant. Couldn't we use "xreports" in v6 as well?

Other points I have noticed (perhaps already reported):
* advanced search:
 * [ ] filters (bbox, activities, event_type, etc.) do not work => no result when enabling a filter
 * filter sliders
  * [ ] ``nb_participants`` and ``nb_impacted`` should have no unit and a step of 1
  * [ ] ``severity``: I think when the values will be translated they will be too long to fit under the slider. Perhaps you should consider using alternate (shorter) values?
 * [ ] As for outings, it would make sense to sort reports according to their date (newest first)
 * [ ] nice to have: filter on user id (as for outings), eg. http://localhost/reports#u=377
* detail view:
 * [ ] date should be displayed in the doc title (as for outings)
 * [ ] reporter is not shown
 * no opinion about the positions of the fields, I'll let the association give a feedback.
 * [ ] personal info about the reporter are not shown (I don't remember who should be able to see them: only moderators and the reporter?)
 * [ ] associated document types are maybe not correct. For instance should we really be able to associate a WP (in v5 one cannot, except climbing sites - by the way @tsauerwein perhaps another migration change as for outings associated to climbing sites?)
 * [ ] "access" seems not a very relevant section title for date and elevation :P 
 * [ ] history page is OK but diffing 2 versions is KO
 * [ ] archived versions are KO as well
* editing form
 * [ ] ``nb_participants`` and ``nb_impacted`` fields should be on the same line. Width of both field should be consistent.
 * [ ] ``age`` and ``gender`` fields should also be on the same line.
 * [ ] ``age`` field is a bit too wide. Suffix would rather be "year old" than "years"?
 * [ ] ``autonomy`` should be placed between ``author_status`` and ``activity_rate``.
 * [ ] It looks a bit weird that the map is placed in its own step. I would rather place this section in step 1, after the "general info" section (a bit like for outings).
 * [ ] last step: having a ``description`` section and then a ``report description`` is not very consistent :P In v5, the second section is entitled ``factors``.
 * [ ] preview is missing
 * [ ] the "save" button cannot be clicked whereas all mandatory attributes are set.

c2corg_ui/static/js/constants.js
```diff
@@ -36,6 +36,7 @@ app.constants = {
     'outings' : ['title' , 'lang', 'date_start', 'routes', 'activities'],
     'images': ['image_type'],
     'profiles': [],
+    'reports': ['title', 'lang', 'activities', 'event_type'],
```
asaunier: ``longitude`` and ``latitude`` are required as well.

c2corg_ui/templates/base.html
```diff
@@ -60,7 +60,7 @@
     <div class="page-header">
       <button class="btn btn-default menu-open-close header"><span class="glyphicon glyphicon-align-justify"></span></button>
       <div class="quick-search">
-        <app-simple-search dataset="wrcb"></app-simple-search>
+        <app-simple-search dataset="wrcbx"></app-simple-search>
```
asaunier: To be discussed/confirmed, but as for outings, I am not sure we want to have the xreports in the simple search.

c2corg_ui/templates/helpers/view.html
```diff
@@ -247,6 +247,29 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
     </section>
 </%def>
 
+<%def name="show_associated_reports(document, type='reports')">\
```
asaunier: Do we really need this ``type`` argument? Isn't it always set to ``reports``? Then you can remove this argument and replace it by ``reports``.

c2corg_ui/templates/i18n.html
```diff
@@ -439,3 +439,75 @@
...
+<span translate>more_than_3m</span>
+
+## autonomies
+<span translate>non_autonomous</span>
+<span translate>autonomous</span>
+<span translate>initiator</span>
+<span translate>expert</span>
+
+## avalanche_levels
+<span translate>level_1</span>
```
asaunier: I'm afraid those strings are a bit too generic. Using ``avalanche_risk_1`` for instance would be less confusing (including for translators).

asaunier:
@dkocich Is this PR ready? If not please prefix it "WIP". Thanks.

dkocich:
I am not sure about the renaming of `severity` and `avalanche level` fields when we all migrated now and this would require changes to common/API part again and maybe cause some errors later...

Currently, I show only associated users but no reporter (= author). I wonder where an author can associate other users. Should it be available in the editing form only (similar like in the 3rd step of [outing creation](http://c2corgv6.demo-camptocamp.com/outings/add))? 

I am still fixing date search, view map icon problem and editing issue.

asaunier:
Xreports are not outings. It is not possible to associate users (participants).
They are personal documents like personal articles.

> Currently, I show only associated users but no reporter (= author).

That should be the opposite => show the reporter, no associated users.

> I am not sure about the renaming of severity and avalanche level fields when we all migrated now 

What do you mean?

dkocich:
F.e. xreport ID 744553 have two associated users and 2 other moderators made all the changes (visible in version history). Usually, there is only one associated user visible. In DB, there are 200 assoc. between user - xreport.
image

asaunier:
OK, my mistake. I have looked at v5 and it is actually possible to associate users like you would do for an outing => same behaviour than outings then.

For most xreports, I think only the reporter is associated. There are about 200 xreports. So about 200 user-xreport associations.

asaunier:
@dkocich Are you aware that some of the xreports attributes should not be made public, visibles only for moderators and the authors (associated users?). I think it concerns those attributes:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/xreport.py#L64-L79
(not sure where it is described in the google sheet).

It means those attributes should be editable normally. But not added to the detail view mako template. I guess an additional asynchronous request (as for profiles) would be necessary to display the info to authorized people. I would suggest not showing those attributes at all for a first version of the UI.

Does the API already handle this special case?

dkocich:
@asaunier code for them is in the template of view / edit pages but the data is not available in the GET request. The API returns anonymized schema to all unauthorised users so they cannot see any details.

https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/xreport.py#L90-L94
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/xreport.py#L47-L51

c2corg_ui/templates/xreport/helpers/view.html
```diff
@@ -0,0 +1,153 @@
...
+          <span translate class="value-title">activities</span><br>
+        % for activity in xreport.get('activities') :
+          <span class="route-activity icon-${activity}" uib-tooltip="{{mainCtrl.translate('${activity}')}}"></span>
+        % endfor
+      </article>
+    % endif
+
+    % if xreport.get('event_type'):
+      <article class="value activities">
+          <span translate class="value-title">event_type</span><br>
```
ginold: i think there is no more such structure as `value-title  > br > value` (it was). Now you should remove the `br` and add a `:` after `<span translate class="value-title">event_type</span>:`.

dkocich: I see  `.value, .value-title {  font-weight: bold; }`  still in Less file... I will add `:`

dkocich:
@asaunier 
- not working date filter is hidden now
- currently, there is no option to add/edit report (UI buttons, search option, URL routes are commented)
- assoc. search field is hidden and it is not possible to add assoc. to anyone (including moderators)
- button to delete associations is hidden on the report detail page, so any assoc. cannot be deleted

Right now reports should be read only to everyone. Tests pass locally. Can we merge it, what do You think?

asaunier:
* [ ] In the advanced search page, with the map filter on (OK if disabled), an error prevents the results from showing up. The server logs say ``elasticsearch.exceptions.RequestError: TransportError(400, 'search_phase_execution_exception', 'No mapping found for [date] in order to sort on')``. Perhaps simply because my ES index is not up to date?!
*[ ] Would it be possible to show the icon in the detail view map? They are available in the advanced search map but not in the detail view.

c2corg_ui/static/js/documentediting.js
```diff
@@ -249,6 +249,10 @@ app.DocumentEditingController.prototype.successRead = function(response) {
   if (this.modelName === 'image') {
     data['date_time'] = new Date(data['date_time']);
   }
+  // xreports date in editing form
```
asaunier: Is it not possible to add this to the ``XreportEditingController``?
If yes, since we are a bit in a hurry, that could be done later (to add to the todo list?).

dkocich:
Search works for me without any errors. I am not sure what is missing for the icon to work how it should (there is only text visible without icon picture). Can't it be added separately?

requirements.txt
```diff
@@ -23,6 +23,6 @@ zope.interface==4.2.0
 # c2corg_common project
 # for development use a local checkout
 # -e ../v6_common
-git+https://github.com/c2corg/v6_common.git@910bb28
+git+https://github.com/c2corg/v6_common.git@a6ce113
```
asaunier: This commit hash is already outdated. https://github.com/c2corg/v6_common/commits/master.
Is your branch rebased?

asaunier:
For the map icon, I suspect something missing in the feature collection here https://github.com/c2corg/v6_ui/blob/773dee014e8e913987b0ff87d7ad0f74ed74f86a/c2corg_ui/templates/xreport/view.html#L47-L67 Not sure what?

asaunier:
Here are a few fixes I was about to commit (but I had a conflict, perhaps my branch was outdated):

```
diff --git a/c2corg_ui/static/js/map/map.js b/c2corg_ui/static/js/map/map.js
index f69c9b1..0a3fc65 100644
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -411,9 +411,9 @@ app.MapController.prototype.createStyleFunction_ = function() {
           case 'waypoints':
           case 'images':
           case 'profiles':
+          case 'xreports':
             return this.createPointStyle_(feature, resolution);
           case 'routes':
-          case 'xreports':
           case 'outings':
             return this.advancedSearch ?
               this.createPointStyle_(feature, resolution) :
diff --git a/c2corg_ui/templates/xreport/view.html b/c2corg_ui/templates/xreport/view.html
index dedd0fb..417759d 100644
--- a/c2corg_ui/templates/xreport/view.html
+++ b/c2corg_ui/templates/xreport/view.html
@@ -86,7 +86,7 @@ other_langs, missing_langs = get_lang_lists(xreport, lang)
     <div class="icons-header routes">${get_activities(xreport, 'top')}</div>
     <span class="title">${show_locale_title(locale)}</span>
     ${show_badge(xreport, 'xreport')}
-  </h1>x
+  </h1>
   <div class="outing-date" ng-cloak>${show_fulldate(xreport['date'])}</div>
 </header>
```

asaunier:
@dkocich I'll test and review the PR this morning. I let you know.

c2corg_ui/static/js/map/map.js
```diff
@@ -413,6 +413,8 @@ app.MapController.prototype.createStyleFunction_ = function() {
           case 'profiles':
             return this.createPointStyle_(feature, resolution);
           case 'routes':
+          case 'xreports':
+            return this.createPointStyle_(feature, resolution);
```
asaunier: This is not correct => you have changed the switch for routes and duplicated the ``return this.createPointStyle_(feature, resolution);`` whereas it already exists 2 lines above.

Please cancel this change and simply add ``case 'xreports':`` after line 413.

c2corg_ui/templates/helpers/view.html
```diff
@@ -234,9 +234,32 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
...
+      <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#associated-xreports">
+        <span class="glyphicon glyphicon-map-marker"></span>
+        <span translate>Associated xreports</span><span class="glyphicon glyphicon-menu-down"></span>
+      </h3>
+      <div class="associated-documents collapse in" id="associated-xreports">
+        <div class="list-item" ng-repeat="doc in ${associations}" ng-class="{'new-item': doc['new']}">
+          <app-card app-card-doc="doc"></app-card>
+          % if type == 'xreports' and document['doctype'] != 'xreports':
+            <app-delete-association child-doctype="xreports" parent-id="${document['document_id']}" child-id="doc.document_id"
+              ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
```
asaunier: I don't think that any authenticated user should be able to remove xreports associations.
It's probably the same than for other personal documents like articles or images. See for instance https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/view.html#L199-L215

That could be fine-tuned after golive but at least change the ``ng-if`` condition so that only moderators can remove the association.

c2corg_ui/templates/xreport/helpers/view.html
```diff
@@ -0,0 +1,153 @@
...
+          </p>
+        % endif
+
+      </span>
+    </div>
+  % endif
+</%def>
+
+<%def name="get_xreport_personal(xreport)">\
+% if xreport.get('author_status') or xreport.get('previous_injuries') or xreport.get('gender') or xreport.get('autonomy') or xreport.get('age'):
```
asaunier: indentation

c2corg_ui/templates/xreport/view.html
```diff
@@ -0,0 +1,194 @@
...
+
+  <div class="view-details-informations col-xs-12 informations">
+    <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#document-informations">
+      <span translate>Information</span><span class="glyphicon glyphicon-menu-down"></span>
+    </h3>
+    <section id="document-informations" class="collapse in">
+      ${get_xreport_general(xreport)}
+      % if not version:
+        ${get_xreport_participants(xreport)}
+        ${get_xreport_personal(xreport)}
```
asaunier: I don't think you can directly plug the personal info in the mako template. Those data are not public and should only be made available to the xreport author and moderators. But the UI server knows nothing about the current user rights.
It means those data should be added using an additional asynchronous API request depending on the user's permissions (a bit like for profiles).
For the golive, it's OK to simply not show those data at all (comment the line out).

c2corg_ui/templates/xreport/view.html
```diff
@@ -0,0 +1,194 @@
...
+  % else:
+    <meta name="robots" content="index,follow">
+  % endif
+</%block>
+
+<%block name="moduleConstantsValues">
+  <%
+      if xreport['geometry']:
+          geometry = xreport['geometry']['geom_detail'] \
+              if xreport['geometry']['geom_detail'] else xreport['geometry']['geom']
```
asaunier: I think you could write
```
geometry = xreport['geometry']['geom_detail'] or xreport['geometry']['geom']
```

c2corg_ui/templates/xreport/view.html
```diff
@@ -0,0 +1,194 @@
...
+  % else:
+    null
+  % endif
+  );
+
+  module.value('documentData', {
+    "document_id": ${xreport.get('document_id')},
+    "lang": "${lang}",
+    "type": "x",
+    "topic_id": ${dumps(locale.get('topic_id'))}
```
asaunier: Comments are disabled on xreports.

c2corg_ui/templates/xreport/view.html
```diff
@@ -0,0 +1,194 @@
...
+% if xreport.get('areas'):
+  <article class="location-static text-center">
+    ${show_areas(xreport)}
+    % if xreport.get('elevation'):
+      <p class="text-center"><b>${xreport['elevation']} m</b></p>
+    % endif
+##     TODO SHOW COORDINATES ON THE YELLOW PAPER IN THE HEADER MAP
+##     % if geometry4326:
+##       <p>${round(geometry4326.x)} °E / ${round(geometry4326.y)} °N</p>
+##     % endif
```
asaunier: @tsauerwein has fix the reprojection problems for waypoints. See https://github.com/c2corg/v6_ui/pull/1085
You can probably do the same for xreports.
OK to do it after golive if not enough time.

c2corg_ui/templates/xreport/view.html
```diff
@@ -0,0 +1,194 @@
...
+          ${show_associated_waypoints(xreport)}
+          ${show_associated_routes(xreport)}
+          ${show_associated_articles(xreport)}
+          ${show_associated_outings(xreport)}
+        </section>
+      </span>
+    </div>
+  % endif
+
+  ${get_comments()}
```
asaunier: No comment in xreports. The "comments" float button should then be removed as well.

asaunier:
My mistake: @HSERGERAERT has confirmed that comments should be enabled on xreports as well. I will update my previous comments => no changes about comments.

c2corg_ui/templates/xreport/view.html
dkocich: Thanks for comments. I fixed it - map marker is there, coord. in yellow box also. Personal data are commented now. That should be all now.

asaunier:
Thanks for the changes! Please merge :)

Issue #865, opened 2016-11-06
Bugs with Articles


fbunoz:
- [x] Filter on article category return always no result.
- [x] Filter on article name return always no result.
- [x] Association tool doesn't work for article : no article can be found using the tool, on any page.
- [x] On an article page, the linked outings are not listed.

stef74:
label association is for Camptocamp Association ... not for relationship between doc

dkocich:
I am not sure why this does not work. On the other testing server that we have, the problems 1, 2 and 3 are OK now.

asaunier:
> On the other testing server

Which one? The "dev" one?
We don't even know what server @fbunoz has tested :P

asaunier:
I give the answer myself (just tested it): OK on http://c2corgv6.demo-camptocamp.com/articles but KO on http://www.demov6.camptocamp.org/articles
The first one is much more up-to-date than the second one. => updating the checklist above.

asaunier:
> On an article page, the linked outings are not listed.

Not that critical. Moving to 6.1

fbunoz:
> We don't even know what server @fbunoz has tested

I tested with http://www.demov6.camptocamp.org server, the only one which must be used for testing by the Association.

asaunier:
> I tested with http://www.demov6.camptocamp.org server, the only one which must be used for testing by the Association.

That's correct indeed :)
But it's worth checking on the other server as well (at least for people like you). This way you can ignore problems that are already fixed :)

stef74:
@fbunoz @shortoland could you test ?

shortoland:
@stef74 Not yet ready for test

stef74:
@fbunoz @shortoland please test

fbunoz:
All the points are fixed.

PR #915, opened 2016-11-17
Add posibility to assoc. routes / outings from articles

dkocich:
Just a small addition to addassociation.js ... Assoc. are visible, links work correctly and articles are visible/working in other documents also.

Related to https://github.com/c2corg/v6_ui/issues/865


asaunier:
Just a minor question, else looks good to me.
Please don't close the issue when merging the PR, rather label it "ready for demo".

PR #918, opened 2016-11-17
new checkbox style

ginold:
I had to do it because I could not watch it anymore and there were display problems, that you probably  have experienced. This PR aims to display checkboxes more uniformly and to change their style - much more plus joli now!

screenshot from 2016-11-17 14-22-56


asaunier:
The main change of this PR is the way the checkboxes are rendered, right?
Have you tested on several pages and with various browsers?

OK for me, even though I don't think we should spend much time on such styles :)

ginold:
tested on ff, chrome and ie 11 !

I will merge soon if no objections.

asaunier:
Please merge, thanks!

PR #919, opened 2016-11-17
fix image url on view page

### LINKED ISSUE #910: Invalid image url on home feed
On the home feed, the image url are invalid in the card : 
```background-image: url(//sos.exo.io/c2corg_demov6_active1479025574_280720615MI.jpg)```
instead of
```background-image: url(//sos.exo.io/c2corg_demov6_active/1479025574_280720615MI.jpg)```

It is not stable : sometime the url is correct, sometime it misses a / , sometime there is 2 slash ( // ) !

ginold:
should fix #910

### DIFF
--- a/c2corg_ui/templates/image/preview.html
+++ b/c2corg_ui/templates/image/preview.html
@@ -39,7 +39,7 @@ ${show_preview_warning()}
 
     <div class="view-details-description col-xs-12 description">
       <span class="lead text-center">
-        <img class="image" src="${image_url}/${image['filename']}" alt="${locale['title']}">
+        <img class="image" src="${image_url}${image['filename']}" alt="${locale['title']}">
       </span>
     </div>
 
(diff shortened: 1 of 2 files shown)


asaunier:
What about https://github.com/c2corg/v6_ui/blob/3966406acd21d7b89a477e865c92469b5064d229/c2corg_ui/templates/image/preview.html#L42 ?

asaunier:
But wouldn't the problem be that ``image_url`` should not have a trailing slash?
If it does it's not consistent with ``image_backend_url`` (no trailing slash):
https://github.com/c2corg/v6_ui/blob/383b49c458e0c554ca375ec12156837a7962cf59/config/default#L9-L10

ginold:
I don't know which one is the problem :P as long as it works everywhere else, it's ok to remove the additional slash from the view pages, IMHO. And it's clearer to write `image_url + filename `rather than `image_url + '/' + filename`.

Issue #593, opened 2016-09-05
Poor loading notifications


asaunier:
As for now when authenticating/registering/resetting/changing the account data, there is nothing indicating something is going on in the background.

A (partial) PR addresses this problem but needs more work I think: https://github.com/c2corg/v6_ui/pull/457

asaunier:
This issue is a bit critical I think: for instance when you try to register, nothing tells you that a request is pending: no loading indicator, the submit button remains active => if the response is not fast (it generally isn't) you just want to click several times on the button.

There's probably a similar problem with several other write actions.

dkocich:
Maybe something like this can be used? https://chieffancypants.github.io/angular-loading-bar/

asaunier:
> Maybe something like this can be used? https://chieffancypants.github.io/angular-loading-bar/

That looks great indeed!
Would be useful for https://github.com/c2corg/v6_ui/issues/456#issuecomment-247940897 as well

tsauerwein:
I've set up angular-loading-bar for another project, and it worked out of the box.

When we set it up, we have to make it sure that it plays nicely with already existing loading indicators (for example for the feed or advanced search).

asaunier:
See also https://github.com/c2corg/v6_ui/pull/925

PR #925, opened 2016-11-18
Add styles for ng-cloak to the base.html to apply them correctly

dkocich:
I was wondering what is the problem with loading and showing some angular variables like `{{resCounter}} {{filtersCtrl...}}`, some DIVs appear and then disappear (they are visible to moderators, etc.). This should be probably hidden. We already use `Ng-cloak` ([docs](https://docs.angularjs.org/api/ng/directive/ngCloak)) in the code for that elements, but for some reason, the styles for it are not applied (they are contained inside in angular js files which are loaded after the page loads???).

I added the styles to base.html so they are loaded immediately and now pages are loading correctly (but it should be probably added somewhere else? I didn't find any angular related .less style files).

Sometimes this feature like it is now can be actually useful for debugging (it works with `?debug` URL now, but that is not available on demo for testers), so feel free to discuss / merge.

Before:
image

After:
image

### DIFF
--- a/c2corg_ui/templates/base.html
+++ b/c2corg_ui/templates/base.html
@@ -38,6 +38,11 @@ closure_library_path = settings.get('closure_library_path')
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/build.css')}">
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/build-print.css')}" media="print">
 % else:
+    <style>
+      [ng\:cloak], [ng-cloak], [data-ng-cloak], [x-ng-cloak], .ng-cloak, .x-ng-cloak {
+        display: none !important;
+      };
+    </style>
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/deps.css')}">
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/build.min.css')}">
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/build-print.min.css')}" media="print">
--- a/c2corg_ui/templates/base.html
+++ b/c2corg_ui/templates/base.html
@@ -8,7 +8,7 @@ closure_library_path = settings.get('closure_library_path')
 %>\
 <%namespace file="helpers/common.html" import="show_title"/>\
 <!DOCTYPE html>
-<html <%block name="pagelang"></%block> ng-app="app" ng-controller="MainController as mainCtrl">
+<html ng-cloak <%block name="pagelang"></%block> ng-app="app" ng-controller="MainController as mainCtrl">
   <head>
     <%block name="pagetitle"><title ng-bind="mainCtrl.page_title('Home')">${show_title('Home')}</title></%block>
     <meta charset="utf-8">


asaunier:
You're right, having all those temporary expression appearing at loading time is a bit ugly. If you have found a solution, that's really cool, thanks!

No idea if there is a more relevant place to put those styles than in ``base.html``. I think it's OK. @ginold OK for you?

ginold:
I would add this part of code to helpers.less or to the main c2corg_ui.less . 

I was telling myself that those tags should be hidden by default, by I didn't realize that it was an issue - i thought it 'is the way it is'.

Other than that - :+1:

asaunier:
> I would add this part of code to helpers.less or to the main c2corg_ui.less

Isn't the problem related to the fact that styles in CSS files are loaded too late? The temporary expressions are then displayed in the meantime.  If we add those ngCloak styles to the HTML page they are loaded faster, aren't they?

ginold:
That's a good question, and that would make sense.
If @dkocich could do a little performance comparison with this block of code inside base.html and a less file - I would appreciate :P

dkocich:
I tried to put it to` c2corg_ui.less` and recompile my project already in the morning. I think it did not work, but I was not sure if it is problem with the browser cache / Less minification (or it is already too late to use when it lies somewhere in the built CSS file)... When I put it to the `base.html`, it was immediately ok.

I added extra commit - tried to add ng-cloak to page main DIV (`<html ng-cloak>`) and do maybe 10 tests with the articles page (there are no map tiles loading, etc.). Every time the load time was around 2,5s instead of previous 3,5s, so the page is now faster when it don't have to restyle everything few times.

asaunier:
I have the feeling that adding the styles in ``base.html``, as you suggested, is the best solution.

Concerning ``<html ng-cloak>``, does this mean it is useless to add ``ng-cloak`` to any other element?

dkocich:
In docs they say:

> The directive can be applied to the <body> element, but the preferred usage is to apply multiple ngCloak directives to small portions of the page to permit progressive rendering of the browser view.

I don't know what can be a problem later (I could not find any problems now). I think that if we will keep adding ng-cloak where possible and later we will see some problem, it will be easy to just remove it from the root HTML div (delete one commit).

If I leave PR without 2nd commit - (having ng-cloak on root), there are still some things jumping around on the page (text are retranslated from initial values, icons in the sidebar are showing up, etc.). With that, we are sure that everything is covered everywhere on the page.

asaunier:
OK, I'll test your branch this week end, I think. I let you know. Thanks anyway!

asaunier:
I have just tested the PR (with ``<html ng-cloak>``.

Pros:
* no more ugly ``{{ foobar }}`` in any page => really cool!
* pages are rendered neatly

Cons:
* pages take a bit longer to be render, since they are only rendered when they are ready

=> For the first loaded page, the screen can remain blank a few seconds
For following pages, it's not that bad because the previous page remains displayed in the mean time (no blank page). If we had the angular-loading-bar (see https://github.com/c2corg/v6_ui/issues/593#issuecomment-258136632) would the loading bar be displayed when loading a new page. If yes, it could be acceptable (would work a bit like in github actually).

Anyway the changes look very interesting!

gottferdom:
I totally agree with @asaunier!

dkocich:
Does that mean I should merge @asaunier ?

asaunier:
What about adding the angular-loading-bar to this PR?

@tsauerwein said it worked out of the box in another project, then I guess it's not a big deal to add it. And it would be a good way to test if this would solve https://github.com/c2corg/v6_ui/issues/593 and https://github.com/c2corg/v6_ui/issues/456

tsauerwein:
> What about adding the angular-loading-bar to this PR?

I would recommend to do it in a separate PR. And like I said, the problem will be to make it play nicely with the other loading indicators.

asaunier:
OK, than @dkocich OK for merging your PR, thanks!

PR #928, opened 2016-11-18
Card attributes updated

ginold:
related to #398

+ added icons for the elevations/diffs 

Associated cards:
+ add outing quality
~~remove outing elevation from associated cards -> show only in the home feed~~ outing's elevation max everywhere
+ add route height_diff_difficulties (=  difficulties denivelation?)
+ add route height_diff_up (= denivelation?)

Search suggestions:
+ add 'height_diff_up' for routes

- Outing Search: Mobile: max elevation should already be available

@loicperrin, @shortoland please tell me if when the above attributes are added if there is still something missing.

### DIFF
--- a/c2corg_ui/static/partials/cards/outings.html
+++ b/c2corg_ui/static/partials/cards/outings.html
@@ -10,10 +10,11 @@
     <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{::locale.lang}})</span>
     <br>
     <div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
-         uib-tooltip="{{'quality'| translate}} : {{::doc['quality'] | translate}}" tooltip-placement="left"></div>
+         uib-tooltip="{{'quality'| translate}} : {{::doc['quality'] | translate}}" tooltip-placement="left">
+    </div>
 
-    <span class="elevation-max" ng-if="::doc['elevation_max']" uib-tooltip="{{'elevation_max' | translate}}">
-      <span translate class="value-title glyphicon glyphicon-sort-by-attributes"></span>
+    <span class="elevation-max"  uib-tooltip="{{'elevation_max' | translate}}">
+      <span class="value-title glyphicon glyphicon-sort-by-attributes rotate-180"></span>
       <span class="value">{{::doc['elevation_max']}}&nbsp;m</span>
     </span>
 
--- a/c2corg_ui/static/partials/suggestions/waypoints.html
+++ b/c2corg_ui/static/partials/suggestions/waypoints.html
@@ -1,11 +1,16 @@
-  <span class="suggestion-text"> 
-    <span class="suggestion-title"> 
-      <span ng-bind-html="highlight(label, _query)"></span> 
-    </span><br> 
-    <span class="suggestion-info"> 
-      <span ng-if="::elevation">{{::elevation}}m</span>
-    </span> 
-  </span> 
-  <span class="icon-container"> 
-    <span class="icon-{{::waypoint_type}}"></span> 
+  <span class="suggestion-text">
+    <span class="suggestion-title">
+      <span ng-bind-html="highlight(label, _query)"></span>
+    </span><br>
+
+    <div class="suggestion-info flex" ng-if="::elevation">
+      <div uib-tooltip="{{'elevation' | translate}}">
+        <span class="glyphicon glyphicon-sort-by-attributes rotate-180"></span>
+        <span>{{::elevation}}m</span>
+      </div>
+    </div>
+
+  </span>
+  <span class="icon-container">
+    <span class="icon-{{::waypoint_type}}"></span>
   </span>
--- a/c2corg_ui/static/partials/cards/routes.html
+++ b/c2corg_ui/static/partials/cards/routes.html
@@ -5,20 +5,20 @@
   <div class="list-item-info">
     <div class="flex wrap-row">
       <span class="elevation-max" ng-if="::doc['elevation_max']" uib-tooltip="{{'elevation_max'| translate}}">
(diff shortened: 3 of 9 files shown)


c2corg_ui/static/partials/cards/outings.html
```diff
@@ -10,10 +10,13 @@
...
     <br>
     <div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
-         uib-tooltip="{{'quality'| translate}} : {{::doc['quality'] | translate}}" tooltip-placement="left"></div>
+         uib-tooltip="{{'quality'| translate}} : {{::doc['quality'] | translate}}" tooltip-placement="left">
+    </div>
 
-    <span class="elevation-max" ng-if="::doc['elevation_max']" uib-tooltip="{{'elevation_max' | translate}}">
-      <span translate class="value-title glyphicon glyphicon-sort-by-attributes"></span>
+    <!--elevation-max is shown only in the home feed-->
+    <span class="elevation-max" ng-if="::doc['elevation_max'] && cardCtrl.currentLocation == '/'"
```
asaunier: Is it not possible to test ``window.location.pathname == '/'`` directly here?

Anyway is this condition robust enough?
For instance feed cards are also used in the profile pages (with a URL != ``/``)...

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -10,10 +10,13 @@
...
+          uib-tooltip="{{'elevation_max' | translate}}">
+      <span translate class="value-title glyphicon glyphicon-sort-by-attributes rotate-180"></span>
```
asaunier: What does this 180° rotation?
Is the ``translate`` directive really useful here? Ditto for the other span with ``rotate-180``.

asaunier:
A small concern about the ``/`` test. Other code changes looks good to me.
No idea about the relevancy of adding those attributes to cards but I guess that has been discussed with @loicperrin and @shortoland in the issue, so OK for me.

c2corg_ui/static/partials/cards/outings.html
ginold: rotation rotates the glyphicon icon, because it's upside down originally :P

oops, forgot to remove the translate directive!

c2corg_ui/static/partials/cards/outings.html
ginold: I don't think it is the "angular way" with writing pure JS inside ng-if, and after a quick look on stackoverflow I didn't find that it was possible.

In this case it seems to be robust enough - elevation-max is shown only in the home feed, and the only page where `window.location.pathname == /` is the home feed page.

ginold: BTW it is the same for mainCtrl.isPath => it checks for `location.pathname == / `for home page.

asaunier: > In this case it seems to be robust enough - elevation-max is shown only in the home feed, and the only page where window.location.pathname == / is the home feed page.

What about the profile feed? It should work the same way than the home page feed.
For me testing the URL is not a robust way.

But why do we need to have the elevation max only in the feed?

ginold: The info is in this sheet: https://docs.google.com/spreadsheets/d/1W2kEAyGPc7cw9yKe3hyYIMTNgUDLPsFUD0ccTS8jI5Y/edit?pref=2&pli=1#gid=0   -> elevation max on homepage only (but you are probably right that in profile feed probably too).

So I could test for  "/" and "/profiles/". Or what other testing would you suggest? this would seem robust enough to me, i cannot think of other cases where it would bring an error.

asaunier: > So I could test for "/" and "/profiles/".

Again, this is not robust. What if the feed is integrated somewhere else? We won't add all the possible URLs in the test :) A robust way would be that the cards know (via a directive attribute?) that they are used in a feed.

Anyway I think we should simply have this ``elevation_max`` info in all outings cards, feed or not.
By the way it seems this attribute is already available in the cards of the advanced search!?

So why don't you drop this condition and reuse what is done for the advanced search cards?

asaunier: > I would not have any testing if Shortoland didn't say in #398 that

I have added a comment https://github.com/c2corg/v6_ui/issues/398#issuecomment-262476481
=> please drop this exception and display the attribute all the time. Thanks :)

ginold: > Get up, stand up, stand up for your right :musical_note: 

thanks! :P

ginold:
Udated: outing's elevation-max is now everywhere! <3

Question: I think I might be outdated but - is it no more allowed to look for outings in the search tool on the top ? :P

c2corg_ui/static/partials/suggestions/waypoints.html
```diff
@@ -1,11 +1,16 @@
...
-    <span class="icon-{{::waypoint_type}}"></span> 
+  <span class="suggestion-text">
+    <span class="suggestion-title">
+      <span ng-bind-html="highlight(label, _query)"></span>
+    </span><br>
+
+    <div class="suggestion-info flex" ng-if="::elevation">
+      <div uib-tooltip="{{'elevation' | translate}}">
+        <span class="glyphicon glyphicon-sort-by-attributes rotate-180"></span>
+        <span>{{::elevation}}m</span>
```
asaunier: missing non-breaking space before the unit I think

asaunier:
A minor comment else looks good to me (I have only looked at the code).

> is it no more allowed to look for outings in the search tool on the top ?

It has never been possible I think. Simple-searching an outing is not very relevant. Users search for routes or waypoints and then look at the related outings. Or check the last outings using the home feed or the advanced search.

Issue #912, opened 2016-11-16
Ad banner integration in the sidemenu


desnoes:
The advertisment picture overlap the community section
see
![pub accueil](https://cloud.githubusercontent.com/assets/19892089/20360210/1f28e7ae-ac33-11e6-9726-e3f296c17dac.jpg)

asaunier:
Actually it only happens when the banner starts overlapping the menu. If you shrink a little more the screen in height, the banner is removed.

I think it worked correctly until the "version ..." link was added in the footer of the menu column. There's probably some CSS to update.

In fact this "version" link probably takes a bit too much space or could be placed somewhere else (no idea where!?) or maybe be removed before the ad banner when the height is too small?

stef74:
@desnoes @shortoland could you test ?

asaunier:
@stef74 The issue is marked as "ready for demo" (not "fixed and ready to be tested"), which means a PR has been merged but that the fix is not available in the demo yet.

stef74:
Ok it''s misunderstanding of the label .....

desnoes:
The banner has disappeared !
![pub](https://cloud.githubusercontent.com/assets/19892089/20573034/44b510d0-b1ae-11e6-9db3-148e9c4c6c78.jpg)

HSERGERAERT:
Je l'aie
[image: Images intégrées 1]

Hervé

Coordinateur Projet V6
*Camptocamp Association*
Tél. +33 6 2563 5290

2016-11-23 18:54 GMT+01:00 desnoes <notifications@github.com>:

asaunier:
> The banner has disappeared !

The banner is not shown when there is not enough space. Your screen is too small :P

desnoes:
Is a screen size of 17.3 inch is considered as small :)

asaunier:
It's rather a question of resolution.
Perhaps there are too many things in the sidemenu...

tsauerwein:
One thing to consider: The ad provider is paying the association to show the ad. But if there is not enough space, and the ad is not shown, the ad would still be loaded and count as "shown" even if it is not visible. This might be a problem.

desnoes:
@mgerbaux : please be informed of the problem identified by Tobbias.

ginold:
When I have time I will try to always show the ad, whatever the size available, by shrinking the ad itself.

asaunier:
> always show the ad, whatever the size available, by shrinking the ad itself.

[...]

PR #931, opened 2016-11-18
Fix hiding ad if height is too small

tsauerwein:
Related to https://github.com/c2corg/v6_ui/issues/912


ginold:
the best would be to shrink it proportionally when the top of the ad hits the menu instead of hiding it.

PR #934, opened 2016-11-19
Improve perceived performance of advanced search

tsauerwein:
When navigating around in the map, new results are constantly queried. While the request itself is asynchronously, updating the DOM (showing the results) is blocking. During that time you can not navigate the map, which feels disturbing.

The aim of this PR is to improve the performance of updating the DOM, and to reduce the number of DOM updates. This is done by:

- Caching the compiled elements for the cards.
- Canceling previous XHR requests, when a new request is made.
- Debouncing the requests. A request is only made if the filters/bbox have not changed in the last 700ms.

This is still not perfect, but it improves the perceived performance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ LIBS_JS_FILES += \
     c2corg_ui/static/lib/angular-bootstrap/ui-bootstrap-custom-tpls-1.3.2.min.js \
     node_modules/angular-gettext/dist/angular-gettext.min.js \
     node_modules/angular-ui-date/dist/date.js \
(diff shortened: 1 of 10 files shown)


asaunier:
I have not tested the branch yet, but it looks very good, thanks for the improvement!!

tsauerwein:
Thanks for the review!

asaunier:
I have just tested the branch, it works great and that's true it's really useful to be able, for instance, to check several checkboxes in a row without triggering a request for each changed checkbox :)

PR #939, opened 2016-11-19
Use one-time-bindings

tsauerwein:
To reduce the number of Angular watchers, this PR proposes to use one-time-bindings. I didn't check all partials, I focused on the card partials because they are used intensively in the advanced search and the feed.

For more information about one-time-bindings, see here: https://www.binpress.com/tutorial/speeding-up-angular-js-with-simple-optimizations/135

The last commit uses [track by](https://docs.angularjs.org/api/ng/directive/ngRepeat#tracking-and-duplicates) for the `ng-repeat` in the advanced search and the feed. See also: http://www.codelord.net/2014/04/15/improving-ng-repeat-performance-with-track-by/

### DIFF
--- a/c2corg_ui/static/js/card.js
+++ b/c2corg_ui/static/js/card.js
@@ -26,7 +26,7 @@ app.cardDirective = function($compile, $templateCache) {
     controllerAs: 'cardCtrl',
     bindToController: true,
     scope: {
-      'doc': '=appCardDoc'
+      'doc': '<appCardDoc'
     },
     link: function(scope, element, attrs, ctrl) {
       element.html(template(ctrl.type));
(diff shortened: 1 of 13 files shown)


tsauerwein:
I just pushed two more commits to optimize the feed. Still ok for merging? :)

c2corg_ui/static/js/card.js
```diff
@@ -26,7 +26,7 @@ app.cardDirective = function($compile, $templateCache) {
     controllerAs: 'cardCtrl',
     bindToController: true,
     scope: {
-      'doc': '=appCardDoc'
+      'doc': '<appCardDoc'
```
asaunier: Didn't know this one. What's the difference?

tsauerwein: Here is a good explanation: https://toddmotto.com/one-way-data-binding-in-angular-1-5/

When using a "one-way-binding" (note the difference to a "one-time-binding"), the changes inside the directive are not propagated to the outside. For objects (like in our case) it does not make a big difference though.

Issue #941, opened 2016-11-19
WP creation 'climbing_indoor': 'elevation' shall not be requested


desnoes:
It is not possible to complete the creation of the document. The 'elevation' is required while it is not specified in the 'required' section (https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_waypoint.py)

![sae](https://cloud.githubusercontent.com/assets/19892089/20455289/bfe4b87c-ae58-11e6-88af-406468dc8be4.jpg)

ginold:
yes it is, look closely: https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_waypoint.py#L16

```
DEFAULT_REQUIRED = [
    'locales',
    'locales.title',
    'geometry',
    'geometry.geom',
    'elevation'
]
```

asaunier:
@ginold It's not. Look closely:
https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_waypoint.py#L94-L99
:p 

What you are pointed at are the default required fields for waypoints but for "climbing_indoor" the default setting is not used.

ginold:
Nice one! :p so @desnoes probably means that elevation is not required for 'climbing_indoor' only (missing info in the issue description).

desnoes:
'elevation' is not considered for the 'climbing_indoor' WP and only for this WP.

stef74:
@mgerbaux could you test ?

stef74:
@desnoes could you test ?

desnoes:
the bug has not been fixed. Elevation is still required and it is not possible to complete the creation of this WP.

----- Mail original -----
De: "stef74" <notifications@github.com>
À: "c2corg/v6_ui" <v6_ui@noreply.github.com>
Cc: "desnoes" <mathieu.desnoes@free.fr>, "Mention" <mention@noreply.github.com>
Envoyé: Vendredi 2 Décembre 2016 10:18:25
Objet: Re: [c2corg/v6_ui] WP creation 'climbing_indoor': 'elevation' shall not be requested (#941)



@desnoes could you test ?

desnoes:
i retested and the bug disappeared !


----- Mail original -----
De: "mathieu desnoes" <mathieu.desnoes@free.fr>
À: "c2corg/v6_ui" <reply@reply.github.com>
Cc: "Mention" <mention@noreply.github.com>, "c2corg/v6_ui" <v6_ui@noreply.github.com>
[...]

Issue #791, opened 2016-10-25
Document attributes


asaunier:
This issue lists problems with missing/misplaced/incorrectly entitled attributes.
It might also be addressed in combination with the documents organisation issue https://github.com/c2corg/v6_ui/issues/449

Some points are fixed by https://github.com/c2corg/v6_ui/pull/851

## Waypoints
- [ ] ~~attribute 'Proéminence' really relevant?,~~ see https://github.com/c2corg/v6_ui/issues/788
- [ ] missing attribute `rock_types`, see https://github.com/c2corg/v6_ui/issues/789
- [x] for waypoint type `shelter`, field 'capacity_staffed' should not be displayed (because it is never staffed anyway), see https://github.com/c2corg/v6_ui/issues/798
- [ ] remove 'elevation' from waypoint type 'climbing_indoor'
## Routes
- [x] attribute `external_resources` (text field for book and web resources) is not displayed either in the detail view pages or in the editing form. OK in API https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/route.py#L232 and in common https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py#L8 => must be added after https://github.com/c2corg/v6_ui/blob/addbe79c1076f4291e1293cbc0dee16f77f4abb2/c2corg_ui/templates/route/view.html#L129

desnoes:
Waypoint 
- [ ] 'paragliding_takeoff' and 'paragliding_landing': field 'ground_types' is missing
- [ ] 'weather_station' : display issue in the 'Types & Styles' section. 2 fields are overlapping. Is the title 'Types & Styles' correct ?

asaunier:
> Is the title 'Types & Styles' correct ?

@desnoes See https://github.com/c2corg/v6_ui/issues/705#issuecomment-256511717

asaunier:
More WP attribute issues: https://github.com/c2corg/v6_ui/issues/825

asaunier:
Additional points moved from https://github.com/c2corg/v6_ui/issues/411#issuecomment-247082807 (perhaps already addressed?):

- [ ] Rocher haute montagne: field 'longueur de l'itinéraire is missing.ex: http://c2corgv6.demo-camptocamp.com/routes/736982/de/essai-alpha-3-climbing

[...]

PR #952, opened 2016-11-21
Fix misplaced and missing attributes

ginold:
related to #791 
related to #941

+ remove 'capacity_staffed' for waypoint type 'shelter'
+ add 'route_length' for waypoint type 'rock_climbing'
+ remove 'difficulties_height' from route type 'mountain_biking', 'hiking' and 'snowshoeing' (as from https://docs.google.com/spreadsheets/d/1p7-RakklDot9gFnciEGox75833zU0ZuC5pRAoLZKziY/edit#gid=1418510514 but on https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py it's still present -> which one is the right one?)

+ add 'ground_types' for waypoint type 'paragliding_landing' and 'paragliding_takeoff'
+ let save waypoint type 'climbing_indoor' without elevation

### DIFF
--- a/c2corg_ui/static/js/documentediting.js
+++ b/c2corg_ui/static/js/documentediting.js
@@ -488,6 +488,9 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
       hasError = (!doc['lonlat'] || (doc['lonlat'][field] === null || doc['lonlat'][field] === undefined));
     } else if (field === 'date_start') {
       hasError = (doc['date_start'] === null || doc['date_start']  === undefined);
+    } else if (field === 'elevation' && doc['waypoint_type'] === 'climbing_indoor') {
+      // waypoint climbing indoor is the only one that does not require 'elevation'
+      continue;
     } else {
       hasError = (!doc[field] || doc[field] === null || doc[field] === undefined);
     }
--- a/c2corg_ui/templates/route/edit.html
+++ b/c2corg_ui/templates/route/edit.html
@@ -275,8 +275,7 @@ updating_doc = route_id and route_lang
 
             ## DIFFICULTIES HEIGHT
             <div id="difficulties_height-group" class="data half form-group" ng-class="{ 'has-error': editForm.difficulties_height.$touched && editForm.difficulties_height.$invalid,'has-success': route.difficulties_height }"
-                 ng-if="route.activities.indexOf('via_ferrata') > -1 || route.activities.indexOf('ice_climbing') > -1 || route.activities.indexOf('rock_climbing') > -1 || route.activities.indexOf('snowshoeing') > -1
-                    || route.activities.indexOf('mountain_climbing') > -1 || route.activities.indexOf('snow_ice_mixed') > -1 || route.activities.indexOf('hiking') > -1 || route.activities.indexOf('mountain_biking') > -1">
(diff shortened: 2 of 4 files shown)


ginold:
@asaunier you may review if you have time.

c2corg_ui/static/js/documentediting.js
```diff
@@ -488,6 +488,9 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
       hasError = (!doc['lonlat'] || (doc['lonlat'][field] === null || doc['lonlat'][field] === undefined));
     } else if (field === 'date_start') {
       hasError = (doc['date_start'] === null || doc['date_start']  === undefined);
+    // waypoint climbing indoor is the only one that does not require 'elevation'
+    } else if (field === 'elevation' && this.scope['waypoint'] && this.scope['waypoint']['waypoint_type'] === 'climbing_indoor') {
```
asaunier: It would perhaps be more explicit to write
```
} else if (type === 'waypoints' && field === 'elevation' && this.scope['waypoint']['waypoint_type'] === 'climbing_indoor') {
  // waypoint climbing indoor is the only one that does not require 'elevation'
  continue;
}
```

Is ``doc`` the same thing than ``this.scope['waypoint']``? Could we use ``doc['waypoint_type']`` instead of ``this.scope['waypoint']['waypoint_type']``?

asaunier:
A suggestion about ``elevation`` vs ``climbing_indoor`` waypoints else OK for me (I have just looked at the code, not checked they match the issue => I'll let the association check when the demo is updated).

> remove 'difficulties_height' from route type 'mountain_biking', 'hiking' and 'snowshoeing' (as from https://docs.google.com/spreadsheets/d/1p7-RakklDot9gFnciEGox75833zU0ZuC5pRAoLZKziY/edit#gid=1418510514 but on https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py it's still present -> which one is the right one?)

The reference is https://docs.google.com/spreadsheets/d/1p7-RakklDot9gFnciEGox75833zU0ZuC5pRAoLZKziY/edit#gid=1418510514

I see the attribute is tagged "expert" in the google docs, which is confusing and must be ignored anyway.

If https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py is not consistent, please update this file as well, thanks.

ginold:
I updated the commons, requirements.txt in UI and APi and with `doc['waypoint_type']`.

c2corg_ui/static/js/documentediting.js
```diff
@@ -488,6 +488,9 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
       hasError = (!doc['lonlat'] || (doc['lonlat'][field] === null || doc['lonlat'][field] === undefined));
     } else if (field === 'date_start') {
       hasError = (doc['date_start'] === null || doc['date_start']  === undefined);
+    } else if (field === 'elevation' && doc['waypoint_type'] === 'climbing_indoor') {
```
asaunier: Don't you need to test that the current document is a waypoint (``type === 'waypoints'``)?

asaunier: OK, I have tested the following code
```
doc = {}
doc['a'] === 'bou' /* return false */
```
So your code works if ``doc`` is not a waypoint.

Then please merge, thanks.

asaunier: > because field === 'elevation' is only true if type is 'waypoints'

As for now. But who knows if a document type with an ``elevation`` attribute will not be added in the future? For instance... accident reports! https://github.com/c2corg/v6_api/blob/reports/c2corg_api/models/report.py#L35

Your assumption makes the test not robust :P 

But it's OK because ``doc['waypoint_type'] === 'climbing_indoor'`` will return ``false`` for the accident reports.

Issue #937, opened 2016-11-19
WP creation 'Paragliding' does not work


desnoes:
The cotation is a number (1 to 5) and the processing is waiting for a string.

![paragliding](https://cloud.githubusercontent.com/assets/19892089/20454850/e14e5f84-ae4b-11e6-95ac-0ea54650d8ce.jpg)

desnoes:
It is ok, i close the issue

PR #954, opened 2016-11-21
fix paragliding_rating handling

ginold:
related to #937


asaunier:
Could it be a problem with the API instead?
Shouldn't the API consider those values as strings instead of integers (even though it would be strings actually containing integers)?

tsauerwein:
The paragliding ratings, like all other ratings, are enums. That why it has to be a string.

But why is the string converted to an integer? I thing the problem is here:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/edit.html#L485

Instead of `ng-options="rat as rat for rat in [1, 2, 3, 4, 5]"` is should rather be `ng-options="rat as rat for rat in ['1', '2', '3', '4', '5']"`. Or you could generate the options like for other enums.

asaunier:
> Or you could generate the options like for other enums.

Definitely! It's not an option to hard code the values. They must always be retrieved from https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py

ginold:
oh sorry, I didn't notice that those were not loaded from attributes.py. It has to be some old code :P

ginold:
updated with ${paragliding_ratings}, works fine.

Issue #959, opened 2016-11-21
Outing creation on a new route: route not associated and cannot be associated


loicperrin:
I have created a new route: http://www.demov6.camptocamp.org/routes/822428/fr/test
If I create an outing on this route, it is not associated to the outing (it is OK if the route already contains an outing).
And if I try to add the route manually, I cannot find it (maybe it due to issue https://github.com/c2corg/v6_api/issues/285).

fjacon:
I have an Internal Server Error when I try to add an outing to a route I just created.

tsauerwein:
The outing has been created: https://www.demov6.camptocamp.org/outings/824705/fr/test-fred-itineraire-1

And the API returns the outing without error: http://api.demov6.camptocamp.org/outings/824705

There seems to be a problem on the UI server side.

ginold:
```
 File "/home/agina/NetBeansProjects/C2C.org/v6_ui/c2corg_ui/templates/base.html", line 149, in render_body
    vcRecaptchaServiceProvider.setSiteKey('${settings["recaptcha_site_key"]}');
  File "/home/agina/NetBeansProjects/C2C.org/v6_ui/c2corg_ui/templates/outing/view.html", line 37, in render_moduleConstantsValues
    <%
TypeError: 'NoneType' object is not subscriptable
```

and line 37

```
  <%
      geometry = outing['geometry']['geom_detail'] \
          if outing['geometry']['geom_detail'] else outing['geometry']['geom']
  %>
```

asaunier:
It seems the outing has no geometry at all whereas a default geometry should come from the associated routes or WP, shouldn't it?

tsauerwein:
> It seems the outing has no geometry at all whereas a default geometry should come from the associated routes or WP, shouldn't it?

The outing does not have geometry, because the route does not have one. And the route does not have a geometry, because it does not have a main waypoint.

asaunier:
OK thanks for precising.
I guess there is the same problem with the associated route? It has no geom at all?

Of course we could change the code mentioned by @ginold but I am not sure it is the real problem.
The real problem is that outings or routes may have no geometry.

If the route has no geometry and no main WP, would it be possible to fallback and use geometries of associated WP?

tsauerwein:
> If the route has no geometry and no main WP, would it be possible to fallback and use geometries of associated WP?

That would be possible.

But as long as there are waypoints without geometry, we have to expect routes and outings without geometry.

[...]

PR #960, opened 2016-11-22
fix create outing from route

ginold:
related to #959

`ngeoLocation.hasFragmentParam` stopped to work correctly, maybe some ngeo updates have changed the method?

### DIFF
--- a/c2corg_ui/static/js/outingediting.js
+++ b/c2corg_ui/static/js/outingediting.js
@@ -71,8 +71,8 @@ app.OutingEditingController = function($scope, $element, $attrs, $http,
 
   // allow association only for a new outing to existing route
   if (ngeoLocation.hasFragmentParam('r')) {
-    var urlParam = {'routes': ngeoLocation.getFragmentParam('r')};
-    appApi.getDocumentByIdAndDoctype(urlParam['routes'], 'r', appLang.getLang()).then(function(doc) {
+    var routeId = parseInt(ngeoLocation.getFragmentParam('r'), 10);
+    appApi.getDocumentByIdAndDoctype(routeId, 'r', appLang.getLang()).then(function(doc) {
       this.documentService.pushToAssociations(doc.data['routes'].documents[0], 'routes', true);
     }.bind(this));
   }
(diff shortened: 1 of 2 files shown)


tsauerwein:
`ngeoLocation.hasFragmentParam` still should work fine. The problem is that the link (e.g. http://www.demov6.camptocamp.org/outings/add?r=822428) is wrong. The link should be `...#r=....`.

ginold:
It is true that it works with `#` instead of `?`, however I like this code more than the previous.
Any opinion, @asaunier ?

tsauerwein:
The reason we are using fragment parameters is that otherwise a new cache entry is created for each document. Fragment parameters are not send to the server, so the request to get the edit page will be the same for all documents of the same type. And this will one create one entry for each document type in the cache.

asaunier:
> Any opinion, @asaunier ?

We have to be consistent and use ``#`` instead of ``?``.

ginold:
ok! I replaced `?` with `#` and I'm keeping the code (still works fine and looks better).

c2corg_ui/static/js/outingediting.js
```diff
@@ -69,10 +69,11 @@ app.OutingEditingController = function($scope, $element, $attrs, $http,
    */
   this.differentDates;
 
+
   // allow association only for a new outing to existing route
-  if (ngeoLocation.hasFragmentParam('r')) {
-    var urlParam = {'routes': ngeoLocation.getFragmentParam('r')};
-    appApi.getDocumentByIdAndDoctype(urlParam['routes'], 'r', appLang.getLang()).then(function(doc) {
+  if (ngeoLocation.getParam('r')) {
```
tsauerwein: Shouldn't you be using `getFragmentParam`?

ginold:
I corrected it in the second you wrote this comment :P

ginold:
@asaunier  ok for merging?

c2corg_ui/static/js/outingediting.js
```diff
@@ -69,10 +69,11 @@ app.OutingEditingController = function($scope, $element, $attrs, $http,
    */
   this.differentDates;
 
+
```
asaunier: ?

Issue #836, opened 2016-11-03
Permissions to add/remove associations in a detail view page


asaunier:
As for now it seems that:
- [x] anyone (if auth) may add or remove a route association in an outing, even though they are not associated to the outing or moderators (clearly not OK)
- [x] anyone (if auth) may remove an association from a collaborative document (ditto for personal docs I guess). Is it OK?

What about outing associations in image detail view pages? See https://github.com/c2corg/v6_ui/pull/820

See also https://github.com/c2corg/v6_ui/issues/694

fbunoz:
> anyone (if auth) may remove an association from a collaborative document (ditto for personal docs I guess). Is it OK?

It is not OK.
Only moderators can remove an association from a collaborative document.
Or we can allow removing an association few minutes after a member add an association.

stef74:
@fjacon @ginold When it is planned ?

ginold:
I'm working on it right now, it should be `ready for demo` today or tomorrow.

stef74:
@ginold Wgat is the status ?

ginold:
Waiting for review and for answers to a few questions. Maybe you can answer those questions in the meantime? see the description of https://github.com/c2corg/v6_ui/pull/961

loicperrin:
I cannot test this because I am moderator and impossible to create a new account for testing (see #1115 )

fbunoz:
I tested with my non modo account : it seems correct, I can't edit or remove an association on an outings of another user, on a route or an article.

PR #961, opened 2016-11-22
Associations permissions

ginold:
related to #836

+ hide the association tool if user is not associated to the outing or isn't moderator
+ don't let the user delete an association from a collaborative document, unless he's a moderator

So for example you cannot delete any association from waypoints, routes, books, collaborative articles and collaborative images. Only if you are moderator - it is allowed.

Still few unclear questions: 

+ Outing is a **non collaborative** document - therefore the associated articles could be deleted -> should it be the case? (i think not)
+ If i'm the author or participant of an outing should I be able to delete any associations I want?
+ what about associations in personal/non-collaborative articles and images?
+ others?

please test and tell me if it's going in the right direction! I think it's pretty confusing with all the exceptions we have to handle in the same time in multiple places.

### DIFF
--- a/c2corg_ui/static/js/auth/authservice.js
+++ b/c2corg_ui/static/js/auth/authservice.js
@@ -112,17 +112,16 @@ app.Authentication.prototype.hasEditRights = function(doctype, options) {
   if (this.isModerator()) {
     return true;
   }
-
   if (doctype === 'outings') {
     return this.hasEditRightsOuting_(options['users']);
-  }
-
-  if (doctype === 'images') {
+  } else if (doctype === 'images') {
     return this.hasEditRightsImage_(options['imageType'], options['imageCreator']);
-  }
-
-  if (doctype === 'profiles') {
+  } else if (doctype === 'profiles') {
     return this.userData.id === options['user_id'];
+  } else if (doctype === 'articles') {
+    return this.hasEditRightsArticle_(options['articleType'], options['authorId']);
+  } else if (doctype === 'xreports') {
+    return false;
   }
 
   return true;
@@ -131,20 +130,16 @@ app.Authentication.prototype.hasEditRights = function(doctype, options) {
 
 /**
  * Checks if the current user has rights to access/edit the outing.
- * @param {!Array<string> | !string} users
+ * @param {string | Array<string>} users
  * @return {boolean}
  * @private
  */
 app.Authentication.prototype.hasEditRightsOuting_ = function(users) {
-  var userid = this.userData.id;
-  var u;
+  var u = typeof users === 'string' ? JSON.parse(users) : users;
 
-  if (typeof users === 'string') {
-    u = JSON.parse(users);
-  }
   if (u) {
     for (var i = 0; i < u.length; i++) {
-      if (userid === u[i].document_id) {
+      if (this.userData.id === u[i].document_id) {
         return true;
       }
     }
@@ -164,6 +159,22 @@ app.Authentication.prototype.hasEditRightsImage_ = function(imageType, creator) 
     return true;
   } else {
     return this.userData.id === creator;
+  }
+};
+
+
+/**
+ * Checks if the current user has rights to access/edit the article.
+ * @param {string} articleType
+ * @param {string} authorId
+ * @return {boolean}
+ * @private
+ */
+app.Authentication.prototype.hasEditRightsArticle_ = function(articleType, authorId) {
+  if (articleType === 'collab') {
+    return true;
+  } else {
+    return this.userData.id === parseInt(authorId, 10);
   }
 };
 
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -1,38 +1,16 @@
 <%!
-    from c2corg_ui.templates.utils import get_attr
+    from c2corg_ui.templates.utils import get_attr, is_collaborative, is_personal, get_doc_type
     from json import dumps, loads
     from shapely.geometry import asShape
     from math import floor
-    from c2corg_ui.views.document import get_slug, is_collaborative
+    from c2corg_ui.views.document import get_slug
 %>
 
 <%def name="get_licence(doc)">\
   <%
-      doctype = doc['type']
-      collaborative = False
-      personal = False
-      copyright = False
-
-      if doctype == 'o' or doctype == 'u' or doctype == 'x':
-          personal = True
-      elif doctype == 'w' or doctype == 'r' or doctype == 'b' or doctype == 'a':
-          collaborative = True
-
-      elif doctype == 'c':
-          if doc['article_type'] == 'personal':
-              personal = True
-          else:
-              collaborative = True
-
-      elif doctype == 'i':
-          if doc['image_type'] == 'personal':
-              personal = True
-          elif doc['image_type'] == 'copyright':
-              copyright = True
-          else:
-              collaborative = True
+      personal = is_personal(doc)
+      collaborative = is_collaborative(doc) if not personal else False
   %>
-
   <div class="name-icon-value" ng-click="detailsCtrl.openTab($event)">
     <h4><span translate>Licence</span><span class="glyphicon glyphicon-menu-right"></span></h4>
     <div class="name-icon">
@@ -68,14 +46,6 @@
       </article>
     </span>
   </div>
-</%def>
-
-<%def name="show_attr(obj, key, parse=True)">\
-  <%
-      attr = get_attr(obj, key, md=parse, bb=parse)
-      attr = attr if attr else ''
-  %>
-  ${attr | n}\
 </%def>
 
 <%def name="show_attr(obj, key, parse=True)">\
@@ -207,11 +177,30 @@
   % endif
 </%def>
 
+<%def name="test_association_rights(document)">\
+  % if is_collaborative(document):
+    userCtrl.auth.isAuthenticated()
+    && userCtrl.auth.isModerator()
+  % elif is_personal(document):
+    userCtrl.hasEditRights(
+        '${get_doc_type(document['type'])}',
+        {'users': ${dumps(document['associations']['users'])},
+      % if document['type'] == 'c':
+        'articleType': '${document['article_type']}',
+        'authorId': '${document['author']['user_id']}'
+      % elif document['type'] == 'i':
+        'imageType': '${document['image_type']}',
+        'imageCreator': '${document['creator']['user_id']}',
+      % endif
+    })
+  % endif
+</%def>
+
+
 <%def name="show_associated_articles(document, type='articles')">\
   <%
     articles = document['associations']['articles']
     associations = 'detailsCtrl.documentService.document.associations.' + type
-    isCollaborative = is_collaborative(document)
   %>
    ${add_seo_association_links(articles, 'articles')}
     <section ng-show="${associations}.length > 0">
@@ -223,12 +212,8 @@
         <div class="list-item" ng-repeat="doc in ${associations}" ng-class="{'new-item': doc['new']}">
           <app-card app-card-doc="doc"></app-card>
           % if type == 'articles':
-            <app-delete-association child-doctype="articles" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.auth.isAuthenticated()
-              % if isCollaborative:
-                && userCtrl.auth.isModerator()
-              % endif
-              "></app-delete-association>
+              <app-delete-association child-doctype="articles" parent-id="${document['document_id']}" child-id="doc.document_id"
+                ng-if="${test_association_rights(document)}"></app-delete-association>
           % endif
         </div>
       </div>
@@ -240,7 +225,6 @@
   <%
     books = document['associations']['books']
     associations = 'detailsCtrl.documentService.document.associations.' + type
-    isCollaborative = is_collaborative(document)
   %>
    ${add_seo_association_links(books, 'books')}
     <section ng-show="${associations}.length > 0">
@@ -252,11 +236,8 @@
         <div class="list-item" ng-repeat="doc in ${associations}" ng-class="{'new-item': doc['new']}">
           <app-card app-card-doc="doc"></app-card>
           % if type == 'books':
-            <app-delete-association child-doctype="books" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
-          % endif
-          % if isCollaborative:
-            && userCtrl.auth.isModerator()
+              <app-delete-association child-doctype="books" parent-id="${document['document_id']}" child-id="doc.document_id"
+                ng-if="${test_association_rights(document)}"></app-delete-association>
           % endif
         </div>
       </div>
@@ -269,7 +250,6 @@
     routes = document['associations']['all_routes']['documents'] if type == 'all_routes' else document['associations']['routes']
     associations = 'detailsCtrl.documentService.document.associations.'
     associations += 'all_routes.documents' if type == 'all_routes' else 'routes'
-    isCollaborative = is_collaborative(document)
   %>
     ${add_seo_association_links(routes, 'routes')}
     <section ng-show="${associations}.length > 0">
@@ -282,15 +262,11 @@
           <app-card app-card-doc="doc"></app-card>
           % if type == 'routes':
             <app-delete-association child-doctype="routes" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.auth.isAuthenticated()
+              ng-if="${test_association_rights(document)}
               % if document['type'] == 'o':
-                && userCtrl.hasEditRights('outings', ${ {"users": dumps(document['associations']['users']) }})
-                && ${associations}.length > 1
+                  && ${associations}.length > 1
               % endif
-              % if isCollaborative:
-                && userCtrl.auth.isModerator()
-              % endif
-              "></app-delete-association>
+            "></app-delete-association>
           % endif
         </div>
         % if type == 'all_routes' and document['associations']['all_routes']['total'] > len(routes):
@@ -308,7 +284,6 @@
 <%def name="show_associated_waypoints(document, type='waypoints', showDelete=True)">\
   ## Possible values for type: waypoints, waypoint_children
   <%
-      isCollaborative = is_collaborative(document)
       associations = 'detailsCtrl.documentService.document.associations.' + type
   %>
   ${add_seo_association_links(document['associations'][type], 'waypoints')}
@@ -328,15 +303,12 @@
            ng-repeat="doc in ${associations} | orderBy: 'waypoint_type'">
            <app-card app-card-doc="doc"></app-card>
         % if showDelete:
-        <app-delete-association child-doctype="${type}" parent-id="${document['document_id']}" child-id="doc.document_id"
-          ng-if="userCtrl.auth.isAuthenticated()
-          % if document['doctype'] == 'routes':
-              && ${associations}.length > 1
-          % endif
-          % if isCollaborative:
-            && userCtrl.auth.isModerator()
-          % endif
-        "></app-delete-association>
+            <app-delete-association child-doctype="${type}" parent-id="${document['document_id']}" child-id="doc.document_id"
+              ng-if="${test_association_rights(document)}
+              % if document['doctype'] == 'routes':
+                  && ${associations}.length > 1
+              % endif
+            "></app-delete-association>
         % endif
       </div>
     </div>
@@ -380,13 +352,8 @@
         % endif
         <div class="list-item outings" ng-repeat="doc in detailsCtrl.documentService.document.associations.${'recent_outings.documents' if isRecentOutings else 'outings'}">
           <app-card app-card-doc="doc"></app-card>
-          % if not isRecentOutings and document['doctype'] == 'images':
-            <%
-                options = {"imageType": document["image_type"], "imageCreator": document["creator"]["user_id"]}
-            %>
             <app-delete-association child-doctype="o" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.hasEditRights('outings', ${options})"></app-delete-association>
-          % endif
+              ng-if="${test_association_rights(document)}"></app-delete-association>
         </div>
         % if isRecentOutings and total > len(outings) :
           <%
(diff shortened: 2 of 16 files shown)


ginold:
@asaunier  can you have a look please?

asaunier:
> Outing is a non collaborative document - therefore the associated articles could be deleted -> should it be the case? (i think not)

It would make sense that associated users may remove associations, the same way they can add associations.

> If i'm the author or participant of an outing should I be able to delete any associations I want?

Yes, see my answer above.

> what about associations in personal/non-collaborative articles and images?

It would be consistent that associations in any personal document are managed the same way, see outings.

> others?

?
For other documents, ie. collaborative docs, the current position of the client is that only moderators can remove their associations.

c2corg_ui/templates/outing/view.html
```diff
@@ -127,7 +127,7 @@ <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" d
   % if not version:
   <div class="view-details-associations col-xs-12  associations">
     <span class="lead">
-      <div ng-show="userCtrl.auth.isAuthenticated()" class="add-association">
+      <div ng-show="userCtrl.auth.isAuthenticated() && userCtrl.hasEditRights('outings', ${ {"users": dumps(outing['associations']['users']) }})">
```
asaunier: OK, makes sense. 
Just curious: the ``add-association`` class is no longer needed?

c2corg_ui/templates/helpers/view.html
```diff
@@ -70,6 +70,13 @@
   </div>
 </%def>
 
+<%def name="show_attr(obj, key, parse=True)">\
```
asaunier: Why do you duplicate this function? It's already defined a few lines below.

c2corg_ui/templates/helpers/view.html
```diff
@@ -216,7 +224,11 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
           <app-card app-card-doc="doc"></app-card>
           % if type == 'articles':
             <app-delete-association child-doctype="articles" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+              ng-if="userCtrl.auth.isAuthenticated()
+              % if isCollaborative:
```
asaunier: Why not simply write
```
% if is_collaborative(document):
```
?
I don't understand why you have introduced the ``isCollaborative`` variable? Ditto for the other changes below.

To avoid multilines in the ``ng-if`` you could write instead ``${'&& userCtrl.auth.isModerator()' if is_collaborative(document) else ''}``.

asaunier: But your test seems a bit weird to me: if the document is collaborative, only moderators can remove the associations, but if the document is not collaborative (ie. personal), any authenticated user can remove the associations!?

c2corg_ui/views/document.py
```diff
@@ -431,6 +431,24 @@ def get_slug_for_title(title):
...
+
+    elif type == 'c':
+        if doc['article_type'] == 'collab':
+            return True
+
+    elif type == 'i':
+        if doc['image_type'] == 'collaborative':
+            return True
+
+    else:
```
asaunier: You don't need ``else`` here. Simply write ``return False`` as the last line of the function. If the conditions above are true this last line will not be executed.

Actually the code is incorrect with ``else``, because for images and articles, if the type check are not ``True`` no value is returned.

Please also remove all the extra and useless blank lines.

c2corg_ui/views/document.py
```diff
@@ -431,6 +431,24 @@ def get_slug_for_title(title):
     return slug
 
 
+def is_collaborative(doc):
```
asaunier: Since this function is only used in the templates, please moved it to https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/utils/__init__.py

c2corg_ui/templates/outing/view.html
asaunier: Concerning the ``hasEditRights`` here, please see issue https://github.com/c2corg/v6_ui/issues/1036

ginold: Class: I removed it and saw no changes. I also removed the dedicated css code.

c2corg_ui/templates/helpers/view.html
ginold: That was one of my questions you answered above!

ginold:
Who can edit associations of a personal article ? 
The persons who are in `associations.users` ? (or rather THE only person, i haven't seen more than 1)

Who can edit associations of books? there are no 'personal' books?  -> moderators only

Who can edit associations of personal images? -> moderators + author (+ associated users? but I haven't seen any).

asaunier:
> Who can edit associations of a personal article ?
The persons who are in associations.users ? (or rather THE only person, i haven't seen more than 1)

It's not possible to associate users to a personal article. I guess only the creator of the article is listed in ``associations.users``.
Then yes, it makes sense to test if the current visitor is in ``associations.users`` to figure out if they should be allowed to remove associations.

> Who can edit associations of books? there are no 'personal' books? -> moderators only

Yes, all books are collaborative documents.

> Who can edit associations of personal images? -> moderators + author (+ associated users? but I haven't seen any).

Author + moderators (same than for personal articles). I don't think it is possible to associate a user to an image.

tsauerwein:
> It's not possible to associate users to a personal article. I guess only the creator of the article is listed in associations.users.
> Then yes, it makes sense to test if the current visitor is in associations.users to figure out if they should be allowed to remove associations.

Note that the API does not automatically add an association for the creator of an article. And I am not sure if this is the case for all existing articles?

ginold:
It would maybe make sense to display the author of an article, as listed in `associations.users`? (if so, I will create a dedicated issue)

tsauerwein:
We could return the creator of an article, like for images. Because I still don't think that there is an association between an article and the creator for each article. To confirm.

ginold:
In fact that would be great. I created an article and I cannot edit it, there is no way to tell who can edit it - no associated users, no authors.

c2corg_ui/templates/helpers/view.html
```diff
@@ -215,8 +188,25 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
...
-            <app-delete-association child-doctype="articles" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+              <app-delete-association child-doctype="articles" parent-id="${document['document_id']}" child-id="doc.document_id"
+                ng-if="userCtrl.auth.isAuthenticated()
+              % if is_collaborative(document):
+                  && userCtrl.auth.isModerator()
+
+              % elif is_personal(document):
+                  && userCtrl.auth.isModerator()
+                  || userCtrl.hasEditRights
```
ginold: I think i can remove `userCtrl.auth.isModerator()` wherever there is `userCtrl.hasEditRights` because it also calls isModerator() inside itself.

ginold:
Still WIP, a small update with articles 'author' is still needed here (UI, not API).

But from what i've tried, it should work like this:

+ moderators can add and remove associations from any document, collaborative or personal
+ a simple user can add associations to all collaborative documents
+ a simple user cannot remove any associations from any collaborative document
+ a simple user can add associations to a personal document if he's the owner (image, article) or participant (outing)
+ a simple user can remove associations from a personal document if he's the owner (image, article) or participant (outing)

This one is SUPER HARD to test: check all document types with all possible other type of associated documents inside, with documents having multiple `document_types` (image_type, article_type), and all that in moderator and normal-user-mode. As a consequence, I'm afraid I could forget a special case, an exception or something.

When the PR is ready, if someone could test it as well I would be thankful :P

c2corg_ui/templates/helpers/view.html
```diff
@@ -45,7 +19,7 @@
     </div>
     <span class="detail-text accordion">
       <article class="value">
-        % if personal:
+        % if is_personal(doc):
```
asaunier: You have already run ``is_personal`` a few lines above.
I would recommend to use variables to store the results once for all at the beginning of the function:
```
personal = is_personal(doc)
collaborative = is_collaborative(doc) if not personal else False
```

c2corg_ui/templates/helpers/view.html
asaunier: As you suggest it, it makes sense to remove ``userCtrl.auth.isModerator()`` here since ``userCtrl.hasEditRights`` already test for moderators.

c2corg_ui/templates/helpers/view.html
```diff
@@ -239,8 +229,23 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
         <div class="list-item" ng-repeat="doc in ${associations}" ng-class="{'new-item': doc['new']}">
           <app-card app-card-doc="doc"></app-card>
           % if type == 'books':
-            <app-delete-association child-doctype="books" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+              <app-delete-association child-doctype="books" parent-id="${document['document_id']}" child-id="doc.document_id"
+                ng-if="userCtrl.auth.isAuthenticated()"></app-delete-association>
+          % endif
+          % if is_collaborative(document):
+              && userCtrl.auth.isModerator()
```
asaunier: I don't understand this code: to what will ``&& userCtrl.auth.isModerator()`` be added?

c2corg_ui/templates/helpers/view.html
```diff
@@ -239,8 +229,23 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
...
+          % elif is_personal(document):
+              && userCtrl.auth.isModerator()
+              || userCtrl.hasEditRights
+                  (
```
asaunier: ``hasEditRights(``

c2corg_ui/templates/helpers/view.html
ginold: probably a bad copy-paste

c2corg_ui/templates/helpers/view.html
ginold: As you can see, this part is repeated 4 times for different document types, I think I should find a way to write a small function that will output this.

c2corg_ui/templates/helpers/view.html
```diff
@@ -264,11 +269,25 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
...
               % if document['type'] == 'o':
-                && userCtrl.hasEditRights('outings', ${ {"users": dumps(document['associations']['users']) }})
-                && ${associations}.length > 1
+                  && ${associations}.length > 1
+              % endif
+              % if is_collaborative(document):
+                  && userCtrl.auth.isModerator()
+              % elif is_personal(document):
+                  && userCtrl.auth.isModerator()
+                  || userCtrl.hasEditRights
```
asaunier: ``hasEditRights(``

c2corg_ui/templates/helpers/view.html
asaunier: > As you suggest it, it makes sense to remove ``userCtrl.auth.isModerator()`` here since ``userCtrl.hasEditRights`` already test for moderators.

In addition it would the test a bit less confusing. Because 
```
userCtrl.auth.isAuthenticated() && userCtrl.auth.isModerator() || userCtrl.hasEditRights()
```
probably means "(isAuth AND isModerator) OR hasEditRights".
OK it would work because "hasEditRights" also tests if "isModerator" (and "isAuth"?). But that looks like a potentially buggy condition...
Same problem for the other conditions a bit further.

c2corg_ui/templates/image/view.html
```diff
@@ -116,7 +116,11 @@ <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" d
   % if not version:
   <div class="view-details-associations col-xs-12 associations">
     <span class="lead">
-      <div ng-show="userCtrl.auth.isAuthenticated()" class="add-association">
+      <div class="add-association" ng-show="userCtrl.auth.isAuthenticated()
+        % if image['image_type'] == 'peronal':
```
asaunier: typo! ``peronal`` => ``personal``

c2corg_ui/templates/image/view.html
```diff
@@ -116,7 +116,11 @@ <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" d
...
+            && userCtrl.hasEditRights('images', {'iimageType': 'personal', 'imageCreator': ${image['creator']['user_id']})
```
asaunier: Typo again: ``iimageType`` => ``imageType``

c2corg_ui/templates/utils/__init__.py
```diff
@@ -26,3 +26,52 @@ def get_attr(obj, key, md=True, bb=True):
...
+
+
+def is_collaborative(doc):
+    type = doc['type']
+
+    if type == 'r' or type == 'w' or type == 'a' or type == 'b':
+        return True
+    elif type == 'c':
+        if doc['article_type'] == 'collab':
+            return True
```
asaunier: Could be written
```
elif type == 'c':
    return doc['article_type'] == 'collab'
elif type == 'i':
    return doc['image_type'] == 'collaborative'
return False
```

c2corg_ui/templates/utils/__init__.py
```diff
@@ -26,3 +26,52 @@ def get_attr(obj, key, md=True, bb=True):
...
+      doctype = doc['type']
+
+      if doctype == 'o' or doctype == 'u' or doctype == 'x':
+          return True
+      elif doctype == 'c':
+          if doc['article_type'] == 'personal':
+              return True
+      elif doctype == 'i':
+          if doc['image_type'] == 'personal':
+              return True
```
asaunier: Could be written
```
elif type == 'c':
    return doc['article_type'] == 'personal'
elif type == 'i':
    return doc['image_type'] == 'personal'
return False
```

ginold:
Can anyone recall me if there are some restrictions concerning the removal of an association:

` routes, all_routes` or `waypoints, waypoint_children` -> is there any that you should not be able to remove, or you can remove any of them?

asaunier:
Any association can be removed except that:
* outings must at least have one associated route and at least one associated user
* routes must at least have one associated WP

ginold:
I'm asking because when you look at https://github.com/c2corg/v6_ui/blob/adbd24b30c0338fd24206e6f22f39665d9094fd6/c2corg_ui/templates/helpers/view.html#L271

you can see that there is the condition `% if type == 'routes'` for `app-delete-association`, as if it were all_routes, you should not be able to delete? -> some old code?

asaunier:
I think it is not allowed to un-associate a route from a WP page (and not possible to associate a route from a WP page either).

Because in WP pages you may also see routes associated to associated WP (which means "with no direct association link").

ginold:
And is it possible to remove an outing in general? I don't think so?

asaunier:
You mean "to deassociate an outing from a route"?
It's not possible from the route page. You can only deassociate a route from the outing page (if there is at least one associated route left).

ginold:
I mean I'm a moderator and should I be able to unassociate an outing from any doc?

asaunier:
I misread your question: it's not possible to associate an outing to another outing. So not possible to remove an outing-outing association either.

If you are a moderator you can do whatever action if the rules described here https://github.com/c2corg/v6_ui/pull/961#issuecomment-264167074 are followed.

ginold:
Ok, I think I got it (hopefully) : you should not be able to unassociate Recent Outings because it's not a 'real' association - it's just a recent outing that is linked, but not really directly ?

ginold:
That's what I understand from https://github.com/c2corg/v6_ui/blob/adbd24b30c0338fd24206e6f22f39665d9094fd6/c2corg_ui/templates/helpers/view.html#L393

asaunier:
If you cannot add an association in a given page, you cannot remove it either.

> you should not be able to unassociate Recent Outings because it's not a 'real' association - it's just a recent outing that is linked, but not really directly ?

Outings shown in WP pages are not directly associated to the WP:
http://www.demov6.camptocamp.org/waypoints/823915/fr/calanque-du-devenson-face-sw
There are routes in between.

Outings shown in route pages are directly associated to the routes.
Since we don't want that an outing can be dissociated from all its routes, it's not possible to dissociate an outing from the route detail view. 
You have to remove the route association from within the outing detail view (where you have all associated routes listed).

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -131,20 +128,17 @@ app.Authentication.prototype.hasEditRights = function(doctype, options) {
...
  * Checks if the current user has rights to access/edit the outing.
- * @param {!Array<string> | !string} users
+ * @param {string} users
  * @return {boolean}
  * @private
  */
 app.Authentication.prototype.hasEditRightsOuting_ = function(users) {
-  var userid = this.userData.id;
   var u;
+  typeof users === 'string' ? u = JSON.parse(users) : u = users;
```
asaunier:
```
var u = typeof users === 'string' ? JSON.parse(users) : users;
```

c2corg_ui/templates/helpers/view.html
```diff
@@ -200,6 +177,25 @@ <h4 translate>No maps associated</h4>
   % endif
 </%def>
 
+<%def name="delete_association_rights(document)">\
```
asaunier: This function name is not consistent: it does not delete the association rights. By the way it could be used for something else then removing associations (perhaps for adding associations?).
``test_association_rights``?

asaunier:
@ginold please run "make check" in your instance. It seems you have errors in your code. For instance:
```
  File "/home/alex/Desktop/v6_ui/c2corg_ui/templates/profile/helpers/view.html", line 34, in render_show_profile_float_buttons
    <%
TypeError: %d format: a number is required, not dict
```

ginold:
I got rid of of the profile problem, now i have this one :
```
  File "/home/agina/NetBeansProjects/C2C.org/v6_ui/c2corg_ui/templates/helpers/view.html", line 190, in render_test_association_rights
    'authorId': '${document['author']['user_id']}'
KeyError: 'author'
```

Is it maybe because there was a recent update of the API that now returns author, but it's not up to date with the UI tests/models ? It's just a make check error, everything works fine while browsing.

tsauerwein:
> It's just a make check error,...

You can update the test data here: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/tests/views/data/article.json

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -131,22 +130,16 @@ app.Authentication.prototype.hasEditRights = function(doctype, options) {
 
 /**
  * Checks if the current user has rights to access/edit the outing.
- * @param {!Array<string> | !string} users
+ * @param {string} users
```
asaunier: This type seems incorrect.
The first line of the function is
```
var u = typeof users === 'string' ? JSON.parse(users) : users;
```
=> this means that ``users`` is not always a string. It seems according to 
```
for (var i = 0; i < u.length; i++) {
```
that it may be an array.

``@param {string|Array.<string>} users`` seems to pass with Closure.

asaunier:
As a normlal contributor: OK for the removal tool => shown only where expected.

But there are a few problems remaining with the association tool shown or not shown as expected:
* [ ] association tool is missing in a personal article I have just created (with no moderator rights) => it's fine that the tool is not available for personal articles of other contributors, but it should appear when the current user is the author of the article. 
* [ ] in an personal article (+other doc types), if no doc is associated and if the association tool is not shown, one can see a blank strip => should be removed.
* [ ] in a personal image (eg. /images/736882/fr/sortie-l5-apres-le-cade) I have able to add association

@ginold Do you think those problems could be fixed today? If not, I would suggest to merge the PR and create a new issue/PR for the problems above. 
Please not the JSDoc issue as well (minor but easy to fix).

ginold:
I think I can handle that!

asaunier:
Anyway thanks a lot for the great work :)

ginold:
I updated the PR, it should fix the issues you listed.

+ in personal documents, if no associations and no rights -> don't show the empty block
+ in personal documents, if no associations and rights -> show the association tool

c2corg_ui/templates/helpers/view.html
```diff
@@ -200,6 +177,39 @@ <h4 translate>No maps associated</h4>
...
+        'authorId': '${document['author']['user_id']}'
+      % elif document['type'] == 'i':
+        'imageType': '${document['image_type']}',
+        'imageCreator': '${document['creator']['user_id']}',
+      % endif
+    })
+  % endif
+</%def>
+
+<%def name="check_if_any_associations(doc)">\
```
asaunier: Argh
Then you get in the HTML something like
```
ng-if="true || userCtrl.hasEditRights( 'articles', { 'articleType': 'personal', 'authorId': 169325 })"
```
Berk! :P 

Please rather create a real python function here https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/utils/__init__.py

and call it in the mako helper to figure if the block should added or not.

c2corg_ui/templates/image/view.html
```diff
@@ -114,9 +114,14 @@ <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" d
 
   ## ASSOCIATIONS
   % if not version:
-  <div class="view-details-associations col-xs-12 associations">
+  <div class="view-details-associations col-xs-12 associations"
+       ng-if="${check_if_any_associations(image)} || userCtrl.hasEditRights( 'images', { 'imageType': '${image['image_type']}', 'imageCreator': ${image['creator']['user_id']} })">
```
asaunier: Not OK for me.

```
<div class="view-details-associations col-xs-12 associations"
  % if not check_if_any_associations(image):
    ng-if="userCtrl.hasEditRights()"
  % endif
```

c2corg_ui/templates/outing/view.html
```diff
@@ -125,9 +125,10 @@ <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" d
 
   ## ASSOCIATIONS
   % if not version:
-  <div class="view-details-associations col-xs-12  associations">
+  <div class="view-details-associations col-xs-12  associations"
+         ng-if="${check_if_any_associations(outing)} || userCtrl.hasEditRights('outings', ${ {'users': dumps(outing['associations']['users']) } } )">
```
asaunier: See above.

asaunier:
> in personal documents, if no associations and no rights -> don't show the empty block

Are you sure it works all the time. With http://localhost:6555/articles/794178/fr/himalaya-du-nepal-treks-et-expes-un-autre-regard-les-heureuses-experiences I can still see the blank stripe (with ``ng-if="true || ..."``).

> in personal documents, if no associations and rights -> show the association tool

ok

ginold:
are sure your api is up to date? it was updated yesterday with 'author' and it will determine if your user id == author id. I don't see anything in the associations on the link you sent.

asaunier:
My API was indeed outdated. I will test again a bit later (doing the rebase for my PRs). This way you have time to make the changes I have suggested :)

ginold:
I did bring some corrections. If you think there are some more issues and this is urgent, please check it and update (this should not be longer than 2 mins for you... :muscle: ).  It worked pretty much ok on my tests but I was a little rushing (gotta really go now -> train !).

That said, schönes Wochenende! and thank you for reviewing!

asaunier:
@ginold thanks for your patience! :)
I have rebased your branch to fix the error with Travis and added a few minor improvements.
I merge the PR because it might be useful for other issues.

PR #963, opened 2016-11-22
[WIP] Add angular loading bar

dkocich:
It still needs some work and optimisation. It looks like the loader bar does not work correctly with current implementation of ng-cloak which hides it. It looks that now the loader bar also loads too late or is restricted only to some requests instead of all available.
Related to https://github.com/c2corg/v6_ui/issues/593 , https://github.com/c2corg/v6_ui/pull/925

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ LIBS_JS_FILES += \
     node_modules/jquery/dist/jquery.min.js \
     node_modules/bootstrap-slider/dist/bootstrap-slider.min.js \
     node_modules/angular/angular.min.js \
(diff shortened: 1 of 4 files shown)


asaunier:
> It looks like the loader bar does not work correctly with current implementation of ng-cloak which hides it.

Perhaps our current use of ``ng-cloak`` is not optimal anyway: white pages now appear when browsing among pages. No suprise since ``ng-cloak`` is added on the ``html`` element, but I thought that when I have tested this change, the previous page was removed only when the new one had finished loading.

Edit: actually it works like this with FF, not with Chromium (blank pages in the meantime).

PR #985, opened 2016-11-24
configuration buttons mobile layout

brunobesson:
Move text to the right of the buttons

(note that I have tested it in the browser but cannot test it with compiled CSS, so I might have a bug, but it's easy to test)

### DIFF
--- a/less/homepage.less
+++ b/less/homepage.less
@@ -128,15 +128,12 @@
     }
     @media @phone {
       width: 100%;
-      .float-button-container {
-        width: 120px;
-      }
       .float-buttons, .float-button {
         margin-right: 5px;
       }
       .float-button-text {
-        width: 120px;
-        margin-top: 5px;
+        display: inline-block;
+        width: 80px;
       }
     }
   }


gottferdom:
Before
before

After
after

asaunier:
@ginold a review for you :)

@gottferdom That looks nice to me indeed! Perhaps it's not very consistent with the positioning of the labels of the usual float buttons (placed under the icons). But anyway those feed buttons are rendered differently in bigger screens, so perhaps it's not a problem.

gottferdom:
@asaunier yes, indeed, and I have opened another issue for this case. The difference being that the text is quite long (on two lines for both buttons), and there are only 2 buttons. So it make sense to me to do it this way.
The default case, with 4 buttons also has problems (see #989).
The 3 buttons when editing (preview, OK, cancel) do not have text.
Maybe I am missing other case.
So there is no consistent behaviour either :laughing:

asaunier:
Looks good to me. I'm OK for merging. @ginold OK for you?

ginold:
If it doesn't impact for the bad the other float-buttons (view page, editing page) in desktop/mobile mode and in other languages - then go ahead !

PR #990, opened 2016-11-24
Discourse menu

ginold:
fixes https://github.com/c2corg/v6_forum/issues/41 

I should probably do it in the v6_forum repository, but I think it was better to add those files to the UI.
I cannot test it on the phone - it simply does not appear, I don't know if I have to set this style to default in order to make it appear.

The menu is _nearly_ the same, some icons were missing.

The only concern I have are the social-links icons: on the UI we use a font with those images, here I get them from random sites -> I have no idea how to add a custom font, the only way i would know in this case is to download it with `@import url()` but it would mean we have to export our font to some CDN or something.

screenshot from 2016-11-24 11-57-29
screenshot from 2016-11-24 11-57-39


asaunier:
> The only concern I have are the social-links icons: on the UI we use a font with those images, here I get them from random sites

That's not an option :P 
What if those sites remove the icons?

What about simply adding those PNG files to https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/img and then getting them from the UI?

Concerning the CSS files, what about using Less as usual and include a built CSS file? You could do the same than for the "print" styles. See https://github.com/c2corg/v6_ui/tree/master/less-print and https://github.com/c2corg/v6_ui/blob/master/Makefile#L217-L223

asaunier:
> Concerning the CSS files, what about using Less as usual and include a built CSS file?

By the way, can't you use the usual menu styles from https://github.com/c2corg/v6_ui/blob/master/less/sidemenu.less and integrate them in the ``less-discourse``  files?

As much as possible we have to avoid duplicating styles among the UI and the forum!
Menu style changes should be done at a single place, not 2.

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,117 @@
+## HEAD
+<link href="//netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap-glyphicons.css" rel="stylesheet">
+
+## En-tete
+<aside id="sidemenu">
+  <div class="page-menu">
+    <a href="//www.demov6.camptocamp.org">
```
asaunier: Can't we avoid hard coded URLs?
It means those links will have to be hand-changed when setting the menu in the prod forum.
Or at least to have a single place ("baseURL"?)?

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,117 @@
...
+    <div class="menu-open-btn">
+      <i class="glyphicon glyphicon-align-justify"></i>
+    </div>
+  </div>
+  <ul class="nav">
+    <li class="main-menu">
+      <a href="//www.demov6.camptocamp.org">
+        <i class="glyphicon glyphicon-home"></i>
+        <span class="lang en menu-text">Home</span>
+        <span class="lang fr menu-text">Accueil</span>
```
asaunier: How does the i18n work here? Only the current lang span is shown?
Is it the system suggested by @gberaudo in the initial demo menu?
I guess using angular-gettext is too complicated here?

If we stick to this system, translations for the other languages are needed.

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,117 @@
...
+    </li>
+    <li class="main-menu">
+      <a href="//www.demov6.camptocamp.org/routes">
+        <i class="glyphicon glyphicon-arrow-right"></i>
+        <span class="lang en menu-text">Guidebook</span>
+        <span class="lang fr menu-text">Topoguide</span>
+      </a>
+    </li>
+    <li class="main-menu menu-selected">
+      <a href="//forum.demov6.camptocamp.org/">
```
asaunier: Here you could use a relative path to avoid this hard coded URL.

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,117 @@
...
+        <a href="//www.demov6.camptocamp.org/articles/106728">
+          <span class="lang en">Content license</span>
+          <span class="lang fr">Licences de contenu</span>
+        </a>
+      </li>
+    </ul>
+
+    <div class="social-networks">
+      <div class="icons">
+        <a href="https://twitter.com/camptocamporg"><img src="https://c866088.ssl.cf3.rackcdn.com/assets/twiiter40x40.png"></a>
```
asaunier: As I said in a comment before, it's not an option to use icon from other sites.

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,117 @@
+## HEAD
+<link href="//netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap-glyphicons.css" rel="stylesheet">
```
asaunier: Isn't it already imported in the CSS file?

By the way where are the CSS files included? In a config attribute in Discourse?

asaunier: I have no admin rights :P 

So you mean both HTML and CSS codes have to be pasted in a form in the admin interface, right?

c2corg_ui/static/menu-discourse/menu.html
ginold: If i use the static images, for ex `...static/4c851e7/img/facebook.svg` will the middle `4c851e7` not change often? I mean, if it does change it will not impact the displaying of the image?

asaunier: The "uuid" part is not mandatory. Both following URLs work:
http://www.demov6.camptocamp.org/static/img/logo.svg
http://www.demov6.camptocamp.org/static/4120b0e/img/logo.svg

c2corg_ui/static/menu-discourse/menu.html
ginold: exactly.
screenshot from 2016-11-24 13-17-44

ginold:
>  By the way, can't you use the usual menu styles ?
no, sidemenu.less is written with LESS whereas discourse only accepts CSS -> i have to rewrite and adapt.

asaunier:
> no, sidemenu.less is written with LESS whereas discourse only accepts CSS -> i have to rewrite and adapt.

Why not writing "less" files as usual and build the final CSS file with Makefile in the UI? Then you copy-paste it to the Discourse interface.
Have a look at the "print styles" targets in the Makefile.

ginold:
It's already finished so..no need :P and there are a few parts of code that we don't need in discourse anyway.

asaunier:
I think we need it because we don't want to have duplicated styles. If changes are done in the sidemenu in the UI it will require to make the changes in the discourse menu as well. Not a big deal if some styles are not used in the discourse sidemenu. 

It's not complicated, you have the example of the print styles:
https://github.com/c2corg/v6_ui/blob/master/Makefile#L217-L223

asaunier:
The ad banner is missing. It is required, in the forum too.
See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/sidemenu.html#L54-L56

c2corg_ui/static/menu-discourse/menu.html
ginold: I could create a variable BaseURL and have something like

```
    $('a').click(function() {
        $(this).attr('route') ? window.location.href = BaseURL/route :  window.location.href = BaseURL;
    });
```
Angular ng-href would be nice... ;p

asaunier: Would it work with the standard ``<base>`` HTML tag?
https://developer.mozilla.org/fr/docs/Web/HTML/Element/base

asaunier:
What about the discourse logo at the top of your screenshots? It can be removed.
Is it possible that the sidemenu starts at the top of the screen instead of under the discourse header?
In phone screens, the c2corg logo is not displayed at all?

c2corg_ui/static/menu-discourse/menu.html
ginold: it does seem to work. The question is if we prefer to suddenly use this element tag (that we never used anywhere else) in a separate file, or to have `var baseUrl = '//www.demov6.camptocamp.org/';` in the same HTML file?

ginold: > There can be at maximum one <base> element in a document, and it must be inside the <head> element.

So no, there is a dedicated "head" tab in discourse.

There is no problem with my code if i specify that i want to replace links only in `#sidemenu a`.

mgerbaux:
>  Is it possible that the sidemenu starts at the top of the screen instead of under the discourse header?
Sure it would be better if the menu was (or appeared to be) the same on the whole site camptocamp.org.
Moreover, the discourse header make a waste of about 2 cm on the left menu, that will offer less opportunity to display advertising banner on small screens.

fjacon:
question : can we replace the discourse logo by another image ?

mgerbaux:
> question : can we replace the discourse logo by another image ?

@fbunoz customized it on the demo, waiting the final menu : https://forum.demov6.camptocamp.org/latest

ginold:
Pico bello waiting to be merged!
Hopefully you can test: http://forum.demov6.camptocamp.org/?preview-style=e55c4247-d233-4af4-b0a4-b1e52a401f60

+ Added all locales
+ Links' href are created based on a `baseUrl`
+ Advertisement
+ Changed two icons
+ Readme file

The only thigs:

- The logo in the Header redirects to the current page `/` whereas the logo in the sidemenu redirects to `camptocamp.com/routes` -> which one should be left?

- The social-website-icons will display after this PR is merged (it adds the PNG versions of the images).

asaunier:
Nice job :)
(I'm having a look at the code)

> The logo in the Header redirects to the current page / whereas the logo in the sidemenu redirects to camptocamp.com/routes -> which one should be left?

I don't understand. You mean the C2C logo in the top left corner? In your demo it opens http://www.demov6.camptocamp.org/ , not http://www.demov6.camptocamp.org/routes

asaunier:
Can't we get rid of the global horizontal scrollbar?

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,190 @@
...
+        <span class="lang ca menu-text">Associació</span>
+        <span class="lang es menu-text">Asociación</span>
+      </a>
+    </li>
+  </ul>
+
+
+  <footer>
+    <div class="advertisement">
+      <iframe src="#" class="src" url="static/4120b0e/affiz.html" width="160" height="320" frameborder="0" scrolling="no"></iframe>
```
asaunier: Do you really need that ``4120b0e`` in the iframe URL?

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,190 @@
...
+    </div>
+
+    <ul class="footer-nav">
+      <li>
+        <a href="#" url="articles/106727/en/contact-us" class="menu-link">
```
asaunier: Why using the "en" URL for this article? Please simply use ``articles/106727``.

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,190 @@
...
+          <span class="lang en">Contact</span>
+          <span class="lang fr">Contact</span>
+          <span class="lang it menu-text">Contatto</span>
```
asaunier: Why don't you have ``menu-text`` for the "en" and "fr" labels?

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,190 @@
...
+    </div>
+
+  </footer>
+
+  <script>
+
+    var locale = I18n.locale;
+    var menu = document.getElementById('sidemenu');
+    var translations = menu.querySelectorAll('.lang');
+    var baseUrl = '//www.demov6.camptocamp.org/';
```
asaunier: Do you think `` var baseUrl = '//www.demov6.camptocamp.org/';`` could be placed at the beginning of the file? This way it would be more obvious what the ``baseUrl`` is and make admin users easily figure out if it is OK.

c2corg_ui/static/menu-discourse/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The less file (`sidemenu.less`) was written to suit both the UI and Discourse, even though we won't use the exact same lines in both cases, but the code stays mostly the same.
+
+The files in Discourse should be changed under **the hamburger button** -> **admin** -> **customize (personnaliser)** -> **CSS/HTML** -> and choose from the left list **discourse menu** or create a new style.
+
+
+## LESS -> CSS
+In order to compile the usual sidemenu.less file into pure CSS, use the command:
+
+`./node_modules/.bin/lessc less/sidemenu.less  > sidemenu.css`
```
asaunier: Why not using the makefile as usual?

c2corg_ui/static/menu-discourse/menu.html
tsauerwein: Note that `4120b0e` will change when a new version of the UI is deployed!

c2corg_ui/static/menu-discourse/menu.html
ginold: "because i'm only a human after all, and i make mistakes" :notes:

c2corg_ui/static/menu-discourse/README.md
ginold: you mean you prefer to add an additional activity to `make less` that will output this? 
I preferred to use this one instead of make less -> was muuuch faster.

ginold:
>  The hamburger button is not the same than in the UI.

It is indeed not the same because there already is a hamburger icon in discourse (and not sure how to change it).

c2corg_ui/static/menu-discourse/README.md
asaunier: > you mean you prefer to add an additional activity to make less that will output this?

Yes

For me it's not consistent to use the way you suggest: all styles files should be built the same way and with the same command.
If you have made changes to the sidemenu, it means that the makefile will have to be run for the UI anyway. Than the forum admin can simply get the forum menu styles by opening  http://www.demov6.camptocamp.org/static/4120b0e/build/build-forum-menu.min.css (or another name) in their browser and copy-paste it in the admin interface (or perhaps it will be possible in the futur to simply link the CSS file).

Your way is more complicated, you have to find what command to run, then run it (and you need to have the environnement to run it, with up-to-date code etc.).

Using the makefile is not difficult to implement and way more consistent/convenient.

Please do use the makefile.

ginold:
> The logo in the sidemenu is bigger in the forum than in the UI. Is there a reason?
Sidemenu entries have not the same heights in the forum and in the menu.

I don't think so:
screenshot from 2016-11-29 12-50-08


Maybe the CSS I use has changed a little and is not available on you instance. But it is the exact same CSS on the UI and Discourse.

ginold:
> Not the same button to close the sidemenu in mobile screens then in the UI sidemenu. Is there a reason?
The hamburger button is not the same than in the UI.

In the UI, the hamburger and the closing arrow are the same element, but it moves and transforms on click. There is already a Hamburger icon in Discourse so I wanted to add a slightly different icon to mark the difference. It would be best to change the discourse's icon and have no changes in our menu.

asaunier:
> I don't think so:

Your menu is no longer available in http://forum.demov6.camptocamp.org/ (@fbunoz have you made changes recently?)
so I cannot check again but yes, the sizes were definitely not the same when I have tested your demo. Perhaps you made changes that impacted the UI sidemenu as well but the demo styles were outdated?

fbunoz:
I didn't chang anything in the demo forum today.
Yesterday, the menu used was my draft menu.

ginold:
Thanks to @asaunier I forced myself to bring the following updates:

+ the hamburger button of discourse is overridden with an ellipsis
+ we have back our usual hamburger icon and its animation in the menu on discourse
+ the logos in the menu and in the header redirect to the same url (had to override with JS the default discourse's behavious)
+ no horizontal scroll bar
+ other minor upgrades

@asaunier I let you write the make less function to output the sidemenu.less -> sidemenu.css (don't forget to update the readme file in menu-discourse).

ginold:
http://forum.demov6.camptocamp.org/?preview-style=e55c4247-d233-4af4-b0a4-b1e52a401f60

c2corg_ui/static/menu-discourse/menu.html
```diff
@@ -0,0 +1,196 @@
...
+        <a href="https://plus.google.com/104270548458536561874/posts"><img src="#" url="static/img/social/google-plus.png" class="src"></a>
+        <a href="https://www.facebook.com/camptocamp.org/"><img src="#" url="static/img/social/facebook.png" class="src"></a>
+      </div>
+    </div>
+
+  </footer>
+</aside>
+
+<script>
+    var baseUrl = '//www.demov6.camptocamp.org/';
```
asaunier: Actually I meant "at the top of the file" :P 
So that one may immediately see it and fix it. Else it's hard to figure out there's a ``baseUrl`` set somewhere in the file.

less-print/c2corg_ui.less
```diff
@@ -1,13 +0,0 @@
-@import "viewdetails.less";
```
asaunier: Why have you removed this file?
It makes the makefile crash!

asaunier:
I have pushed a new commit d39243c that add targets in the Makefile so that it also generates the CSS for the Discourse menu. Not that complicated :P 

The generated ``build-discourse.min.css`` (minified) and ``build-discourse.css`` are added to the usual CSS location in ``c2corg_ui/static/build`` (as for the screen and print styles of the UI).

In addition I think you should remove https://github.com/c2corg/v6_ui/blob/discourse-menu/c2corg_ui/static/menu-discourse/menu.css (by the way in the README you name it ``sidemenu.css`` instead of ``menu.css`` - ditto for the HTML file) and move the README and the HTML file to the ``less-discourse`` dir I have created: no use that those files are made available in the website.

less-print/c2corg_ui.less
ginold: I was playing around..i didn't even notice it had an impact :P I will bring it back.

asaunier:
Thanks for addressing the points I have reported at https://github.com/c2corg/v6_ui/pull/990#issuecomment-263530296

The results is now really really cool and the forum is now really well integrated to the website. And it works well on all screens. Thank you very much :)

I have noticed a few minor problems though :P 
* [ ] Icons are not the same in UI or forum for "Forum" (!), "BD accidents" and "Camptocamp Association"
* [ ] The entry title should be "Camptocamp Association", not "Association"
* [ ] Footer links are a bit different, + a small vertical shift (perhaps because the social icons are missing?)

sans titre

ginold:
> Icons are not the same in UI or forum for "Forum" (!), "BD accidents" and "Camptocamp Association"

They will be when we merge this pr, it brings some minor changes to the UI menu too. I changed the forum icon to 'comment' and the association icon to a person :P

brunobesson:
Looks like the arrow to close the menu on mobile or tablet is not placed identically (on bottom tather than on top)

c2corg_ui/static/menu-discourse/menu.css
```diff
@@ -0,0 +1,391 @@
+#sidemenu {
```
asaunier: It's better not to commit this file since it is generated.

asaunier:
> Looks like the arrow to close the menu on mobile or tablet is not placed identically (on bottom tather than on top)

@brunobesson You're right. In Chromium the button is placed at the top but not in FF.

ginold:
> It's better not to commit this file since it is generated.

Maybe it's a good idea to not commit this generated file, but in there there are a few lines dedicated to discourse only (compilation of discourse css + sidemenu css). 

I will delete the comiled sidemenu.css part and leave only the discourse thing.

less-discourse/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+## LESS -> CSS
+In order to compile the usual sidemenu.less file into pure CSS, use the command:
+
+`make less`
+
+Now open `c2corg_ui/static/build/build-discourse.css` (or `build-discourse.min.css`) and copy the content of the file into the **CSS** tab, after: 
+
+`///// ADD SIDEMENU.CSS BELOW ////`
```
asaunier: Do you still need this?

ginold: Yes, read my comment just above (or what follows in the readme file).

less-print/c2corg_ui.less
```diff
@@ -11,3 +11,4 @@
 a:after {
   display: none;
 }
+
```
asaunier: ?

less-discourse/menu.css
```diff
@@ -0,0 +1,31 @@
+body {
```
asaunier: I don't understand why you need this file?
Why not using those styles in ``less-discourse/discourse.less``?

less-discourse/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The files in Discourse should be changed under **the hamburger button** -> **admin** -> **customize (personnaliser)** -> **CSS/HTML** -> and choose from the left list **discourse menu** or create a new style.
+
+
+## LESS -> CSS
+In order to compile the usual sidemenu.less file into pure CSS, use the command:
+
+`make less`
+
+Now open `c2corg_ui/static/build/build-discourse.css` (or `build-discourse.min.css`) and copy the content of the file into the **CSS** tab, after:
```
asaunier: That looks weird. One should not edit generated files!

less-discourse/menu.css
ginold: aah ok now i see :P i didn't think of this that way. I will update in 2 min.

ginold: I was wondering what those files were in the first place...then you ask me why i deleted one? because i thought it was of no use :stuck_out_tongue_closed_eyes:  my bad!

ginold:
UPDATED:

+ moved the baseUrl to the top of the HTML so it's easily seen 
+ translations corrections
+ moved files to less-discourse
+ readme updated

- the few pixels difference is probably due to the lack of social-icons
- after merging, both the UI and forum menu will have the same icons

less-discourse/README.md
```diff
@@ -0,0 +1,31 @@
+
+### Menu on the Discourse Forum
+
+The goal was to rebuild the exact same menu in both UI and on the Discourse Forum.
+Only a few icons change (to differenciate between the already-too-similar-ones on Discourse) and the HTML for the menu on Discourse had to be adapted.
```
asaunier: I don't think this section is useful, it's rather a PR comment than something useful for developers that want to change the sidemenu.

less-discourse/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+If you make any changes to the `sidemenu.html` or `sidemenu.less`, make sure you update the Discourse's menu too.
```
asaunier: ``sidemenu.html`` does not exist. You have named it ``menu.html``. Perhaps you should rename it.
Using the full paths would help to figure out what files you are talking about.

asaunier:
A few comments about the README, else OK for me. Thanks for the changes.

ginold:
Thanks for the review.
Could you have a look in the logs why there is an error? I quite don't understand why (or maybe again everyone has this error).

less-discourse/README.md
ginold: indeed i meant the original sidemenu template, nothing to do with menu.html. I will use full paths.

asaunier:
> or maybe again everyone has this error).

Yes, same problem with https://github.com/c2corg/v6_ui/pull/1038
It's not related to your PR.

ginold:
I activated the menu on the Forum. The only issue I see is the mentioned arrow that displays a little differently on FF. I can investigate later, or better focus on other priority issues.

Should I wait before merging to fix the problem with snyk ?

asaunier:
> Should I wait before merging to fix the problem with snyk ?

Tests pass in my instance, I think it's safe to merge the PR. Thanks!

asaunier:
> The only issue I see is the mentioned arrow that displays a little differently on FF. I can investigate later, or better focus on other priority issues.

Please open an issue (in the forum repo).

jmarcmod:
Beau travail.
Je ne sais pas si c'est un dommage collatéral, mais depuis qu'il y a cette colonne menu de gauche, le hamburger de droite est réduit à un bâton (au moins sur écran 1920 px).
J'imagine qu'il faut créer un ticket dédié ?

asaunier:
Comments about the new Discourse menu can be reported in this issue: https://github.com/c2corg/v6_forum/issues/56

ginold:
@jmarcmod 
Comme vous l'aurez remarqué, le hamburger de droite a changé d'icône pour ne pas porter à confusion entre notre hamburger du menu de droite et le leur, de Discourse. Deux casi mêmes hamburgers qui font deux choses différentes n'étaient pas le bon choix. :+1:

jmarcmod:
OK, merci (et désolé pour le dérangement).

PR #994, opened 2016-11-24
[WIP] Correct close preview button positioning

brunobesson:
Again, tested only directly in the browser, but it is pretty straightforward.
Removing useless margin-left property, adjusting margin-top.

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -68,8 +68,12 @@
 <%def name="show_preview_container()">\
   <div class="ng-hide">
     <div id="preview-container">
-      <div id="preview-container-content"></div>
-      <button class="btn btn-warning" type="button" ng-click="previewModalCtrl.close()" translate>Close</button>
+      <div class="modal-body">
+        <div id="preview-container-content"></div>
+      </div>
+      <div class="modal-footer">
+        <button class="btn btn-warning" type="button" ng-click="previewModalCtrl.close()" translate>Close</button>
+      </div>
     </div>
   </div>
 </%def>
--- a/less/documentedit.less
+++ b/less/documentedit.less
@@ -633,8 +633,7 @@ ul.uib-datepicker {
 
   button {
     float: right;
-    margin-top: 10px;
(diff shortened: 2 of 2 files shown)


ginold:
maybe there was a second button that has been removed in the meantime.
However i find 2px of top margin very little.

besson-genostar:
I was surprized by your comment, because there is not a lot of place to put the button below. So I checked again. The problem is with the `#preview-container-content` havaing a 95% height. So the place for the button varies depending on screen. I will propose something else using `calc` property.

besson-genostar:
Mmmh... Why are the preview content and button directly added in `.modal-content`? They should be in `.modal-body` and `.modal-footer`, respectively. Then we have to ensure the scrolling.
I'll work on it and make a proposal, but this cannot be done right now. I'll keep you posted.

gottferdom:
(arf, sorry, again a confusion between my work account the other one, hope you followed!)

gottferdom:
(note to myself: it is apparently not easy to ensure correct modal height with scrollbar for body without js: http://codepen.io/dimbslmh/full/mKfCc/)

asaunier:
Perhaps a screenshot?

gottferdom:
Work is not finished yet!

asaunier:
OK, then please prefix your PR with ``[WIP]`` (for "work in progress" - perhaps we could create a label for this?).

brunobesson:
Nope, I haven't touched it for a while. It will certainly require some javascript to handle maximized modal size.
But maybe we could close it and open an issue for this, and think more about how the preview is shown: maybe a modal is not the best solution?

asaunier:
I let you close the PR (and open an issue) if you think it's better to do so.

> think more about how the preview is shown: maybe a modal is not the best solution?

My first idea was to open the preview whenever the user clicks on the save button (previewing would then be mandatory before actually saving). But it's a bit more complicated then having a modal preview that is a bit more explicit (that's not the saved page) and you can easily go back to the attribute you were editing before previewing.

brunobesson:
I'm closing this PR, as it is not working the way it should. Opening #1387 instead.

PR #995, opened 2016-11-24
Useless details icon width on mobile

brunobesson:
I don't know why this rule exists. Whether when viewing a document or previewing an edit, it produces useless space between icon and title. Simply removing it is IMHO the best options

### DIFF
--- a/less/viewdetails.less
+++ b/less/viewdetails.less
@@ -57,6 +57,9 @@
       .title {
         max-width: 70%;
         margin-right: 10px;
+        @media @phone {
+          max-width: none;
+        }
       }
       .icons-header {
         font-size: 1.5em;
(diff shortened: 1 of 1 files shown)


asaunier:
I'm not sure to understand what the problem is. Perhaps you could provide a screenshot?

gottferdom:
Before:
before
After:
after

This is difficult to explain the whole difference with only 2 screenshots. Still, the idea is to occupy more space on mobile.

gottferdom:
I have also corrected the title max-width property on mobile.

asaunier:
I would prefer @ginold reviews this PR since he's the specialist. But from my perspective (based upon the screenshots) your change makes sense, thanks.

asaunier:
Tested in my instance, titles are indeed more efficiently placed. I have noticed no side effects.

Issue #923, opened 2016-11-17
The geonames search does not work in https


asaunier:
When using an advanced search with an HTTPS URL, 
https://www.demov6.camptocamp.org/routes#bbox=-593070%252C3794863%252C1393069%252C7405136
the geonames-based recentering tool does not work ("not result") because the service is not available in HTTPS.

Actually it is available but as a premium webservice (not free): http://www.geonames.org/commercial-webservices.html

tsauerwein:
Possible solutions:
* The association is willing to pay for the premium service.
* We set up a proxy.
* We use an alternative service, e.g. [Photon](http://photon.komoot.de/) or [OSMNames](http://osmnames.org/api/)

stef74:
The cost of premium service is 720€/ years.

So what is the time needed to setup the two other solution ?

tsauerwein:
> So what is the time needed to setup the two other solution ?

<0.5 d

HSERGERAERT:
What is the advantage right now to work with https:// if this function is working with http:// ?

asaunier:
It's mandatory to use https service when the calling page is https. Else the browser complains or even blocks the requests.

HSERGERAERT:
Sure. Do we have strong argues to use an https:// starting page instead of the http://, at least until this issue is solved. This issue is linked to milestone v6.1, I have a bad feeling if the recenter on is not working at go-live

tsauerwein:
In 2016 there are little reasons not use HTTPS everywhere (for example see here if you need arguments: https://developers.google.com/web/fundamentals/security/encrypt-in-transit/why-https). Changing the geo search provider is a fairly simple change, this should not block us from using HTTPS.

stef74:
@tsauerwein I agree with you but starting HTTPS post live is not a big issue ... currently many point are not ready for live ... so priority is not HTTPS.

asaunier:
Please not the "recenter on my position" button  (on the map) does not work if not in https.

tsauerwein:
Tiens, cadeau : https://github.com/c2corg/v6_ui/pull/1001 :)

asaunier:
@tsauerwein OH-MY-GOD This is sooooooo cool!!

The service works in https indeed. But I have also figure out it returns results not only for cities but also for places, regions or mountains (actually all data available in OSM I guess).
[...]

PR #1001, opened 2016-11-24
Replace GeoNames with Photon

tsauerwein:
See http://photon.komoot.de/

Closes https://github.com/c2corg/v6_ui/issues/923


asaunier:
Works fine in my instance, thanks :)

Here are the few minor problems I have noticed (I think it's OK if this PR does not address them now):
* the results are often too wide and do not fit in the map (unless the screen is quite wide).

photon

* when the search takes a bit of time, there is no loading indicator (as with geonames I guess). Perhaps fix if we add the angular-load-bar tool?
* the recentering is done on points. It's OK for places but does not work that good with areas such as regions or countries (I guess it zooms on the center point of the area). Test for instance with "Valais", "Savoie", "France" etc.
* perhaps not all the results are relevant. For instance addresses or city POIs:

capture d ecran 2016-11-24 a 23 28 24

tsauerwein:
I added a loading indicator and zoom to an extent if one is provided for a search result.

I would propose to handle your other comments later.

PR #1003, opened 2016-11-25
Add missing preview warning for outings

asaunier:


tsauerwein:
The warning is there for all other document types?

asaunier:
> The warning is there for all other document types?

Yes

Thanks for the review!

PR #1009, opened 2016-11-25
Float icons mobile dropdown

brunobesson:
Currently position looks a bit strange

before

This PR changes it to the following (never mind the second `versions` item, it was for debugging - also ignore the tooltip, it's not relevant):

after

Note: this is OK only if we always have the plus button on the right.

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -67,12 +67,8 @@
 
 <%def name="show_preview_container()">\
   <div class="ng-hide">
-    <div class="modal-body">
-      <div id="preview-container">
-        <div id="preview-container-content"></div>
-      </div>
-    </div>
-    <div class="modal-footer">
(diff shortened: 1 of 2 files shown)


ginold:
Good! I noticed it earlier but had no time to have a look. I didn't even know there was such thing as `unset` :P but it seems to be  not supported by IE and older android browsers (4.4) http://caniuse.com/#search=unset 
Would it not work the same with `initial` or `inherit` ?

brunobesson:
I didn't pay attention to the support of `unset` property back then (my bad, I usually do it). I'll check that this evening and let you know, but it is certainly possible to use another property. From https://developer.mozilla.org/en/docs/Web/CSS/unset, it is a combination of `inherit` and `initial`. `initial` is probably the best choice.

brunobesson:
I tried using `initial` instead and it works great. Now merging.

PR #1012, opened 2016-11-25
Feed inner items styling

brunobesson:
* same border for images and outings/waypoints/other cards
* add margin around boxes

Before:
before
After:
after

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -67,12 +67,8 @@
 
 <%def name="show_preview_container()">\
   <div class="ng-hide">
-    <div class="modal-body">
-      <div id="preview-container">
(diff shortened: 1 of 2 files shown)


ginold:
:+1:  for border color.

I was testing with a bigger padding and margins when I was creating the cards, but I didn't like it :P but it's a question of taste - as it pleases you (and the association) !

besson-genostar:
I propose we merge this PR for now, as it at least add consistent look&feel. We can always change the layout later.
Do you agree with merging?

brunobesson:
I propose we merge this PR for now, as it at least add consistent look&feel. We can always change the layout later.
Do you agree with merging?

ginold:
yes no problem. Just squash your commits into a single one or remove the 2 first.

PR #1015, opened 2016-11-26
User menu mobile dropdown

brunobesson:
* remove useless button height so that dropdown is positioned correctly vertically (before, was stuck to button, now behave like lang dropdown)
* on mobile, right aligned menu

Note: this time I tested it locally :wink:

### DIFF
--- a/less/c2corg_ui.less
+++ b/less/c2corg_ui.less
@@ -824,8 +824,6 @@ app-simple-search {
   }
 
   #user-menu {
-    height: 33px;
-
     .icon-user {
       display: none;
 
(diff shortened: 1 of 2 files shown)


less/c2corg_ui.less
```diff
@@ -846,6 +844,12 @@ app-simple-search {
     a.logout {
       cursor: pointer;
     }
+    @media @phone {
+      .dropdown-menu { /* right-aligned user menu on mobile */
+        left: auto;
```
ginold: ... or you could add `dropdown-menu-right` to do exactly the same thing without CSS. I did it recently for 
https://github.com/c2corg/v6_ui/blob/f349f2c0bed61d6ea09a5809e6f5cf84ef3d16ab/c2corg_ui/static/js/lang.js#L26

brunobesson:
I didn't want to use the `dropdown-menu-right` class as it would mean that on tablet and desktop, the menu is also aligned right, which I think we do not want. What's your opinion on this?

ginold:
I had to see for myself how it looked and I think you are right :+1: Keep it as is.

But careful with the 2 first commits ! ;)

PR #1017, opened 2016-11-26
Header label alignment

brunobesson:
Ensure 'browse site in' label is vertically centered which wasn't the case because of default label bottom-margin.

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -67,12 +67,8 @@
 
 <%def name="show_preview_container()">\
   <div class="ng-hide">
-    <div class="modal-body">
-      <div id="preview-container">
-        <div id="preview-container-content"></div>
-      </div>
-    </div>
-    <div class="modal-footer">
+    <div id="preview-container">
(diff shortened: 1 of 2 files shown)


ginold:
from what I can see, `correct modal structure` and `revert correct modal structure` exclude each other, bringing no changes -> those commits should be deleted? (try `git rebase -i HEAD~3` and delete the two lines with the commits to delete, save, run `git rebase origin/master` and re-push the branch).

brunobesson:
@ginold yes I know I have a problem with those f*** commits I made mistakenly. I already discussed this with @asaunier but still I cannot manage to clean up all this, sorry. Following your procedure is not working (it might have before, but now thinks are getting a little bit fucked up on my fork).
If you can help on the subject, it will be appreciated a lot.
However, maybe we can circumvent this by simply squashing my PRs?

asaunier:
> However, maybe we can circumvent this by simply squashing my PRs?

That would work too. I don't really know what happens with your fork.
Since @ginold has approved your PR, please "squash and merge" it, thanks.

Issue #966, opened 2016-11-23
Outing - Block Associated route still display an empty area when closed


dribagnac:
When I am displaying an outing, for eg 
http://www.demov6.camptocamp.org/outings/821396/fr/pointe-de-chalune-par-le-col-de-la-ramaz-depuis-sommand#/&gid=1&pid=4

   I want to close the block _Associated route_, a small DIV is still displayed (div class="associated-documents collapse")
![screenshot-www demov6 camptocamp org 2016-11-23 11-15-20](https://cloud.githubusercontent.com/assets/995825/20558100/caf1202a-b16e-11e6-9f93-197793214e53.png)

ginold:
probably because `.associated-documents` has a `padding-top`, which does not play too well with bootstraps `.collapse `

brunobesson:
It's both the `padding-bottom: 10px` and the `border: 1px solid #eee` properties.
They are defined in https://github.com/c2corg/v6_ui/blob/master/less/viewdetails.less#L325

But don't you think we could remove them? The visuals are OK for me without (it's lighter, I don't like having too many imbricated borders).

asaunier:
That doesn't look like a critical problem. I have postponed it to 6.1.

brunobesson:
See PR #1018

The boxes collapse nicely, plus IMHO the design is lighter and more attractive. When the content needs scrollbars, I find it properly integrated too.

stef74:
@brunobesson you PR fix this issue ?

brunobesson:
Yes, it does. You can close the issue.

PR #1018, opened 2016-11-26
Associated data design

brunobesson:
References #966 

Now, with simple inner card and scrollable content:
after

(and of course, collapses nicely now)

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -67,12 +67,8 @@
 
 <%def name="show_preview_container()">\
   <div class="ng-hide">
-    <div class="modal-body">
-      <div id="preview-container">
-        <div id="preview-container-content"></div>
-      </div>
-    </div>
-    <div class="modal-footer">
+    <div id="preview-container">
+      <div id="preview-container-content"></div>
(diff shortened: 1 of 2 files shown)


asaunier:
Do you really need 3 commits for such a change?
OK the commit can be squashed when merging, but it's better to drop useless commits/changes before the review :)

brunobesson:
I know, I should remove those useless commits... But I do not master git enough yet to do that....

But when you're OK, I'll do the squashing, promise!

asaunier:
I notice that the 2 first commits are provided in many of your recent PRs. Correct?
Are they needed in all those PR?
If yes perhaps it's better to keep the commits distinct and to drop them when they have been integrated by the first merged PR. Else you might have to fix conflicts I think.

brunobesson:
It's a commit I mistakenly made on master branch, then cancelled. Those two commits are useless. I need to have a look on how to rewrite history for this, I'll try tomorrow.

ginold:
Have a quick look if it does impact the associated documents on editing pages, and if it does - if 
that is what we want.

Else, it looks smoother indeed!

brunobesson:
I have check there is no impact on editing pages, so merging this PR.

Issue #977, opened 2016-11-23
Timezones are not handled correctly


tsauerwein:
I just created an outing and the feed shows "... created ... an hour ago".

stef74:
For me is not blocking issue

tsauerwein:
I should have provided more details. The problem is more fundamental. In v5 all date times were stored without time zone. When setting up the data model for v6, we did the same. Luckily the problem was revealed by our demo server which seems to use a different time zone as system time. We should make sure that we store date times with a time zone. Otherwise we will have problems when the site is used from a different time zone or simply with winter/summer time changes.

I would consider that as blocking.

dkocich:
Storing the time zone preference in the user profile can be maybe a solution? They can fill it during registration or after in profile settings. UTC time could be adjusted on UI according to user's current setting.

tsauerwein:
> Storing the time zone preference in the user profile can be maybe a solution? They can fill it during registration or after in profile settings. UTC time could be adjusted on UI according to user's current setting.

To start with I would simply take the default timezone of the users system. For example like here: https://css-tricks.com/localizing-times-in-javascript/

stef74:
@tsauerwein Ok for "blocking"

stef74:
@tsauerwein When do you think this issue is fixed ?

tsauerwein:
@stef74 Today/tomorrow.

tsauerwein:
Fixed with https://github.com/c2corg/v6_api/pull/536 and https://github.com/c2corg/v6_ui/pull/1020.

The best way to test this is by creating a custom profile for Chrome, see http://stackoverflow.com/a/39533934/119937

PR #1020, opened 2016-11-26
Show times in local system time zone

dkocich:
See problem https://github.com/c2corg/v6_ui/issues/977

Work in progress - feel free to adjust when API will be fixed

Use `moment-timezone` with `angular-moment` to show time correctly - see docs for usage https://github.com/urish/angular-moment#amlocal-filter

### DIFF
--- a/c2corg_ui/static/partials/slideinfo.html
+++ b/c2corg_ui/static/partials/slideinfo.html
@@ -1,4 +1,3 @@
-<!--if you modify this template, please update utils.js createSimpleImageSlide()-->
 <div id="{{image_id}}-slide">
   <h2 class="image-title" ng-bind="locales[0].title"></h2>
 
@@ -39,4 +38,3 @@
     <button class="pswp__button info glyphicon glyphicon-info-sign" img-id="{{image_id}}"></button>
   </div>
 </div>
-
--- a/c2corg_ui/templates/base.html
+++ b/c2corg_ui/templates/base.html
@@ -96,6 +96,7 @@ closure_library_path = settings.get('closure_library_path')
     <script src="${request.static_path('%s/typeahead.js/dist/typeahead.bundle.js' % node_modules_path)}"></script>
     <script src="${request.static_path('%s/ng-infinite-scroll/build/ng-infinite-scroll.js' % node_modules_path)}"></script>
     <script src="${request.static_path('%s/moment/moment.js' % node_modules_path)}"></script>
+    <script src="${request.static_path('%s/moment-timezone/builds/moment-timezone-with-data.js' % node_modules_path)}"></script>
     <script src="${request.static_path('%s/angular-moment/angular-moment.js' % node_modules_path)}"></script>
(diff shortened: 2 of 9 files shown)


c2corg_ui/templates/base.html
```diff
@@ -149,6 +151,10 @@
            vcRecaptchaServiceProvider.setSiteKey('${settings["recaptcha_site_key"]}');
          });
 
+         angular.module('app').constant('angularMomentConfig', ['moment-timezone'], {
+           timezone: 'Europe/London'
```
tsauerwein: Would it be possible to get the default timezone of the users system?

package.json
```diff
@@ -21,6 +21,7 @@
     "bootstrap": "3.3.6",
     "bootstrap-slider": "7.1.0",
     "closure-util": "1.15.1",
+    "moment-timezone": "^0.5.9",
```
tsauerwein: Please use "0.5.9" instead of "^0.5.9" to have a reproducible build.

c2corg_ui/templates/base.html
dkocich: According to [SO comment](http://stackoverflow.com/a/25284017/4114514), it is possible to just set it as `value `instead of `constant `and it should be handled automatically. Later it is probably not needed to set this part at all, but I would leave it here now until we fix the issue.

tsauerwein: If you do not set `timezone`, it will take the system timezoneß That's what we want. So why do you set it? :)

package.json
dkocich: Ok. This is how `Npm install --save `adds package version by default. That led me to look at other packages -  `snyk `and `angular-moment` are also set like that, When I looked at `angular-moment` release history, we are few versions behind (beta5 added bugfix of `amTimeAgo` we want to work with now + we use latest `moment `and `moment-timezone` packages which look like they not supported by currently used older version of `angular-moment`?)
https://github.com/urish/angular-moment/releases

tsauerwein: The problem with `^` is that you don't know which version you will get, because it installs the last major version. If you are trying to install the project, and in the mean time a new version has been released, you project might not work any more.

Feel free to update `angular-moment`.

c2corg_ui/templates/base.html
dkocich: Because I think that I had some problems on Thursday and this was the latest WIP commit where I tested different settings

tsauerwein:
Ready to review. Thanks for the initial work on this, @dkocich! :+1: 

https://github.com/c2corg/v6_api/pull/536 must be merged, before we can merge this PR.

c2corg_ui/static/partials/slideinfo.html
```diff
@@ -1,4 +1,3 @@
-<!--if you modify this template, please update utils.js createSimpleImageSlide()-->
```
asaunier: Should we remove this ``createSimpleImageSlide()`` function?

tsauerwein: There is no such function. According to @ginold we can remove this comment.

tsauerwein:
Thanks for the review!

PR #1021, opened 2016-11-26
Progress bar color

brunobesson:
* make progress bar colors be the same as those used elsewhere in the site
* use defined variable, that's what less is made for


asaunier:
Perhaps a small screenshot? :)

brunobesson:
There you go:
snapshot1

ginold:
:+1: 

But carful with the 2 unused commits :)

Issue #331, opened 2016-04-28
Elevation/timestamp are removed from GPX tracks


fjacon:
* [ ] Make sure elevations/dates are not removed from GPX tracks uploaded in the routes/outings editing forms.

asaunier:
I am not sure it is a UI issue. See https://github.com/c2corg/v6_api/issues/262

asaunier:
I have removed the "alpha 3" milestone from this issue because the issue cannot be fixed on the UI side until elevation/timestamp are supported on the API side (it's not as far as I know).

asaunier:
The milestone of this issue cannot be v6.1 because it impacts the data.

If c2corg/v6_api#262 is not fixed at golive, all migrated geometries will be migrated as 2D and the elevations and times will be lost.
Ditto the elevations of uploaded GPX track in v6 will be lost.

Please note that anyway, drawing a route on the map will only be 2D, unless we set an elevation service to add the elevation to the new points.

tsauerwein:
The API now supports 3D/4D geometries. But the UI does only send geometries with 2 dimensions. The other dimensions are removed.

PR #1023, opened 2016-11-26
Simplify with simplify.js

tsauerwein:
So far we were using the simplify function of ol3, which removed the 3rd and 4th dimension. This PR proposes to use [simplify.js](http://mourner.github.io/simplify-js/), which keeps the dimensions.

Closes https://github.com/c2corg/v6_ui/issues/331

### DIFF
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -29,6 +29,7 @@ goog.require('ol.style.Icon');
 goog.require('ol.style.Stroke');
 goog.require('ol.style.Style');
 goog.require('ol.style.Text');
+goog.require('app.map.simplify');
 
 
 /**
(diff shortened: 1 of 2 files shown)


c2corg_ui/static/js/map/map.js
```diff
@@ -841,9 +842,7 @@ app.MapController.prototype.addTrackImporter_ = function() {
   dragAndDropInteraction.on('addfeatures', function(event) {
     var features = event.features;
     if (features.length) {
-      features.forEach(this.simplifyFeature_);
-      this.showFeatures_(features, true);
-      this.scope_.$root.$emit('mapFeaturesChange', features);
+      this.handleFeaturesUpload_(null, features);
```
asaunier: Good catch!

c2corg_ui/static/js/map/simplify.js
```diff
@@ -0,0 +1,191 @@
+/*
```
asaunier: Is a double ``*`` needed here?

PR #1025, opened 2016-11-26
Single day outing

brunobesson:
Add pointer on the cross used to remove second outing date (was default one)

At the moment the structure of the less files is not really clear to me (IMHO it could be structure more) so I'm not sure about where to place my CSS code. Class name could be discussed too.

Thanks for your feddback.

### DIFF
--- a/c2corg_ui/templates/outing/edit.html
+++ b/c2corg_ui/templates/outing/edit.html
@@ -79,7 +79,7 @@ updating_doc = outing_id and outing_lang
                     <button type="button" class="btn btn-default" ng-click="openDateEnd = true"><span class="glyphicon glyphicon-calendar"></span></button>
                   </span>
                 </div>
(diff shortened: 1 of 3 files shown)


asaunier:
> Add pointer on the cross used to remove second outing date (was default one)

Nice!

> At the moment the structure of the less files is not really clear to me (IMHO it could be structure more) so I'm not sure about where to place my CSS code.

Don't know much about this part. @ginold what do you think?

ginold:
The placement in documentediting.less is ok.

If you don't want to create a new class, writing `.date.glyphicon-remove` should be enough. Or create a helper cass `.pointer: {cursor: pointer}` which can be reused by adding "pointer" on any element.

Warning: I've seen the two first commits about "correct modal structure" in other branches - you probably created a new branch from an existing one with commits inside. Be sure to delete those commits that have nothing to do with this PR.

brunobesson:
Indeed, I think using a `.pointer` generic class is quite a nice solution. I'll change this this evening and squash my PR (as discussed in other threads :confounded:)

PR #1032, opened 2016-11-28
Do not show field access_period when editing a climbing_outdoor waypoint

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/705#issuecomment-261408302

> in WP/site, the field after "Accès" should have a title in edition mode, and the title in display mode should be changed with a new Transifex label (access_period is used with another sense)

Actually climbing_outdoor WP already have a ``best_periods`` attribute (a list of checkboxes with months). Then a text-field for ``access_period`` is redundant => this PR makes sure it is not shown for climbing_outdoor WP.
Nothing to do in the detail view page: if not set, the attribute is not displayed.

### DIFF
--- a/c2corg_ui/templates/waypoint/edit.html
+++ b/c2corg_ui/templates/waypoint/edit.html
@@ -747,7 +747,7 @@ from c2corg_common.attributes import default_langs, climbing_outdoor_types, publ
             </div>
 
             ## ACCESS PERIOD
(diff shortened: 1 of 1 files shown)


ginold:
:+1:

asaunier:
Thanks @ginold for reviewing!

fbunoz:
The access_period is used for climbing outdoor wp to describe the details
of forbidden period, sector or route.
Then, it must be visible, but the title should be "Restrictions d'accès".

asaunier:
Argh
Please create an issue.

Issue #999, opened 2016-11-24
Comments display a 404 page in image view


arnaud-morvan:
See: http://www.demov6.camptocamp.org/images/821567/fr/pointe-de-chalune

This is missing in image view template: 
```
module.value('documentData', {
...
 "topic_id": ${dumps(locale.get('topic_id'))}
```
Specific issue about comments in image view from #972

asaunier:
The PR is not merged. So it's not "ready for demo" yet :P

PR #1034, opened 2016-11-29
add padding to discourse comments + fix cut sidemenu tooltips

ginold:
Adds a little padding around the comments and hides the iframe with glichy discourse menu when no comments.
screenshot from 2016-11-29 12-26-14

### DIFF
--- a/less/sidemenu.less
+++ b/less/sidemenu.less
@@ -14,7 +14,6 @@
   height: 100%;
   font-size: 1em;
   transition: 0.3s;
-  overflow-y: auto;
 
   @media (max-width: @tablet-max) {
     width: @menu-width-tablet;
@@ -48,6 +47,8 @@
   }
 
   @media (max-height: 400px){
+    overflow-y: auto;
+
     ul li {
       font-size: 0.9em;
       >a{
--- a/less/viewdetails.less
+++ b/less/viewdetails.less
@@ -574,11 +574,12 @@
 
     #discourse-comments {
       text-align: center;
+      padding: 20px;
+      background-color: #fbfaf6;
 
       .no-thread {
         display: inline-block;
         background: #CACACA;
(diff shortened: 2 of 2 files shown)


asaunier:
No opinion about hidding the iframe when there is no comment. Don't we need to show the "post a comment" button?

Anyway adding a padding as you suggest looks good to me :)

ginold:
screenshot from 2016-11-30 10-07-29

It is to avoid the above ;p button will still be displayed.

asaunier:
Isn't the problem that such a page is displayed?
Before hidding it, we should make sure there is not another problem!?
@arnaud-morvan that looks normal to you?

asaunier:
@ginold could you give an example URL where this iframe error can be tested?
I don't see such a problem in the demo with WP, routes, outings or images.

Perhaps it's related to https://github.com/c2corg/v6_ui/issues/999 ?

ginold:
I admit I don't remember where I encountered it :P but I will give a link when i see it next time.

asaunier:
I think you'd better remove the iframe changes from this PR and create a new PR with them (or maybe only if the problem shows up again).
The changes about the padding make sense, it's a bit a pity that they are not merged yet.

ginold:
I will remove the iframe thing and merge as soon as it passes.

ginold:
I added a commit to fix the tooltip cut in as from https://github.com/c2corg/v6_forum/issues/56

asaunier:
What's the connection with the comments?
OK for this one but it's really better to make dedicated PRs.

Could you reword the first commit message since it is no longer related to the iframe?

Then please merge when Travis has passed.

asaunier:
Reword...

ginold:
what the heck, it didn't work ;/

Issue #996, opened 2016-11-24
Image carousel does not work in Safari


desnoes:
The images of a route are not displayed correctly (vertically instead of horizontally) with Safari. It seems to be ok with chrome.
Outing: http://www.demov6.camptocamp.org/routes/818885/fr/les-verts-par-le-pas-de-monthieu-et-lancheron
Result with Safari:
![ezcn3te](https://cloud.githubusercontent.com/assets/19892089/20599416/5b2f39bc-b24e-11e6-8995-ef43ebe32c2f.png)

asaunier:
Seems OK with Firefox and Chromium.

Have you tested other routes? What about outings or waypoints?

ginold:
if you remove the property `float: left;` from the image slide, the slides will align themselves as on the image.
#strange

desnoes:
This problem was reported by Christian (cf. http://www.camptocamp.org/forums/viewtopic.php?pid=2231539#p2231539). I did not observe it directly.

desnoes:
I've asked Christain to test with outing, WP.. He oberves the same problem. The image are displayed vertically. He had not this bug when the initial beta version was released.
I copy his answer (in french):
"J'ai fait le test sur sortie, itinéraires et points de passage, sur tous les documents j'ai les photos en colonne masquées par les cadres information et autres.
Je n'avais pas ça à la sortie de la version Béta."

asaunier:
I have the same problem, testing with Safari 10.0.1
> Error: Syntax error, unrecognized expression: [class^="embedded_"

It's a jquery error related to 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/viewdetails.js#L366

asaunier:
Christian has reported he no longer has the problem with Safari. I have also tested it right now: the carousel and slideshow are OK.

PR #1037, opened 2016-11-29
fix loading embedded images

ginold:
small typo fix


asaunier:
That fixes the Safari problem, right?
https://github.com/c2corg/v6_ui/issues/996

PR #1039, opened 2016-11-29
Package upgrades recommended by snyk

asaunier:
I have run snyk wizard (see https://github.com/c2corg/v6_ui/wiki#snyk) hoping that it would solve the error in Travis of https://github.com/c2corg/v6_ui/pull/1038 and https://github.com/c2corg/v6_ui/pull/990).

Here are its suggested changes. Some package version are not fixed, I guess it's not what we want!?

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "moment-timezone": "0.5.9",
     "ng-infinite-scroll": "1.3.0",
     "photoswipe": "4.1.1",
-    "slick-carousel": "1.6.0",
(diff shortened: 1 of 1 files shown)


asaunier:
Obviously it does not fix the problem :P 

But at least it should address this vulnerability reported by the weekly snyk report:
> Cross-site Scripting (XSS) vulnerability in jquery

asaunier:
I think there have been new recommendations from snyk in the meantime. I will create a new PR later.

PR #1043, opened 2016-11-30
Delete old book sitemap HTML

dkocich:
I think that after the change of sitemaps it is no longer required. It is not available also for other document types.


tsauerwein:
Yes, good catch! Feel free to merge.

PR #1055, opened 2016-12-01
Reword 'add padding...' commit

ginold:
not sure if it's the way to do it  :P?


asaunier:
Don't waste time with this. 
Your PR has been merged. I don't see the point of adding new commits for doing the same thing then what is already in master.

ginold:
I thought it would replace the commits, just with other names.

asaunier:
I don't think so!?
Honestly it's playing with the fire. I would prefer you close this PR.

Issue #945, opened 2016-11-19
Impossible to create a route on new waypoints


mgerbaux:
After creating a new waypoint, it is impossible to create an associated route
- there is no link  in the waypoint page to add a new route
- if using the route creation interface, the recently created waypoint is not proposed by the association tool (problem of indexation ?)

mgerbaux:
@stef74 : should it put for milestone GoLive ? (I think yes)

stef74:
For the first point an issue allready exist

asaunier:
> For the first point an issue allready exist

https://github.com/c2corg/v6_ui/issues/198

asaunier:
> there is no link in the waypoint page to add a new route

will be fixed by #1060

> if using the route creation interface, the recently created waypoint is not proposed by the association tool (problem of indexation ?)

Tested on the demov6 and it works. It was probably an indexing problem.

loicperrin:
Fixed

PR #1060, opened 2016-12-01
Add a route creation button in waypoint detail views

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/198 and https://github.com/c2corg/v6_ui/issues/945

For waypoint detail view pages (not other pages that have an associated-routes section):
* the associated-routes section is always visible, even though there is no associated route yet
* an "add a route to this WP" button is added at the end of the associated routes list
* when the button is clicked, the route addition form is open with the source WP already associated

sans titre
sans titre2

No change for other document types.

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -251,7 +251,11 @@
     associations += 'all_routes.documents' if type == 'all_routes' else 'routes'
   %>
     ${add_seo_association_links(routes, 'routes')}
-    <section ng-show="${associations}.length > 0">
+    <section
+      % if document['doctype'] != 'waypoints':
+        ng-show="${associations}.length > 0"
+      % endif
+    >
       <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#associated-routes">
         <span class="glyphicon glyphicon-map-marker"></span>
         <span translate>Associated routes</span><span class="glyphicon glyphicon-menu-down"></span>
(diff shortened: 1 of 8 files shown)


c2corg_ui/templates/waypoint/view.html
```diff
@@ -154,16 +154,18 @@ <h3 class="heading informations" ng-click="mainCtrl.animateHeaderIcon($event)" d
...
       <span class="lead">
         <section>
           ${show_associated_routes(waypoint, 'all_routes')}
         </section>
       </span>
     </div>
 
-    % if 'recent_outings' in waypoint['associations'] and waypoint['associations']['recent_outings']['total'] > 0:
-      <div class="view-details-associations col-xs-12  associations">
+    % if waypoint['associations']['recent_outings']['total'] or \
```
ginold: Im not sure of those python conditions in this places (speaking of all the view.html) -> if there are no outings, articles or books and if you just add one with the association tool, it will not appear because this part of the html has never been rendered. I'd say we can remove this condition and do check this condition in angular with `ng-show`.

asaunier: You're definitely right! I have replaced this change by an ``ng-show`` condition as you suggest it. See a4244bc

c2corg_ui/templates/helpers/view.html
```diff
@@ -276,6 +280,14 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
             <button class="gray-btn btn show-more" translate>show more</button>
           </div>
         % endif
+        % if document['doctype'] == 'waypoints':
+          <div class="list-item vertical-align" protected-url-btn
+               url="${request.route_path('routes_add') + '#w=' + str(document['document_id'])}">
+            <a>
+              <button class="btn orange-btn" translate>add a new route to this waypoint</button>
```
tsauerwein: Start with a capital letter?

asaunier: It would make sense but I wanted to be consistent with the "add a new outing to this document" button here https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/view.html#L348

Issue #981, opened 2016-11-23
Uncomplete document https URLs are redirected to http instead of https


fbunoz:
I test the demo using https.
When I clic on a partial url, there is a redirection to the full url. But the redirection changed https to http.
Try with this url : https://www.demov6.camptocamp.org/images/821218
Or with the first lik in the description of this doc : https://www.demov6.camptocamp.org/waypoints/38970/fr/mont-pelvoux-pointe-puiseux
The redirection is also used with the link to the profile in the menu.

stef74:
We probably start in HTTP because some issue with HTTPS (geoname for ex).

asaunier:
By "partial URL", I guess you mean a document URL with no lang and slug, which indeed redirects to the full URL with lang and slug.

Indeed https://www.demov6.camptocamp.org/images/821218 redirects to http://www.demov6.camptocamp.org/images/821218/fr/versant-w-des-pics-de-berlanche-et-du-grand-charnier-d-allevard instead of https://www.demov6.camptocamp.org/images/821218/fr/versant-w-des-pics-de-berlanche-et-du-grand-charnier-d-allevard

There is probably an error (missing scheme?) in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L355-L398

asaunier:
The problem is around here:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L356-L369

A ``Location`` header is generated but with a relative path (``location = self.request.route_path``) and most likely the scheme is omitted. We probably have to replace ``self.request.route_path`` by ``self.request.route_url``. If it is not enough, we might have to add a ``_scheme``.

It's pretty hard to test because except the demo we have no https instance available.

asaunier:
It seems PR #1062 has not fixed the problem: the https scheme is still lost when redirecting to the slug URL.

Perhaps ``self.request.scheme`` https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L349 is not the relevant data because SSL is only used to reach the frontend (haproxy? gunicorn?) and is not propagated to pyramid, ``self.request.scheme``  would then always be "http".

Perhaps we should test if the original scheme is passed through an HTTP header?

tsauerwein:
This information should be passed on in a `X-Forwarded-Proto` header, see http://docs.gunicorn.org/en/stable/deploy.html

[...]

Issue #842, opened 2016-11-04
Support of HTTPS


asaunier:
The demo is now available with HTTPS: https://www.demov6.camptocamp.org/

Please report here the HTTPS-related issues:
* [x] the API URL should use HTTPS as well (else the feed and other API requests fail) => config
* [x] images should be loaded with HTTPS as well => config
* [x] links to the forum should use HTTPS => config
* [x] the ad banner should use HTTPS => probably some config to fix in the ad server

According to @mfournier 
> I suggest to enable HSTS then, which will force (modern) browsers to switch to https.

asaunier:
I have made the change in the adserver + made changes in the UI and API demo configs (not tested!). @arnaud-morvan looks good to you?

asaunier:
We have to make sure that

https://github.com/c2corg/v6_ui/pull/844#issuecomment-258438908
> The urls are generated with route_url. But we will have to test if this works correctly.

works as expected.

stef74:
We can probably also use HTTP2 ...

stef74:
Warning with HSTS need to have full validate certificate (for none production instance also), exception cannot be possible in browser

asaunier:
I have merged both PRs because anyway we have no other way to test if they do the job then to do it in the beta demo (no other instance available with https) and we cannot wait for the golive to test it.

For ``route_url`` I am quite confident because anyway I think we have replaced them by ``route_path`` that creates relative URLs.

stef74:
@asaunier close and add task to force HTTPS in trello ?

asaunier:
No idea if this part should be done during golive.
@mfournier what do you think?

PR #1062, opened 2016-12-01
Make sure slug redirects use the current scheme

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/981

It works with my localhost (in http - at least ``scheme`` is correctly detected) but I have no https instance for testing. I wonder if it is possible to write some unit tests for this?

### DIFF
--- a/c2corg_ui/views/document.py
+++ b/c2corg_ui/views/document.py
@@ -353,12 +353,15 @@ class Document(object):
             key_v1, key_v2, caching.CACHE_VERSION)
 
     def _redirect(self, id, lang, slug=None, is_lang_set=False):
+        scheme = self.request.scheme
         if slug is None:
-            location = self.request.route_path(
-                self._API_ROUTE + '_view_id_lang', id=id, lang=lang)
+            location = self.request.route_url(
+                self._API_ROUTE + '_view_id_lang',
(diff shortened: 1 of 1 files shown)


asaunier:
> Maybe you could do this change on the demo server to make sure it works properly.

You mean "by hand"? Not sure I know how to do it :P 

@arnaud-morvan is a code update planned soon (except at golive :P )?

asaunier:
I don't remember exactly why we have replaced ``route_url`` by ``route_path``. Could it cause a wrong hostname to be used here?

tsauerwein:
> I don't remember exactly why we have replaced route_url by route_path. Could it cause a wrong hostname to be used here?

In general `route_path` seems to works, all other links are generated that way (except for the sitemap). But it does not seem to work properly for the redirect in this case.

tsauerwein:
Which makes me wonder, should we generate the sitemap urls with https? That would make sense.

asaunier:
> Which makes me wonder, should we generate the sitemap urls with https? That would make sense.

Sure. Especially since https URLs will be the standard ones (see https://github.com/c2corg/v6_ui/issues/842) and because it seems https pages are better ranked by search engines (?).

PR #1064, opened 2016-12-01
create or edit document route association layout

asaunier:
Replace PR https://github.com/c2corg/v6_ui/pull/1022 by @brunobesson 
I have rebased the PR and removed useless commits.

I had a quick look in FF and it seems OK but I am not very sure what I have to test.

@ginold OK for you?

>* do not specify a padding different from the other boxes
>* do not use % value (rendering problem on firefox + we do not want a variable padding, do we?)
>* simply CSS

### DIFF
--- a/c2corg_ui/templates/route/edit.html
+++ b/c2corg_ui/templates/route/edit.html
@@ -97,7 +97,7 @@ updating_doc = route_id and route_lang
           <h5 class="full" translate ng-show="editCtrl.documentService.document.associations.routes.length == 0 && editCtrl.documentService.document.associations.waypoints.length == 0">
             You can add associated routes and waypoints by searching in the field.
           </h5>
-          <div class="content route-associations">
+          <div class="content">
             ## wr = waypoints, routes
             <app-simple-search parent-id="route.document_id"
               app-select="editCtrl.documentService.pushToAssociations(doc)" dataset="wr"></app-simple-search>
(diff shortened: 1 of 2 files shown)


brunobesson:
OK. The initial idea was that the associated routes section in outing edition wasn't properly layed out. It used `width: 100%` and `padding: 2%` properties wich made it not aligned with other sections (route description, duration and all).
But I checked today on the demo, and the style doesn't appear to be applied anymore. Maybe the structure changed a little bit? Currently there isn't a combination of classes `content` and `route-associations` on the div anymore. Thus, the visual bug I observed disappeared.

Still, we can safely remove this rule which is now useless and apply the PR.

ginold:
Indeed this rule has no impact and can be removed. You could also therefore remove the class from the html at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/route/edit.html#L100

asaunier:
> You could also therefore remove the class .route-associations from the html at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/route/edit.html#L100 (.content is still needed).

Done at 71205a3

Let me know (or add a commit) if other changes are needed,

asaunier:
@ginold @brunobesson is there still some work to do with this PR? May we merge it?

brunobesson:
We can merge. Maybe there could be some more improvements, but anyway it can be done at a later time.

ginold:
:+1:

stef74:
So is too late for the demo ?

brunobesson:
It's not a problem as, in the end, we only remove unnecessary code.

Issue #1073, opened 2016-12-02
Add new outing box layout in feed


brunobesson:
On mobile, the position is OK, but on tablet and desktop, there is a missing top margin.
![toto](https://cloud.githubusercontent.com/assets/2234024/20828160/6168a1c8-b876-11e6-9455-320d4eae87ff.png)

tsauerwein:
I also think the additional box around the button is not required.

ginold:
I was going to do that, but if you can, you have to change/replace :

```
.action-line  {     
    padding-top: 5px;
    padding-left: 10px;
} 

#feed .list-item {
    margin: 10px;
}

.feed-list-item {
    padding-bottom: 2px;
}
```

![screenshot from 2016-12-02 10-57-44](https://cloud.githubusercontent.com/assets/7814311/20829847/2be6c70c-b87e-11e6-8bf2-83501d80cc11.png)

brunobesson:
Feel free to do it!
Otherwise, I can work on it, but this may not be ready for GO live (dunno if I'll have time to work on it this week-end)

PR #1076, opened 2016-12-02
fix feed-cards padding and add article/xreports icons

ginold:
related to #1073

adds uniform margins to the feed cards.

In addition, I added icons for xreports and articles. It was just pure CSS in the same file.
screenshot from 2016-12-02 11-38-57


ginold:
I will merge is no objections.

Issue #1078, opened 2016-12-02
Outing association tool


asaunier:
* [ ] as for now it is not possible to associate a participant from the outing detail view, one needs to edit the outing to do so => it should be possible to associate users from the outing detail view.

PR #1079, opened 2016-12-02
[WIP] Add associated users on outing view page and not only during edit mode

dkocich:
Association is saved but it is not visible correctly. Page has to be restyled

Close https://github.com/c2corg/v6_ui/issues/1078

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -315,6 +315,29 @@
     </section>
 </%def>
 
+<%def name="show_associated_users(document)">\
+  <%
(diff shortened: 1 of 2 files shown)


asaunier:
I had not paid attention that users were not listed as cards in the outing detail view (which is a bit not consistent...).

Thanks for the PR but I think we should think a bit more about how associated users are shown in outings. As for now they are also listed in the "info" section (as well as text-described participants).

ginold:
maybe the association-tool should let the user look for users, but those should only appear in the info block. That would mean that this part (participants) should be rewritten with a `ng-repeat` instead of a mako `for`.

PR #1080, opened 2016-12-02
Fix Travis build

tsauerwein:
Because of a problem with the Travis container Java was no longer available (see https://github.com/travis-ci/travis-ci/issues/6959).


asaunier:
Thanks!!! :)
Please merge when Travis is satisfied of itself.

Issue #1007, opened 2016-11-25
When editing outing with no own geometry, new georef is not taken into account


brunobesson:
To reproduce:
* create an outing without trace
* edit it and add trace on the map (with or without other modifications)

I expect the trace to be used as geometry, but this is not the case.

asaunier:
I have updated the title of the issue because:
* it seems possible to update the detailed geometry of an outing if it already had one. See for instance: http://www.demov6.camptocamp.org/outings/823122/fr/pointe-d-ardens-vallon-d-ardens-voie-normale (I have moved some points of detailed geom created from the uploaded GPX file)
* if the outing has no detailed geometry, changes in the map of the editing form are not taken into account. I have tested with http://www.demov6.camptocamp.org/outings/824693/fr/test

fbunoz:
I see the same behavior that asaunier : 
- I create an outing, with a geometry which I created with the tool on the edit form.
Then, the geometry is saved, and it taken account for areas.
I can modify the outing and the geometry, and it is saved correctly.
- I create a geometry without a geometry.
I modify the outing, I add a geometry using the tool => the geometry is not saved.

loicperrin:
Seems to be fixed.
@brunobesson @fbunoz can you confirm?

brunobesson:
Yes, it works. We can close the issue.

However, I think there are 2 other issues we can open:

* removing an existing geometry does not seem possible currently? (one can only reset to previous geometry). @loicperrin @asaunier what do you think?
* if I edit or create an outing, add a geometry from GPX file, then reinitialize it an select a file again, it is does not work anymore > #1119

stef74:
I close please open the new issue.

brunobesson:
@stef74 please read carefully my comment:

* I have created an issue for the second point
* I requested feedback on whether or not the first point should be an issue. Because in v5 the ability to clear a geometry is reserved to moderators, and moderator tools are targeted for a 6.X release, I'm not sure it is pertinent to open an issue.

asaunier:
> removing an existing geometry does not seem possible currently? (one can only reset to previous geometry).

Not sure this feature is available or even listed yet. Feel free to open an issue, it will then be discussed possibly for a future release.

brunobesson:
I created #1120 for the first point.

PR #1083, opened 2016-12-02
Load non-point geometries correctly in edit form

tsauerwein:
Related to https://github.com/c2corg/v6_ui/issues/1007


tsauerwein:
Thanks for testing and the review!

Issue #20, opened 2015-11-05
WP show page crashing because of reprojection


asaunier:
The WP detail page returns a 500 Server Error with the following message in the logs:

> [Thu Nov 05 14:50:00.508357 2015] [wsgi:error] [pid 8146] [client 10.27.71.4:46254] Truncated or oversized response headers received from daemon process 'c2corg_ui-ui': /var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/apache/app-c2corg_ui.wsgi

Only when serving pages with Apache/mod_wsgi. OK with pserve.

The crash is triggered by the use of the reprojection function "transform" in the template:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L6

Actually using this function directly in the Pyramid view also makes the page crash. I have tried to isolate the guilty line and it seems it is related to `pyproj.Proj`:

```
source_proj = pyproj.Proj(init=source_epsg)
```

https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/document.py#L78
If this line or the next one (defining the destination projection) is used (even though no reprojection is finally done), we get the error.

tsauerwein:
Something to test (`threads=1`):

    WSGIDaemonProcess c2corg_ui-{instanceid} python-path={site_packages} threads=1

in https://github.com/c2corg/v6_ui/blob/master/apache/wsgi.conf.in

asaunier:
I have added ``threads=1`` to https://github.com/c2corg/v6_ui/blob/master/apache/wsgi.conf.in#L2 in the main dev instance but I still get the same error when loading the WP detail page (with the transform function).

gberaudo:
The issue does not appear on my machine (Apache + installed ui).
libapache2-mod-wsgi 4.3.0-1build1
apache2 2.4.12-2ubuntu2

On demo:
libapache2-mod-wsg 4.3.0-1
apache2 2.4.10-10+deb8u3

asaunier:
Perhaps an issue with the versions of pyproj/apache2/libapache2-mod-wsgi on the dev server since it works on your computer.

apache2 and libapache2-mod-wsg versions are the ones quoted in 
https://packages.debian.org/fr/jessie/libapache2-mod-wsgi
https://packages.debian.org/fr/jessie/apache2

Perhaps we could try using another version of python-pyproj.

https://packages.debian.org/fr/jessie/python-pyproj mentions 1.8.9-1.1
[...]

PR #1085, opened 2016-12-03
Show coordinates in 4326 in map note

tsauerwein:
Closes https://github.com/c2corg/v6_ui/issues/20

Reprojects and formats the coordinates with OpenLayers.

c2corg_coords

### DIFF
--- a/c2corg_ui/templates/waypoint/view.html
+++ b/c2corg_ui/templates/waypoint/view.html
@@ -22,10 +22,6 @@ from json import dumps
 waypoint_id = waypoint['document_id']
 other_langs, missing_langs = get_lang_lists(waypoint, lang)
 waypoint['doctype'] = 'waypoints'
-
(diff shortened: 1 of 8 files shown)


c2corg_ui/templates/waypoint/view.html
```diff
@@ -97,8 +93,8 @@
   % if waypoint.get('elevation'):
     <p class="text-center"><b>${waypoint['elevation']} m</b></p>
   % endif
-  % if geometry4326:
-    <p>${round(geometry4326.x)} °E / ${round(geometry4326.y)} °N</p>
+  % if geometry:
+    <p ng-bind="'${round(geometry.x)}/${round(geometry.y)}' | coordinate"></p>
```
tsauerwein: This is not visible to the user, it's only the input for the filter.

Issue #1013, opened 2016-11-25
Page to list users followed by the current user


asaunier:
Most likely accessible form the user menu,
The API is already ready.

PR #1087, opened 2016-12-04
Add list of followed users

asaunier:
Close https://github.com/c2corg/v6_ui/issues/1013

capture d ecran 2016-12-04 a 01 02 34

The list of cards is initially fed using the API ``users/following`` service. It is possible to add new followed users using the search tool. When clicking an "unfollow" button, the user card remains but the button is switched to "follow".

~~Perhaps a small bug to solve: even though the cards have ``app-card-disable-click="true"``, clicking on a user card sometimes reacts and opened a link that does not exist with ``undefined`` in it.~~ => I have finally made the card clickable, opening the user's profile.

Btw a question: we now have many templates (``account.html``, ``preferences.html``, etc.) and routes (``/account``, ``/preferences``, etc.) at the root of templates and URLs. Should we move them in a ``users`` directory (ie. ``users/account.html`` and ``/users/account``, etc.)?

### DIFF
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -150,6 +150,7 @@ def main(global_config, **settings):
     config.add_route('account', '/account')
     config.add_route('preferences', '/preferences')
     config.add_route('mailinglists', '/mailinglists')
+    config.add_route('following', '/following')
 
     config.add_route('sitemap_index', '/sitemap.xml')
     config.add_route('sitemap', '/sitemaps/{doc_type:[a-z]{1}}/{i:\d+}.xml')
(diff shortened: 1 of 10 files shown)


HSERGERAERT:
Exactly my expectation, well done

asaunier:
Thanks @tsauerwein 
What do you think about my question
> we now have many templates (``account.html``, ``preferences.html``, etc.) and routes (``/account``, ``/preferences``, etc.) at the root of templates and URLs. Should we move them in a ``users`` directory (ie. ``users/account.html`` and ``/users/account``, etc.)? 

Anyway that's not critical and could be done in a dedicated PR. Merging.

ginold:
>  clicking on a user card sometimes reacts and opened a link that does not exist

FYI: if you wanted to fix that, you probably should add a condition to: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js#L134

asaunier:
@ginold Thanks for the tip. Finally links are useful to open the users' profiles from this list page as well.

Issue #1088, opened 2016-12-04
robots.txt is not made available with nginx


asaunier:
A ``robots.txt`` file is available, see the apache config at https://github.com/c2corg/v6_ui/blob/master/apache/wsgi.conf.in#L7

But it seems missing in the actual demov6/prod config, using nginx instead of Apache.

asaunier:
The syntax seems to be
```
  location /robots.txt {
    alias /path/to/basedir/c2corg_ui/static/robots.txt;
  }
```
Don't know if it is possible not to hard-code ``/path/to/basedir``?

stef74:
#1152

asaunier:
The PR is merged but not available in prod yet.

Issue #639, opened 2016-09-20
TODO just before Go Live


asaunier:
- [ ] migration: check there is no image duplicate: https://github.com/c2corg/v6_api/issues/363#issuecomment-249872137
- [ ] check API keys
- [ ] update Google Analytics key
- [ ] update discourse domain name in posts, see https://github.com/c2corg/v6_forum/issues/32#issuecomment-248318602
- [ ] reactivate discourse google indexing in admin settings / security
- [ ] make sure the `robots.txt` and `robots meta` are OK, see https://github.com/c2corg/v6_ui/pull/784, revert "noindex" change in meta "robots" 7f91d79
- [ ] Link sitemap in `robots.txt`: http://stackoverflow.com/questions/14196801/sitemap-url-in-robots-txt
- [ ] switch the snow forecast mailing lists system to the v6 DB
- [ ] Revert https://github.com/c2corg/v6_api/commit/7e0fd123ce50ad15db6915de6ee581b8a919f1aa (see https://github.com/c2corg/v6_api/issues/363)
- ...

stef74:
In this ticket action is needed few day before go live or only action during the Go Live ?

asaunier:
During go live.
Or at least when the next upgrade is the production, not the demo.

Some changes can be done now if we are sure that the Basic Auth will not be removed on the demo before the go live: robots.txt etc.

asaunier:
Closing the issue, all tasks have been moved to the GoLive Trello board.

PR #1089, opened 2016-12-04
SEO settings for golive

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/639

There is still a problem with ``robots.txt`` not served by nginx. See https://github.com/c2corg/v6_ui/issues/1088


asaunier:
Thanks @tsauerwein 

Does anyone know if the demo will be updated and made public? If yes, it is better to way before merging this PR.

asaunier:
Merging the PR before forgetting it :)
I don't think the demo will go public 2 days before the prod anyway.

Issue #1068, opened 2016-12-02
Add a revert action in the wiki 


asaunier:
In v5 it is possible to revert changes in a document by opening the document history, opening a document version before the changes to cancel and editing from this version (the old version is loaded in the editing form). Saving the form (even without any change) creates a new version based upon the data of the old version.
This feature is available to any contributor (not only moderators). Of course a warning is displayed in the editing form saying that the data are out of date.

In v6 we could have another approach if something is more convenient (a "restore this version" button?) but as for now I think the v5 approach is pretty efficient and would require no API change. I am not sure but I think this is also the approach used on wikipedia.

TODO:
* [x] make the edit button available in the archived version and open a special editing route with the archive id. For instance http://www.camptocamp.org/routes/edit/45297/fr/22
 * [ ] personal documents: only show edit button if user has edit rights on the document (or limit tool to moderators?)
* [x] Open the archive in the editing form:
 * [x] Add a pyramid route for such URLs with pretty much the same view
 * [x] except that the editing form is fed using a archive API request. Associations are not provided for such request and are anyway not relevant to restore an archive (associations history is not connected to document history). For instance: http://api.demov6.camptocamp.org/routes/390305?l=fr vs http://api.demov6.camptocamp.org/routes/390305/fr/633669
 * [x] add a warning in the editing form
 * [ ] Ignore associations (including for enabling the save button when associated WP etc. are mandatory)

fbunoz:
> I think the v5 approach is pretty efficient

Yes : 
- It allows to do some changes on the old version before saving the revert.
- It allows to see and copy text with tags from the old version, if we don't want to revert all the text but only a little part.

asaunier:
Pull requests for this issue:
* UI: https://github.com/c2corg/v6_ui/pull/1669
* API: https://github.com/c2corg/v6_api/pull/671

PR #1101, opened 2016-12-04
[WIP] Add possibility to revert to an archived document version

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1068

### DIFF
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -149,6 +149,8 @@ def main(global_config, **settings):
     config.add_route('xreports_index', '/xreports')
     config.add_route('xreports_add', '/xreports/add')
     config.add_route('xreports_edit', '/xreports/edit/{id:\d+}/{lang}')
+    config.add_route('xreports_edit_archive',
+                     '/xreports/edit/{id:\d+}/{lang}/{version:\d+}')
     config.add_route('xreports_history', '/xreports/history/{id:\d+}/{lang}')
     config.add_route('xreports_archive',
                      '/xreports/version/{id:\d+}/{lang}/{version:\d+}')
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -74,6 +74,8 @@ def main(global_config, **settings):
     config.add_route('waypoints_index', '/waypoints')
     config.add_route('waypoints_add', '/waypoints/add')
     config.add_route('waypoints_edit', '/waypoints/edit/{id:\d+}/{lang}')
+    config.add_route('waypoints_edit_archive',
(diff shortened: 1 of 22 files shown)


tsauerwein:
Something to test: If a document has associations, check that these associations still exist after the revert.

asaunier:
Indeed, that's something I needed to manage:
* hide association tools when editing an archive version
* remove ``associations`` from the posted data

What happens if there is no ``associations`` attribute in the posted data? API crashes? No change to ``associations``? All associations are removed?

tsauerwein:
> What happens if there is no associations attribute in the posted data? API crashes? No change to associations? All associations are removed?

The associations should not be changed.

asaunier:
Remaining tasks:
* [ ] add view/route for xreports
* [ ] check user permissions before displaying the "Restore" button in the archived versions views: auth users for collaborative docs / associated users or moderators for personal docs
* [ ] ignore associations (don't show/send them + ) when editing an archived version (including for enabling the save button when associated WP etc. are mandatory)

asaunier:
> check user permissions before displaying the "Restore" button in the archived versions views: auth users for collaborative docs / associated users or moderators for personal docs

The first versions of this PR only showed the revert button to moderators. Commit 1401f96 uses function ``userCtrl.hasEditRights()`` which is already used to decide if the editing button should be shown in the standard view pages. Unfortunately some data such as assocations (including associated users) or creator user_id are not provided for archives. As a result the editing right permissions check does not work for personal documents (OK for collaborative docs). On the other hand since personal documents are editable only by their authors or associated users, they are much less likely to be vandalized. Moderators still have access to the revert function for such documents.

asaunier:
Another "problem": the "revert" button is not shown (except for moderators) for protected documents... except that the "protected" attribute is the one that comes with the edited archived version!

Which means that if the document was not protected at the time of the archive, the "revert" button is shown even though the document is currently protected. On the other hand, if the document is no longer protected but was protected at the time of the archive version, the "revert" button is not shown (except for moderators).

Please note though that the API actually does a "protected" status check (based on the last version) check and will not accept a change of a protected document anyway. So it's only a UI problem.

There is no simple solution because the updated ``protected`` attribute is not available when viewing an archive. This attribute should be added by the API in the metadata provided along with the current archive document data:
https://github.com/c2corg/v6_api/blob/53ac2dc5c460fe33bcf3336b4b023cc5a451f405/c2corg_api/views/document_version.py#L74-L79
but since it is not critical and quite unlikely, perhaps we could live with this approximation.

asaunier:
Replaced by https://github.com/c2corg/v6_ui/pull/1669

Issue #1099, opened 2016-12-04
Error 500 with an outing


asaunier:
http://c2corgv6.demo-camptocamp.com/outings/93037/fr/voie-plume => Internal Server Error
(ditto with the same outing on demov6):

```
[Sun Dec 04 22:55:14.695016 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/outing/view.html", line 135, in render_body
[Sun Dec 04 22:55:14.695022 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]     ${show_associated_routes(outing)}
[Sun Dec 04 22:55:14.695029 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/helpers/view.html", line 280, in render_show_associated_routes
[Sun Dec 04 22:55:14.695036 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]     ${add_seo_association_links(routes, 'routes')}
[Sun Dec 04 22:55:14.695048 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/helpers/view.html", line 273, in add_seo_association_links
[Sun Dec 04 22:55:14.695056 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]     <%def name="show_associated_routes(document, type='routes')">\\
[Sun Dec 04 22:55:14.695078 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_ui/c2corg_ui/templates/helpers/view.html", line 189, in render_add_seo_association_links
[Sun Dec 04 22:55:14.695085 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303]     <%
[Sun Dec 04 22:55:14.695091 2016] [wsgi:error] [pid 20886] [remote 10.27.71.4:22303] KeyError: 'title_prefix'
```

dkocich:
[...]

PR #1103, opened 2016-12-05
Fix assoc. card links problems with title / title-prefix for route

dkocich:
Now documents where associated routes miss `title/title_prefix` are visible and broken routes can be easily identified. It is not possible to view detail of route this route (201889) because it requires adjusting many things on UI to see it. It is probably better to fix them in DB than using UI.

Close https://github.com/c2corg/v6_ui/issues/1099

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -165,7 +165,11 @@
       % for doc in documents:
         <%
             locale = doc['locales'][0]
-            title = locale['title_prefix'] + ' : ' if doctype == 'routes' and locale['title_prefix'] is not None else ''
+            if 'title' in locale:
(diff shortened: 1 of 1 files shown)


asaunier:
```
locale['title_prefix'] = 'TitlePrefixError'
locale['title'] = 'TitleError'
```
looks weird to me.

Isn't it a problem that should be fixed in the API instead?

dkocich:
Wells, it points out easily in which documents the problem can be, rest of page works without error and we know what to fix on API/DB. But yes, just an empty string is better probably.

asaunier:
I would prefer we change nothing in the UI and that the API actually returns the ``title`` and ``title_prefix`` attributes (possibly empty or ``None``).

asaunier:
I think this PR is no longer relevant and could be discarded. Thanks anyway!

dkocich:
Ok.

PR #1107, opened 2016-12-06
Show article author/participants on the view page

dkocich:
On API of articles, we have author field. I think it is useful to show it directly in information window so that user can quickly know who is the origin author/owner of copyrights (and he can get to his page with more similar articles) without making an additional request on the history of the current document.

I am not sure about participants field -
In articles, there are sometimes more participants (f.e. here API http://localhost:6548/articles/147444 UI http://localhost:6558/articles/147444/fr/l-esprit-des-fetes-par-jean-pierre-banville) and it can be also useful to list them on the view page. But when there is only one author (and the same person as associated user), it can be redundant to show both fields.

image

I am not sure if adding new user associations to articles should be possible only on creation or editing or not at all. I think we miss both cases now and NO association user-article is now not created when an article is created.

### DIFF
--- a/c2corg_ui/templates/article/helpers/view.html
+++ b/c2corg_ui/templates/article/helpers/view.html
@@ -31,7 +31,7 @@
       <p><span class="value-title" translate>article_type</span>: <span class="value" x-translate>${article['article_type']}</span></p>
     % endif
 
-    % if article.get('author') and article['article_type'] != 'collab':
+    % if article.get('author') and article['article_type'] != 'collab' and article['author']['user_id'] != article['associations']['users'][0]['document_id']:
       <%
           user_id = article['author']['user_id']
           user_lang = article['locales'][0]['lang']
@@ -48,8 +48,8 @@
 </div>
 </%def>
 
-<%def name="get_article_participants(article)">\
-  % if article.get('participant_count') or article['locales'][0].get('participants') or article['associations']['users']:
+<%def name="get_article_associated_users(article)">\
+  % if article['associations']['users']:
     % if len(article['associations']['users']) > 1 or not article['associations']['users'][0]['document_id'] == article['author']['user_id']:
       <div class="name-icon-value" ng-click="detailsCtrl.openTab($event)">
         <h4><span translate>associated users</span><span class="glyphicon glyphicon-menu-right"></span></h4>
@@ -64,11 +64,10 @@
                 <%
                     user_id = user['document_id']
                     user_lang = user['locales'][0]['lang']
-                    is_last_participant = loop.last and not article['locales'][0].get('participants')
+                    is_last_associated_user = loop.last
                 %>
-                <a href="${request.route_path('profiles_view', id=user_id, lang=user_lang)}">${user['name']}</a>${'' if is_last_participant else ', '}
+                <a href="${request.route_path('profiles_view', id=user_id, lang=user_lang)}">${user['name']}</a>${'' if is_last_associated_user else ', '}
               % endfor
-              ${article['locales'][0].get('participants') or ''}
             </p>
           % endif
 
--- a/c2corg_ui/templates/article/helpers/view.html
+++ b/c2corg_ui/templates/article/helpers/view.html
@@ -31,7 +31,7 @@
       <p><span class="value-title" translate>article_type</span>: <span class="value" x-translate>${article['article_type']}</span></p>
     % endif
 
-    % if article.get('author'):
+    % if article.get('author') and article['article_type'] != 'collab':
       <%
           user_id = article['author']['user_id']
           user_lang = article['locales'][0]['lang']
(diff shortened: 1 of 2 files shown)


asaunier:
It indeed makes sense to display the author's name in the detail view of personal article (I think it has even been mentioned in an issue).

Please note that:
* "participants" is not the correct name for articles => "author"
* personal articles are... personal :) Which means there is only one author (not to mention moderators that would have made changes but they don't need to be listed as authors) => only display the original article creator as author.
* no need to mention the author twice. In your screenshot, there is a "participants" section and an "author" attribute => making the author name clickable would be enough.

fbunoz:
>  only display the original article creator as author.

It's better to display the associated user(s), as a moderator can create the article and associate the text author.
A personal article should be associated to only one user, but it may be several users after a merge or other moderation operation. And I think it is not a problem to have several author (it is like an outing).

dkocich:
Link with an example I posted is a personal article and it contains 2 associated users. The first one is an author, the other one did not change the document - he is only associated to doc. (an article can be written about other associated users, etc.)... (they both are not moderators in this case)
=> I think it is fine to show both author field and then the article's associated users even if it is a personal article.

So maybe it can be somehow like this now?
* 1 personal art. - if author is same as only ONE associated user => show only author field and participants should be hidden (will be visible after adding new associations)
* 2 personal art. - if there are muliple assoc. users => show both author and assoc. users fields
* 3 collab art.- show only associated users (if there are any), there is no author field (they can be listed in history)

1 image 2 image 3 image

fbunoz:
> Link with an example I posted is a personal article and it contains 2 associated users. The first one is an author, the other one did not change the document - he is only associated to doc. (an article can be written about other associated users, etc.)... (they both are not moderators in this case)

It typically the case where a moderator creates the article by copying the text from the original author, and associates the author. But he forgot to dissociate himself.

For the article, the creator is not important (he can be a moderator) and it needn't be displayed.
Only the associated user should be displayed as the author.

>  1 personal art. - if author is same as only ONE associated user => show only author field and participants should be hidden (will be visible after adding new associations)

OK.
But if there is only one associated user but he isn't the creator, he must be shown as the author.

>   2 personal art. - if there are muliple assoc. users => show both author and assoc. users fields

The author must be one of the associated users.
I think for a personal article it's better to display the associated user list in the author field (like for 1 assoicated user), and the "participant" field should not be displayed.

>    3 collab art.- show only associated users (if there are any), there is no author field (they can be listed in history)

OK.

stef74:
@dkocich need review or some extra work ?

c2corg_ui/templates/article/helpers/view.html
```diff
@@ -31,10 +31,49 @@
...
+          <p translate>associated users</p>
+        </div>
+        <span class="detail-text accordion">
+          % if article.get('associations').get('users'):
+            <p>
+              % for user in article['associations']['users']:
+                <%
+                    user_id = user['document_id']
+                    user_lang = user['locales'][0]['lang']
+                    is_last_participant = loop.last and not article['locales'][0].get('participants')
```
asaunier: This concept of "participants" looks weird to me. There is no ``participants`` attribute in the article model.

dkocich: That is really little weird. I renamed it to assoc. users and cleaned code little bit.

c2corg_ui/templates/article/helpers/view.html
```diff
@@ -31,10 +31,48 @@
       <p><span class="value-title" translate>article_type</span>: <span class="value" x-translate>${article['article_type']}</span></p>
     % endif
 
+    % if article.get('author') and article['article_type'] != 'collab' and article['author']['user_id'] != article['associations']['users'][0]['document_id']:
```
asaunier: Instead of ``article['article_type'] != 'collab'``, why not testing ``article['article_type'] == 'personal'``? It's more straightforward, isn't it?

I don't understand the full condition anyway: if the article is personal, we should show the author's name. Why do you test if ``article['author']['user_id'] != article['associations']['users'][0]['document_id']``?

c2corg_ui/templates/article/helpers/view.html
```diff
@@ -31,10 +31,48 @@
...
+
     % if article.get('quality'):
       <p><span class="value-title" translate>quality</span>: <span class="value" x-translate>${article['quality']}</span></p>
     % endif
 
     </span>
 </div>
 </%def>
+
+<%def name="get_article_associated_users(article)">\
```
asaunier: Is is really possible to associate users to an article?

dkocich: Currently, there is no function to do it but it is allowed in v6_common/associations and old data contain associated users also.

c2corg_ui/templates/article/helpers/view.html
dkocich: > But if there is only one associated user but he isn't the creator, he must be shown as the author
I think this is the reason why I added it. It is confusing when to show what. Should I change/delete it to keep more simple?

asaunier:
@dkocich My questions are quite minor, I suggest that you merge and will see at usage if a different approach is needed. Thanks.

Issue #974, opened 2016-11-23
Create new SVG icon for reports?


dkocich:
To add reports to the UI (https://github.com/c2corg/v6_ui/pull/905) I probably need some new icon for them to put reports in the map. Nothing is in the folder right now.

Right now I am using [waypoints/misc.svg](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/img/documents/waypoints/misc.svg). Something similar can be maybe used just with an exclamation mark instead of the mountain?. 
![image](https://cloud.githubusercontent.com/assets/6165660/20563377/9e7cfe1e-b188-11e6-8583-002c9422986e.png)

asaunier:
See also https://github.com/c2corg/v6_ui/issues/390#issuecomment-257816638

A possibility would be to use the v5 icon: https://github.com/c2corg/camptocamp.org/blob/master/web/static/images/svg/xreports.svg

PR #1108, opened 2016-12-06
Xreports advanced - adding/editing xreports

dkocich:
Work in progress..
Related to icon issue https://github.com/c2corg/v6_ui/issues/974

Continue of 1st merged part here https://github.com/c2corg/v6_ui/pull/905

Stil TODO from the previous discussion:

by me:
- [x] From-To date calendar in advanced filter does not work correctly (results are not filtered) ... fixed https://github.com/c2corg/v6_api/pull/584

By @asaunier
* advanced search:
* [ ] ~~``severity``: I think when the values will be translated they will be too long to fit under the slider. Perhaps you should consider using alternate (shorter) values?~~
  * [x] ~~LATER nice to have: filter on user id (as for outings), eg. http://localhost/reports#u=377 ... opened https://github.com/c2corg/v6_api/pull/585~~
* detail view:
* [x] history page is OK but diffing 2 versions is KO
* [x] associated document types are maybe not correct. For instance, should we really be able to associate a WP (in v5 one cannot, except climbing sites - by the way @tsauerwein perhaps another migration change as for outings associated with climbing sites?)

* other:
- [x] _fix `XreportEditingController` as mentioned in_ https://github.com/c2corg/v6_ui/pull/905#discussion_r90652076
* [x] _Fix remove association rights mentioned in_ https://github.com/c2corg/v6_ui/pull/905#discussion_r90834720 ?
* [x] _Do not put personal data into mako template - use separate authorized request https://github.com/c2corg/v6_ui/pull/905#discussion_r90835831 _
note - similar approach should be used to edit xreport (load part of form with private data if I am authorized) ... (**requires change in API also?**)

New issues?:
* [x] new created Xreports are not associated to the author by default (but that is probably OK - same for articles now)


dkocich:
Hello @asaunier, maybe you can see and test this branch late? Thanks.

c2corg_ui/templates/xreport/filters.html
```diff
@@ -29,87 +29,91 @@
...
+            values-list="['${"','".join(severities) |n}']"></app-slider>
+        </div>
 
-      <div class="col-xs-12 col-sm-6 filter">
-        <label translate>nb_impacted</label><br>
-          <app-slider filter="ximp" filters-list="filtersCtrl.filters" step="1" unit="" min="0" max="25"></app-slider>
-      </div>
+        <div class="col-xs-12 col-sm-12 filter">
+
+  ##         <div class="col-xs-12 col-sm-6 filter">
```
asaunier: Do you really need this commented-out section?

asaunier:
Could you please change the ``Reports`` titles in
* https://github.com/c2corg/v6_ui/blob/xreports-advanced/c2corg_ui/templates/xreport/index.html#L5
* https://github.com/c2corg/v6_ui/blob/xreports-advanced/c2corg_ui/templates/xreport/index.html#L25
* https://github.com/c2corg/v6_ui/blob/xreports-advanced/c2corg_ui/templates/404.html#L45
?

They are confusing. Outings are also reports... I don't know what title is more appropriate. Perhaps simply "Xreports"?

asaunier:
Problems I have noticed when testing in my instance:

### Editing

* it should not be possible to select a date in the future when creating/editing an xreport (already OK in the advanced search filters)
* fields "number of persons" and "number of impacted persons" could be on the same line, no on dedicated lines
* it is not possible to associate users. The "Associated participants" section only has a title but no field or association tool.
* the association tool below the map is confusing: there is no title or tip explaining what it is used for:
capture du 2017-01-16 18-52-17
* adding content to one of the "factors" fields prevent to click on the save or preview buttons

dkocich:
Thanks for the review.
> adding content to one of the "factors" fields prevent to click on the save or preview buttons

I added restriction to write at least +-10  characters to prevent having only one-word texts inside each section (tell me if I should remove it). Saving and other functions should work fine.

asaunier:
> I added restriction to write at least +-10 characters to prevent having only one-word texts inside each section

Why not the same for description and summary then?
And nothing tells you that you should at least write 10 characters.
I think it is confusing and that we should not bother about the length of the content: it's better to remove those conditions.

dkocich:
> the association tool below the map is confusing: there is no title or tip explaining what it is used for:

with no associations
alt text
after adding of associations
alt text

> it is not possible to associate users. The "Associated participants" section only has a title but no field or association tool.

It should work now correctly during creation of the xreport.
alt text

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -367,6 +383,11 @@ app.Authentication.prototype.needAuthorization = function(method, url) {
     return true;
   }
 
+  if (url.indexOf('/xreports') !== -1 && this.isAuthenticated()) {
```
asaunier: @dkocich Could you explain what this case is used for?

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -242,6 +242,9 @@ app.DocumentEditingController.prototype.successRead = function(response) {
   if (this.modelName === 'image') {
     data['date_time'] = new Date(data['date_time']);
   }
+  if (this.modelName === 'xreport') {
+    data['date'] = new Date(data['date']);
+  }
```
asaunier: This should be moved to a ``filterData()`` method in ``app.XreportEditingController``. As in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/editing/imageediting.js#L58-L69

This will also fix the conflict with this file.

c2corg_ui/static/js/auth/authservice.js
dkocich: I had a problem when I wanted to show private data in the form on the EDIT page  (of my own Xreport). They were loaded only always when authorized on View page http://localhost:6558/xreports/edit/828344/en?debug
Now I fixed it little bit in the last commit.

c2corg_ui/views/xreport.py
```diff
@@ -39,6 +64,47 @@ def render_page(xreport, locale):
 
         return self._get_or_create_detail(id, lang, render_page)
 
+    def _get_xreport_info(self, id, lang, old_api_cache_key=None):
```
asaunier: This method seems to be used nowhere!?

c2corg_ui/views/xreport.py
```diff
@@ -20,6 +22,29 @@ def redirect_to_full_url(self):
 
     @view_config(route_name='xreports_view')
     def detail(self):
```
asaunier: This method behaviour seems to be the same than for the profile data: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/profile.py#L21-L71

but I notice significant differences, is there a reason for this?

For instance profile method has a ``load_data`` function:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/views/profile.py#L64-L71
```
        def load_data(old_api_cache_key=None):
            not_modified, api_cache_key, document_and_locale = \
                self._get_profile_info(id, lang, old_api_cache_key)
            return not_modified, api_cache_key, document_and_locale

        return self._get_or_create(
            (id, lang), cache_document_detail, load_data, render_page,
            self._get_cache_key)
```

whereas the xreport detail function finishes with
```
return self._get_or_create_detail(id, lang, render_page)
```

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -379,12 +374,9 @@ app.Authentication.prototype.needAuthorization = function(method, url) {
...
       url.indexOf('/profiles') !== -1 ||
-      url.indexOf('/personal-feed') !== -1) {
-    return true;
-  }
-
-  if (url.indexOf('/xreports') !== -1 && this.isAuthenticated()) {
-    // forward auth header if user is authenticated
+      url.indexOf('/personal-feed') !== -1 ||
+      url.indexOf('/xreports/edit') !== -1 ||
+      url.indexOf('/xreports/')) {
```
asaunier: Isn't a test value missing here?

asaunier:
Concerning the associated users, I have noticed the current contributor (that creates the report) is not directly associated, as when creating an outing. That would make sense.

c2corg_ui/views/xreport.py
dkocich: I think that on profiles page, there is Activity feed with cached documents which are requested and rendered many times after scrolling down. I am querying one document once.

c2corg_ui/static/js/auth/authservice.js
```diff
@@ -363,7 +374,9 @@ app.Authentication.prototype.needAuthorization = function(method, url) {
       url.indexOf('/users/mailinglists') !== -1 ||
       url.indexOf('/users/following') !== -1 ||
       url.indexOf('/profiles') !== -1 ||
-      url.indexOf('/personal-feed') !== -1) {
+      url.indexOf('/personal-feed') !== -1 ||
+      url.indexOf('/xreports/edit') !== -1 ||
+      url.indexOf('/xreports/') !== 1) {
```
asaunier: Are you sure of ``!== 1``? Shouldn't it be ``-1``?
If yes, ``/xreports/`` should also work for ``/xreports/edit``?

c2corg_ui/views/xreport.py
asaunier: The feed shown in the profile pages are handled using the feed directive. There is no feed caching managed by the profile UI services. The UI server side caching is not even aware feed items are loaded through the profile data template.

I don't think there should be a difference in the code of profile's ``profile`` and xreport's ``detail`` views.

c2corg_ui/static/js/auth/authservice.js
dkocich: Yes. It works fine when only `/xreports/` is used here.

dkocich:
> Concerning the associated users, I have noticed the current contributor (that creates the report) is not directly associated, as when creating an outing. That would make sense.

Now he is and card shows automatically but when I save the Xreport he is not visible on API in associations http://api.demov6.camptocamp.org/xreports/828362 Any clue?
This can be tested only with adding of new Xreport when did not merge PR that allows adding of users while updating Xreport.

dkocich:
> dkocich: ...card shows automatically but when I save the Xreport he is not visible on API in associations...

@asaunier So this was a problem only when I was using the demo server. On localhost with current master everything works fine. Users can be added while creating or editing Xreport.

asaunier:
@dkocich I have tested the branch in my instance.

### Editing form

The editing form seems OK but

* [ ] it is not possible to make a preview, the request is 500 because of some error in the mako templates:
```
  File "/home/alex/Desktop/v6_ui/c2corg_ui/templates/xreport/preview.html", line 32, in render_body
    ${get_xreport_participants(xreport)}
  File "/home/alex/Desktop/v6_ui/c2corg_ui/templates/xreport/helpers/view.html", line 105, in render_get_xreport_participants
    <%
KeyError: 'locales'
```
* [ ] using the association tool, waypoints results are returned but nothing appears when selecting such a result (OK for routes) => waypoints should not appear here.
* [ ] it is not possible to associate an article (not article results are returned).

### Detail view

It is possible to associate routes and articles => OK
* [ ] But also waypoints => it should not be possible.


### General

* [ ] I have not understood how to associate an outing whether in the editing form or in the detail view.

dkocich:
> ...it is not possible to make a preview...

There was a small problem with locales of associated users which prevented to create a preview (it should be fine now).

>  using the association tool, waypoints results are returned but nothing appears when selecting such a result (OK for routes) => waypoints should not appear here.

I am little confused here. So when I create a new report - waypoint association is allowed on API ([document_schemas.py](https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_associations.py#L37)), but when I want to go back in 2 minutes and add it - then it is not allowed on API ([model/associations.py](https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/association.py#L384)). Is it a mistake (waypoints should not be allowed anywhere)? If it is ok to allow it only during Xreport creation I will adjust dataset search for creating/editing of Xreport. But maybe we should make sure that waypoint is required (or allow to add it in some time interval after creation) ~~(OR allow to add it later which is something what you did not want as was said in a discussion yesterday)~~

> it is not possible to associate an article (no article results are returned).

Fixed. It works now in both View+Edit pages.
image

> I have not understood how to associate an outing whether in the editing form or in the detail view.

It is allowed on View page now.
I see that it is allowed probably only on the article page View page. Maybe an option to add it also from the edit page can be implemented later? 
image

asaunier:
I don't understand your question about waypoint association.
It should not be possible to associate waypoint to an xreport, whereas it is being created or edited, neither in the detail view.
See https://github.com/c2corg/v6_api/pull/606/files

I have tried to run the updated branch in my instance but the build fails:
```
/home/alex/Desktop/v6_ui/c2corg_ui/static/js/editing/xreportediting.js
  55:9  error  Expected indentation of 10 space characters but found 8  indent
```

asaunier:
I have just tested the new version version of the branch: the code looks good to me and the behaviour of the UI seems OK. =D

So for me it's OK to merge the PR (perhaps you could squash the commits first or when merging?).

dkocich:
Thanks, finally! :) Ok, I will squash some commits together and merge.

asaunier:
Thanks for your hard work @dkocich !

asaunier:
@dkocich Are there issues that can now be marked as "ready for demo"?

dkocich:
I couldn't find any in PR references and issues log + many unfinished things were not allowed currently on UI => No issues to be marked now...

Issue #849, opened 2016-11-05
Creating new documents using the simple search / association tool


asaunier:
When searching for an association, if there is no result, the association tool offers buttons to create new associated documents:
<img width="602" alt="capture d ecran 2016-11-05 a 09 43 27" src="https://cloud.githubusercontent.com/assets/1192331/20028838/96e1b618-a33c-11e6-9843-5162ffbf7205.png">

There is still in issue though, because such buttons are only shown if no result can be found, which does not happen that often because the search engine finds a bit too many results I think and because for routes, if the name is quite common and used for other unrelated routes (eg. "face N"), you'll find a bunch of unrelevant results.
Perhaps the "create a new ..." tool should appear after each set of results anyway?

asaunier:
See also https://github.com/c2corg/v6_ui/issues/1086

asaunier:
Moving here the related issue https://github.com/c2corg/v6_ui/issues/1086

<img width="650" alt="capture d ecran 2016-12-03 a 17 31 40" src="https://cloud.githubusercontent.com/assets/1192331/20860693/94b5f324-b97e-11e6-81e2-ab0d32a1d0fd.png">

* Why "no result" when there are articles found? => rename this title or even remove it 
* Why showing it before the articles list anyway?

I have the feeling this "create a new..." block is not OK:
* [ ] since it appears even when there are results, it should be renamed to "no match?". Or we could simply remove the title.
* [ ] Perhaps the "create" button should be right at the end of the list of every document type (matching the dataset, for instance ``dataset="wcb"``) => if no result for a given doc type, show the doc type title + "no match" + the create button?
* [ ] Never show the "create" buttons when the association tool is within an editing form. See https://github.com/c2corg/v6_ui/issues/1059.
* [ ] Btw the "create" buttons are not really relevant in the association tool because clicking on such a button will indeed create a document but not associate it to the previous document.
* [ ] It is not relevant to show "create" buttons for document types that are not listed in the tool's ``dataset``.

At short term, perhaps we should simply disable the "create" buttons in the association tool because they would require a lot of work to make the behaviour consistent.

asaunier:
Minimal changes have been made in #1109 
The issue should remain open anyway but could be postpone to another milestone.

stef74:
Ok for 1st approach

PR #1109, opened 2016-12-06
"No result" buttons in simple search suggestions

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/849

* [x] No need for outings and xreports creation buttons in simple search
* [x] "empty" title is renamed to "No match?" (because it appears also when there are results). Remains "no results" in association context to at least gives a feedback (see next point).
* [x] Show creation buttons only in the simple search context, not in an association context. This would make sense but would require much more work and we don't have time before golive.
* [ ] Article results should appear before the "no match?" section.

The last point is still open. I have not figure out why articles appear AFTER the "no match" section:
capture d ecran 2016-12-03 a 17 31 40

Anyone has a tip?

### DIFF
--- a/c2corg_ui/static/js/search/simplesearch.js
+++ b/c2corg_ui/static/js/search/simplesearch.js
@@ -295,10 +295,11 @@ app.SimpleSearchController.prototype.createDataset_ = function(type) {
  * @private
  */
 app.SimpleSearchController.prototype.createAndInitBloodhound_ = function(type) {
+  var limit = 7;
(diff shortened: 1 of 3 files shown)


asaunier:
I don't know how to make sure article results are always shown before the creation buttons and would like to merge as is except if someone knows how to do it.
Please review.

tsauerwein:
Not sure how we could change the order. Your other changes look good to me. Feel free to merge.

asaunier:
Thanks for your review and your changes. Merging.

Issue #1041, opened 2016-11-29
Poor error message when login fails


asaunier:
At least a translation problem.

<img width="300" alt="capture d ecran 2016-11-29 a 23 49 35" src="https://cloud.githubusercontent.com/assets/1192331/20732341/8f8a392a-b68e-11e6-8315-3893be2ada59.png">

loicperrin:
So what is the problem in this case (to find the right translation)?
The user has entered a wrong username or a wrong password?

asaunier:
> The user has entered a wrong username or a wrong password?

Yes.

Actually there is already a translation for "Login failed:" in Transifex.
https://www.transifex.com/camptocamp-association/c2corg_ui/viewstrings/#fr/v6/96458637?q=Login%20failed

And the message should be translated here (or perhaps only marked as to be translated?):
https://github.com/c2corg/v6_ui/blob/6d385f9fb2fc96325a52e3dbbf2e1b62efd0c381/c2corg_ui/static/js/apiservice.js#L379

loicperrin:
So it is OK in Transifex, the modification has to be made in the API code.

stef74:
@asaunier why do you remove "association" ? Its a translation issue (it's my understanding).

asaunier:
I don't think it is a translation issue. The string "Login failed:" is indeed translated in Transifex. It's rather a code problem.

loicperrin:
I think there are other error messages with account creation that cannot be translated in Transifex.

PR #1111, opened 2016-12-06
Fix missing translation in error message

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1041

sans titre

### DIFF
--- a/c2corg_ui/static/js/alertservice.js
+++ b/c2corg_ui/static/js/alertservice.js
@@ -89,7 +89,7 @@ app.Alerts.prototype.addError = function(msg) {
  * @export
  */
 app.Alerts.prototype.addErrorWithMsg = function(msg, errors) {
-  var content = msg + '<br>' + this.formatErrorMsg_(errors);
+  var content = this.filterStr_(msg) + '<br>' + this.formatErrorMsg_(errors);
   var timeout = 5000;
   this.addLoading_(timeout);
   this.alerts_.push({


asaunier:
Actually I have the feeling a bigger fix would make sense (later?) and that there are unconsistancies in the code.

For instance:
* ``Login failed:`` (with colons) as the main message. see https://github.com/c2corg/v6_ui/blob/6d385f9fb2fc96325a52e3dbbf2e1b62efd0c381/c2corg_ui/static/js/apiservice.js#L379
* ``Login failed`` (no colons) in the error description returned by the API and used at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/alertservice.js#L144

We probably don't need both lines since they basically say the same thing. Not sure other error messages such as https://github.com/c2corg/v6_ui/blob/6d385f9fb2fc96325a52e3dbbf2e1b62efd0c381/c2corg_ui/static/js/apiservice.js#L365 are also duplicated.

Perhaps we could simply rely on the API answer: https://github.com/c2corg/v6_api/blob/40de9fe2d5fcac8359898aaeae22db9acd117304/c2corg_api/views/user.py#L456
Instead of ``name`` "user" could we use a more explicit string to be translated? For instance
```
request.errors.add('body', 'Incorrect username or password', 'Login failed')
```

asaunier:
As explained above, I think more changes would make sense but I am not sure to have time before golive. The proposed changes in this PR at least make the main error messages translated.
Please review.

Issue #1046, opened 2016-11-30
Enable client-side resizing when uploading an image


asaunier:
There is an option in ng-file-upload, ``ngf-resize``. See https://github.com/c2corg/v6_ui/issues/929#issuecomment-262923465

asaunier:
ng-file-upload config is here: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/imageuploader.html#L1-L12

> And what would be considered as a "too big image" and what would be the "correct, smaller size"?
> We can take V5 values : size < 8192 x 2048 px AND weight < 2 Mo

It is possible to specify for what dimensions an image should be resized:
```
ngf-resize-if="$width > 1000 || $height > 1000" or "resizeCondition($file, $width, $height)"
```
I don't know if it is possible to test the image weight? Perhaps using ``$file``?

At least it is possible to add limitations/validations and refuse images that would be too big (or too small...). See the validations section in the reference at https://github.com/danialfarid/ng-file-upload#file-select-and-drop
For instance ``ngf-min-size, ngf-max-size, ngf-max-total-size="100" in bytes or "'10KB'" or "'10MB'" or "'10GB'"``  or ``ngf-min-height, ngf-max-height, ngf-min-width, ngf-max-width="1000" in pixels``.

Possible resizing options:
```
ngf-resize="{width: 100, height: 100, quality: .8, type: 'image/jpeg', 
               ratio: '1:2', centerCrop: true, pattern='.jpg', restoreExif: false}" 
               or resizeOptions() // a function returning a promise which resolves into the options.
    // resizes the image to the given width/height or ratio. Quality is optional between 0.1 and 1.0), 
    // type is optional convert it to the given image type format.
    // centerCrop true will center crop the image if it does not fit within the given width/height or ratio. 
    // centerCrop false (default) will not crop the image and will fit it within the given width/height or ratio 
    // so the resulting image width (or height) could be less than given width (or height).
    // pattern is to resize only the files that their name or type matches the pattern similar to ngf-pattern.
    // restoreExif boolean default true, will restore exif info on the resized image.
```

See the demo at https://angular-file-upload.appspot.com/

ginold:
Exactly what i was looking for! I will start to work on this today.

tsauerwein:
Please test if the resizing doesn't freeze the browser when trying to add a lot of images at once (especially on slow machines).

asaunier:
[...]

PR #1113, opened 2016-12-06
Resize too big images before uploading them

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1046

In v5, client-side resizing is managed at https://github.com/c2corg/camptocamp.org/blob/72e777075f2e6260b1d2ef563ab0f90c383038ba/web/static/js/plupload.wrapper.js#L135-L144

* png and jpg images <2M will get resized only if they exceed c2c limits (8192x2048)
* images >2M will be resized to max 4096x1024
* svg and gifs won't be resized anyway

I have not fully succeeded in completing such a complicated pattern but the current PR at least adds the support of a simpler condition that might be already quite enough (at least as a first version?). Perhaps the resized dimensions should be confirmed. @lbesson @fbunoz @mgerbaux what do you think? See a962e23

It seems possible to pass as ``ngf-resize`` a function "returning a promise which resolves into the options" (see https://github.com/danialfarid/ng-file-upload#file-select-and-drop) that could probably be used to switch among resized dimensions depending on the file size but I have not figured out how to write such a promise function. Here is a piece of code that could be used as base:
```
/**
 * @param {Object} file
 * @return {Object}
 * @export
 */
app.ImageUploaderController.prototype.resizeOptions = function(file) {
  if (file.size > app.ImageUploaderController.MAX_IMAGE_SIZE) {
    return {'width': 4096, 'height': 1024, 'quality': 1.0};
  } else {
    return {'width': 8192, 'height': 2048, 'quality': 1.0};
  }
};
```
See also https://github.com/danialfarid/ng-file-upload/blob/052ea1942dfbe5cb9cb14b2caeb35fec15c3040d/src/model.js#L82-L99

### DIFF
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -387,6 +387,23 @@ app.ImageUploaderController.prototype.selectOption = function(object, property, 
 };
 
 
+/**
+ * @param {Object} file
+ * @param {number} width Width of uploaded image.
+ * @param {number} height Height of uploaded image.
+ * @return {boolean}
+ * @export
+ */
+app.ImageUploaderController.prototype.resizeIf = function(
+    file, width, height) {
+  if (file.type === 'image/jpeg' || file.type === 'image/png') {
+    return file.size > 2 * 1024 * 1024 /** 2 MB */ ||
+      width > 4096 || height > 2048;
+  }
+  return false;
+};
+
+
 app.module.controller('AppImageUploaderController', app.ImageUploaderController);
 
 
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -153,52 +153,6 @@ app.ImageUploaderController = function($scope, $uibModal, $compile, $q, appAlert
 
 
 /**
- * We have to use a secondary controller for the modal so that we can inject
- * uibModalInstance which is not available from the first level controller.
- * @param {Object} $uibModalInstance modal from angular bootstrap
- * @constructor
- * @ngInject
- * @returns {app.ImageUploaderModalController}
- */
-app.ImageUploaderModalController = function($scope, $uibModalInstance) {
-
-  /**
-   * @type {!angular.Scope}
-   * @private
-   */
-  this.scope_ = $scope;
-
-  /**
-   * @type {Object} $uibModalInstance angular bootstrap
-   * @private
-   */
-  this.modalInstance_ = $uibModalInstance;
-
-  $scope.$on('modal.closing', function(event, reason, closed) {
-    this.scope_['uplCtrl'].abortAllUploads();
-  }.bind(this));
-};
-
-
-/**
- * @export
- */
-app.ImageUploaderModalController.prototype.close = function() {
-  this.modalInstance_.close();
-};
-
-
-/**
- * @export
- */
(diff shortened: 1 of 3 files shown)


lbesson:
I don't know exactly how resizing works in v6 (and currently don't have time to look at it), but in v5, this is done using canvas. A few remarks:
* "c2c limits" where already there before I updated the upload code. This is some protection against odd formats, but it could probably be dropped (I mainly set it to prevent the image to be rejected by symfony, without questioning the "c2c limits")
* 4096x1024 was used as a compromise between quality and file size. (The goal is not to resize an image to something close to 2M, but to have a good looking image) Most of the time, the image was "good enough", but using canvas for resizing had some edge cases where these limits produced an image >2M, also it is too small for exotic formats like panoramas

Client-size resizing via canvas is quite quick, so we could make a better algorithm that would do several tries to find the best compromise depending on each image. But of course, doing something simple in the beginning (and maybe keep it simple afterwards) should be ok.

asaunier:
Thanks @lbesson for the explanations.

I have the feeling that we should have a simpler config (at least first). I also think it is a bit a pity to resize images higher than 1024px.
Server side, the resizing is currently set to a max height of 1500px for BI images:
https://github.com/c2corg/v6_images/blob/master/c2corg_images/__init__.py#L7
In addition, since more and more devices have vertical display (eg. tablets), it's not that relevant to make a big distinction between width and height, except perhaps for panoramas. 

That's why I have set the check (and resizing params) to w=4096 and h=2048.
See https://github.com/c2corg/v6_ui/pull/1113/commits/a962e237a77beae05d5c21ff697420b70c56d34d

Does that make sense?
No idea if with such a config we would prevent downloading 2MB+ images?!

lbesson:
maybe it would be easy to detect the orientation of the image by simply looking at the longest side.

Anyway, the 2MB restriction was only because one the server side, php was configured (well, it's default value) to refuse files larger that 2MB. And no resizing was performed by the server (aka file >2MB => error, files <2MB but bigger than c2c limits => error -- both errors being very frustrating for the user).

You seem to have resizing both on client and server side ?

One underlying question is whether it has sense to have quite big images (say >2MB) hosted on c2c :)

tsauerwein:
> It seems possible to pass as ngf-resize a function "returning a promise which resolves into the options" (see https://github.com/danialfarid/ng-file-upload#file-select-and-drop) that could probably be used to switch among resized dimensions depending on the file size but I have not figured out how to write such a promise function.

The problem is that this function gets a list of files, but you only can return a single resize-parameter object, which is used for all files.

asaunier:
> You seem to have resizing both on client and server side ?

The original is not resized server side. It's hosted as is.
Resized versions (BI, MI, SI) are indeed created, as in v5 actually.

Client side resizing is not enable yet, at least until this PR is merged :P 

> One underlying question is whether it has sense to have quite big images (say >2MB) hosted on c2c :)

I guess not because it would take some extra disc space.
On the other hand, perhaps in the future screens will be huge with big resolution and regenerating the BI images to bigger formats would be interesting.

I would say the main interest of client-side resizing is too reduce the upload-time and avoid the timeout problems we have encountered.

asaunier:
@tsauerwein Thanks for the review.

As discussed together it's probably safer to wait that the prod is gone live before merging the PR (and make real-size testing in the demo).

Issue #1117, opened 2016-12-07
Image slideshow is "moody"


tsauerwein:
Going forward and backward in the image slideshow does not work properly. When clicking on the next arrow, the next image flashes up, but then the current image is shown again.

fbunoz:
I see also this behavior, which is a bit random : sometime it's work correctly, sometime not.

More over, on desktop it would be better that a click on a large part of right or left image goes to the next/perv image, instead of clicking only on the small arrow. the keyboard arrow work also, but sometime we use the keyboard, or the mouse.

tsauerwein:
> More over, on desktop it would be better that a click on a large part of right or left image goes to the next/perv image, instead of clicking only on the small arrow. the keyboard arrow work also, but sometime we use the keyboard, or the mouse.

Please open a separate issue for that.

asaunier:
> More over, on desktop it would be better that a click on a large part of right or left image goes to the next/perv image, instead of clicking only on the small arrow. the keyboard arrow work also, but sometime we use the keyboard, or the mouse.

Please note it is possible to drag the image with the mouse cursor to the right or to the left. It's 2016, people no longer click, they drag :P

PR #1121, opened 2016-12-07
Disable history function in slideshow

tsauerwein:
[PhotoSwipe](http://photoswipe.com/documentation/options.html) has a history function, that is it changes the URL for example to `/#&gid=1&pid=1`. The idea is that you share this URL and the slideshow will open with the right image.

Which is a good idea, but it does not work right now and it creates the problem described in https://github.com/c2corg/v6_ui/issues/1117 in build mode. I am proposing to disable the history function for now. 

Closes https://github.com/c2corg/v6_ui/issues/1117


asaunier:
Indeed URLs with ``/#&gid=1&pid=1`` simply load the parent page, not the slideshow.
OK for removing the history since it seems not to work anyway.

ginold:
whoa, I would never have thought that it had an impact. I remember having this problem some time ago, but since a while I haven't observed it.
:+1:

PR #1122, opened 2016-12-07
Update xreports entry in Discourse menu

asaunier:
Xreports entry in the sidemenu is not the same in UI and in Discourse.


asaunier:
Anyone for this easy review?

Issue #690, opened 2016-10-05
Problems with columns layout


asaunier:
- [x] sorting problem:

There a problem with the columns layout we have just reintroduced in cards listings (eg. advanced search).

For instance with the outings: most recent outings appears in the leftmost column from top to bottom, then the next outings appear in the next column (top to bottom). See for instance:

![sans titre](https://cloud.githubusercontent.com/assets/1192331/19112228/f1c9d142-8b03-11e6-8364-630190f21dc8.png)

This is not very intuitive: users will most likely expect to see the most recent outings on top of the list, whatever the column.

In addition, for the home page feed, if infinite scrolling is used (or with pagination such as a "see more feed" button), older feed items will be added at the bottom.

asaunier:
* [ ] In FF only one card column is displayed (instead of 3 as with Chromium):
![sans titre](https://cloud.githubusercontent.com/assets/1192331/19151662/30f57c66-8bce-11e6-9cd2-9b63edc3ce23.png)

loicperrin:
- [ ] Another problem with columns layout: sometimes the first cards of the column are cut. Example:
![issue690](https://cloud.githubusercontent.com/assets/19718055/19622604/866d568c-98ad-11e6-9422-769fb1b2b081.jpg)

ginold:
What browser are you using?

stef74:
I reproduce in FF

loicperrin:
FF 49.0.2 (last version).
Seems to be OK with Chrome.

stef74:
I reproduce on FF

The first point for me is fixed
​

2016-10-24 9:17 GMT+02:00 ginold <notifications@github.com>:

asaunier:
> The first point for me is fixed

It's not. In the outings advanced search page, outings are still order from the the newest to the oldest vertically (columns) but not horizontally (rows). Outings at the top of the list should be the newest, whatever the column.

@ginold do we have a solution for this?

ginold:
not yet, I will have to look deeper into this problem.

ginold:
It seems that it cannot be fully solved with pure CSS (http://stackoverflow.com/questions/5234749/css-floating-divs-at-variable-heights). I managed to do something like this which is not that bad: 

![screenshot from 2016-11-03 14-18-35](https://cloud.githubusercontent.com/assets/7814311/19967438/7e3e3d92-a1d0-11e6-9456-971893414c40.png)

[...]

PR #1124, opened 2016-12-07
Layout columns

ginold:
related to #690 
related to #1138 
related to #1139

After a few unsuccessful attempts to work with `columns` and `ellipsis` (= hide a text too long and replace last letters with"..." -> we're in 2016 and it's impossible to do it for more than a single line with pure CSS!! -.- only chrome supports a special property to do it, how surprising...), I decided to go back again to `flexbox` which we had at the beginning and that we decided to throw away because all the cards had the same height per row of cards, and often it looked weird with one card full and the next one nearly empty.

Now I came back to this solution and time has passed, the cards have more data and it doesn't look that bad. However, I had to make a lot of adaptations and the cards have changed the layout, hopefully for the best:

**outings** when in the feed, the author info is redundant with "xxx has created an outing"
screenshot from 2016-12-07 09-24-35

**articles** you can show the last activities if more than 4 (anyway always showing them all was unreadable...). Could totally be used for outings, routes and others too, but i have never seen more than 3 activities.
screenshot from 2016-12-07 13-46-00

**routes**
screenshot from 2016-12-07 09-23-39

**areas**
screenshot from 2016-12-07 09-25-43

**xreports**
screenshot from 2016-12-07 14-10-17

**waypoints**
screenshot from 2016-12-07 09-25-01

**books**
screenshot from 2016-12-07 09-25-23

**users** in outing editing
screenshot from 2016-12-07 14-18-04

All in all, the cards look good IMHO and are sorted from left to right.
I didn't test the Images because they don't load at all and I can't upload, but it should be ok, there is very little to be shown.

Still looking for typos, it may happen after editing 19 files :P

### DIFF
--- a/c2corg_ui/static/partials/cards/articles.html
+++ b/c2corg_ui/static/partials/cards/articles.html
@@ -1,21 +1,35 @@
 <a ng-href="{{cardCtrl.createURL()}}" ng-init="doc = cardCtrl.doc; locale = cardCtrl.locale;">
   <div class="list-item-info">
 
-    <span class="list-item-title">{{::locale.title}}</span>
-    <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{::locale.lang}})</span>
-    <div class="value">{{::doc['article_type'] | translate }}</div>
+    <div class="list-item-title-lang">
+      <span class="list-item-title">{{::locale.title}}</span>
+      <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{::locale.lang}})</span>
+    </div>
+
+    <b class="article-type margin-t-sm margin-b-sm"><span class="green-text glyphicon glyphicon-tag margin-r-sm"></span>{{::doc['article_type']| translate }}</b>
 
     <div class="article-categories">
-      <span ng-repeat="category in ::doc.categories" class="value">{{::category | translate }}{{$last ? '' : ', '}}</span>
+      <b ng-repeat="category in ::doc.categories" class="value">{{::category | translate }}{{$last ? '' : ', '}}</b>
     </div>
 
-    <div class="article-activities">
-      <span ng-repeat="activity in ::doc.activities" class="route-activity icon-{{::activity}}"
-            uib-tooltip="{{ cardCtrl.translate(activity) }}"></span>
+    <div class="activities flex wrap-row margin-t-sm">
+      <span ng-repeat="activity in ::doc.activities | limitTo: 4"  class="activity icon-{{::activity}}"
+            uib-tooltip="{{ cardCtrl.translate(activity)}}"></span>
+
+      <span class="flex flexend-align show-more-activities"
+            ng-if="doc.activities.length > 4 && !cardCtrl.remainingActivities"
+            ng-click="$event.preventDefault(); cardCtrl.remainingActivities = true">
+        ...(+ {{::doc.activities.length - 4}})
+      </span>
+
+      <span ng-if="cardCtrl.remainingActivities" ng-repeat="activity in ::doc.activities | limitTo: (4 - doc.activities.length)"
+            class="activity icon-{{::activity}}" uib-tooltip="{{ cardCtrl.translate(activity)}}">
+      </span>
     </div>
 
-    <br>
     <div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
-         uib-tooltip="{{'quality'| translate}} : {{::doc['quality'] | translate}}" tooltip-placement="left"></div>
+         uib-tooltip="{{'quality'| translate}} : {{::doc['quality'] | translate}}" tooltip-placement="left">
+    </div>
+
   </div>
 </a>
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -6,12 +6,12 @@
   <section ng-show="${model}.associations.${type}.length > 0" class="associated-section">
     <label translate>Associated waypoints</label>
     <article class="associated-documents">
-      <div class="list-item"
+      <div class="list-item waypoints ${'to-route' if model == 'route' else ''}"
            ng-repeat="wp in ${model}.associations.${type} | orderBy: 'waypoint_type'"
            ng-class="{'main-waypoint': ${model}.main_waypoint_id === wp.document_id, 'new-item': wp['new']}">
         <app-card app-card-doc="wp" app-card-disable-click="true"></app-card>
         % if model == 'route':
-          <div class="set-main-waypoint" ng-if="${model}.main_waypoint_id === wp.document_id" ng-init="${model}.main_waypoint_title = wp.locales[0].title" translate>main waypoint</div>
+          <b class="set-main-waypoint green-text" ng-if="${model}.main_waypoint_id === wp.document_id" ng-init="${model}.main_waypoint_title = wp.locales[0].title" translate>main waypoint</b>
           <button class="btn btn-sm btn-default set-main-waypoint" translate
                   ng-click="${model}.main_waypoint_id = wp.document_id; ${model}.main_waypoint_title = wp.locales[0].title"
                   ng-if="${model}.main_waypoint_id != wp.document_id">
--- a/c2corg_ui/static/js/card.js
+++ b/c2corg_ui/static/js/card.js
@@ -93,6 +93,12 @@ app.CardController = function(gettextCatalog, appUrl, imageUrl) {
   this.doc;
 
   /**
+   * @type {boolean}
+   * @export
+   */
+  this.remainingActivities = false;
+
+  /**
    * @type {string}
    * @public
    */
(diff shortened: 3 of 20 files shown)


c2corg_ui/static/partials/cards/articles.html
```diff
@@ -1,21 +1,35 @@
...
 
-    <div class="article-activities">
-      <span ng-repeat="activity in ::doc.activities" class="route-activity icon-{{::activity}}"
-            uib-tooltip="{{ cardCtrl.translate(activity) }}"></span>
+    <div class="activities flex wrap-row margin-t-sm">
+      <span ng-repeat="activity in ::doc.activities | limitTo: 4"  class="activity icon-{{::activity}}"
+            uib-tooltip="{{ cardCtrl.translate(activity)}}"></span>
+
+      <span class="flex flexend-align show-more-activities"
+            ng-if="::doc.activities.length > 4" ng-hide="cardCtrl.remainingActivities"
```
ginold: i swear it didn't work with `ng-if="::doc.activities.length > 4 && !cardCtrl.remainingActivities"` :(

lbesson: probably a mix between the "once" (::) at the beginning and having the whole expression on one line (didn't check though..)

ginold: indeed it worked! I think i had this error once upon a time. Thx!

asaunier:
> outings when in the feed, the author info is redundant with "xxx has created an outing"

True, but that was already the case before this PR anyway :)

The problem with positioning the outing author at the bottom of the card is that the card takes even more height than before, whereas people were already complaining the cards were to high. See https://github.com/c2corg/v6_ui/issues/1058

Before:

sans titre

After:

sans titre

For me the previous layout was better, at least in the feed.

ginold:
The problem is you never know how many activities there are, how long the user name is, how long the title so it's for me impossible to define a work-all-around-on-all-device-widths styling.

When everything is on its own line, it's ok.

stef74:
for feed 
feed

for outings 
outings

asaunier:
@stef74 card templates are the same in the feed, in advanced search pages or associated documents lists.

@ginold How does it work for images? It seems that cards on the same line all have the same height.

sans titre

What happens if images have different formats (vertical/horizontal, etc.)?

ginold:
> How does it work for images?

As I mentioned above, i haven't done nothing for images because they wouldn't load (and i could not upload any neither)

stef74:
@asaunier hummm the layout of outing in feed or in card (based on screenshoot of this PR) are not the same ...

stef74:
For me show only one activites in outing is correct (if you had a + if more than one).

ginold:
maybe 90% of the outings have 1-2 activities, but i would like to have one with 3-4 and then what? i will not work anymore. I've tested this :P

asaunier:
> hummm the layout of outing in feed or in card (based on screenshoot of this PR) are not the same ...

Yes they are. In the feed there is also a frame around cards, and the width might not be the same, but the content is the same.

> As I mentioned above, i haven't done nothing for images because they wouldn't load (and i could not upload any neither)

@ginold you should use the image backend URL from demov6 in your instance. For both your UI and API, add to your config file:
```
export image_backend_url = https://images.demov6.camptocamp.org
export image_url = https://sos.exo.io/c2corg_demov6_active/
```

ginold:
I have just done that. I will post an update soon.

ginold:
In the meantime maybe you will find a suggestion for the outing cards :P I agree i liked the previous card style more, but just in the feed - it looks quite good in the lists. 
As for the date, do you think it would be a good idea to put the date with the title on the same line, like in the view-page?

stef74:
For me date on same line that title is good.

c2corg_ui/templates/helpers/edit.html
```diff
@@ -6,12 +6,15 @@
   <section ng-show="${model}.associations.${type}.length > 0" class="associated-section">
     <label translate>Associated waypoints</label>
     <article class="associated-documents">
-      <div class="list-item"
+      <div class="list-item waypoints"
+           % if model == 'route':
+              style="min-height: 115px;"
```
asaunier: Can we use a class instead?

ginold:
Images: 

screenshot from 2016-12-07 16-46-26

With the following question (and answer): Yes @asaunier, the cards will be as high as the highest in the row, it's the way it is. Sometimes there is text, sometimes not. As for the images, would we prefer the title to align in the middle (like above), top or bottom ?  we have to choose the lesser bad :P

screenshot from 2016-12-07 16-57-19

screenshot from 2016-12-07 16-57-44

asaunier:
The rendering of the images is nice, thanks.
Top vertically aligned titles looks better to me.
What about vertical images, they are somehow cut as "Mont Dolent" or "David dans ses oeuvres" above, right?

ginold:
Yes, all the images have a fixed height here (so the centering may vary).

ginold:
Thanks for the compliments and the review!

I'm having a hard time to decide too. Maybe it's not unacceptable to have a slightly different styling for the feed?

loicperrin:
I have not read all this in details, but seems to be much better. Good job Thanks !
I am eager to see the result on demo.

> Maybe it's not unacceptable to have a slightly different styling for the feed?

IMHO, it is better to have different styles for feed and for result pages, if it is optimized in both cases.

asaunier:
I would say the cards don't have to be exactly the same in the feed or in the advanced search / association but it should quite consistent so that users may easily recognize outings cards.

In the feed, cards are a bit more horizontal than in the advanced search where they are a bit more square or vertical. The content should be quite the same even though the elements could be placed a bit differently.

ginold:
So how about I change only the outings cards for the feed? there is much more horizontal space than in the listings and therefore I can better use the space given.

asaunier:
What do you suggest? A dedicated outing card for the feed? Some customized CSS for outing cards in the feed?
CSS adaptations sounds acceptable to me if the card layout is still quite consistent. Could you make a proposal?

ginold:
CSS changes for the feed will be enough. Working on that!

asaunier:
OK,
perhaps it could be the opportunity to have all feed card metadata on the same line:  "time" info ("5 minutes ago") to the right of "XYZ has created an outing"?
This would help optimizing the height of feed cards.

ginold:
So here it is, a little different CSS for outing cards in the feed:
screenshot from 2016-12-08 11-36-07


To avoid that: 
screenshot from 2016-12-08 11-31-37

You will notice that on webkit/chrome/safari the long user names are cut if the name is too long, or if it has 2 members (for ex "Alexandre Saunier"),the last name will go on the line beneath. It's not supported on FF -> the name can be cut in the middle of "Sau-" and the rest goes below.

screenshot from 2016-12-08 11-43-42

stef74:
Could you write on same line the metadata of feed (as @asaunier mentionned in his last comment).

ginold:
Oops i missed that one. So metadata on the same line, but as you can see the "xxx has created an outing" is of variable length and the date can not always fit -> should it go on the line below in that case, or should the two **always** be on the same line?

ginold:
Here it goes on the same line if there is enough space, and goes below it not.
screenshot from 2016-12-08 12-13-43

asaunier:
@ginold I have just tested your branch in Chrome and FF, and looked a bit at the code.

It looks fine to me except that I think you should not change the user icon. The "old" user icon is still used in many places, for instance the user association seach tools and the user cards.
We need to have the same icon everywhere. It's probably easier to stick to the "old" icon.

asaunier:
In the standard outing cards, if it is easy to do, I would suggest to show the outing creator name at the bottom of the card. A bit like the author names in book cards.

ginold:
I haven't uploaded the last modifications yet. Could you answer the question above about metadata on the same/or-not line ?

>  I would suggest to show the outing creator name at the bottom of the card

it is already like this - look at the first screenshot at the top of this issue. The layout only changes in the feed.

Ok for the icon !

asaunier:
> Here it goes on the same line if there is enough space, and goes below it not.

That looks great to me, except that when a second line is needed, the line-height is a bit too big (the 2 lines should be closer from each other).

ginold:
updated 

+ old user-icon 
+ action metadatas on the same line when enough space and smaller line-height between the two

asaunier:
@ginold That's cool thanks :)

2 things:
* date is perhaps the most important information in an outing card (with the title) => please move it at the beginning of its line, before elevation etc.
* minor but it seems the author's name is not always stuck at the bottom of the cards:
sans titre

ginold:
Dear firefox and mozzilla.. you created the internet standards and can't catch up with chrome :(

ginold:
moved the date :+1: 

as for the alignment, I can't figure it out... probably some browser differences. Hopefully it's not very very recurrent. What I noticed is that when you delete the quality div from the card, the author-name will be aligned as it should. But I have no idea what it has to do with it, because it has a position absolute, so it shouldn't care (like on chrome). Can we go with it like this?

brunobesson:
IMHO, as it is one of the most requested change, I would go for it. It's a huge improvement, we can make more corrections and adjustments afterwards.

asaunier:
OK for merging as is when Travis is OK.

Issue #1127, opened 2016-12-07
Infinite loop when logging out from forum.


arnaud-morvan:
When we logout from the forum, we are redirected to https://www.camptocamp.org/auth-sso?logout
And browser loop on refreshing the page.

asaunier:
Same problem on https://www.demov6.camptocamp.org/

But it works normally with the http URL: https://www.demov6.camptocamp.org/

When logging out from https://forum.demov6.camptocamp.org/
one is redirected to http://www.demov6.camptocamp.org/auth-sso?logout (http, not https)
Shouldn't it redirect to https://www.demov6.camptocamp.org/auth-sso?logout (https!)?

tsauerwein:
> Shouldn't it redirect to https://www.demov6.camptocamp.org/auth-sso?logout (https!)?

Isn't this configured in the Discourse settings?

asaunier:
Perhaps this is not related to the infinite loop. As reported by @arnaud-morvan in prod it loops on https://www.camptocamp.org/auth-sso?logout

Perhaps in prod we have the https page because http redirections to https are forced (HSTS?)?

tsauerwein:
> As reported by @arnaud-morvan in prod it loops on https://www.camptocamp.org/auth-sso?logout

I also have the infinite loop on localhost. PR is coming.

tsauerwein:
> Perhaps in prod we have the https page because http redirections to https are forced (HSTS?)?

In prod it works correctly, the redirect is made to https://www.camptocamp.org/auth-sso?logout

arnaud-morvan:
I've changed the settings in discourse to use HTTPS for login/logout

tsauerwein:
Thanks!

fbunoz:
> I've changed the settings in discourse to use HTTPS for login/logout

I had always the problem on the demo after each demo update.
I changed the url to add the https, and it worked correctly.
But at the next update, it was returned to http, and I should make the change again.
I don't know where was the default config defined, but it was wrong.

PR #1129, opened 2016-12-07
Do not redirect after logout from Discourse

tsauerwein:
The problem is that the referrer is not set. Instead simply show the login page.

Closes https://github.com/c2corg/v6_ui/issues/1127


asaunier:
> Instead simply show the login page.

Redirecting to the forum or the UI homepage would also make sense.
But OK for the login page (this way one can authenticate with another user).

PR #1152, opened 2016-12-08
Add a route for robots.txt

asaunier:
@mfournier has suggested that the ``robots.txt`` is fully managed by the UI and that we don't rely on some config in haproxy to alias ``/robots.txt`` and ``/static/robots.txt`` as we previously did it with the apache conf (that is embedded in the UI files).

I have then followed the recommendations of the Pyramid documentation here: http://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/static_assets/files.html#serving-a-single-file-from-the-root


asaunier:
> Can we get rid of the apache config?

You mean the full ``apache/`` directory?
I guess @mfournier would recommend to do so and work in a docker based environment :P 
But in the mean time we still need it to make the old dev demo work.

PR #1167, opened 2016-12-09
Don't include Bootstrap CSS twice

tsauerwein:
The CSS for Bootstrap is already included in our LESS files:
https://github.com/c2corg/v6_ui/blob/master/less/c2corg_ui.less#L2

This should resolve the problem with 404 requests to /static/5ff028d/fonts/glyphicons-halflings-regular.ttf 

But please test properly if the layout is not broken somewhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,6 @@ LIBS_JS_FILES += \
 
 # CSS files of dependencies that are concatenated into a single file
 LIBS_CSS_FILES += \
-    node_modules/bootstrap/dist/css/bootstrap.min.css \
     node_modules/bootstrap-slider/dist/css/bootstrap-slider.css \
     node_modules/slick-carousel/slick/slick.css \
     node_modules/photoswipe/dist/photoswipe.css \


asaunier:
I have tested the branch and have seen no layout problem except that now the "follow" / "unfollow" buttons in the list of followed users are no longer shown inside the user cards. I will have a look if this can be fixed easily.

sans titre

ginold:
it's rather due to my pr on cards rather bootstrap. Or when you where testing my branch it worked fine?

asaunier:
@ginold You're right, it's not related to this branch, I have the same problem when using the master branch.
I am not really sure how this can be fixed.
The association removal icons are placed correctly in cards, but they use an absolute positioning. If I do the same for those "follow" buttons instead of ``float:right``, they overlap the content of the card (user's name). Would you have an idea? Or perhaps you could fix it? 

Then as far as I can see, the present change has no visible impact on the layout. OK for merging.

tsauerwein:
Thanks for testing!

Issue #1162, opened 2016-12-08
Map position is lost when changing doc type


loicperrin:
On routes result page, I center the map on the area where I want to find routes.
If I want to see WP/Access, I lost the previous map position when changing "route" to "WP".

asaunier:
A new page is loaded when changing the document type, all filters are indeed reset.
It could be possible tough when changing the page to get the bbox parameter if any and to add it to the new page URL before loading it.

loicperrin:
OK

fbunoz:
On Desktop and mobile (FF) : when changing the doc type, the map is zoomed out (one zoom step), whereas the bbox is correct. Then, the results are in a virtual rectangle smaller than the visible map : 

![map-extend](https://cloud.githubusercontent.com/assets/3902555/22621826/d0ff42e2-eb2c-11e6-940c-319efa1456fa.png)

PR #1188, opened 2016-12-12
Better doctype selector in advanced search pages

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1162

TODO

* [x] propagate the current bbox or selected areas if relevant
* [x] propagate selected activities if relevant

### DIFF
--- a/c2corg_ui/static/js/doctypeselector.js
+++ b/c2corg_ui/static/js/doctypeselector.js
@@ -1,0 +1,144 @@
+goog.provide('app.DoctypeSelectorController');
+goog.provide('app.doctypeSelectorDirective');
+
+goog.require('app');
+
+
+/**
+ * Directive managing the user mailinglists.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.doctypeSelectorDirective = function() {
+  return {
+    restrict: 'E',
+    bindToController: {
+      'defaultType': '@appDoctype'
+    },
+    controller: 'appDoctypeSelectorController',
+    controllerAs: 'doctypeCtrl',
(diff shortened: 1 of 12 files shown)


asaunier:
This PR has just been updated: it adds a new ``app-doctype-selector`` directive that replaces the mako helper ``add_doctype_selector``.

The apparent behaviour is the same except that parameters ``bbox``, ``a`` (areas) or ``act`` (activities) are propagated to the target advanced search page if available and relevant.

Please review :)

brunobesson:
I tried to checkout your branch, but when switching from outings to waypoints in the combobox, I get a 

    Uncaught ReferenceError: doctypeCtrl is not defined at HTMLSelectElement.onchange

related to `doctypeselector.html`

asaunier:
The branch you use is outdated. This is the error that came with the original version of the branch, a month ago.

To get the very last version of the branch you may type:
```
git fetch
git checkout doctype-selector-directive
git reset --hard origin/doctype-selector-directive
```

brunobesson:
OK... I don't understand why because I did a git fetch. Maybe I only forgot to pull updates to the branch? Whatever. Seems to be working now. I might not be the best fitted for a code review, but my tests show it is working as adverted.

asaunier:
Probably because I did a "push force" (push -f) since my branch was rebased. I think a "git fetch" is not enough to get the final branch since the branch had diverged.
Perhaps in that case a "git fetch" or "git pull" reports the divergence?

asaunier:
@brunobesson Thanks for the review and testing! I'll merge the PR this afternoon if no one else objects in the meantime or ask for more time.

PR #1194, opened 2016-12-12
Debug left menu for forum

fbunoz:
- Keep menu visible on mobile : all z-index >= 1000
- Isolate footer rules for #sidemenu only, to avoid conflict with discourse footer.

### DIFF
--- a/less/sidemenu.less
+++ b/less/sidemenu.less
@@ -10,7 +10,7 @@
   position: fixed;
   top: 0;
   white-space: nowrap;
-  z-index: 1002;
+  z-index: 1100;
   height: 100%;
   font-size: 1em;
   transition: 0.3s;
@@ -190,7 +190,7 @@
   display: none;
   margin: auto;
   position: relative;
-  z-index: 2;
+  z-index: 1002;
   width: 40px;
   height: 40px;
   background-color: white;
@@ -262,7 +262,7 @@
   top: 5px;
   position: fixed;
   left: 100px;
-  z-index: 31;
+  z-index: 1031;
   font-size: 17px;
   padding-bottom: 2px;
   padding-top: 7px;
@@ -282,7 +282,7 @@
   }
 }
 
-footer {
+#sidemenu footer {
   font-size: 0.8em;
   position: absolute;
   width: 100%;
@@ -367,7 +367,7 @@ footer {
   display: none;
   margin-left: 10px;
   margin-bottom: 6px;
-  z-index: 34;
+  z-index: 1034;
   width: 80px;
 
   @media @tablet {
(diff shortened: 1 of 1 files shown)


asaunier:
I have tested the branch in my UI instance (no available Discourse to test though).
It seems OK for the menu itself.

But now the menu is display above all modal window, for instance the image upload panel, the association removal confirmation or the preview modal:

capture d ecran 2016-12-12 a 22 55 58

capture d ecran 2016-12-12 a 22 59 05

capture d ecran 2016-12-12 a 22 59 29

fbunoz:
I tested on topoguide, but without modal box...
It is not correct, the z-index changes must be done only for Discourse.
It is possible to do that with less ?

brunobesson:
Less is not a miracle tool it's only a CSS preprocessor, so I don't understand your question. What do you want to achieve?

BTW, is 1000 an arbitrary value, or related to z-indices in discourse? Maybe the simplest solution would be to use appropriate values.

fbunoz:
> is 1000 an arbitrary value, or related to z-indices in discourse?

On forum, the hamburger menu is not shown on mobile because a z-index is 2 in topoguide, and which must be higher than 1000 to have correct behavior.
But as other z-index are 31, 34, ..., to keep the behavior we must change it to 1031, 1034,
But only for discourse CSS, not for topoguide.

brunobesson:
>which must be higher than 1000
What component has a z-index of 1000?

asaunier:
Perhaps you should use smaller ``z-index`` values to make sure the menu is shown below the modal elements.
Or perhaps the ``z-index`` of modal elements should be increased?

fbunoz:
Sorry for the noise.
The better solution is not to change the z-index in left menu CSS, but to add rules in forum CSS to overload the z-index values by the values adapted for the forum.
I'll commit a change to remove the z-index changes.

asaunier:
> I'll commit a change to remove the z-index changes.

I have tested the branch: with your additional change the z-index conflicts are solved, I see no special problem.

2 remarks:
* it would be better to squash the commits (one commit is enough for a one-line change :) )
* it would make more sense to move the ``.footer`` inside the ``#sidemenu`` block.

@ginold what do you think?
Is there a reason for not having the ``.footer`` selector inside ``#sidemenu``?

asaunier:
@fbunoz I have noticed that when you enlarge images posted in the forum, eg. here https://forum.camptocamp.org/t/v6-vos-questions-les-reponses/173468/133
the image modal is shown below the left menu:
capture d ecran 2016-12-14 a 11 26 16

ginold:
I do remember that indeed an element in Discourse has a z-index of 1000 = `.d-header`! so when i wanted the hamburger button to be on the top of it, it had to have a +1000 value.

asaunier:
@fbunoz Is this PR still relevant? It seems you have made other changes in the forum CSS in the meantime?!

Issue #1170, opened 2016-12-09
Broken buttons layout in followed users list


asaunier:
Not in prod yet, but can be seen in demov6.

The "follow" / "unfollow" buttons in the list of followed users are no longer shown inside the user cards. I will have a look if this can be fixed easily.

![sans titre](https://cloud.githubusercontent.com/assets/1192331/21048073/9a8aa798-be0d-11e6-8472-551a7a99de2d.png)

Buttons should be shown inside the cards to take less space and because they are overlapped by the cards of the following line.

PR #1171, opened 2016-12-09
[WIP] Move un/follow buttons into user cards

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1170

Moving the buttons inside the card element helps a bit but is not enough: moving the button at the bottom of the card is still needed.

sans titre

I don't have time to investigate more now.

Issue #1146, opened 2016-12-08
On user's profile page, add link to add him to ones friends


brunobesson:
Currently, there are many missing _shortcuts_ in V6 to ease or simplify actions.
One of them could be adding a button to add/remove a user to your friends list.

tsauerwein:
Apparently the star button is not clear enough. What could we do to make it more obvious?

![image](https://cloud.githubusercontent.com/assets/266474/21009056/ec469ade-bd44-11e6-9a62-4b406d7933f1.png)

brunobesson:
You're right, I missed it!
Don't have an answer yet, it's too new to make any suggestion.

tsauerwein:
Maybe display it on the same line as the user name?

asaunier:
In Twitter, the follow button contains a text, not an icon. It's probably clearer.
But it means the button should be moved somewhere else.

PR #1199, opened 2016-12-13
fix user follow-btn in cards

ginold:
this PR can be deleted #1171
fixes https://github.com/c2corg/v6_ui/issues/1170
screenshot from 2016-12-13 10-28-50

I put small buttons without text because, as usual, it can create conflicts with long names and translations of always variable length in different languages -> big cards.

### DIFF
--- a/c2corg_ui/templates/following.html
+++ b/c2corg_ui/templates/following.html
@@ -14,12 +14,10 @@ from c2corg_common.attributes import mailinglists
     <div class="flex wrap-row areas" ng-show="flCtrl.following.length > 0">
       <div class="list-item users" ng-repeat="user in flCtrl.following">
         <app-card app-card-doc="user"></app-card>
-        <button class="btn orange-btn" ng-click="flCtrl.toggle(user.document_id)"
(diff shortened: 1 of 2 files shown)


asaunier:
Nice, but actually wanted to use plain text button, because icons are not that clear.
See https://github.com/c2corg/v6_ui/issues/1146 that is a good example that icons are not always enough.

My goal with those user cards and the "follow"/"unfollow" buttons was to do something like on Twitter, which much more obvious:

sans titre

ginold:
there is no problem with that,  but it would look like this:
screenshot from 2016-12-13 11-08-08
or like this
screenshot from 2016-12-13 11-07-22

You never know the length of the name, nor of the translated string. If one or the other is ok, please select one.

stef74:
Do you think is ready to add in hotfix milestone ?

asaunier:
@ginold Let's go for your initial proposal (with the icons). I have tested, it works fine and it's quite intuitive anyway.

ginold:
:+1: ! so ok for merging?

asaunier:
Yes, please merge when Travis is OK.
Thanks :)

PR #1200, opened 2016-12-13
Prevent navigating to last page in avanced search

tsauerwein:
Related to https://github.com/c2corg/v6_api/issues/578:

> When making a request to an advanced-search view with an offset > 10000 (for [example](http://localhost:6544/routes?bbox=457507%252C5594424%252C559627%252C5730176&offset=30000)), ElasticSearch complains.

This PR proposes to remove the "go to last page" button (and for consistency reasons also the "go to first page" button). This is similar to the behavior with Google: When you have millions of results, you also can't jump to the last page of these results.

### DIFF
--- a/c2corg_ui/static/js/search/pagination.js
+++ b/c2corg_ui/static/js/search/pagination.js
@@ -54,6 +54,12 @@ app.PaginationController = function($scope, ngeoLocation) {
   this.total = 0;
 
   /**
+   * @type {boolean}
+   * @export
+   */
+  this.showGoToLastPage = true;
+
+  /**
    * @type {number}
    * @export
    */
@@ -81,6 +87,11 @@ app.PaginationController.prototype.handleSearchChange_ = function(event,
     features, total, recenter) {
   this.total = total;
   this.offset = this.location_.getFragmentParamAsInt('offset') || 0;
+  // don't show the "Go to last page" button when doing a full-text-search
+  // (the full result set has to be iterated in ElasticSearch when doing
+  // requests with large offsets)
+  this.showGoToLastPage = !(this.location_.hasFragmentParam('q') &&
+    this.location_.getFragmentParam('q') !== '');
 };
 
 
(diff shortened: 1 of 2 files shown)


asaunier:
OK for removing the "got to last" button, since it causes ES offset errors.

But why removing the "go to first" button?
If you have clicked several times on the "next page" button and finally want to go back to the initial list of results, you'll have to click again several times on the "previous page" button, whereas you could simply click on the "first page". And this button does not cause ES error, right? :P 
I would keep the "first page" button.

tsauerwein:
I don't have a strong opinion on this, but I mainly removed it for consistency reasons. It looks in-harmonic to have a "go to first page" but no "go to last page" button. And I tend to say that these pagination buttons should not be that important. Usually you would try to refine your search instead of flipping through tens of pages.

asaunier:
I feel like it's such a pity to remove a feature that works. And the "first page" button works all the time. And is useful as I have explained it.

Actually what about only hidding the "last page" button when there are too many results (>10000?)?

This "last page" button is useful for instance when a user wants to have a look at their own outings list. If you don't have this "last page" button you have no way to go back to your first outings.

tsauerwein:
Ok, I restored the "go to first page" button.

> Actually what about only hidding the "last page" button when there are too many results (>10000?)?

10000 is a hard limit by ElasticSearch to avoid bad performances. But the performance will also be bad if you "only" have 9000 results.

> This "last page" button is useful for instance when a user wants to have a look at their own outings list. If you don't have this "last page" button you have no way to go back to your first outings.

I would propose to revisit this once the "outings for a user" page exists.

asaunier:
> 10000 is a hard limit by ElasticSearch to avoid bad performances. But the performance will also be bad if you "only" have 9000 results.

Ok. Then 5000 :)

> I would propose to revisit this once the "outings for a user" page exists.

It does. For instance mine: https://www.camptocamp.org/outings#u=377
Accessible from the profile: https://www.camptocamp.org/profiles/377/fr > button "Show this user's outings"

tsauerwein:
In fact this is only an issue when doing FTS queries. I modified the PR so that the button is only hidden when searching with a query string.

So, you can still see [your first outing](https://www.camptocamp.org/outings/58920/fr/mont-chauffe-traversee-des-aretes)! ;)

Issue #1186, opened 2016-12-12
Cards improvements


asaunier:
* [x] Cards should be a bit wider but less high => use 2 columns instead of 3 for standard screens. 3 columns (or more) only for very big screens. Perhaps also increase a bit the threshold at which only one column is used.
* [x] As much as possible, we should avoid having too many lines
* [x] Reduce the font size to about 75%

stef74:
For infos in search v5 we can see around 12 document in one screen (1366x786) see : http://www.pre-prod.dev.camptocamp.org/outings to compare. In v6 we can also see 12 but some infos is missing.

And 24 in home ... but it's not the topic here.

asaunier:
> For infos in search v5 we can see around 12 document in one screen (1366x786) see : 

Here is what I can see in my browser for the v5 preprod site (I have removed your URL because I am not sure it should be made public):

<img width="400" alt="capture d ecran 2016-12-12 a 17 24 27" src="https://cloud.githubusercontent.com/assets/1192331/21106882/0df83a7c-c090-11e6-9257-e878b6fb5134.png">

Issue #1187, opened 2016-12-12
Add possibility to enlarge or hide map in the advanced search pages


asaunier:
Perhaps a duplicate of https://github.com/c2corg/v6_ui/issues/428 but I prefer to create a new issue to make it clearer.

* [x] No change for mobile screens: the map is hidden by default and shown full page on request.
* [x] Keep the default behaviour as is: half page map by default for desktop screen.
* [x] Add button to enlarge the map to the full page (perhaps keep the menu and header visible)
* [x] Add a button to hide the map

mgerbaux:
Same evolution should be planned for georeferenced document pages (with a vertical enlargement of the map)

asaunier:
> Same evolution should be planned for georeferenced document pages (with a vertical enlargement of the map)

Perhaps you're right but that's not what should be done in this issue. If needed it should be discussed in a dedicated issue. I have then made obvious this is not a task required for this issue, thanks :)

fbunoz:
> it should be discussed in a dedicated issue.

Done : https://github.com/c2corg/v6_ui/issues/1191

ginold:
Continuing the discussion from #1191 

**Still a question: where to put the toggle list/map buttons?**


Normal

![screenshot from 2016-12-13 15-42-32](https://cloud.githubusercontent.com/assets/7814311/21144409/ca5070d0-c14a-11e6-8711-c498b7b0889f.png)

enlarge map (if anyone knows the function to update the map on resize it would be cool)

![screenshot from 2016-12-13 15-42-03](https://cloud.githubusercontent.com/assets/7814311/21144419/d4678658-c14a-11e6-8240-07dfe9efb193.png)

enlarge list

![screenshot from 2016-12-13 15-42-17](https://cloud.githubusercontent.com/assets/7814311/21144432/d9755486-c14a-11e6-9906-c17a8784bdc3.png)

tsauerwein:
Not an answer to your question: But the map looks like you want to call [ol.Map#updateSize](http://openlayers.org/en/master/apidoc/ol.Map.html#updateSize) after having changed the size of the map.

asaunier:
> Still a question: where to put the toggle list/map buttons?

Several possibilities:
* use the same approach than in a mobile screen => buttons in the bottom right corner of the map.
 * pros: consistent with the mobile
 * cons: not that visible and perhaps there are too many items on the map (attribution icon for instance)
[...]

PR #1201, opened 2016-12-13
reduce card font-size + larger cards treshold

ginold:
related to #1186
screenshot from 2016-12-13 12-53-33

+ 3 cards per row for >2000px screens width
+ 2 cards per row default
+ 1 card per row for < 800px screens width
+ 90% card font size (super small imho)

Should this change be applied only to the advanced search lists? or also to associated documents in views/editing pages?

### DIFF
--- a/c2corg_ui/static/partials/cards/outings.html
+++ b/c2corg_ui/static/partials/cards/outings.html
@@ -18,14 +18,16 @@
       </span>
     </div>
 
-    <div class="activities">
-      <span ng-repeat="activity in ::doc.activities" class="activity icon-{{::activity}}"
-            uib-tooltip="{{ ::cardCtrl.translate(activity)}}"></span>
-    </div>
+    <div class="activity-author">
+      <div class="activities">
+        <span ng-repeat="activity in ::doc.activities" class="activity icon-{{::activity}}"
+              uib-tooltip="{{ ::cardCtrl.translate(activity)}}"></span>
+      </div>
 
-    <div class="outing-author flex baseline-align" ng-if="!feed">
-      <span class="icon-user"></span>
-      <span class="author-name">{{::doc['author']['name']}}</span>
+      <div class="outing-author flex baseline-align" ng-if="!feed">
+        <span class="icon-user"></span>
+        <span class="author-name">{{::doc['author']['name']}}</span>
+      </div>
     </div>
 
     <div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
--- a/less/lists.less
+++ b/less/lists.less
@@ -172,7 +172,7 @@
   .set-main-waypoint {
     margin: 0 0 10px 10px;
   }
-  .outing-author, .article-activities {
+  .article-activities {
     .calc(width, "100% - 40px");
   }
   // .list-item.waypoints
@@ -267,29 +267,6 @@
       .calc(top, "100% - 60px");
     }
 
-    @media @phone {
-
-      .list-item-info {
-        padding-left: 25%;
-      }
-      .outing-author {
-        top: 35%;
-        left: 10px;
-        width: 20%;
-        text-align: center;
-        display: initial;
-        position: absolute;
-
-        .icon-user {
-          float: none;
-          width: 100%;
-        }
-      }
-      .activities {
-        margin-bottom: 0;
-        margin-top: 3px;
-      }
-    }
   } // end outing cards
 
   // .list-item.highlight
--- a/less/lists.less
+++ b/less/lists.less
@@ -8,15 +8,7 @@
   @media (max-width: @screen-sm-min) {
     padding-bottom: 50px;
   }
-  .list-item {
-    flex-basis: 32%;
-    @media (max-width: 1200px){
-      flex-basis: 48%;
-    }
-    @media (max-width: 730px) {
-      flex-basis: 100%;
-    }
-  }
+
   & + .no-results {
     display: none;
   }
@@ -55,7 +47,6 @@
   }
   @media @phone {
     border-width: 1px;
-    font-size: 1.2em;
   }
 
   a {
@@ -68,7 +59,7 @@
     padding: 10px;
 
     @media @phone {
-      font-size: .8em;
+      font-size: 1.1em;
       padding: 7px;
     }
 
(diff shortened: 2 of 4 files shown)


stef74:
to be more consistent apply also in association

Could you provide a screenshoot on outings please

stef74:
For me the goal is to have 2 lines per card ... if you have same number of line as previsous no interesset

stef74:
push icon activity at the right

ginold:
Ok, changes work everywhere:

View page:

screenshot from 2016-12-13 13-17-13


Editing page: 

screenshot from 2016-12-13 13-17-31


Outings advanced search:

screenshot from 2016-12-13 13-17-53

Icons on the right would look like this (weird imho):

screenshot from 2016-12-13 13-19-19

Is this ok for you @stef74 ? @asaunier  ?

stef74:
For the last screenshot could you try to one column ... and move icon at same level of texte. Keep in mind they are missing some infos.

ginold:
having the icon next to the text is a very bad idea because there might be 2-4 activities which take the whole card width, and the title can be really long too. 
It would maybe work for most of the route cards with short title and 1 activity, but not for the others.

stef74:
something like this
test_card

So probaly enlarge again the card. Currently thet missing some element so some space are need in card.

ginold:
as i said, __maybe__ it would work with short titles and 1 activity, but with more than that you have to expect glitches, getting the card larger is not the best solution.

screenshot from 2016-12-13 13-49-23
screenshot from 2016-12-13 13-49-49

stef74:
You can consider the right of card is column with icon as you do for outing in the feed. Like this 
sans titre

So we need to work on 1 column

stef74:
Cou can reduce black badge ... many complain that is too big.

stef74:
2871c50c-c13b-11e6-9b57-51ef4295f60e

ginold:
That sounds pretty difficult and challenges completely the actual layout. 

In any case, if we want to pursue this change, I would prefer to do it in a separate pull request because it has nothing to do with the issue it tries to solve.

desnoes:
For the outings, it would be good to put on the same line: date, author and region (area). I think there is enough space. This would same one line.

stef74:
Ok do that, you proposition is not bad ... but insufficient. Could you move cativty icon on same line than user for outing ?

stef74:
if is simple align author on right if not push as this need to test different case and resolution

ginold:
It looks ok on different resolutions:

**large screen**

screenshot from 2016-12-13 14-29-42


**actual mobile**

screenshot from 2016-12-13 14-30-47


but i think we should also change to 

screenshot from 2016-12-13 14-33-49

stef74:
:+1:  could you push on demo when it's merged ... so other user can test.

asaunier:
This PR must be reviewed before merging :)
@ginold is it ready for review?
If yes I will have a look.

ginold:
Yes it's ready. It brings the changes  I listed above + outing card style like on the 1st and last screenshot of my previous comment.

asaunier:
The code is OK for me.

I have tested the branch in my instance/browser:
* for doctypes with no geom (ie. no map in the advanced search), the cards are too wide IMO (ditto for areas):
sans titre
* in document detail view pages, since there is no map, the routes and WP cards are wider too (perhaps too wide?). For instance for WP "La Meije":
sans titre

* when there is a map, the card width is good (see screenshots above) but perhaps the threshold to the single column layout is a bit too small:
sans titre
* image cards are also wider and we only have 2 columns as for the other doctype. OK for me (what about portrait images?) but one has to be aware of this change too. See the prod for the current size.
sans titre

ginold:
so maybe:
+ waypoints cards could have +1 more card per row ?
+ lists with no maps could have +1 more card per row?
+ will make the treshold for 1 col bigger (is 800px now) ?

asaunier:
Makes sense to me, yes.

asaunier:
> * waypoints cards could have +1 more card per row ?
> * lists with no maps could have +1 more card per row?

Just to make sure:
* if I have a 10000px wide screen, how many columns will I see? 2 (3 for no-map content) ? Or as many columns as it fits in the available width?
* related to https://github.com/c2corg/v6_ui/issues/1187 => when hidding the map, will it increase the number of columns or only the width of cards?

stef74:
Ok with your last recommendation.
Is better way is now to move in demo ... to have more deep test with many resolution (assoication can do that).

ginold:
> Or as many columns as it fits in the available width?

I think there is no such "as many as it fits" -> we have to define the max.
I hardly see people having a 10'000px wide screen but i can add a complementary case to have 4 columns :P

As for, #1187  i would prefer to handle it in <-- that pr directly if needed, not here.

asaunier:
What I mean is that the system must be flexible: having much wider cards because the screen is larger or the map is hidden does not bring much gain. The idea is to be able to see more results (ie. more rows, more columns) when there are more space available.

stef74:
a maths formula can help to dertermine how many colum is needed ... nb_colum = with_screen modulo ( optimal_width_of_card)
The optimal_width depend of doc type

ginold:
the width is always a % - > 33% = 3 cards per row etc. I will define for up to 4 cards per row. I think it's impossible to have such logic in CSS as you suggest, maybe could be in JS.

asaunier:
@ginold let me recap:
* for outings and routes we need wider cards: generally displayed in 2 cols when the map is shown (1 if the screen is a bit too narrow)
* if there is enough space (wider screen, mo map or hidden map): 3 cols, maybe 4.
* for other document types: since the cards have less content (?), 3 cols by default is OK, but of course we should be also able to see more cols (4 or 5?) when the screen is wide or when the map is hidden.
* for images? No clue if it's better to have wider cards or to keep the current width. What do you think?

Since it is important that at least a first version go to prod soon, I would recommend to use a simple CSS approach first and to use a more flexible JS approach later if needed.

Sounds good to you?

ginold:
that's what i've done :+1: 

There are up to 4 cards per row on super large screens (+2100px) for all doctypes and +1 more card per row than the default - for lists that have no map (articles, areas...) and for waypoints. 
Should also work for associated documents on editing and view pages.

asaunier:
@ginold I have tested with FF and Chrome on my laptop. As far as I can see, it looks good to me, thanks!

Please merge.

PR #1208, opened 2016-12-13
Toggle map and toggle list in advanced search pages

ginold:
related to #1187
a proposal to solve the above issue. 
I'm not convicted by the placement of the buttons, but I can hardly tell where is the best place to put them, nor if it's very "intuitive". But the way it works is decent. 

Works for outings only (for now).

### DIFF
--- a/c2corg_ui/static/js/maincontroller.js
+++ b/c2corg_ui/static/js/maincontroller.js
@@ -17,7 +17,13 @@ goog.require('app.coordinate');
  * @export
  * @ngInject
  */
-app.MainController = function(gettextCatalog, appApi, appAuthentication, authUrl) {
+app.MainController = function($scope, gettextCatalog, appApi, appAuthentication, authUrl) {
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
 
   /**
    * @type {angularGettext.Catalog}
@@ -30,7 +36,6 @@ app.MainController = function(gettextCatalog, appApi, appAuthentication, authUrl
    * @export
    */
   this.appApi = appApi;
-
 
   /**
    * @type {app.Authentication}
@@ -91,4 +96,12 @@ app.MainController.prototype.isPath = function(path) {
  */
 app.MainController.prototype.animateHeaderIcon = function(e) {
   app.utils.animateHeaderIcon(e);
+};
+
+
+/**
+ * @export
+ */
+app.MainController.prototype.resizeMap = function() {
+  this.scope_.$root.$emit('resizeMap');
 };
--- a/c2corg_ui/static/js/map/map.js
+++ b/c2corg_ui/static/js/map/map.js
@@ -250,6 +250,8 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
 
   // advanced search mode
   if (this.advancedSearch) {
+    this.scope_.$root.$on('resizeMap', ngeoDebounce(this.resizeMap_.bind(this), 300, true));
+
     if (this.location_.hasFragmentParam('bbox')) {
       this.enableMapFilter = true;
       var bbox = this.location_.getFragmentParam('bbox');
@@ -887,6 +889,15 @@ app.MapController.prototype.resetFeature = function() {
     features.push(feature);
   }
   this.scope_.$root.$emit('mapFeaturesChange', features, true /* isReset */);
+};
+
+
+/**
+ * @private
+ */
+app.MapController.prototype.resizeMap_ = function() {
+  this.map.renderSync();
+  this.map.updateSize();
 };
 
 
--- a/c2corg_ui/static/js/main.js
+++ b/c2corg_ui/static/js/main.js
@@ -15,6 +15,7 @@ goog.require('app.XreportEditingController');
 goog.require('app.OutingFiltersController');
 goog.require('app.RouteEditingController');
 goog.require('app.XreportFiltersController');
+goog.require('app.MapController');
 goog.require('app.addAssociationDirective');
 goog.require('app.advancedSearchDirective');
 goog.require('app.alertsDirective');
--- a/c2corg_ui/templates/helpers/list.html
+++ b/c2corg_ui/templates/helpers/list.html
@@ -8,3 +8,30 @@
     <option translate value="${request.route_path('areas_index')}" ng-selected="{{mainCtrl.isPath('areas')}}">Areas</option>
   </select>
 </%def>
+
+<%def name="toggle_map_btns()">\
+  <button class="btn btn-primary no-map-btn show-list" ng-show="noList"
+          ng-click="noList = false; noMap = false; mainCtrl.resizeMap();">
+    <span  class="glyphicon glyphicon-th-large"></span>
+    <span translate>Show list</span>
+  </button>
+  <button class="btn btn-primary no-map-btn" ng-show="!noList"
+          ng-click="noMap = !noMap; noList = false;">
+    <span  class="glyphicon glyphicon-eye-close"></span>
+    <span translate>Hide map</span>
+  </button>
+</%def>
+
+
+<%def name="toggle_list_btns()">\
+<button class="btn orange-btn btn-default no-list-btn"
+        ng-show="!noMap" ng-click="noList = !noList; noMap = false; mainCtrl.resizeMap();">
+  <span class="glyphicon glyphicon-eye-close"></span>
+  <span translate>Hide list</span>
+</button>
+<button class="btn orange-btn btn-default no-list-btn"
+        ng-show="noMap" ng-click="noList = false; noMap = false; mainCtrl.resizeMap();">
+  <span class="glyphicon glyphicon-globe"></span>
+  <span translate>Show map</span>
+</button>
+</%def>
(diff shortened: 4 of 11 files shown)


asaunier:
That's pretty nice indeed :)

I must confess I was a bit confused first because I expected buttons to hide the map and one to make it full page but your system rather has a button to hide the map (shown on the map) and one to hide the results list (consequently showing the map full page - the button is placed in the list, which is logical). 

See below the initial state of the map (the buttons are the orange and blue ones with a slashed-eye icon):

sans titre

When the "hide map" button on the map is clicked, here is how it looks: in addition to the "hide list" button of the first step, there is a "show map" button:
* click on the "show map" button => restore the map normal size
* click on the "hide list" button => show the map full page
sans titre

Here is what happens when the map is full page (the map itself is distorted - additional work is needed to fix this) => the "hide map" button is still shown on the map + there is a new "show list" button that restore the initial layout (both list AND map).

sans titre

asaunier:
> the map itself is distorted - additional work is needed to fix this

@ginold To fix this we have to use the trick  @tsauerwein has suggested:

> call [ol.Map#updateSize](http://openlayers.org/en/master/apidoc/ol.Map.html#updateSize) after having changed the size of the map

I think it means clicking on the buttons should trigger angular events that will be catch by some handler to add to the ``map`` directive, eventually running ``updateSize()``.
Perhaps some event is already triggered when changing the map container size?

I am a bit concerned because changing the map size might trigger a map extent change:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/map.js#L273-L275
Consequently a new API request with a new set of results, as if you had panned or zoomed the map.

ginold:
I think you were right @asaunier and i found that two buttons to do things were too much and unnecessarily complicated - now 1 button at a time, 1 on map and 1 on list, changes action depending on the state.

It might seem confusing at first, but very logical after a while.

There are now 3 states:

### when you load the page, default
screenshot from 2016-12-16 16-24-57

### when you click on "hide list" on the list
screenshot from 2016-12-16 16-25-06


When you reclick on "show list", it will go back to the initial state.

### when you click on "hide map" on the list
screenshot from 2016-12-16 16-25-19


@tsauerwein I managed to update the map thanks to `updateSize()` however this looks a little staggering - the map only updates after it has finished resizing and is not smooth (just a visual issue, not a big deal). 

I trigger the function after the animation has been finished (setTimeout with 300ms).

I added a function to the `MapController` which listens to an emitted event "updateMapResize". The only issue I have is calling `mapCtrl.updateMapResize()` directly in the buttons template (helpers/list.html), I admit that I not always understand when a controller can be called from anywhere, like the `MainController`.

Thats why i'm emitting this event within the `mainController` which is caught further in the `mapController`, which is probably not the way to do it. Or is it?

brunobesson:
I like the proposal! :+1: 

I'm just wondering if we could remove the text for the map buttons? Maybe using a tooltip, or with a button that resize on hover?
I fear we'll add some more buttons later (slope and other layers added on top) and I would like it to use a little space as possible.

ginold:
> Maybe using a tooltip, or with a button that resize on hover?

Yes we can, not a problem. But at least it is clear with a text, in addition when it is the same button, in the same place but that only changes the function and icon. I think it will not be noticeable if there were only a tooltip.

stef74:
@ginold what is the remaining job ?

asaunier:
> /home/travis/build/c2corg/v6_ui/c2corg_ui/static/js/map/map.js
  903:2  error  More than 2 blank lines not allowed  no-multiple-empty-lines

Come on @ginold Please always run ``make check`` before pushing a PR.

c2corg_ui/static/js/maincontroller.js
```diff
@@ -92,3 +97,9 @@ app.MainController.prototype.isPath = function(path) {
 app.MainController.prototype.animateHeaderIcon = function(e) {
   app.utils.animateHeaderIcon(e);
 };
+/**
+ * @export
+ */
+app.MainController.prototype.updateMapResize = function() {
+  this.scope_ .$root.$emit('updateMapResize');
```
asaunier: Extra blank space inside after ``this.scope_``.

c2corg_ui/static/js/maincontroller.js
```diff
@@ -92,3 +97,9 @@ app.MainController.prototype.isPath = function(path) {
 app.MainController.prototype.animateHeaderIcon = function(e) {
   app.utils.animateHeaderIcon(e);
 };
```
asaunier: Missing blank line

brunobesson:
@asaunier which text editor do you use? E.g. for atom there are plugins to apply eslint on the fly when editing. I assume it is the same in webstorm (never used sadly) and others.

c2corg_ui/static/js/map/map.js
```diff
@@ -237,6 +237,8 @@ app.MapController = function($scope, mapFeatureCollection, ngeoLocation,
    */
   this.view_ = this.map.getView();
 
+  this.scope_.$root.$on('updateMapResize', this.updateMapResize_.bind(this));
```
asaunier: Please move this line inside the ``if (this.advancedSearch) {`` block, a few lines below (line 250) => we don't need to add this listener whenever a map is created. Only for advanced search pages (perhaps for some other maps later but that still hypothetical => we will make the change later if needed).

c2corg_ui/static/js/map/map.js
```diff
@@ -887,5 +889,16 @@ app.MapController.prototype.resetFeature = function() {
   this.scope_.$root.$emit('mapFeaturesChange', features, true /* isReset */);
 };
 
+/**
+ * @private
+ */
+app.MapController.prototype.updateMapResize_ = function() {
+  setTimeout(function() {
```
asaunier: Please rather use ``ngeo,Debounce`` as it's done at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/map.js#L275-L278

c2corg_ui/static/js/maincontroller.js
asaunier: The event name is not very consistent. What about ``resizeMap`` instead?
The function name is not very consistent either. What about simply ``resizeMap``?

In addition if find it a bit weird to add such a function to the main controller. This function is not used in any page. What about adding it to the advanced search controller instead? https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/advancedsearch.js

Just a silly question: isn't it possible to call ``mapCtrl`` in the toggle buttons? If yes, do we really need such a function?

asaunier:
> which text editor do you use?

@brunobesson I use VIM. I don't know what editor @ginold uses.
Anyway it's a good practise to always run the tests (linter, compilation, unittests) before pushing a commit because it helps catching errors (not only coding standard errors :P ) before someone spends some time reviewing.

asaunier:
Except the code changes suggested above, I have played a bit with those toggle buttons in my instance and the result is really cool :)
There are less buttons, so it's more intuitive and straightforward.
The map extent parameter remains the same when resizing the map, so the results list remains consistents. Perhaps a minor default is that when you use the map full size and then shrink it to show the list, items on the map may be no longer visible. But I think it's acceptable.

Thanks for the great work @ginold If you could take care of the remaining code problems, it would be excellent :)

c2corg_ui/static/js/maincontroller.js
ginold: This is WIP so I let myself some consistencies :p 

See my comment above: I could not access the `mapCtrl` in those buttons probably because they are outside the directive whereas the `mainCtrl` can be accessed anywhere. This was one of my concerns and i'm not sure how to deal with it correctly.

asaunier: OK, than in the advanced search controller? I'm not very found with having very specific pieces of code in the main controller.

ginold: Me neither. Maybe I should create a special directive for those buttons...but it would be another 50 lines of code and a new file instead of 3 lines in the mainCtrl ;p

c2corg_ui/static/js/map/map.js
ginold: I initialized `ngeo.debounce` and wrote 
```
  this.ngeoDebounce_(function() {
    this.map.renderSync();
    this.map.updateSize();
  }.bind(this), 300, true);
```

but the functions are not even being called. Any ideas?

ginold: ok nevermind, i wrote it in the wrong order ;p works ok now!

ginold:
I updated the PR - the only problem I'm aware of is the placement of the code in the mainCtrl which can seem to be not the best place for it.

asaunier:
> the only problem I'm aware of is the placement of the code in the mainCtrl which can seem to be not the best place for it

Placing the function in the advanced search controller doesn't work?

asaunier:
After the changes made by @ginold this PR looks good to me.

In my opinion it could be improved so that the maps can be made full size in other pages too (detail view pages). I wonder for instance if some of the buttons should not be part of the map partial https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/map/map.html instead of having them in a mako template helper.

But such changes could be done when working on this feature.

asaunier:
I have just rebased the PR to fix a conflict with changes made in the meantime + added the toggle map buttons to the xreports as well.

I will merge the PR when Travis is OK.

stef74:
:+1:

Issue #1202, opened 2016-12-13
Card error in outing editing form


asaunier:
When opening the outing creation form (no problem when updating an outing or creating a route for instance), JS erros are displayed in the console:
```
Can't interpolate: {{cardCtrl.createURL()}}
TypeError: Cannot read property 'lang' of undefined
```

Perhaps a problem with the author's user card? This is the only card existing when opening the outing creation form.

asaunier:
It's indeed a problem with the current user's card. In outing creation mode, the user's data do not contain a ``locales`` property and ``cardCtrl.createURL()``.

But actually we don't want to create links for cards in editing forms. See the condition here:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js#L140
```
if (loc.indexOf('/edit/') === -1 && loc.indexOf('/add/') === -1) {
```

The problem is that the outing creation URL is ``outings/add``, not ``outings/add/``. So the test is not correct for document creation pages.

I see the test was previously
```
if (loc.indexOf('edit') === -1 && loc.indexOf('add') === -1) {
```
before https://github.com/c2corg/v6_ui/commit/e1e0339a599cc357500eb702ab51c567a680e42f#diff-245677ec3c01f6fb52aa9163ef37f992
@ginold Is there a reason for adding the slashes?
Do you think it is OK to rather write
```
if (loc.indexOf('/edit/') === -1 && loc.indexOf('/add') === -1) {
```
?

PR #1210, opened 2016-12-13
Fix createURL condition

asaunier:
Close https://github.com/c2corg/v6_ui/issues/1202
I have played around in my instance and could not observe side-effects after this change.

@ginold looks good to you?

### DIFF
--- a/c2corg_ui/static/js/card.js
+++ b/c2corg_ui/static/js/card.js
@@ -137,7 +137,7 @@ app.CardController.prototype.translate = function(str) {
 app.CardController.prototype.createURL = function() {
   var loc = window.location.pathname;
   // Don't create links on edit and add pages.
-  if (loc.indexOf('/edit/') === -1 && loc.indexOf('/add/') === -1) {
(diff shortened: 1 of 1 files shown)


ginold:
Looks good, but i'm afraid I added the slash to the left and right of the words because it can happen, that the path contains `/add...something` and the url will not be generated.

asaunier:
> i'm afraid I added the slash to the left and right of the words because it can happen, that the path contains a title /add...something and the url will not be generated, for example in the associated cards.

It's another demonstration that testing the URL is not a robust way to figure out if something should be activated or not :P 

At least in the pyramid routes I see no other ``/add*`` URLs than the one for creating a new document: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/__init__.py

I will test again and see if there are problems with the associated cards and will merge if not.

ginold:
It should be pretty rare that for example a route name will have add and the url will be /routes/213123/adda-aida-della-ruta :P

PR #1219, opened 2016-12-15
Printing style for detail views

flob38:
Related to #211 

What is done  : 
- Reduce font size
- Reduce margin
- Display area under the title
- Do not display the picto at the right of the title
- Do not display the headers of the collapsed boxes
- Do not display the arrows to collapse boxes
- Do not display the borders
- Underline the headers
- Reduce the orientation picto

_**Before :**_ 
image
With a lot of blank page in the first page
image

_**After :**_ 
image
The headers (Remarques, Matériel) don't appear even if the box is extanded but it's because of a bug in my copied html page on Firefox (class "closed" is not added when the box is collapsed, works well on the site). Works well on both screen and print when I test with Chrome.
image

Several tests with images but there is a bug with the diaporama. However, the bug is also here on the screen. (I work with saved pages from the site, so I hope the bug is not related to CSS)
image
image

If someone could test and confirm (or not) that the bugs I have found are not related to CSS problem but to the "extract" I have made from the site.

### DIFF
--- a/less-print/viewdetails.less
+++ b/less-print/viewdetails.less
@@ -55,3 +55,11 @@
 .heading {
 	text-decoration: underline;
 }
+
+.heading.closed{
+ 	display: none;
+}
+
+.heading .glyphicon{
+	display: none;
+}
--- a/less-print/viewdetails.less
+++ b/less-print/viewdetails.less
@@ -4,14 +4,13 @@
 }
 
 .view-details-title {
-	display: block;
 	font-size: .75em !important;
 	margin-top: 0px;
 	margin-bottom: 10px;
 }
 
-.description , .view-details-info {
-	display: block !important;
+.view-details-info , .view-details-section .view-details-description {
+	float: none;
 }
 
 .view-details-section {
@@ -42,7 +41,6 @@
 
 .view-details-title, .view-details-description, .view-details-informations , .view-details-info {
 	border: none !important;
-	display: block;
 	width: 100%;
 }
 
@@ -62,16 +60,11 @@
 	text-decoration: underline;
 }
 
-.heading.closed {
+.heading.closed, .heading .glyphicon {
  	display: none;
 }
 
-.heading .glyphicon {
-	display: none;
-}
-
 .location-static {
-	position: static;
 	float: none;
 	transform: none;
 	max-width: none;
--- a/less-print/viewdetails.less
+++ b/less-print/viewdetails.less
@@ -1,9 +1,9 @@
 .view-details-associations, .view-details-comments, .view-details-photos,
 .scroll-top-btn, .view-details-map, .badge, .icons-header {
-  display: none !important;
+ 	display: none !important;
 }
 
-.view-details-title{
+.view-details-title {
 	display: block;
 	font-size: .75em !important;
 	margin-top: 0px;
@@ -22,7 +22,7 @@
 	width: 100%;
 }
 
-.lead{
+.lead {
 	padding: 0 10px !important;
 }
 
@@ -31,7 +31,7 @@
 }
 
 .view-details-section .view-details-informations .accordion {
-    padding: 0;
+	padding: 0;
 }
 
 .document-summary {
@@ -62,11 +62,11 @@
 	text-decoration: underline;
 }
 
-.heading.closed{
+.heading.closed {
  	display: none;
 }
 
-.heading .glyphicon{
+.heading .glyphicon {
 	display: none;
 }
 
@@ -79,12 +79,13 @@
 	padding: 5px;
 	margin-bottom: 0;
 	top: 0;
- }
- .location-static li {
-	display: inline;
- }
+}
 
- .view-details-title>h1.routes .title {
-    max-width: 100%;
-    margin-right: 0;
+.location-static li {
+	display: inline;
+}
+
+.view-details-title>h1.routes .title {
+	max-width: 100%;
+	margin-right: 0;
 }
--- a/less-print/viewdetails.less
+++ b/less-print/viewdetails.less
@@ -17,6 +17,9 @@
 .view-details-section {
 	display: block;
 	font-size: .75em !important;
+	margin-left: 0;
+	padding-bottom: 0;
+	width: 100%;
 }
 
 .lead{
@@ -40,6 +43,7 @@
 .view-details-title, .view-details-description, .view-details-informations , .view-details-info {
 	border: none !important;
 	display: block;
+	width: 100%;
 }
 
 #orientation-svg {
@@ -79,3 +83,8 @@
  .location-static li {
 	display: inline;
  }
+
+ .view-details-title>h1.routes .title {
+    max-width: 100%;
+    margin-right: 0;
+}
--- a/less-print/viewdetails.less
+++ b/less-print/viewdetails.less
@@ -1,5 +1,5 @@
 .view-details-associations, .view-details-comments, .view-details-photos,
-.scroll-top-btn, .view-details-map, .location-static, .badge, .icons-header {
+.scroll-top-btn, .view-details-map, .badge, .icons-header {
   display: none !important;
 }
 
@@ -11,7 +11,7 @@
 }
 
 .description , .view-details-info {
-	display: inline-block !important;
+	display: block !important;
 }
 
 .view-details-section {
@@ -37,8 +37,9 @@
 	margin: 0;
 }
 
-.view-details-description, .view-details-informations , .view-details-info {
+.view-details-title, .view-details-description, .view-details-informations , .view-details-info {
 	border: none !important;
+	display: block;
 }
 
 #orientation-svg {
@@ -50,6 +51,7 @@
 	border-bottom: none !important;
 	min-height: 10px !important;
 	max-width: 50%;
+	display: inline;
 }
 
 .heading {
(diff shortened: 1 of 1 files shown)


less-print/viewdetails.less
```diff
@@ -1,5 +1,90 @@
 .view-details-associations, .view-details-comments, .view-details-photos,
-.scroll-top-btn, .view-details-map, .location-static, .badge, .icons-header,
-.logo {
-  display: none;
+.scroll-top-btn, .view-details-map, .badge, .icons-header {
+  display: none !important;
+}
+
+.view-details-title{
```
brunobesson: Please add space before opening bracket.


(@asaunier: not in the short term, but I would like to organize the CSS code more and apply a linter. I am wondering why you chose less over sass or postcss?)

less-print/viewdetails.less
```diff
@@ -1,5 +1,90 @@
...
+	display: block;
+	font-size: .75em !important;
+	margin-top: 0px;
+	margin-bottom: 10px;
+}
+
+.description , .view-details-info {
```
brunobesson: Adding `display: block !important` to `.description` causes the `view-details-photos` to be shown, which is not what we want (and is part of the problem you mentionned)
But I don't understand why the `description` class is added to the div... So I think we should remove the class and your proposal should be fine.
Could you try this and see how it impacts the rendering?

less-print/viewdetails.less
```diff
@@ -1,5 +1,90 @@
...
+	display: block !important;
```
brunobesson: Appart from the comment above, why did you change to `block`? To me it looks like it only adds some unwanted margin.

less-print/viewdetails.less
```diff
@@ -1,5 +1,90 @@
...
+
+.heading {
+	text-decoration: underline;
+}
+
+.heading.closed{
+ 	display: none;
+}
+
+.heading .glyphicon{
```
brunobesson: Maybe use another class? `glyphicon` is a generic class for bootstrap icons. Maybe `.heading .glyphicon-menu-down:last-child`?

less-print/viewdetails.less
```diff
@@ -1,5 +1,90 @@
...
+	display: none;
+}
+
+.location-static {
```
brunobesson: Your solution is nice looking, but I'm wondering whether this information is really needed when printing a route?

less-print/viewdetails.less
flob38: There is a bug with Firefox, with long routes (such as https://www.camptocamp.org/routes/53940/fr/mont-aiguille-la-tour-des-gemeaux) : the Matériel and extras infos parts are cutting the Description part in two : 
image

Yesterday, seems to me that the .description to display: block was solutionning the problem, but no. So there is no reason to have .description to display: block.
I'm working on that.

flob38: I've made some changes : 
- Firefox displaying bug is solved (on my tests)
- No more thumbnails
- Small modifications about the h3 balise and images which were split on two pages

It take a lot of time to make tests with different details views (routes, outings, articles) and different contents (short or long description, images embedded or not, headers, lists...).
I think there could be a bug with Chrome : if content takes a line more than the first page, this line may be not displayed.
Maybe the way texts are broken can by improved.

image
image

image
image
image

Even the home page is better :) (only tested with Chrome). I havn't done anything to improve the printing of the home page, but I wanted to see how it was : 
image

flob38:
Were is the comment I have written? Must have forgotten to save...

I have no time to redo the previews, but the result is quite good (Firefox and Chrome, no test done with Safari or Opera or another browser).
Thumbnails aren't displayed, text is in the good order.
I have left the area because I think it's an important information when you print a route (even more if it's for a to-do list a long time ahead).

Just to see, I have tried the home page (only with Chrome) and it is better too.

This time I save :)

brunobesson:
@flob38 someone, maybe you (?) must have clicked on the cross to delete it, because I received the email notification.

Nevertheless, all of this looks good to me. As Camptocamp SA is still the main entity working on the code, I would like one of them to review your proposal before squashing it into the trunk. @ginold could you do this?

less-print/viewdetails.less
```diff
@@ -1,5 +1,94 @@
...
+
+.lead {
+	padding: 0 10px !important;
+}
+
+.heading {
+	padding: 5px 10px !important;
+}
+
+.view-details-section .view-details-informations .accordion {
```
brunobesson: OK, I start another review only for compacting your code. With less and most other preprocessors, you can nest rules.
E.g., you could write lines 32 to 34 inside block at line 16 the following way:

    .view-details-section {
      display: block;
      font-size: .75em !important;
      margin-left: 0;
      padding-bottom: 0;
      width: 100%;
      
      .view-details-informations .accordion {
        padding: 0;
      }
    }

less-print/viewdetails.less
```diff
@@ -1,5 +1,94 @@
...
+	min-height: 10px !important;
+	max-width: 50%;
+	display: inline;
+}
+
+.heading {
+	text-decoration: underline;
+}
+
+.heading.closed, .heading .glyphicon {
```
brunobesson: Could be put inside `.heading` block above

less-print/viewdetails.less
```diff
@@ -1,5 +1,94 @@
...
+	float: none;
+	transform: none;
+	max-width: none;
+	min-height: 0;
+	padding: 5px;
+	margin-bottom: 0;
+	top: 0;
+}
+
+.location-static li {
```
brunobesson: Could be put inside block above

flob38:
I'm not sure of this :

.heading {
padding: 5px 10px !important;
text-decoration: underline;
&.closed, .glyphicon {
display: none;
}
}

Is it correct?

2016-12-15 21:51 GMT+01:00 Bruno Besson <notifications@github.com>:

brunobesson:
>Is it correct?

Yes. For readibility, you could also put the part after the comma on another line, like this:

    .heading {
      padding: 5px 10px !important;
      text-decoration: underline;
      
      &.closed,
      .glyphicon {
        display: none;
      }
    }

flob38:
Which is the better organisation between having cascades of classes but
duplicate identical properties in those cascades and agglomerated classes
with same properties?

.view-details-title, .view-details-informations {
border: none !important;
width: 100%;
}

.view-details-description{
border: none !important;
width: 100%;
.embedded_right {
page-break-inside: avoid;
}
}

.view-details-info {
float: none;
border: none !important;
width: 100%;
}

.view-details-section {
display: block;
font-size: .75em !important;
margin-left: 0;
padding-bottom: 0;
width: 100%;
.view-details-informations .accordion {
padding: 0;
}
.view-details-description {
float: none;
}
}

And

.view-details-info , .view-details-section .view-details-description {
float: none;
}

.view-details-section {
display: block;
font-size: .75em !important;
margin-left: 0;
padding-bottom: 0;
width: 100%;
.view-details-informations .accordion {
padding: 0;
}
}

.view-details-title, .view-details-description, .view-details-informations
, .view-details-info {
border: none !important;
width: 100%;
}

.view-details-description .embedded_right {
page-break-inside: avoid;
}

2016-12-15 22:22 GMT+01:00 Bruno Besson <notifications@github.com>:

brunobesson:
>Which is the better organisation

Vast subject :wink:
I would say what's make more sense. The goal is not to make it as concise a possible (there are other tools for this) tu to organize the code, factorize what can be and make it easier to understand. Thus, the second option looks better, doesn't it?

flob38:
Yes, aesthetic is definitely a good way to validate things. It's the same
at work (concrete design for buildings) :)

2016-12-15 22:53 GMT+01:00 Bruno Besson <notifications@github.com>:

brunobesson:
Let's wait for @ginold review. But he already works all day on the code, I don't think he does this at night too :wink:

asaunier:
@flob38 I have not tested your changes yet but:
* thanks anyway :)
* could you use 2-spaces indentation instead of tabs? All other CSS/Less files use 2-spaces indentations. Tabs are prohibited in all files (except in the Makefile).

> I am wondering why you chose less over sass or postcss?

@brunobesson No idea. I think that less is the usual tool we use at the SA.

flob38:
@asaunier What is the reason of the banishment of the tabs? Changes are done.

asaunier:
Thanks for the change :)

> What is the reason of the banishment of the tabs? 

In CSS styles we use 2-spaces indentation, tabs are bigger and create mess in the files.
In JS and HTML files we also use 2-spaces indentation.
In python files, 4 spaces (because Python expects 4 spaces anyway).

Anyway using tabs is generally a bad idea, mainly because they are often not interpreted the same way by editors and may result in messy or hard-to-read files. Ditto when viewing the files in github. See for instance  https://www.quora.com/Why-would-a-coder-use-spaces-over-tabs

Finally the main reason for me is that it's important to be consistent and to use always the same way to indent files, ie. always the same characters.

flob38:
Ok, thanks for explanation/explications.

2016-12-19 22:12 GMT+01:00 Alex Saunier <notifications@github.com>:

brunobesson:
@asaunier @ginold can we squash and merge this PR?

asaunier:
@flob38 I am testing your changes in the demo website. As far as I can see it looks very good. Thanks for the great work!

flob38:
Thanks :)
But I'm sure there could be still some work to do : possibility to print
the map, maybe some photo for the outings...

2016-12-22 21:46 GMT+01:00 Alex Saunier <notifications@github.com>:

asaunier:
Sure. But that's the same for all the new c2corg site.
That's my new motto: Rome wasn't built in a day :P

Issue #1165, opened 2016-12-09
Route creation: select the first waypoint as main waypoint by default


brunobesson:
Currently, no waypoint is by default the main waypoint, thus users create routes without prefixed name.

Several possibilities:
* select first waypoint as main waypoint by default. It should then removed if needed
* if no waypoint is selected when creating route, select highest one. But this causes a problem if we do not want a main waypoint, it would requires to edit afterwards
* if no waypoint is selected, ask question whether or not to associate one. Bad UI IMHO

desnoes:
fixed on demo. There is a small traduction issue, i have translated the string ' unset main waypoint' in transiflex.

fbunoz:
The problem came back with the last prod update.

asaunier:
I have just tested with FF 50.1 on the demo and observed no problem: http://www.demov6.camptocamp.org/routes/828342/fr/un-itineraire-pour-main-wp-par-defaut
(I have created a route and only associated the WP => it has been made the main WP directly with no additional action, as expected).

Changes propaged to production recently seem to have no relationship with this feature:
> Image title should not be required when editing an image 
Incorrect WP longitude validation 
Add pagination buttons at the bottom of advanced search results 
Image upload: copyright should be allowed only for moderators 
Map scale design Ergo / Design 
Outing creation: the route search engine is confusing 
fixed WP 'Access': display errorfor the field "Type de transport en commun" 
Comments in topoguide
Add possibility to enlarge or hide map in the advanced search pages 
Show link "more results" only for the main simple search tool 
Sitemap in prod have an error
Cropped dropdown in following page 
No close outing end date button on mobile
Lien profil topoguide -> profil forum

desnoes:
I didn't notice any problem since the last update.
@fbunoz: would you check again? If you do not observe the problem, please close the issue.

desnoes:
I have juste tested on desktop (Chrome) on both the demo and the site, it works as expected.
@fbunoz and @stef74 : would you please check quickly ? Otherwise i will close this issue.

stef74:
what is the test ?

desnoes:
According to Bubu, there is a bug in the selection of a WP by default. Alex and i think it works fine.

----- Mail original -----
De: "stef74" <notifications@github.com>
À: "c2corg/v6_ui" <v6_ui@noreply.github.com>
Cc: "desnoes" <mathieu.desnoes@free.fr>, "State change" <state_change@noreply.github.com>
[...]

PR #1221, opened 2016-12-15
set 1st wp as main - route editing

ginold:
related to #1165

+ if creating/editing a route - the first waypoint is set to main by default : if you have 2 waypoints and delete the 1 which was the main wp -> the last remaining wp will become main. If you have no waypoints at all, the 1st will be main. 
+ possibility to unset the main wp.

screenshot from 2016-12-15 10-51-28

### DIFF
--- a/c2corg_ui/static/js/documentservice.js
+++ b/c2corg_ui/static/js/documentservice.js
@@ -124,16 +124,22 @@ app.Document.prototype.hasAssociation = function(type, id) {
  * @param {appx.SimpleSearchDocument} doc
  * @param {string=} doctype Optional doctype
  * @param {boolean=} setOutingTitle
+ * @param {?boolean} editing
  * @export
  */
 app.Document.prototype.pushToAssociations = function(doc, doctype,
-    setOutingTitle) {
+    setOutingTitle, editing) {
   var associations = this.document.associations;
   doctype = doctype || app.utils.getDoctype(doc['type']);
   setOutingTitle = typeof setOutingTitle !== 'undefined' ?
     setOutingTitle : false;
   doc['new'] = true;
   associations[doctype].push(doc);
+
+  // when creating/editing a route, make the first associated wp a main one
+  if (editing && associations.waypoints.length === 1) {
+    this.document['main_waypoint_id'] = doc['document_id'];
+  }
 
   // When creating an outing, the outing title defaults to the title
   // of the first associated route.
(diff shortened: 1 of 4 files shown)


c2corg_ui/static/js/documentservice.js
```diff
@@ -124,17 +124,23 @@ app.Document.prototype.hasAssociation = function(type, id) {
...
+    setOutingTitle, editing) {
   var associations = this.document.associations;
   doctype = doctype || app.utils.getDoctype(doc['type']);
   setOutingTitle = typeof setOutingTitle !== 'undefined' ?
     setOutingTitle : false;
   doc['new'] = true;
   associations[doctype].push(doc);
 
+  // when creating a route, make the first associated wp a main one
+  if (editing && associations.waypoints.length === 1) {
```
brunobesson: Your comment and the code look the opposite to me. Shouldn't it be `if (!editing ...`?

ginold: I maybe missed something - when i did this, i thought this should be available only while creating/editing of a route, right? For now, it is the only place to define a route's main waypoint. 
Or should it also happen when an admin removes the last waypoints from a route-view page? (more difficult to do, if so)

ginold:
>  do we really want this? I don't understand the purpose but surely I am missing something

I think it should be the normal UI behavior - like when there is no waypoints and you add a 1st one and it will become the main waypoint, the same when you delete all the waypoints but leave only 1 -> it will be main too. You can always unset it.

c2corg_ui/static/js/documentservice.js
brunobesson: No I think I misunderstood what you did. I thought you distinguished between editing / creating with this variable.
Because to me, you shouldn't apply the main waypoint automatically to a route you are editing: if the route has no main waypoint, it's a determinate choice. Plus, it is error prone: imagine you only want to correct a sentence in a route without main point. When commiting your changes, you need to uncheck the automatically added waypoint. This is not the expected behaviour, is it?

brunobesson:
>I think it should be the normal UI behavior

OK. I'm not sure it is needed, but I have no real opposition to doing this. My real concern remains the distinction between creating and editing.

c2corg_ui/static/js/documentservice.js
ginold: I'm not sure to understand - for example, if you edit a route that has waypoints but none of them is a main one, and you only want to edit a sentence in the description of the route, it will have no impact on the waypoints.

The change occur only when you add/remove associated waypoints, or if you click on "set/unset main waypoint". Editing the route other attributes and saving them will do nothing to the associations nor to the `main_waypoint_id`.

brunobesson: OK my bad then!
I agree with your choices.
(this is not a code review though, sorry)

ginold: In other words, it is the same behaviour when creating or editing - the first and only associated waypoint will be a main waypoint, that you can always unset. If you edit a route with for example one waypoint that is not a main one, it will have no impact untill you choose to click on "set as main waypoint" or delete this waypoint and add a new one -> which will become a main one, automatically.

ginold:
@tsauerwein do you have any reservations about this PR? If not i will merge today.

Issue #1014, opened 2016-11-26
In the image upload tool, pre-select the activities


fbunoz:
After uploading an image on an outing or a route in the upload tool, the image activities should be pre-selected with the outing or route activities.
For now, no activity is pre-selected, and almost no-one will take the time to select the activities for all his images. Filtering on image activity will return no V6 image (only old V5 images).

desnoes:
fixed on demo site

PR #1222, opened 2016-12-15
Preselect activity when uploading image to outing

ginold:
related to #1014

when you upload images to a outing, preselect the activities for all images, being those of the outing.

Seems pretty simple but I've got a problem I cannot solve, maybe @tsauerwein you have an idea. 

1) upload multiple images
2) they do get their activities preselected, being those of the outing
3) when I change the activities of one image, it changes the activities of all of them
4) if I don't use a variable like` 'activities': this.defaultActivities_` but instead `'activities': ['hiking', 'snowshoeing']`, no problem with defining specific activities for every image.

Any ideas?

### DIFF
--- a/c2corg_ui/static/js/documentservice.js
+++ b/c2corg_ui/static/js/documentservice.js
@@ -48,6 +48,7 @@ app.Document = function(appAuthentication, $rootScope) {
     },
     'locales': [{'title': '', 'lang': ''}],
     'type': '',
+    'activities': [],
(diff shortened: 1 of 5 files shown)


tsauerwein:
> when I change the activities of one image, it changes the activities of all of them

Make sure to clone the activities array.

ginold:
`angular.copy(this.defaultActivities_) `did the trick!

Issue #1197, opened 2016-12-13
On creation/edition module, allow to increase textual box size


mgerbaux:
For textual field, boxes are sometimes too small to work properly (for example for article).
![screenshot at 2016-12-13 09 04 57](https://cloud.githubusercontent.com/assets/22116605/21132345/ba65d070-c113-11e6-8648-3c9ab7b591a6.png)
In this screenshot, its obvious that this box could be higher for this screen

In V5; such boxes could be increased at demand

fbunoz:
Moreover, the box should be resizable only for the height and not for the width.
To do that : 

- Remove !important attributs on all rules for height on textarea (there are several rules...).
- Add the rule on textarea : 
```resize: vertical;```

ginold:
exactly what I wanted to do, thanks for pointing it out.
Maybe a good idea would be to implement an automatic height, getting bigger with text? 
It even is the case here, in the comment textarea of github.

AlexisDemongeot:
Just a comment (if Textarea auto height is doing the job): 
Actualy, we can resize textual fields but only the widht : 
![image](https://cloud.githubusercontent.com/assets/24522280/23275687/c94e114c-fa07-11e6-9002-e7f55017b55e.png)

Issue #305, opened 2016-04-13
Text formatting shortcuts buttons


fjacon:
@asaunier  said Having a graphical editor seems to me too difficult because the syntax used for the documents is too complicated and too customized. I think we should restrict to some shortcut buttons, as in v5 or github etc.

fjacon:
agreed.

asaunier:
I have set this issue to milestone 6.1 because without those shortcuts people will have a hard time writting the correct markdown (and bbcode?) syntax.

By the way for such tools it would make sense to rely on an existing JS library.
After a quick search I have found:
* https://www.npmjs.com/package/angular-markdown-editor / https://github.com/ghiscoding/angular-markdown-editor / demo : http://plnkr.co/AFxN7uiwkZlJ9OMZwXey
* https://github.com/GrumpyWizards/wizMarkdown

Other JS libraries:
* https://simplemde.com/ / https://github.com/NextStepWebs/simplemde-markdown-editor (looks nice!)
* https://github.com/lepture/editor

It would be interesting to know if it is possible to extend such libraries and add our own tags (wikilinks, img, L#, etc.).

I also think we need to get rid of BBcode, I don't think we will find a JS editor that supports both markdown and bbcode... See https://github.com/c2corg/v6_ui/issues/131

dkocich:
note - this editor looks fine too http://textangular.com/ https://github.com/fraywing/textAngular

arnaud-morvan:
It seems that angular-markdown-editor ( https://www.npmjs.com/package/angular-markdown-editor ) already do the job for basic markdown.
We should create our own directive to customise the editor (remove/add buttons).
So we have to create externs for bootstrap-markdown.

Here is a snapshot : 
![markdown-editor](https://cloud.githubusercontent.com/assets/6095494/23996829/35c8fa56-0a4f-11e7-974d-11824e4d9746.png)

There is some css to apply (buttons text color).
If we want to keep the preview, we should apply the same styles as in the document view mode, or we should hide the button for a first time.
We should also hide the image button for the moment.

asaunier:
> If we want to keep the preview, we should apply the same styles as in the document view mode, or we should hide the button for a first time.

Isn't this button a bit redundant with the general preview button in the right sidebar?

arnaud-morvan:
Yes !! if we can remove the markdown editor preview button, we by-pass complex stuff about css.

fjacon:
for me it is OK to go on and have it on demo asap

[...]

PR #1232, opened 2016-12-16
Textarea auto height

ginold:
related to  #1197

Auto expands the text areas:

+ allow user to set the height of the text area
+ on focus, expand the text area to the full text height
+ on type, adjust the height to the text
+ on mobile, on focus - scroll into view

The only issue i found is that when you type a lot of text and then delete it, you cannot make the area smaller han what it was, with the triangle tool on the bottom right (at least on chrome - on firefox it still works). But IMHO it's not a big issue compared to the gain.


ginold:
@fbunoz would you like to give your opinion on this?

mgerbaux:
> @mgerbaux would you like to give your opinion on this?
it seems very interesting !

fbunoz:
> on focus, expand the text area to the full text height

I should test with a [long article](https://www.camptocamp.org/articles/edit/128981/fr) to evaluate this behavior.
Maybe it's better to limit the height to 80-90% of the screen heigth, but if the height is reduced when the focus is lost, it can be ok.

Other points seem good (and better than the current behavior).

ginold:
I haven't worked with such a long article, it could be a nice try. And for now the focus isn't lost when clicking on another field, I think it would make too much jumping closing/opening in the same time. But i will give it a try!

tsauerwein:
Relying on `resize: auto` is not possible because it does not work well on mobile?

ginold:
@fbunoz So i find that it works well when you limit the height of the textarea to a certain % and resize it down to what it was when the focus is lost. 

You will have however to scroll you way down to the bottom of the text - I found it was annoying if it scrolled for you because maybe you don't want to edit the bottom but the place where you clicked? It's save to just expand the textarea and scroll the window into view (top of the textarea).

For example:

## Default, nothing open, textareas on the bottom

screenshot from 2016-12-21 11-07-34



## click on one of the textareas -> will expand and scroll into view
screenshot from 2016-12-21 11-07-43

## click on another -> closes the previous, opens the new and scrolls into view
screenshot from 2016-12-21 11-07-56

Are those changes ok for you?

fbunoz:
> You will have however to scroll you way down to the bottom of the text - I found it was annoying if it scrolled for you because maybe you don't want to edit the bottom but the place where you clicked?

Yes, it's better not to auto scroll to the bottom.
But if the user scrolls into the textarea, then he changes to another field, and returns to the first field : is the scroll position returns to the top, or the last position is keeped ?
Because when we edit a routes, we pass from map to the text several times (ideally the map should be always visible on large screen, and foldable/expendable, but it is another enhancement). And we don't want to lose the scroll position in the textarea.

ginold:
>  is the scroll position returns to the top, or the last position is keeped ?

The position will be the top, cursor where you clicked -> this is the default behaviour. Making so that the cursor returns to the last known position in a given textarea is, IMHO, a _super extra_ bonus (and pretty hard to implement) and I'm also confused about it: maybe you mostly want to return to where you last clicked, but maybe also a lot of people would like to edit where they just clicked in the beginning of the list? and it will scroll all the way where it was = annoying. 

I would say that we should go with this change as it is now because it already brings a lot of cool improvements. If something goes bad or needs upgrades, we can do it in the future and go back or add new functionalities.

I suggest we merge this PR and you will be able to test it.

c2corg_ui/static/js/style/autoheight.js
```diff
@@ -0,0 +1,66 @@
+goog.provide('app.autoHeightDirective');
+goog.provide('app.AutoHeightController');
+
+goog.require('app');
+
+
+/**
+ * Apply automatic height depending on text to all textarea tags
```
asaunier: Is it really what we want?
It makes sense for the description attribute but not for the summary that should remain... short.

In addition I find it really confusing to add an implicit change to a standard HTML element: it's then hard to figure out why standard textareas get larger with no apparent reason.

I would recommend to create a brand new attribute directive that could be explicitly be added to the textareas this features should actually be used for. Not to override the standard element as it's done a few lines below:
```
app.module.directive('textarea', app.autoHeightDirective);
```

ginold: > It makes sense for the description attribute but not for the summary that should remain... short. 

If we want the summaries to be short, we should maybe add an explicit text "no loger than x characters" and a `max-length` attribute. The simple fact that a textarea will open according to the text is very handy imho. Have you tested it? if not, you should! :p

:+1:  ok for creating a special attribute to distinguish the textareas

asaunier: There is no explicit length limitation for the summary attribute. But since a good summary should remain short, the default textarea size and behaviour look enough to me.

>  The simple fact that a textarea will open according to the text is very handy imho. Have you tested it?

Not yet but I can admit it can be quite handy (no clue for mobile usage though).

I am not that sure it is really relevant that the textarea height grows with no limit: for a very long piece of text it does not make much difference to scroll inside the textarea than in the whole page. In addition if the textarea gets really high it will become a pain in the neck to reach the formatting toolbar (placed at the top of the textarea?) that we will have to restore one day or another to help users  format the content (see https://github.com/c2corg/v6_ui/issues/305).
Or perhaps this toolbar will always be visible, floating at the top of the visible part of the current textarea?

I wonder if setting a max-height for those textareas is not needed anyway.

ginold: > I am not that sure it is really relevant that the textarea height grows with no limit

there is a limit! a limit that i defined 'à l'oeil' and it's max 90% of the viewport height. See https://github.com/c2corg/v6_ui/pull/1232/commits/5d1e0cecb082acd59e2aa51c904c458f564b2b38#diff-bca22f81eb4368fbe89d6009bfa48f2cR58 

Indeed working with a very long text was annoying when the textarea expanded to the whole lenght. 

I think it's hard to imagine the way it works and say 'i think that...' without actually testing it, because it's a little bit abstract. I suggest you test it - it works pretty well in my opinion! on mobile and on desktop.

stef74:
@asaunier some change are needed ?

asaunier:
@stef74 Yes, see my comments above.
For me it is not acceptable that the behaviour of any textarea is impacted by this change. It should be explicitly enabled.

Issue #1126, opened 2016-12-07
WP creation (weather station and webcam)


desnoes:
I create a WP for a weather station. If i forgot to add the url, it is not possible to complete the creation because this field is missing (normal). However, there is no warning message telling this information. This kind of message is displayed when the altitude is missing.
Minor bug.

Same behavior with a webcam

desnoes:
fixed on demo

PR #1246, opened 2016-12-20
fix error message missing:URL for wp weather_station and webcam

ginold:
related to #1126

also fixes the number of steps for weather_station


tsauerwein:
Looks good to me!

Issue #1135, opened 2016-12-07
Calendar to choose outing date cannot be translated


loicperrin:
The calendar used to choose the date of an outing is in english. Is it possible to translate it?

PR #1255, opened 2016-12-22
translate uib-datepicker

ginold:
related to #1135

According to stackoverflow, the `uib-datepicker`uses angular.$locale to determine translated strings. This `$locale` cannot be changed during runtime, so a special directive was needed - `angular-dynamic-locale` and  a npm package `angular-i18n`.

Lots of things for a small change...

Not working on build mode (debug only) - undefined provider even though i added JS to the compiled JS in makefile and created externs ;/


asaunier:
Closure compiler reports a compilation error:
```
ERR! compile /home/travis/build/c2corg/v6_ui/c2corg_ui/static/js/langservice.js:72: ERROR - variable angularLocalePath is undeclared
ERR! compile   this.angularLocalePath_ = angularLocalePath;
ERR! compile                             ^
```

ginold:
I don't get where I should define it? 
it is instantiated in https://github.com/c2corg/v6_ui/pull/1255/files#diff-fe6024dfed98636d73b840f9d9a8bfc0R137

ginold:
Still a problem with unknown provider on compiled mode.

PR #1259, opened 2016-12-22
Define full-width homepage toolbar on mobile

brunobesson:
Before:
snapshot7

After:
snapshot8

### DIFF
--- a/less/homepage.less
+++ b/less/homepage.less
@@ -128,6 +128,8 @@
     }
     @media @phone {
       width: 100%;
+      right:0;
+      border-radius: unset;
       .float-buttons, .float-button {
         margin-right: 5px;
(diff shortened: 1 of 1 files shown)


asaunier:
@brunobesson Perhaps my eyes have not fully recovered yet, but I don't really see the difference between your screenshots :P

brunobesson:
That's because the contrast is week :wink: 
Magnify the left part.

asaunier:
To be honest I still see everything on my screen a bit blurry. So I can't really tell :P

brunobesson:
When playing with curves to accentuate the difference: snapshot12

ginold:
:+1:  eye of the eagle!

Issue #1261, opened 2016-12-22
Map scale design


brunobesson:
#1240 brought a great scale to the map. Thanks for this! :+1: 
However, I dislike it being on a blue background. Maybe this is more consistent with other buttons in the map, but it is _not_ a button! The map already contains a lot of objects.

Why not simply use black?
![snapshot10](https://cloud.githubusercontent.com/assets/2234024/21441307/b3dc82ea-c897-11e6-970a-4a1c522e8117.png)

Using `text-shadow` property, we could ensure text is always visible even depending on the map background. E.g. `text-shadow: 2px 0 0 #fff, -2px 0 0 #fff, 0 2px 0 #fff, 0 -2px 0 #fff`
![snapshot11](https://cloud.githubusercontent.com/assets/2234024/21441596/24af899e-c899-11e6-9496-e9139461d965.png)

What do you think?

asaunier:
I'm +1 because:
- it's not really a tool that could be used to interact with the map
- there are already so many item on the map. It's getting harder to see the map :P 
- it's not a critical element and it's current layout perhaps attract a bit too much the attention (well, at least you can't say you had not seen it)

What about adding a slight white background shade? It would make it easier to read when the map is dark.

brunobesson:
See PR #1262

ginold:
It was the default style, I didn't reflect on the style. I would just say that with no background, occasionally the scale is hardly readable (see 2nd screenshot for example).

brunobesson:
On the second screenshot, I was applying `text-shadow`. In the PR I tried instead a background as suggested by @asaunier.
Anyway, we can always use another alpha if the one I chose feels to translucent to some. I suggest we go this way, and make further adjustments later on if needed (because IMHO, it's still an improvement).

asaunier:
Since the PR is merged, can we close this issue or at least tag it "ready for demo"?

brunobesson:
I'm going to adjust the `limit search to map` field a little bit more, but yes, do as you usually handle it. @stef74?

brunobesson:
Sorry, the additional commit is not related to this issue, it concerns the overall map buttons and inputs design.

celinestern:
It's ok.
But no easy in first sight. Buttons are once down, once up.

desnoes:
fixed on demo

PR #1262, opened 2016-12-22
Advanced search map buttons

brunobesson:
* Reduce some buttons and inputs size (hope you won't find it to small on mobile, but there are already some small buttons nobody complained of)
* scale rendering (see #1261)

Before:
snapshot13
After:
snapshot12


brunobesson:
@ginold @tsauerwein  OK to merge?

ginold:
I just had a little remark in #1261:

>  I would just say that with no background, occasionally the scale is hardly readable (see 2nd screenshot for example). 

other than that :+1:

brunobesson:
OK, I answered to your comment in #1261. I merge this PR, and we can always adjust coloring later on as needed. Thanks for the feedback!

Issue #1272, opened 2016-12-23
Image upload: copyright should be allowed only for moderators


loicperrin:
When uploading an image (with an account which have no moderators right), I can select "copyright" for its type. This type should be allowed only for moderators.

asaunier:
In addition I think "copyright" is only allowed for books (covers).

Perhaps the list of image types should be filtered here: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/imageuploader.html#L78

celinestern:
Ok, "copyright" choice is not present for "simple" member.

Issue #1112, opened 2016-12-06
Prevent multiple image uploads


asaunier:
After dragging images to the upload panel, it is possible to click several times in a row on the "save" button. This will trigger several identical POST ``images/list`` requests to the API.

It seems that only the first one succeeds, the other returns a 500 error. Because of the image filename unicity constraint I guess. So it probably does not create duplicates in the DB.

In addition there is no indicator showing that the images are being saved.

TODO:
* [ ] show a loading indicator when saving image (using the ``app-loading`` directive?)
* [x] disable the buttons when the "save" button has been pressed

HSERGERAERT:
Tested this morning with outing http://www.demov6.camptocamp.org/outings/828236/fr/bel-oiseau-couloir-sse.
Only the first picture was loaded, the others weren't, no link with image sizing.
I loaded successfully the first one.
Then I moved, and came back later on the outing, it wasn't possible to add pictures. Is-it the same issue ?

asaunier:
@HSERGERAERT I have not been able to reproduce the behaviour you describe on demov6, including with a normal (not moderator) account. No problem to load images, including in several times, opening another page in the meantime etc.

Clicking several times on the "save" button does what I have described above: images are saved only once, the subsequent requests fails with a 500 error, which is actually OK since it's the same image file. Of course it works if I upload several times the same image (but than filenames are different).

HSERGERAERT:
@asaunier : test made right now, it works.

Issue #929, opened 2016-11-18
Problems with image uploading


desnoes:
When i try to upload an image, the upload counter (percentage) goes up to 99% and then go back to 10% and restart. This process loops endlessly and the image is never uploaded. I use FF with version 49.0.2.

mgerbaux:
Same issue (using FF or Chromium)
Images with small weight (100-200 ko) can be uploaded, larger(about 1.5Mo) not (with @asaunier poor wifi connection).

ginold:
I tried to upload big images (one 2.5Mo and one 5.1Mo) simultaneously, on FF (49.0.2) and Chrome with no issues. 

If there is a poor connection - that is nothing special. It happens to me to load pictures on facebook from my phone with bad connection and the loading sometimes never reaches the end...

asaunier:
I have tested to upload  3.7MB image from my place, with a standard connection (actually a pretty low one). It failed all the times (4 times) I have tried with alert  "Error while uploading the image: Timeout": after about 60 seconds the upload seems to be interrupted if it's not over.

At least twice I have noticed the progress bar went until ~60% then fell back to 10% before growing again and finally being interrupted by the time out.

ginold:
I retested with the 5MB image with throttling in google (Regular 2G - 50kb/s) - it took maybe 5 minutes but it made it - but only on my instance, not on demo-c2corg! 

I'm not sure to be able to answer to this issue. @arnaud-morvan, does this problem says something to you ?

mgerbaux:
I try with another PC (49.0.2 /linux) : same issue to upload images (about 1.5Mo, compliant to V5 spec and already uploaded on V5)
My connection is at the moment about 1Mb/s for uploading

fbunoz:
I tried uploading a 5Mb image : it seems that it is uploaded without resizing (firebug notes 5Mb uploaded data).
In these conditions, it is normal the server doesn't accept the file.

arnaud-morvan:
Sorry, but I don't really know how to solve this.
Maybe we could increase the image backend timeout, gunicorn should have an option for that.

mgerbaux:
What are the specificities of the images (size, weight) theoritecally accepted by the server ?

arnaud-morvan:
No restriction on image properties, the problem should come from some timeout with the request.

asaunier:
Is the problem limited to FF? If not we should update the title.

[...]

PR #1282, opened 2016-12-28
Images upload fixes

asaunier:
* Only moderators can upload copyright-typed images => see https://github.com/c2corg/v6_ui/issues/1272
* Client-side resizing was active only for the initial upload panel, not for the additional panel shown after a first upload has been triggered => there was no resizing in that case => perhaps causing https://github.com/c2corg/v6_ui/issues/929#issuecomment-268836270
* Save/Close buttons are disabled after clicking the Save button => prevent multiple requests, see https://github.com/c2corg/v6_ui/issues/1112 It would be nice to show a loading indicator too but not sure we can really use the ``appLoading`` directive here since we don't want to trigger the loading indicator when uploading files, only when saving the images data to the API.

### DIFF
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -152,6 +152,12 @@ app.ImageUploaderController = function($scope, $uibModal, $compile, $q,
    */
   this.resizeOptions = {'width': 2048, 'height': 2048, 'quality': 0.9};
 
+  /**
+   * @type {boolean}
+   * @export
+   */
+  this.saving = false;
+
   this.scope_['activities'] = this.activities;
   this.scope_['types'] = this.types;
   this.scope_['categories'] = this.categories;
@@ -475,7 +481,9 @@ app.ImageUploaderModalController = function($scope, $uibModalInstance) {
  * @export
  */
 app.ImageUploaderModalController.prototype.close = function() {
-  this.modalInstance_.close();
+  if (!this.scope_['uplCtrl'].saving) {
+    this.modalInstance_.close();
+  }
 };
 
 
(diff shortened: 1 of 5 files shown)


asaunier:
faf24da fixes the way to figure out if a document is collaborative => used for instance to figure out the default image types - collaborative or personal - of uploaded images.

c2corg_ui/static/js/imageuploader.js
```diff
@@ -409,6 +417,23 @@ app.ImageUploaderController.prototype.resizeIf = function(
 };
 
 
+/**
+ * @param {Array.<string>} image_types
+ * @return {Array.<string>}
+ * @export
+ */
+app.ImageUploaderController.prototype.filterImageTypes = function(image_types) {
```
tsauerwein: camel case: `imageTypes`?

c2corg_ui/static/partials/imageuploader.html
```diff
@@ -113,6 +113,8 @@
          ngf-keep="'distinct'"
          ngf-fix-orientation="true"
          ngf-allow-dir="true"
+         ngf-resize-if="uplCtrl.resizeIf($file, $width, $height)"
+         ngf-resize="{width: 2048, height: 2048, quality: 0.9}"
```
tsauerwein: Do you maybe want to move these options into a variable in the controller and use `ngf-resize="uplCtrl.resizeOptions"` here and above?

c2corg_ui/static/js/imageuploader.js
```diff
@@ -477,7 +477,9 @@ app.ImageUploaderModalController.prototype.close = function() {
  * @export
  */
 app.ImageUploaderModalController.prototype.save = function() {
+  $('.img-upl-button').addClass('disabled');
```
tsauerwein: Adding and removing the class like this with jQuery is not very Angular!

asaunier: I think I have copied some code found somewhere else in the UI :P 
How would you recommend to do it?

tsauerwein: More Angular-like would be to use [ng-class](https://docs.angularjs.org/api/ng/directive/ngClass) and to store the disabled status in a variable on the controller.

asaunier: Thanks. I will make changes to take your suggestions into account.

asaunier:
@tsauerwein I have made the changes you have suggested. By the way I have realized that setting class ``disabled`` on the buttons does not prevent them to be clicked => I have also used the saving status to figure out whether their handlers should be enabled or not.

I will merge the PR when Travis is OK if you have no other comment.

Issue #1287, opened 2016-12-30
Image title should not be required when editing an image


asaunier:
When editing an existing image, the editing form requires the image title to be provided whereas this attribute is not required when creating the image: the attribute should not be required when updating as well.

AlexisDemongeot:
It works for me

desnoes:
fixed on demo

PR #1289, opened 2017-01-01
Allow empty title in image editting form

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1287

It won't be required to submit image's title in the edit form now.


asaunier:
Thanks @dkocich !
That looked good to me but I have also made additional changes to your PR, removing the "mandatory field" indicators. OK for you?

Issue #1284, opened 2016-12-29
Article association tool


loicperrin:
- [x] It is not possible to associate an article from an outing, where as it is possible from a route (but the outing can be associated from the article)

stef74:
@loicperrin could you add bug label for thie type of issue ;-)

stef74:
If is the bug is critical use in complement the lable blocking or high prio

desnoes:
fixed in demo

PR #1290, opened 2017-01-01
Allow to associate article from outing detail view page

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1284
It is now able to associate articles (and routes) from outing detail view page

### DIFF
--- a/c2corg_ui/static/js/addassociation.js
+++ b/c2corg_ui/static/js/addassociation.js
@@ -88,7 +88,8 @@ app.AddAssociationController.prototype.associate = function(doc) {
   // if the parent doc is a route and the child doc is a waypoint OR
   // if the parent doc is an outing, inverse the IDs.
   if ((parentType === 'routes' && doc['type'] === 'w') ||
-      parentType === 'outings' || parentType === 'images' ||
+      (parentType === 'outings' && (doc['type'] === 'r' || doc['type'] === 'u')) ||
+      parentType === 'images' ||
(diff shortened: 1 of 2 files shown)


c2corg_ui/static/js/addassociation.js
```diff
@@ -88,7 +88,8 @@ app.AddAssociationController.prototype.associate = function(doc) {
   // if the parent doc is a route and the child doc is a waypoint OR
   // if the parent doc is an outing, inverse the IDs.
   if ((parentType === 'routes' && doc['type'] === 'w') ||
-      parentType === 'outings' || parentType === 'images' ||
+      (parentType === 'outings' && (doc['type'] === 'r' || doc['type'] === 'u')) ||
```
asaunier: I am not sure to understand this change. Isn't article type missing here?

dkocich: Outings were swapped to be child in all added associations on the outing view page (but in the case of articles we should keep them as children in its association to outings and there is no need to include them here - for outings only route/user should be swapped). https://github.com/c2corg/v6_common/blob/master/c2corg_common/associations.py#L7-L8

asaunier: OK thanks for the explanations.

To be honest I think this part is getting a bit confusing and not easy to understand. I don't know if we could do something to make it more explicit?

asaunier:
I have merged the PR because we would like to update the demo soon :)

Issue #1300, opened 2017-01-04
Distances for mountainbiking are displayed in meters instead of kilometers


asaunier:
In the detail view:

![capture du 2017-01-04 18-48-55](https://cloud.githubusercontent.com/assets/1192331/21652804/990c1ab8-d2ae-11e6-8c92-59d847baaa61.png)

In the editing form:

![sans titre](https://cloud.githubusercontent.com/assets/1192331/21652815/a0c39416-d2ae-11e6-8a88-efc8d2a9c0a9.png)

desnoes:
edition form: distance is given in meters (to remain consistent with data already in the DB)
display: distance is given in km

asaunier:
OK thanks, so no change then? I think it's already how it works in prod.

asaunier:
PR #1301 has been updated.

desnoes:
it has been fixed on demo

PR #1301, opened 2017-01-04
Fix incorrect units for distance attributes

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1300

I have made sure the unit for those attributes is actually "km" in the google spreadsheet.
I have also fixed a error for attribute ``route_length`` that should be used for activity ``mountain_climbing`` instead of ``rock_climbing``.

### DIFF
--- a/c2corg_ui/static/js/editing/outingediting.js
+++ b/c2corg_ui/static/js/editing/outingediting.js
@@ -123,9 +123,7 @@ app.OutingEditingController.prototype.successRead = function(response) {
  * @public
  */
 app.OutingEditingController.prototype.prepareData = function(data) {
-  // adapt the Object for what's expected on the API side + format the outing
-  this.submit = true;
-  this.formatOuting_(/** @type appx.Outing */ (data));
+  this.formatOuting_(/** @type appx.Outing */ (data), true);
   // Length attributes are stored in meters but shown in kilometers:
   data['length_total'] *= 1000;
   return data;
@@ -133,15 +131,17 @@ app.OutingEditingController.prototype.prepareData = function(data) {
 
 
 /**
- * @param {Object} outing
+ * @param {appx.Outing} outing
+ * @param {boolean=} submit
  * better edit-form checking before saving
  * @private
  */
-app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
-  if (this.submit) {
+app.OutingEditingController.prototype.formatOuting_ = function(outing, submit) {
+  if (submit) {
     // transform condition_levels to a string
     if (typeof outing.locales[0]['conditions_levels'] !== 'string') {
-      outing.locales[0]['conditions_levels'] = JSON.stringify(outing['locales'][0]['conditions_levels']);
+      outing.locales[0]['conditions_levels'] =
+        JSON.stringify(outing['locales'][0]['conditions_levels']);
     }
     if (outing.date_start instanceof Date) {
       outing.date_start = window.moment(outing.date_start).format('YYYY-MM-DD');
@@ -187,7 +187,6 @@ app.DocumentEditingController.prototype.formatOuting_ = function(outing) {
     }
   }
 
-  this.submit = false;
   return outing;
 };
 
--- a/c2corg_ui/static/js/editing/outingediting.js
+++ b/c2corg_ui/static/js/editing/outingediting.js
@@ -105,6 +105,8 @@ app.OutingEditingController.prototype.successRead = function(response) {
     if (!this.differentDates) {
       outing['date_end'] = undefined;
     }
+    // Length attributes are stored in meters but shown in kilometers:
+    outing['length_total'] /= 1000;
   } else {
     this.alerts.addError('You have no rights to edit this document.');
     setTimeout(function() { // redirect to the details-view page
@@ -122,7 +124,10 @@ app.OutingEditingController.prototype.successRead = function(response) {
  */
 app.OutingEditingController.prototype.prepareData = function(data) {
   // adapt the Object for what's expected on the API side + format the outing
+  this.submit = true;
   this.formatOuting_(/** @type appx.Outing */ (data));
+  // Length attributes are stored in meters but shown in kilometers:
+  data['length_total'] *= 1000;
   return data;
 };
 
--- a/c2corg_ui/templates/outing/helpers/view.html
+++ b/c2corg_ui/templates/outing/helpers/view.html
@@ -204,10 +204,10 @@
         <p><b class="value-title" translate>elevation_access</b>:  <span class="value">${outing['elevation_access']}&nbsp;m</span></p>
         % endif
 
-       ${get_document_up_down(outing, 'height_diff')}
+        ${get_document_up_down(outing, 'height_diff')}
 
         % if outing.get('length_total'):
-        <p><span class="value-title" translate>length_total</span>: <span class="value">${outing['length_total']}&nbsp;m</span></p>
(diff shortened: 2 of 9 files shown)


asaunier:
@fbunoz seems to say it's not correct at https://forum.camptocamp.org/t/unite-des-distances-pour-les-itineraires-vtt/175832/4

If I understand correctly, the attribute is in meters and must be set in meters in the editing form, but displayed in kilometers. I find it a bit inconsistent and confusing but it seems data from v5 has been migrated in meters. In the google spreadsheet the attributes are marked in km.

Or perhaps we should use km for editing and displaying but store them in meters in the database?
Again some complexity...

asaunier:
I have updated the PR to:
* take the instructions from https://github.com/c2corg/v6_ui/issues/1300#issuecomment-272205380 into account
* distances are however shown in km in the editing form as well (for routes and outings) but remain stored in meters.
* fix the preview that didn't use the ``prepareData`` overridable method
* add a ``filterData`` method that can be overridden to apply some conversions on the data provided by the API to the editing form ( @dkocich perhaps it will be useful for xreports as well?).

dkocich:
@asaunier I have no attributes with meter units in Xreports (except elevation where it is correct I think)

asaunier:
> I have no attributes with meter units in Xreports

@dkocich I was rather thinking to this addition you made to the document editing controller and that could be moved to the xreports editing controller when this PR is merged:
https://github.com/c2corg/v6_ui/blob/d7b9f81cbe68f0663523025d75e2980271b035c7/c2corg_ui/static/js/editing/documentediting.js#L245-L247

asaunier:
Any objections against this PR?
@tsauerwein has approved it 2 weeks ago but many changes have been done in the meantime, so I would prefer the PR is reviewed and approved again.

brunobesson:
@asaunier are you wishing for a review from association members, or from the SA only? I did have a look at the diff, and it seems consistent to me (although I would have one tiny question, and wonder why you corrected rock_climbing with mountain_climbing without mentioning it although I think this is correct), but I don't now if my review is really valuable in this context.

asaunier:
Up to you: all comments welcome!

c2corg_ui/static/js/editing/outingediting.js
```diff
@@ -122,7 +124,10 @@ app.OutingEditingController.prototype.successRead = function(response) {
  */
 app.OutingEditingController.prototype.prepareData = function(data) {
   // adapt the Object for what's expected on the API side + format the outing
+  this.submit = true;
```
brunobesson: what's the use of this?

c2corg_ui/templates/route/edit.html
```diff
@@ -297,11 +297,11 @@ <h2 class="heading show-phone"><span translate>numbers</span></h2>
             ## ROUTE LENGTH
             <div id="route_length-group" class="data half form-group" ng-class="{ 'has-error': editForm.route_length.$touched && editForm.route_length.$invalid,'has-success': route.route_length }"
                  ng-if="route.activities.indexOf('skitouring') > -1 || route.activities.indexOf('snow_ice_mixed') > -1 || route.activities.indexOf('hiking') > -1 || route.activities.indexOf('snowshoeing') > -1
-                           || route.activities.indexOf('via_ferrata') > -1 || route.activities.indexOf('mountain_biking') > -1 || route.activities.indexOf('rock_climbing') > -1">
```
brunobesson: As you haven't mentioned it in the issue or in this PR, do you confirm using mountain_climbing rather than rock_climbing is a bugfix? (I would agrre)

c2corg_ui/templates/outing/helpers/view.html
```diff
@@ -204,10 +204,10 @@
         <p><b class="value-title" translate>elevation_access</b>:  <span class="value">${outing['elevation_access']}&nbsp;m</span></p>
         % endif
 
-       ${get_document_up_down(outing, 'height_diff')}
+        ${get_document_up_down(outing, 'height_diff')}
 
         % if outing.get('length_total'):
-        <p><span class="value-title" translate>length_total</span>: <span class="value">${outing['length_total']}&nbsp;m</span></p>
+        <p><span class="value-title" translate>length_total</span>: <span class="value">${outing['length_total']/1000}&nbsp;km</span></p>
```
brunobesson: If divided by 1000, how is rounding applied? (yes, I could checkout the branch and test, too, but let's take that as an exercise :wink:)

c2corg_ui/static/js/editing/outingediting.js
asaunier: Thanks for asking :)

That's because ``prepareData`` is now used by the preview request as well: https://github.com/c2corg/v6_ui/blob/3a080ad3536c30eb3b2f166e0d51b9d10e833b5d/c2corg_ui/static/js/editing/documentediting.js#L570-L574 (data sent to the preview service or to the saving/creating services should be consistent since the same template helpers are used to display them).

But ``this.submit`` is set to ``true`` only when submitting the form: https://github.com/c2corg/v6_ui/blob/3a080ad3536c30eb3b2f166e0d51b9d10e833b5d/c2corg_ui/static/js/editing/documentediting.js#L318

and we need it in the outing's ``formatOuting_`` method: https://github.com/c2corg/v6_ui/blob/3a080ad3536c30eb3b2f166e0d51b9d10e833b5d/c2corg_ui/static/js/editing/outingediting.js#L140-L141

I admit it would make sense to move ``this.submit = true;`` directly to the preview service caller in the parent class: https://github.com/c2corg/v6_ui/blob/3a080ad3536c30eb3b2f166e0d51b9d10e833b5d/c2corg_ui/static/js/editing/documentediting.js#L562
=> I think I will make the change.

c2corg_ui/templates/route/edit.html
asaunier: I don't remember precisely because I made the change 2 weeks ago in the initial version of the PR. But I think I double checked in the google spreadsheet, yes.

c2corg_ui/templates/outing/helpers/view.html
asaunier: The data is stored in meters. With this change, this attribute will for instance be displayed as follows:
* 1000 m => 1.0 km
* 12345 m => 12.345 km
* 12400 m => 12.4 km
* 150 m => 0.15 km

Actually you can already test in prod/demo for the route's ``route_length`` since that attribute was already displayed like this in master:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/route/helpers/view.html#L170

c2corg_ui/templates/route/edit.html
brunobesson: OK then.

c2corg_ui/templates/outing/helpers/view.html
brunobesson: OK. Didn't think we would have only 3 digits at most, which is not a problem (and should be very uncommon)

c2corg_ui/static/js/editing/outingediting.js
brunobesson: Thanks for the explanation :wink:

asaunier: > I admit it would make sense to move this.submit = true; directly to the preview service caller in the parent class

Actually I have removed this ``this.submit`` flag because it was outing-specific and could be replaced by an additional argument in the outing's ``formatOuting_`` method. See 08b7274

Issue #1304, opened 2017-01-05
Add more column in card list when screen are large


stef74:
Due to the hiding map, we can easly had one or two column for large screen. I have tested on mys side and the rederendering is better (iwth only 4th column card are too wide).

Probably need to had more than 30 results but it's an other point.

I try to do The PR by myself

brunobesson:
Working, although break points are not optimal IMHO. I wonder if there isn't a problem with the media query used: it applies the same layout with or without the map. Thus, it may be OK without the map, but not when map is shown. @stef74 have you taken this into account?

stef74:
I have observed that when map is hidden the number of column is increasing. So I not challenge more than that ... but I have found is strange that is work as well based only on size of screen. 
We can do a media query on more precise element ?

brunobesson:
>We can do a media query on more precise element ?

Nope.

stef74:
I idea to test the fact map is hidden or not in CSS

brunobesson:
It is not the right way to do it. Instead, there are flex properties that should be able to solve this (so I think, at least, using shrink and grow properties, in addition to max-width, or smth similar). It should be investigated this way.

stef74:
An other way is to use class no-map ... but I think I have the solution. 
flex-basis:-moz-min-content and min-width:250px ... tested on my firefox it's seem to work

I have check the support of this properties

stef74:
https://developer.mozilla.org/fr/docs/Web/CSS/flex-basis#Compatibilité_des_navigateurs

flex based on content not supported by safarari ... with special properties for firefox. I'm not sure is a good idea to implement it. 

I have found an other similar and tested on browser (not on my vm) with fixed width (it's not so bad) with sucess ... in many size and with(out) map. 
flex-basis:  auto;
width:300 px;

We can after remove media query based on flex basis.

stef74:
We can close

PR #1305, opened 2017-01-05
add up to 7 column

stef74:
#1304


asaunier:
Not sure your changes are fully correct.

I have tried to span my browser on both my screen and hidden the map.
I succeed in having 7 columns as expected:

capture du 2017-01-05 17-41-37

but if I continue making the window wider it goes back to 4 columns:

capture du 2017-01-05 17-34-27

stef74:
could you provide me you CSS file generated, I test on my desktop this evening. I have an idea of some conflict ...

asaunier:
> could you provide me you CSS file generated

Here is my non minified (but zipped) built CSS file:
[build.css.zip](https://github.com/c2corg/v6_ui/files/687985/build.css.zip)

stef74:
Ok I have found ... we not manage correctly more than 2700px

stef74:
Rendering on imac 27"
capture d ecran 2017-01-05 a 20 41 46

Need to add more line now ;-) Card are very powerful in large screen !!!!

less/lists.less
```diff
@@ -394,15 +394,24 @@
   }
 }
 
-// + 1 more card per row than usual, max 4
+// + 1 more card per row than usual, max 7
```
asaunier: Is this comment still relevant? I mean that's no longer "1 more card per row", that could be "4 more cards"?

stef74: ok change to "add more" ... max 7

asaunier:
That seems to work fine now.
However I have noticed that in a "normal" screen when hidding the map, 4 cols are now displayed instead of 3 (2 cols in both cases when the map is on), which means the cards are narrower. I think we had make sure the cards (especially the routes and outings) were wide enough, with not too many cols?

Before:

capture du 2017-01-06 12-23-09

After:

capture du 2017-01-06 12-23-52

less/lists.less
asaunier: I don't know. Simpler is better. For instance "up to 7 columns"

stef74:
what is your screen size in your example ? It's just matter of adjustment.

less/lists.less
```diff
@@ -394,15 +394,24 @@
...
+      flex-basis: 14% !important;
+    }
+    @media (max-width: 2400px) {
+      flex-basis: 16% !important;
+    }
+    @media (max-width: 2100px) {
+      flex-basis: 19% !important;
     }
-    @media (min-width: @large-width) {
+    @media (max-width: 1800px) {
```
asaunier: Please note ``@large-width`` is the same here (OK it's a not that readable) since its value is actually 1800px. See https://github.com/c2corg/v6_ui/blob/master/less/variables.less#L7

stef74: sometimes this less file have variable ... sometimes not. For me it's more readable whith direct value ... but I can use existant variable or create some new one.

asaunier:
I have still the feeling that cards are made narrower than before when the map is hidden and that it's a regression considering the fact that changes have actually been made to ensure cards are always wide enough.

See for instance in my screen the same window with/without the map:

capture d ecran 2017-01-11 a 00 30 02

capture d ecran 2017-01-11 a 00 30 18

But OK for me. When merging please squash your commits. Thanks.

stef74:
@asaunier what is you screen size ? I can increase little bit the level to have only 3 cards in you case.

asaunier:
No idea, it was on my 15'' macbook laptop, say a quite standard screen.
But that's not a disaster. I think we can merge as is and fine tune later according to testers/users feedbacks.

Please squash your commits when merging.

Issue #1297, opened 2017-01-04
Missing attributes in the preview


asaunier:
Some attributes are not shown in the preview even though set in the editing form:
* [x]  outing's ``conditions`` (see https://github.com/c2corg/v6_ui/pull/1299)
* [x] outing's ``conditions_levels`` (see https://github.com/c2corg/v6_ui/pull/1307)

desnoes:
fixed in demo

PR #1307, opened 2017-01-05
Fix conditions levels preview

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1297

### DIFF
--- a/c2corg_ui/templates/outing/helpers/view.html
+++ b/c2corg_ui/templates/outing/helpers/view.html
@@ -210,15 +210,12 @@
 
 <%def name="get_conditions(locale, open_tab=False)">\
   <%
-      if type(locale['conditions_levels']) is str:
-          cond = loads(locale['conditions_levels'])
-          conditions_levels_exist = (cond is not Null and len(cond)) and \
-              ( ('level_snow_height_total' in cond[0] and cond[0].get('level_snow_height_total') != '' and cond[0].get('level_snow_height_total') is not None) or \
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/outing/helpers/view.html
```diff
@@ -210,15 +210,12 @@
...
-              ('level_place' in cond[0] and cond[0].get('level_place')  != '') or \
-              ('level_comment' in cond[0] and cond[0].get('level_comment') != '') )
-      else:
-          conditions_levels_exist = False
+      cond = loads(locale['conditions_levels']) if type(locale['conditions_levels']) is str else locale['conditions_levels']
+      conditions_levels_exist = (cond is not None and len(cond)) and \
+          ( ('level_snow_height_total' in cond[0] and cond[0].get('level_snow_height_total') != '' and cond[0].get('level_snow_height_total') is not None) or \
+          ('level_snow_height_soft' in cond[0] and  cond[0]['level_snow_height_soft'] != '' and cond[0].get('level_snow_height_soft') is not None) or \
+          ('level_place' in cond[0] and cond[0].get('level_place')  != '') or \
+          ('level_comment' in cond[0] and cond[0].get('level_comment') != '') )
```
tsauerwein: I am not trying to understand this condition, but I assume you tested carefully.

asaunier: I have not changed this condition, it was written by @ginold (I have replaced ``Null`` by ``None`` though).
My change is related to the fact that those ``conditions_levels`` are passed as a list when previewing instead of a string in detail view => the new code takes also into account the list, not only the string.

Issue #1298, opened 2017-01-04
Associating books from route edition page


brunobesson:
When editing or creating a route, the association tool deosn't return book results, whereas it does in the view page.
Associating books to a route is a common task, which currently needs to be done after edition.

What do you think about adding book results when editing?

brunobesson:
Do you propose I make a PR for this?

brunobesson:
>In the detail view, we have dataset="wrcb"

Thus, we should allow articles, too. Isn't it?

asaunier:
> Thus, we should allow articles, too. Isn't it?

I guess so.

> Do you propose I make a PR for this?

As you want. Not sure the change I suggest is enough but since it is an easy change, it's worth trying.

brunobesson:
I think I misused the tags here and it went undersea. The PR for this issue has been merged, and the code went in production some time ago. Closing.

PR #1310, opened 2017-01-05
Allow to associate books and articles when creating/editing a route

brunobesson:
Relates to #1298

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -43,6 +43,20 @@
   </section>
 </%def>
 
+<%def name="show_editing_associated_books(model, hide_title = False)">\
+  <section ng-show="${model}.associations.books.length > 0" class="associated-section">
+    % if not hide_title:
+      <label translate>Associated books</label>
+    % endif
+    <article class="associated-documents">
+      <div class="list-item" ng-repeat="bk in ${model}.associations.books" ng-class="{'new-item': bk['new']}">
+        <app-card app-card-doc="bk" app-card-disable-click="true"></app-card>
+        <span class="glyphicon glyphicon-trash" ng-click="editCtrl.documentService.removeAssociation(bk.document_id, 'books', $event)"></span>
+      </div>
+    </article>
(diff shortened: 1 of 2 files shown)


brunobesson:
I forgot to test that associated books and articles are effectively saved... and it's not working, sorry. Adding [WIP] prefix.

brunobesson:
:confounded: 
I'm sorry to spam github. I think I'm getting tired, but before I tested creating a new route for a climbing site on the demo website, without it being set as main waypoint, and then got a 400 error when submitting, with the response `no association for the main waypoint : main_waypoint_id`.
But know, it isn't happening anymore (maybe it never was...). So I guess the problem is caused by those new associations.
If I try to submit a new route with additional book in associations, I get the bad request error.

Could someone try and tell me what's happening? Assuming the problem is caused by the new associations, is this then a bug or change to be made to the UI?

brunobesson:
@tsauerwein maybe you can give me advice on whether the problem is API-related or I am missing something? Thx for your review.

dkocich:
It is an API error here https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/route.py#L66
You must set main WP like in the image for route 840251. I successfully added associations to mainWP/WP/routes/books/articles and saved my routes without error. It appears only when you do not have main WP selected.
image

brunobesson:
Coming back to this PR, left appart for a few days. Thanks @dkocich for your feedback. As far as I understand, this PR looks OK for you. The problem with the main waypoint and the creation error is another issue (if it is one).

Thus, we could merge this PR?

dkocich:
I am not sure if this error appears in logs - if yes, to prevent this it is maybe possible to hide save button when there is no Main WP set and show warning that it is required. Otherwise, it was ok for me and I let others decide about merging.

asaunier:
> The problem with the main waypoint and the creation error is another issue

Indeed, I don't really see the connection between this PR about associated books/articles and main waypoints :)

@brunobesson  The PR looks good to me, except that a conflict should be resolved first.

PR #1313, opened 2017-01-06
Hide toggle buttons on mobile

### LINKED ISSUE #1312: Toggle map and list on mobile
In response to #1187, #1208 introduced new buttons to hide the map or the results.
This is not working properly on mobile layout (and there is already a button to toggle between results and map).

@ginold @asaunier could adding class `hidden-xs` to the new buttons be enough to solve this?

brunobesson:
Hide buttons on mobile, they are of no use in this case and only introduce problems.

related to #1312

### DIFF
--- a/c2corg_ui/templates/helpers/list.html
+++ b/c2corg_ui/templates/helpers/list.html
@@ -10,12 +10,12 @@
 </%def>
 
 <%def name="toggle_map_btns()">\
-  <button class="btn btn-primary no-map-btn show-list" ng-show="noList"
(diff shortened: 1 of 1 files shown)


asaunier:
@brunobesson no big deal because I think it was not related to your PR, but when Travis is not OK we generally prefer not to merge the PR. I have asked @tsauerwein and @arnaud-morvan if they had a clue about the error Travis reported. Sorry for my confusing comment above.

brunobesson:
My bad, I thought you were aknowledging it.

PR #1314, opened 2017-01-06
Correct position of thumbnails in feed

brunobesson:
When there are 2 images, ensure correct position.

Before:
certif3

Now:
certif2


asaunier:
I have pushed a fix for the tests. Please rebase your branch so that it takes the change into account and makes Travis pass.

PR #1318, opened 2017-01-06
Icon quality change

stef74:
related to #1317

PR #1328, opened 2017-01-09
[WIP] Svg font quality

stef74:
#1318


brunobesson:
@stef74 I checked out your branch. Is your PR ready? There are too problems for me:

* the icons are not shown in the cards. The class is e.g. `quality draft` whee it should be `quality icon-draft`
* the SVG fonts do not correspond to quality icons. For me, draft looks like base camp (?), empty like an access, etc.

Am I missing something?

stef74:
I have not tested (yesterday take time to install a proper environnement
... too difficult to test in browser directly).

I'm not mastering this topic ... i'm trying to do as otrher svg icon exist
(activites, license etc ...)

I need to change the name of icon to be named icon-xxxx ? in less file is
icon-xxxx .... so I need to change the less or the html ... more simple if
not modify the html.

For the confusing with other icon I supect if I use the same content key of
other icon.

.icon-empty:before { content: "\e900"; } .icon-fine:before { content: "\e901
"; } \e900 is allready use ... I don't know how to change it (is generated
by iconmoon)





2017-01-10 9:24 GMT+01:00 Bruno Besson <notifications@github.com>:

brunobesson:
@stef74 I would like to help, but... are you testing what you are doing? I can't follow with your changes, they seem quite inconsistent to me, sorry.
It's not working at all when I check out your branch.

How can I help you?

stef74:
1st point now can be ok ...

For the second my understanding is in html resultat we need to have : 
<div class="quality ng-scope **icon-**draft" .....

So I need to modify the htlm link to card for add "icon-" before quality, It is probably here : 

<div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
           uib-tooltip="{{'quality'| translate}} : {{doc['quality']| translate}}" tooltip-append-to-body="true"></div>

but I don't know where ... here ? ng-class="**icon-**::doc['quality']" ?

stef74:
@brunobesson I have just recreated the font with icomoon (with new "content" to avoid conflict with other icon user same "content").

less/quality.less
```diff
@@ -0,0 +1,42 @@
+@font-face {
+  font-family: 'icomoon';
+  src:  url('../img/fonts/icomoon.eot?r9fnjw');
```
asaunier: Are you sure of this filenames?
The files are rather named ``quality.oet``, ``quality.ttf``, etc.

asaunier:
I think there is a mistake in the font filenames.
At least I can no longer see the quality icons when testing in my instance, whereas my console reports not found ``icomoon.*`` files.

Your code (even though the commits could be squashed) seems consistent with what is done for social networks icons for instance.
Please note though that it adds files to be downloaded by the end users (all ``quality.*`` files).

That would probably make sense to group the existing SVG fonts (quality, social networks, etc.) so that we don't create a set of font files to download for every icon!?

asaunier:
> #1318

What is the relationship between the current PR and PR #1318 ?
Shall we close #1318 ?

brunobesson:
>What is the relationship between the current PR and PR #1318 ?

@stef74 first proposed #1318 for the same purpose, but he was using SVG sprites. Now he is trying to do the same with font icons. That's why I asked on #1318 if it should be closed.

>That would probably make sense to group the existing SVG fonts (quality, social networks, etc.) so that we don't create a set of font files to download for every icon!?

I agree :+1: But maybe this could/should be another PR, once this one is ready?

stef74:
I test tonight.

For regrouping font I propose to do on other PR ... and it probably and good idea to regroup all icon in one font currently we have : activities,quality,other,license,social,waypoints

stef74:
After hard work ...

capture d ecran 2017-01-11 a 21 20 11

Color is probably the default ...

I'm not sure is godd way I have use so I commit only for outing ... if it's I modify the other template after.

asaunier:
By the way, are you sure those quality icons are explicit enough. When I look at the screen shot I see a small green (why green?) spot of color in the bottom left corner of the cards. That does not seem very obvious what it is. In addition it seems a bit too small, isn't it?

stef74:
@asaunier I have implemented the icon given ... 
The green is probably the default color, but I can change (it seem grey is validated)

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -45,7 +45,7 @@
       </div>
     </div>
 
-    <div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
+    <div class="quality {{'icon'+ doc['quality']}}" ng-if="::doc['quality']"
```
stef74: Angular is completely new for me ... I have reproduce what is use for other icon (ng class is not use except for quality) Ok so I move 'icon' to ng-class something like that : ng-class ="'icon'+::doc['quality']" 
I'm not sure of the syntax ...

asaunier: I have tested in my instance. At least the generated HTML seems correct:
```
<div class="quality iconmedium" ng-if="::doc['quality']" uib-tooltip="Complétude : moyen" tooltip-append-to-body="true"> </div>
```

But the ``-`` is missing in the class name (should be ``icon-medium``). At the moment quality icons are not shown because of this typo.

Using ``ng-class`` is probably better though. But I can't spend time investigating now.

I think it is important that you carefully test your changes before updating the PR.

stef74:
@asaunier @brunobesson I don't found the right syntax to use ng-class to append 'icon-' test many case but not work.

c2corg_ui/static/js/map/map.js
```diff
@@ -451,7 +451,7 @@ app.MapController.prototype.createPointStyle_ = function(feature, resolution) {
```
brunobesson: Why is your PR including stuff related to map icons? I think you mixed different jobs?

stef74: Yes it's a mistake ... I have do some test without sucess and commit too quickly. I'm not able to revert ... i have take time to found how do but ...
I have plan to do proper branch when this one work.

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -45,8 +45,13 @@
...
     </div>
 
-    <div class="quality" ng-if="::doc['quality']" ng-class="::doc['quality']"
+<<<<<<< HEAD
+    <div class="quality" ng-if="::doc['quality']" ng-class="'icon-'+::doc['quality']"
+	 uib-tooltip="{{'quality' | translate}} : {{doc['quality'] | translate}}" tooltip-append-to-body="true">
+=======
+    <div class="quality {{'icon'+ doc['quality']}}" ng-if="::doc['quality']" 
          uib-tooltip="{{'quality' | translate}} : {{doc['quality'] | translate}}" tooltip-append-to-body="true">
+>>>>>>> c70cc577b83813ab84a5611c84760ef2709f9098
```
dkocich: you have an unresolved conflict here

stef74: Yes he bad copy paste but remove localy , but not work ...

c2corg_ui/static/js/map/map.js
stef74: Yes I know ... I need to clean the branch, but now priority is to fix, after I clean ...

stef74:
I'm not able to generate proper icon font ... We have also unicode char ○ ◔ ◑ ◕ ● ... but I don't know how to show unicode base on value of doc['quality']

brunobesson:
@stef74 can we close your PR? I have merged #1391 which does the same.

PR #1329, opened 2017-01-09
Reorder fields in outing

brunobesson:
New ordering:

* route description
* conditions
* avalanches
* timing
* access
* access comment
* hut comment
* personal comments


asaunier:
Looks good, but I think you have forgotten to update the fields order in the preview as well: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/outing/preview.html#L45-L51

brunobesson:
Good catch! I didn't think of it...
I will update my PR according to your comments.

brunobesson:
@asaunier I should be OK now. I tested locally, so a quick syntax check should be enough for review.

PR #1330, opened 2017-01-09
Handle different cases with summary and description

brunobesson:
When there is a summary but no description, display the summary as if it was the description

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -462,16 +462,18 @@
       <span class="lead">
         <div id="document-description" class="collapse in">
           % if locale.get('summary'):
-            <summary class="document-summary">
-              <label translate>summary</label><br>
+            % if locale.get('description'):
+              <summary class="document-summary">
+                <label translate>summary</label><br>
+                ${show_attr(locale, 'summary')}
+              </summary>
+            % else:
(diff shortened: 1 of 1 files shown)


asaunier:
Looks good, but isn't it a bit confusing to make the summary appear like the description when there is no description? Then you don't really understand the difference between those 2 fields.
It would made more sense that users move the summary to the description by hand, wouldn't it?

brunobesson:
I understand your point. The problem is, we currently have specifications on the association side e.g. for climbing sites.
They state that a pitch created for a climbing site should have a short description, in the summary field (https://www.camptocamp.org/articles/822764/fr/saisir-une-sortie-couenne-ou-bloc-sur-la-v6).

Thus we have a lot (or will have) of routes with a short summary and the text "No description for..." which if find ugly. I was trying to address this. I assume @fbunoz would like in the future to copy the content of the summary of each pitch into the climbing site.

On the other side, are there many routes other than pitches that have a summary but no description? I haven't seen it happen since V6 is online.

asaunier:
OK, up to you. Feel free to merge the PR then :)

brunobesson:
I don't want to impose anything. It was pretty obvious to me, but you having another point of view means we should gather more opinions. @stef74 @mgerbaux @desnoes (and maybe others?)

desnoes:
you can merge

Issue #1143, opened 2016-12-08
GPX upload error


stef74:
> J'ai essayer de rentrer une sortie et au moment de valider le tout : please consult the server logs : unexpected error
> 
> J'ai essayer de tout refaire et la fin est la même...

Currently 3 outing was created without issue.

tsauerwein:
In the logs is this error: https://github.com/c2corg/v6_api/issues/580

Would you have more information about what data was provided to create the outing? Especially what was used as geometry?

stef74:
> Entre 11h15 et 11h40 je dirai, j'ai fait plusieurs essais...
> 
> J'ai finalement réussi à remettre la sortie avec le minimum d'info pour voir ce qui bloque, et ça semble être la trace GPS... Pourtant il y a déjà une autre sortie avec trace GPS...
>

stef74:
That confirm geometry is the issue ... ;-)

tsauerwein:
Could we get this GPX file to test?

stef74:
[2016-12-02 Pic Blanc du Galibier.zip](https://github.com/c2corg/v6_ui/files/639177/2016-12-02.Pic.Blanc.du.Galibier.zip)

stef74:
I have tested to add a GPX fil on old outing ... nos issue with this one : 
[chavasse.zip](https://github.com/c2corg/v6_ui/files/639251/chavasse.zip)

jpicheral:
When this file [CimeGollien.zip](https://github.com/c2corg/v6_ui/files/689575/CimeGollien.zip) is uploaded to this outing https://www.camptocamp.org/outings/829358/it/cime-de-gollien-spalla-ovest-per-il-col-gollien the track is displayed on map on the editing page, but it is not displayed on the map of the main outing page

asaunier:
@jpicheral Thanks for the example GPX file!

I notice this file contains 2 waypoints (pretty far away by the way) in addition to the expected ``<trk>`` element:
```
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<gpx xmlns="http://www.topografix.com/GPX/1/1" creator="MapSource 6.15.11" version="1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd">

  <metadata>
    <link href="http://www.garmin.com">
[...]

PR #1331, opened 2017-01-10
Only take tracks into account when uploading GPX files

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1143

By the way I have moved all the editing files to a dedicated directory (as for map and search files for instance).


asaunier:
@tsauerwein @dkocich @arnaud-morvan Could you review this PR?
I would like to prepare a new release this afternoon and have the demo updated tonight.
Thanks!

c2corg_ui/static/js/map/map.js
```diff
@@ -636,11 +636,23 @@ app.MapController.prototype.handleEditModelChange_ = function(event, data) {
 
 
 /**
+ * @param {ol.Feature} feature
+ * @return {boolean}
+ * @private
+ */
+app.MapController.prototype.isLineFeature_ = function(feature) {
```
tsauerwein: Do you maybe want to move this function into `app.utils`? It's the same as you used for the download?

asaunier: Indeed. Since the duplicated is very small, I was not sure it was worth moving it to ``app.utils``. Do you recommend to do so?

tsauerwein: Why not?

PR #1332, opened 2017-01-10
Wind roses simplification

brunobesson:
* Use a wind roses similar to that of MeteoFrance avalanche bulletins. This makes it easier to (un)select
* No pointer cursor when the widget is not interactive
* Max size is 150px instead of 200px

certif4


asaunier:
I will test tonight but at first I really like the new graphic!

desnoes:
very nice! adopted ! I commit on the master branch :)

PR #1333, opened 2017-01-10
Add GPX/KML download buttons in outings/routes detail views

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/678

The buttons only appear if the route/outing has a ``geom_detail`` (aka GPS track).

capture du 2017-01-10 20-49-47

capture du 2017-01-10 20-50-48

### DIFF
--- a/c2corg_ui/static/js/trackdownload.js
+++ b/c2corg_ui/static/js/trackdownload.js
@@ -1,0 +1,105 @@
+goog.provide('app.TrackDownloadController');
+goog.provide('app.trackDownloadDirective');
+
+goog.require('app');
+goog.require('ngeo.Download');
+goog.require('ol.format.GeoJSON');
+goog.require('ol.format.GPX');
+goog.require('ol.format.KML');
+
+
+/**
+ * This directive is used to display a track download button.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.trackDownloadDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppTrackDownloadController',
+    controllerAs: 'trackCtrl',
+    templateUrl: '/static/partials/trackdownload.html'
+  };
+};
+
+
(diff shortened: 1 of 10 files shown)


c2corg_ui/static/js/trackdownload.js
```diff
@@ -0,0 +1,105 @@
...
+
+app.module.directive('appTrackDownload', app.trackDownloadDirective);
+
+
+/**
+ * @param {ngeo.Download} ngeoDownload ngeo download service.
+ * @param {?GeoJSONFeatureCollection} mapFeatureCollection FeatureCollection of
+ *    features shown on the map.
+ * @constructor
+ * @export
```
tsauerwein: This `@export` should not be needed.

c2corg_ui/static/partials/trackdownload.html
```diff
@@ -0,0 +1,2 @@
+<button class="btn btn-primary" ng-click="trackCtrl.downloadGpx()">GPX</button>
```
tsauerwein: I would rather use a link instead of a button.

brunobesson: I was going to make the same comment. The button is too big IMHO, a link would be more consistent with the content around.

asaunier: Links are not OK on mobile device. And it has to be quite obvious. I prefer a button.

I agree that improvements such as styling (color, icon, size) or placement (perhaps it's not the best place to put the buttons?) could be done. But I think it can be done later. I prefer to merge as is since the PR looks good to both of you aside of this :)

I will remove the extra @export and merge when Travis is OK. Thanks for the review/comments!

asaunier: Actually I will remove the extra ``@export`` in another PR because there are a bunch of other controllers that also have it.

brunobesson:
@asaunier I found an issue: the buttons appear in preview, but the line shouldn't be shown.
I opened another issue #1338

certif6

PR #1334, opened 2017-01-11
Wind roses coloring

brunobesson:
Use C2C orange only when editing or for filter (i.e. when interactive).
Use bright green in view page, to be consistent with other items.

Also correct a small typo in SVG file.

(I should have noticed this before merging previous PR :wink:)


asaunier:
By the way I notice the windrose is missing in the route previews. Perhaps something to do in a later PR :P

brunobesson:
>windrose is missing in the route previews

I created #1335 for this.

stef74:
could you please create an issue pour each pr ... its simple to know how issue need to be tested (with correct label).

PR #1346, opened 2017-01-12
Break words in topoguide bullets appropriately

brunobesson:
Use space for breaking, unless the word is too long. Then it can be cut too.

certif7

### DIFF
--- a/less/topoguide.less
+++ b/less/topoguide.less
@@ -41,6 +41,7 @@
     transition: .2s;
     background-color: rgba(117, 117, 117, 0.44);
     margin-bottom: 15px;
+    padding: 2px;
     &:hover {
       cursor: pointer;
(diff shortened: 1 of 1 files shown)


asaunier:
Looks good to me but it does seem to work in Chromium (OK in FF). No idea for other browsers.

brunobesson:
For me it is working on chrome 55. 
Caniuse also states that it is supported<sup>1</sup>: http://caniuse.com/#search=overflow-wrap
The problem may be that there is no padding, thus it seem not cut, but the browser decided there is no need to. We should add some small padding to the `<a>` tag around. Can you check this in chromium?

<sup>1</sup> I had checked support on MDN. Seems like we should use `word-wrap` for compatibility. I will update PR accordingly.

PR #1349, opened 2017-01-12
Gpx download design

brunobesson:
* use smaller buttons
* on mobile, clicking the GPX or KML button won't close the corresponding section too

certif9
certif10

### DIFF
--- a/c2corg_ui/static/js/trackdownload.js
+++ b/c2corg_ui/static/js/trackdownload.js
@@ -72,18 +72,22 @@ app.TrackDownloadController.prototype.downloadFeatures_ = function(
 
 
 /**
+ * @param {goog.events.Event | jQuery.Event} event
  * @export
  */
-app.TrackDownloadController.prototype.downloadGpx = function() {
+app.TrackDownloadController.prototype.downloadGpx = function(event) {
+  event.stopPropagation();
   this.downloadFeatures_(
     new ol.format.GPX(), '.gpx', 'application/gpx+xml');
 };
 
 
 /**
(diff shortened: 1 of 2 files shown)


c2corg_ui/static/js/trackdownload.js
```diff
@@ -74,7 +74,8 @@ app.TrackDownloadController.prototype.downloadFeatures_ = function(
 /**
  * @export
  */
-app.TrackDownloadController.prototype.downloadGpx = function() {
+app.TrackDownloadController.prototype.downloadGpx = function($event) {
```
asaunier: The JS doc of the function should be updated to make this new argument appear.
Perhaps
```
* @param {goog.events.Event | jQuery.Event} $event
```
See for instance https://github.com/c2corg/v6_ui/blob/5cd68ec0a539e281b4fe5aac27b6a07ce6cd8370/c2corg_ui/static/js/viewdetails.js#L445-L453

You can also use a more generic type like
```
* @param {Object} $event
```
but that's not a good practise to use too generic types if you can use a more precise one.

I would also recommend to name the argument ``event`` instead of ``$event`` in the function definition (as for other methods with an event argument).

asaunier:
Else the changes look very good to me, thanks!

c2corg_ui/static/js/trackdownload.js
brunobesson: You're right. I forgot to clean everything once it was working. I'll update the commit accordingly.

brunobesson:
@asaunier if you could have a quick look, I took your feedback into account.

Issue #1350, opened 2017-01-12
Impossible to edit some books


desnoes:
It is impossible to save changes of this edited book: https://www.camptocamp.org/books/edit/114465/fr
I don't see why and when i click on the save button, there is no indication ('Que manque t'il ?')

dkocich:
I set [here](https://github.com/c2corg/v6_ui/blob/4c851e7fab7bd7b5f896798c149fe2e456aaff03/c2corg_ui/templates/book/edit.html#L133) restriction to the length of 9-13 characters to ISBN field and I did not realise that there are also "-" inside (it is longer immediately when edit open). When I make it shorter the save/preview buttons are active again.

brunobesson:
Maybe we could use a more specific validation, e.g. inspired from https://github.com/janantala/angular-isbn?

asaunier:
@dkocich OK thanks for the explanation. Yes ISBN may contain dashes: https://en.wikipedia.org/wiki/International_Standard_Book_Number
What about updating the ``maxlength`` to 17 (13+4 dashes)?

> Maybe we could use a more specific validation, e.g. inspired from https://github.com/janantala/angular-isbn?

@brunobesson That's an option. But is it really worth adding some code to download for this minor validation? It depends if we really need strict ISBN formatting (for bookstore links?).

brunobesson:
>But is it really worth adding some code to download for this minor validation?

Nah, not really. Allowing 17 characters is enough for me.

brunobesson:
It is now working. Closing.

PR #1355, opened 2017-01-13
Fix book edit (ISBN length)

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1350


asaunier:
I have edited your comment to replace "Close" by "Related to" because we don't want to autoclose the issue when merging the PR. Please tag the issue as "ready for demo" instead.

PR #1359, opened 2017-01-13
Do no print GPX download buttons

brunobesson:
Simple and easy!


stef74:
Do you respect the graphical chart ? have your inform association of your change ? Have you use CERFA 20279898 ? 
:+1:

PR #1360, opened 2017-01-13
Download buttons design

brunobesson:
* use bright green for consistency
* no line break before buttons

certif14


asaunier:
> no line break before buttons

Are you sure it works when translated in French, German, etc.?
I had added a line break to make sure we had no button at the end of the first line whereas the other one was on the following line.

brunobesson:
>I had added a line break to make sure we had no button at the end of the first line whereas the other one was on the following line.

Well, let's rather ensure both buttons are always displayed together and line break happens before :wink: (see latest commit)

PR #1361, opened 2017-01-13
Similar visuals for no outing button and new comments button

brunobesson:
This applies when there is no outing for a route yet.

* adjust font size
* center button

Before:
certif16

Now:
certif15

### DIFF
--- a/c2corg_ui/templates/route/view.html
+++ b/c2corg_ui/templates/route/view.html
@@ -146,10 +146,12 @@ other_langs, missing_langs = get_lang_lists(route, lang)
           </div>
           % if len(route['associations']['recent_outings']['documents']) == 0:
             <div class="add-outing">
-              <p class="text-center"><strong translate>no outings yet?</strong><p>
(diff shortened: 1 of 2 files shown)


asaunier:
Thanks for addressing this problem, I was not really happy with the existing layout for this button.
I wonder if this change is enough though.

Shouldn't we rather always show the "Outings" section (shown when there actually are outings) which already contains an "add outing to this route" button and remove the button?
That would be more consistent.

By the way why showing this outing button here when there is no outing and a bit further (in the outing section) when there are outings?

brunobesson:
>That would be more consistent.

Sure, I would agree with that. But this PR wasn't thinking that much, I simply wanted to have a consistent button L&F.
If e.g. @desnoes or @loicperrin agree with this, we could discard this PR and create another one solving the issue the way you propose.

loicperrin:
I agree with alex proposal.

desnoes:
>I agree with alex proposal

idem

brunobesson:
Replaced by #1367

PR #1365, opened 2017-01-14
Increase number of column (associated list) and font-size reduction

stef74:
To be homogeneous:
- increase number of column in associated list (up to 6 .... 7 rendering too small card)
- ajuste font size in card to be homogeneous with search

capture d ecran 2017-01-14 a 12 10 51

### DIFF
--- a/less/lists.less
+++ b/less/lists.less
@@ -381,12 +381,20 @@
 
 .section.associations, .view-details-associations, .list {
   .list-item {
-    flex-basis: 48%;
-    @media (min-width: @large-width){
-      flex-basis: 32%;
+    @media (min-width: 2400px) {
+      flex-basis: 16% !important;
     }
-    @media (min-width: 2100px) {
-      flex-basis: 24%;
+    @media (max-width: 2400px) {
+      flex-basis: 19% !important;
+    }
+    @media (max-width: 2100px) {
+      flex-basis: 24% !important;
+    }
+    @media (max-width: @large-width) {
+      flex-basis: 32% !important;
+    }
+    @media (max-width: @tablet-max) {
+      flex-basis: 48% !important;
     }
     @media (max-width: 900px) {
       flex-basis: 100% !important;
--- a/less/lists.less
+++ b/less/lists.less
@@ -111,7 +111,7 @@
     margin: auto;
     height: 100%;
     width: 100%;
-    font-size: 90%;
+    font-size: 0.84em;
     text-decoration: none;
     color: #919191;
     font-weight: normal;
@@ -381,7 +381,6 @@
 
 .section.associations, .view-details-associations, .list {
   .list-item {
-    font-size: 90%;
     @media (min-width: 2400px) {
       flex-basis: 16% !important;
     }
--- a/less/lists.less
+++ b/less/lists.less
@@ -111,7 +111,7 @@
     margin: auto;
     height: 100%;
     width: 100%;
-    font-size: 0.84em;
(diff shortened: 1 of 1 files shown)


stef74:
This PR can be reviewed

less/lists.less
```diff
@@ -381,14 +381,23 @@
 
 .section.associations, .view-details-associations, .list {
   .list-item {
-    flex-basis: 48%;
-    @media (min-width: @large-width){
-      flex-basis: 32%;
+    font-size: 90%;
```
asaunier: Are ``%`` the good way to set a font size? I though ``em`` were recommended instead.

asaunier:
I prefer to let @brunobesson make the review because I don't know much about CSS.

I notice that the fonts in any cards are quite smaller with this change than in the current demo. I prefer the current size, I think the cards content should have a font-size consistent with the other content's font sizes.

less/lists.less
stef74: sometimes in current you have % sometimes em ...

stef74:
@asaunier could you precise is card of search ... or associated doc ? currently size are not consistant beetween two type of card.

stef74:
I have found the solution to have exact size so I confirm currently : 
12.6px in search
13.6px in docs

I restes my branch tonight to be homogeneous

stef74:
I have fix an tested now font-size is similar. I have also reuse font-size property (not created a new one)

less/lists.less
```diff
@@ -381,14 +381,22 @@
 
 .section.associations, .view-details-associations, .list {
   .list-item {
-    flex-basis: 48%;
-    @media (min-width: @large-width){
-      flex-basis: 32%;
+    @media (min-width: 2400px) {
+      flex-basis: 16% !important;
```
brunobesson: Why are you using `!important`? It works as well without. We should avoid using this property if possible.

I think you did the same in your previous PR. This could be corrected too.

less/lists.less
```diff
@@ -111,7 +111,7 @@
     margin: auto;
     height: 100%;
     width: 100%;
-    font-size: 90%;
+    font-size: 0.84em;
```
brunobesson: If the font size is set to `90%` elsewhere, then let's be homogenous by using the same value (in a perfect world, this would even be a less variable).
I suggest you also use a em-based font size for the search cards. Morover, I would simpify to 0.9em, rather than 0.84.

less/lists.less
stef74: simple copy past ... I remove

less/lists.less
stef74: yes but is not same size ... probably many imbrication of font-size

less/lists.less
```diff
@@ -111,7 +111,7 @@
     margin: auto;
     height: 100%;
     width: 100%;
-    font-size: 0.84em;
+    font-size: 0.90em;
```
besson-genostar: Could be written `0.9em` instead :wink: 
So you decided not to change the other `font-size: 90%` value? It's not that important.

We could also use rems for solving nested sizing, but this would rather be a global rework of CSS.

stef74: Font size need a reproper rework ... I have also plan to rework mediaqueries properly and avoid some test, I can try to do on font-size

brunobesson:
Thanks, you're now free to squash your PR!

stef74:
I'm waiting travis check after I squash and merge

PR #1366, opened 2017-01-16
Use CSS to style snow conditions

brunobesson:
Do not duplicate code

Apart from this, I would like to correct styling:
* we use a color for titles that isn't used anywhere else
* the rounding is different from other sections

But this could be another PR.

### DIFF
--- a/c2corg_ui/templates/outing/helpers/view.html
+++ b/c2corg_ui/templates/outing/helpers/view.html
@@ -256,46 +256,6 @@
 
 <%def name="create_conditions_levels_tables(conditions)">\
   <label translate>conditions_levels</label>
-  <div class="show-only-mobile conditions-levels">
-    % for condition in conditions:
-      <%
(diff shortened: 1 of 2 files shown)


asaunier:
Could you explain more precisely your changes? I don't understand why you have removed all this "conditions levels" code in the template helper.

brunobesson:
Sorry, I thought it was obvious. Currently, 2 sections are generated for snow conditions. One is a table displayed on tablets and desktops, hidden on mobile. The other one is a div displayed on mobile, hidden otherwise.

Thus, we have duplicated content, and half of it is always hidden.

This PR uses CSS on a single content to adapt between mobile and others. Basically, we apply a specific display to the tables, and use data-attributes to repeat the title of each section.

brunobesson:
Well, the alternate CSS for mobile only uses simple divs. It is quite safe. Merging, thus!

Issue #1005, opened 2016-11-25
On route, links to material article are not migrated


HSERGERAERT:
**on v5**, route http://www.camptocamp.org/routes/784566/fr/la-pyramide-traversee-toit-pyramide-petite-traversee
The route has a connection, in material section,  to the article "material" http://www.camptocamp.org/articles/185384/fr/le-contenu-du-sac-alpinisme-rocheux-de-f-a-ad

**on v6** the article exists 
https://www.demov6.camptocamp.org/articles/185384/fr/le-contenu-du-sac-alpinisme-rocheux-de-f-a-ad

but the route has no connection to the article
https://www.demov6.camptocamp.org/routes/784566/fr/la-pyramide-traversee-toit-pyramide-petite-traversee, only "sangles" was migrated

tsauerwein:
When was the association between the route and article created? Because in the v5 database that we made the migration with, this article only has associations to 4 other articles (which are shown).

gottferdom:
I don't think the association exists in the database. In v5, the article link is displayed depending on the route characteristics (activity, difficulty, ...). So this isn't a migration problem.

tsauerwein:
> In v5, the article link is displayed depending on the route characteristics (activity, difficulty, ...).

Can you give more information about how this works?

gottferdom:
The code is here on v5: https://github.com/c2corg/camptocamp.org/blob/72e777075f2e6260b1d2ef563ab0f90c383038ba/apps/frontend/modules/routes/templates/_i18n.php

Basically, if some conditions are met, a link to a specific article is added in gear section. There are several articles (https://github.com/c2corg/camptocamp.org/blob/9a44a3b50ab2681a5e27435e1469c3911a192f6d/deployment/conf.ini-dist#L81)

To translate the logic:
* if activity is snowshoeing or skiing, add article on skitouring gear (id 183333)
* if activity is snow/ice/mixed and global rating <= PD+, add article for easy snow/ice/mixed (185750)
* if activity is mountain climbing and global rating <= AD, then 185384
* if rock climbing and equipment is P1 or P1+ then 185384
* if rock climbing and equipment is P2 or more and global rating <= AD and there isn't also the mountain climbing activity on the route, then 183332
* if ice climbing, then 194479
* if (skitouring or snow/ice/mixed or moutain climbing or snowshoeing) and glacier route is checked, then 185750 (but at anchor #glacier-gear)

In addition to this, there are some strings for each language to display the link.

[...]

PR #1371, opened 2017-01-16
Add gear-specific articles depending on ratings

brunobesson:
References #1005

Currently, we do not respect the anchor for glacier gear (it's not yet possible)

I have 2 questions though:

* is there a way to use comparisons rather than testing if value is in list for ratings?
* can you check the i18n keys are consistent?

### DIFF
--- a/c2corg_ui/templates/route/helpers/view.html
+++ b/c2corg_ui/templates/route/helpers/view.html
@@ -1,5 +1,8 @@
+<%!
+from c2corg_ui.templates.utils import get_route_gear_articles
+%>
 <%namespace file="../../helpers/view.html" import="get_global_rating, get_hiking_mtb_rating, get_skitouring_rating,
-    get_document_min_max, get_document_up_down, show_areas, show_maps, show_track_download"/>
+    get_document_min_max, get_document_up_down, show_areas, show_maps, show_track_download, show_attr"/>
 <%namespace file="../../helpers/orientations.svg" import="show_orientations"/>
 
 ## LOCATION
@@ -252,6 +255,35 @@
         <span translate class="value-title">glacier_gear</span>:
         <span class="value" x-translate>${route.get('glacier_gear')}</span>
       </span>
+    </div>
+  % endif
+</%def>
+
+## SPECIFIC GEAR
+<%def name="get_route_gear(route, locale)">\
+  <%
+    gear_articles = get_route_gear_articles(route)
+  %>
+  % if locale['gear'] or gear_articles:
+    <div class="view-details-info col-xs-12">
+      <h3 class="heading show-phone"
+          ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#${'gear'}">
+        <span x-translate>gear</span><span class="glyphicon glyphicon-menu-down"></span>
+      </h3>
+      <section class="collapse in" id="gear" aria-expanded="true">
+        <div class="lead">
+          % if gear_articles:
+            <ul>
+            % for article_id, name in gear_articles.items():
+              <li><a href="${request.route_path('articles_view_id', id=article_id)}" x-translate>${name}</a></li>
+            % endfor
+            </ul>
+          % endif
+          % if locale['gear']:
+            ${show_attr(locale, 'gear')}
+          % endif
+        </div>
+      </section>
     </div>
   % endif
 </%def>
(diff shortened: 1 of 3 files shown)


brunobesson:
Anyone for a review? :pray:

asaunier:
I had not realized this PR needed a review since the [WIP] prefix is still on :)

brunobesson:
Well. We had a similar discussion before. I don't consider it finished because I am awaiting feedback, but appart from this it is already working. What should I do in that case? :wink:

asaunier:
OK I'll have a look at home.

c2corg_ui/templates/utils/__init__.py
```diff
@@ -82,3 +82,30 @@ def has_associations(doc):
         a.get('all_routes') or a.get('routes') or \
         a.get('recent_outings') or a.get('outings') or \
         a.get('articles') or a.get('xreports') or a.get('books')
+
+
+def get_route_gear_articles(route):
+    articles = []
+    if route.get('activities'):
+        if ('snowshoeing' in route.get('activities') or
```
asaunier: Since ``route.get('activities')`` is used in all conditions, I recommend to set a variable instead ``activities = route.get('activities')``.

Then you can write:
```
if 'snowshoeing' in activities or 'skitouring' in activities:
```

A very pythonic way to write this condition would be to use list comprehension:
```
if any(activity in activities for activity in ['snowshoeing', 'skitouring']):
```
but perhaps it doesn't really bring a readability gain here?
For ``any()`` see https://docs.python.org/3/library/functions.html#any (the "contrary function" is ``all()``).

c2corg_ui/templates/route/helpers/view.html
```diff
@@ -255,3 +258,32 @@
...
+## SPECIFIC GEAR
+<%def name="get_route_gear(route, locale)">\
+  <%
+    gear_articles = get_route_gear_articles(route)
+  %>
+  % if locale['gear'] or gear_articles:
+    <div class="view-details-info col-xs-12">
+      <h3 class="heading show-phone"
+          ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#${'gear'}">
+        <span x-translate>${'gear'}</span><span class="glyphicon glyphicon-menu-down"></span>
```
brunobesson: Copy paste from base `get_document_locale_text` which wasn't cleaned up properly thereafter. Thx

c2corg_ui/templates/utils/__init__.py
brunobesson: I tried to follow the recommendations here: https://www.python.org/dev/peps/pep-0008/
I'll adapt the formatting.

brunobesson:
@asaunier thx for the review. I have made adjustments following your comments.

c2corg_ui/templates/route/helpers/view.html
```diff
@@ -255,3 +258,32 @@
...
+  <%
+    gear_articles = get_route_gear_articles(route)
+  %>
+  % if locale['gear'] or gear_articles:
+    <div class="view-details-info col-xs-12">
+      <h3 class="heading show-phone"
+          ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#${'gear'}">
+        <span x-translate>gear</span><span class="glyphicon glyphicon-menu-down"></span>
+      </h3>
+      <section class="collapse in" id="${'gear'}" aria-expanded="true">
```
asaunier: Is this ``${'gear'}`` correct? Ditto 3 lines above?

brunobesson: Nope. I'll correct it.

c2corg_ui/templates/route/helpers/view.html
```diff
@@ -255,3 +258,32 @@
...
+
+## SPECIFIC GEAR
+<%def name="get_route_gear(route, locale)">\
+  <%
+    gear_articles = get_route_gear_articles(route)
+  %>
+  % if locale['gear'] or gear_articles:
+    <div class="view-details-info col-xs-12">
+      <h3 class="heading show-phone"
+          ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#${'gear'}">
```
brunobesson: This should be simplified to  `data-target="#gear"` too.

Issue #1376, opened 2017-01-17
Media queries optim and fontsize


stef74:
They have many variable or constant in less file for mediaqueries ... not all are homogeneous.
Same thing for font-size.

I propose to refactor this two elements

stef74:
For media queries I propose to define 3 major class and more sub class (needed for column in card)

Mobile : [0,700px]
=> Nos sub class
Tablet : [701,1100px]
Desktop : [1100,*****]
=> SmallDesk : [1100,1700]
=> StandDesk : [1701,2400]
=> BigDesk[2401,****]

Probably need so test and adjust ...

stef74:
For font size is more complexe.

I check font-size in pixel for text only (icon is managed differently)

Standard reference seem to be 20.

https://github.com/c2corg/v6_ui/blob/3f91ced353404d1dc66312beacbff049d3a0ee08/less/images.less#L25
=> Why is 14 

https://github.com/c2corg/v6_ui/blob/c94ed70f73fb876271ec6f1c3cbbc862f8818ca9/less/helpers.less#L50
=> is a specific 

https://github.com/c2corg/v6_ui/blob/c94ed70f73fb876271ec6f1c3cbbc862f8818ca9/less/helpers.less#L115
=> is specific for datpicker

To be homogeneous why not define : 
- mobile 80% 
- tablet 90%
- Desktop 100%

Also to be homogenous we can define globally
- 14 px for text in homepage in Desktop (as is) probably define is the "standard" and for filter and block name in doc ("information)
- 12.6px for card (90% of standard) (as is) + for doc content (little reduction)

Icon need to be managed separatly

PR #1379, opened 2017-01-17
Rework font-size + color

stef74:
related #1376 

First rework on font-size : 
- simplify font-size (global approche for tablet/phone avoid some specific
- remove pixel usage

Your now two standard value : 
- 14px (in desktop) for text
- 0.9em for Card
And some exception (but less than previous)

Some rework is need for Icon ... but not in this PR

 Some size are little changed for : 
- Sometimes some texte are little toot big without very good reason
- Some value are very similar


less/c2corg_ui.less
```diff
@@ -33,8 +33,16 @@ html, body{
...
   background-color: @body-color;
+  @media @phone {
+    font-size: 0.9em;
+  }
+  @media @tablet {
+    font-size: 0.95em;
+  }
 }
 
+
```
brunobesson: why new empty lines?

stef74: to be fixe

stef74:
I have define a font-size based on pixel only in body, and so i Have converted old px value to em value.

brunobesson:
@stef74 what's the status of this PR? You do no seem to have made change since a week. Are you stil working on it?

stef74:
No I need to rework on it ... but probably need to rebase before

stef74:
I have tested on many page ... not observed regression concerning font size (tested in split screen with actual master ... and my branch)

stef74:
Ho my god ....

brunobesson:
:rofl:

stef74:
Nothing happen 😄

stef74:
Color are sometimes managed by naming ... so not redereing at the same on different browser / OS.

Replace by RGB code needed.

stef74:
I have to use variable ... work on it this rainy we.

stef74:
Now card have the same size in all page.
I have tested ... but if someone can checkout and test. 
Some optimization can bed added (all color are not managed by variable ... but we can do progresssivly)

stef74:
@brunobesson @asaunier ready for review. I squash after

asaunier:
I don't have an opinion about the CSS changes themselves. I don't know much about CSS.

Concerning the rendering in my Chrome 56, it looks nice (including cards) except that:
* the "filter on map extent" text on the advanced search map became unnecessarily bigger
* the font-size of document texts are now much smaller than cards contents or figures texts => that's not very consistent, font-sizes should be quite the same:

capture d ecran 2017-02-08 a 22 36 14

stef74:
@asaunier many issue have been fixed including font-size of profile feed.
Could you test again if is correct for you now

asaunier:
I don't have a strong opinion about this PR: most changes are quite light.

Document texts and are a bit smaller: why not.
On the other hand I have noticed the orange section titles were a bit bigger: that's not very useful and consistent.
Card texts and titles are a bit smaller, titles are a bit darker: why not.
Figures attribute titles are gray instead of black: why not.

I have noticed the simple search and association tool typed texts are now light grey instead of black:   black was better.

stef74:
According to you remark we propose this fix : 
- [ ] section title 
- [ ] filter on map extent (could you show a screenshoot ...)
- [ ] incrase size of button text in non connected homepage
- [ ] fix color in search tool

stef74:
Other issue on font-size

stef74:
Ok some issue are fixed

stef74:
@asaunier @brunobesson ok for reviewing

asaunier:
I have not tested the branch but feel free to merge if ready.

Issue #1368, opened 2017-01-16
Add "my outings" link in user's menu


brunobesson:
Some users have difficulties finding their outings list. Should we add another item to the current menu? Or do you think there are already too many items?

We could also think ahead and take statistics page: how will this one be positioned and accessed?

@desnoes could you ask the current task force you're leading?

stef74:
Which menu ?

brunobesson:
>Which menu ?

The user's menu. The one in blue you can access from the top right.

stef74:
Yes ... but it's not simple change ... filtering on user is not available throught search

brunobesson:
What? No. Producing a link to the user's outing page is easy (there is already this link on the user's profile page).

asaunier:
> Some users have difficulties finding their outings list. 

There's a big orange "show this user's outings" button at the top of the profile feed!?
Why would users see the menu you suggest if they cannot see this button in their profile?

@stef74 Yes, it's easy. It seems you have not seen this big orange button I was just talking about :P

brunobesson:
>There's a big orange "show this user's outings" button at the top of the profile feed!?
Why would users see the menu you suggest if they cannot see this button in their profile?

Apparently, it's not obvious to follow the path my profile to see your outings. The problem may not be that htey don't find the profile menu item or don't see the orange button then, but rather that it seems not logical to them to go this way.

asaunier:
That's how it was done in v5!?

stef74:
You are agree the link allready exist ... i don't remeber you have not only the feed in user profil

desnoes:
Maybe, the size font of the message 'Voir les sorties de cet utilisateur' could be increased.

stef74:
@brunobesson when you create and issue than you handle please add association label :-)

fbunoz:
> That's how it was done in v5!?

On V5, in the menu "My C2C", there was 3 items : 

- my outings
- my images
- my xreports

Then we were be able to acces to our outings list from any c2c page, in 1 clic.

asaunier:
OK, feel free to add such a link to the menu.

brunobesson:
OK, although translations have not been updated.

asaunier:
The translation for "My outings" was not available when the v6.0.7 image was created an hour ago. I made the translation in the meantime but it's too late for this release.

stef74:
but ok for prod ?

asaunier:
[...]

PR #1380, opened 2017-01-17
add menu item for user's outings

brunobesson:
Related to #1368, assuming everyone agrees to add this menu item.

@stef74 @asaunier Once this is merged, how do we deal with transifex items?


asaunier:
> Once this is merged, how do we deal with transifex items?

When the PR is merged, Travis will run the i18n extractor (including in the partials) and add the new strings to Transifex.

brunobesson:
>When the PR is merged, Travis will run the i18n extractor (including in the partials) and add the new strings to Transifex.

Nice! This also answers a question that would have arisen in #1005 (could this be a subliminal message :wink:?)

PR #1383, opened 2017-01-18
Update xreports menu link

asaunier:
@mgerbaux has asked that the menu entry for xreports points to the introduction article instead of the advanced search page.


asaunier:
Anyone for this easy review? :)

PR #1390, opened 2017-01-19
Validate LineStrings in uploaded MultiLineString geometry

dkocich:
Related to API issue https://github.com/c2corg/v6_api/issues/580
Fixes UI issue https://github.com/c2corg/v6_ui/issues/1143

Currently, upload for these files work fine - they can be displayed (view/edit pages), downloaded as GPX/KML on View page and opened using GEarth:

- [CimeGollien.zip](https://github.com/c2corg/v6_ui/files/689575/CimeGollien.zip) mentioned in https://github.com/c2corg/v6_ui/issues/1143#issuecomment-270874397
- [2016-12-02 Pic Blanc du Galibier.zip](https://github.com/c2corg/v6_ui/files/639177/2016-12-02.Pic.Blanc.du.Galibier.zip) mentioned in https://github.com/c2corg/v6_ui/issues/1143#issuecomment-265714053

### DIFF
--- a/c2corg_ui/static/js/gpxupload.js
+++ b/c2corg_ui/static/js/gpxupload.js
@@ -5,7 +5,9 @@ goog.require('app');
 /** @suppress {extraRequire} */
 goog.require('ngeo.filereaderDirective');
 goog.require('ol.format.GPX');
-
+goog.require('ol.Feature');
+goog.require('ol.geom.MultiLineString');
+goog.require('ol.geom.LineString');
 
 /**
  * This directive is used to display a GPX file upload button.
@@ -59,6 +61,56 @@ app.GpxUploadController = function($scope) {
 
 
 /**
+ *
+ * @param {Array.<ol.Feature>} features - invalid features that have to be checked
+ * @returns {Array.<ol.Feature>} features - this geometry is valid
+ * @private
+ */
+app.GpxUploadController.prototype.validatedFeatures_ = function(features) {
+  for (var i = 0; i < features.length; i++) {
+    /**
+     *
+     * @type {ol.geom.Geometry}
+     */
+    var geom = /**@type{ol.geom.Geometry}*/ (features[i].getGeometry());
+
+    /**
+     *
+     * @type {ol.geom.GeometryType}
+     */
+    var geomType = geom.getType();
+
+    if (geomType === 'MultiLineString') {
+      var multiLineString = /**@type{ol.geom.MultiLineString}*/ (geom);
+      /**
+       *
+       * @type {Array.<ol.geom.LineString>}
+       */
+      var lineStrings = multiLineString.getLineStrings();
+
+      for (var j = 0; j < lineStrings.length; j++) {
+        if (lineStrings[j].getCoordinates().length === 1) {
+          delete lineStrings[j];
+        }
+      }
+      lineStrings = lineStrings.filter(function(n) {
+        return n != undefined;
+      });
+
+      /**
+       *
+       * @type {ol.geom.MultiLineString}
+       */
+      var newMs = new ol.geom.MultiLineString([]);
+      newMs.setLineStrings(lineStrings);
+      features[i].setGeometry(newMs);
+    }
+  }
+  return features;
+};
+
+
+/**
  * @param {string} gpx GPX document.
  * @private
  */
@@ -68,7 +120,14 @@ app.GpxUploadController.prototype.importGpx_ = function(gpx) {
     featureProjection: 'EPSG:3857'
   });
   if (features.length > 0) {
-    this.scope_.$root.$emit('featuresUpload', features);
+    features = this.validatedFeatures_(features);
+    if (features) {
+      this.scope_.$root.$emit('featuresUpload', features);
+    } else {
+      if (goog.DEBUG) {
+        console.error('Fatal : failed to validate geometry');
+      }
+    }
   }
 };
 
--- a/c2corg_ui/static/js/gpxupload.js
+++ b/c2corg_ui/static/js/gpxupload.js
@@ -7,7 +7,6 @@ goog.require('ngeo.filereaderDirective');
 goog.require('ol.format.GPX');
 goog.require('ol.Feature');
 goog.require('ol.geom.MultiLineString');
-goog.require('ol.geom.LineString');
 
 /**
  * This directive is used to display a GPX file upload button.
@@ -68,24 +67,10 @@ app.GpxUploadController = function($scope) {
  */
 app.GpxUploadController.prototype.validatedFeatures_ = function(features) {
   for (var i = 0; i < features.length; i++) {
-    /**
-     *
-     * @type {ol.geom.Geometry}
-     */
     var geom = /**@type{ol.geom.Geometry}*/ (features[i].getGeometry());
 
-    /**
-     *
-     * @type {ol.geom.GeometryType}
-     */
-    var geomType = geom.getType();
-
-    if (geomType === 'MultiLineString') {
+    if (geom instanceof ol.geom.MultiLineString) {
       var multiLineString = /**@type{ol.geom.MultiLineString}*/ (geom);
-      /**
-       *
-       * @type {Array.<ol.geom.LineString>}
-       */
       var lineStrings = multiLineString.getLineStrings();
 
       for (var j = 0; j < lineStrings.length; j++) {
@@ -97,10 +82,6 @@ app.GpxUploadController.prototype.validatedFeatures_ = function(features) {
         return n != undefined;
       });
 
-      /**
-       *
-       * @type {ol.geom.MultiLineString}
-       */
       var newMs = new ol.geom.MultiLineString([]);
       newMs.setLineStrings(lineStrings);
       features[i].setGeometry(newMs);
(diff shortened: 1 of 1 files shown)


c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +61,60 @@ app.GpxUploadController = function($scope) {
...
+     *
+     * @type {ol.geom.GeometryType}
+     */
+    var geomType = geom.getType();
+
+    // TODO validate LineString? pass now - no problem we know about?
+    // if (geomType === 'LineString') {
+    // }
+
+    if (geomType === 'MultiLineString') {
```
asaunier: Why not rather
```
if (geom instanceof ol.geom.MultiLineString)
```
as in https://github.com/c2corg/v6_ui/blob/7b5e5e79fd7d5a7bd36c0e41cac7e1ee158613dd/c2corg_ui/static/js/map/simplify.js#L183
https://github.com/c2corg/v6_ui/blob/0bf5dd38cd30b5aa94debbbf61737d1bf68f6f75/c2corg_ui/static/js/utils.js#L312
https://github.com/c2corg/v6_ui/blob/ab1514dc90c4ecf6395faa0326a03e1f18fdbb80/c2corg_ui/static/js/editing/documentediting.js#L433
?

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +61,60 @@ app.GpxUploadController = function($scope) {
...
+      var multiLineString = /**@type{ol.geom.MultiLineString}*/ (geom);
```
asaunier: Do you really need to cast ``geom``?

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +61,60 @@ app.GpxUploadController = function($scope) {
...
+      /**
+       *
+       * @type {Array.<ol.geom.LineString>}
+       */
+      var lineStrings = multiLineString.getLineStrings();
+
+      for (var j = 0; j < lineStrings.length; j++) {
+        if (lineStrings[j].getCoordinates().length === 1) {
```
asaunier: Can't you use ``app.utils.isLineFeature`` instead of counting how many coords there are?

c2corg_ui/static/js/gpxupload.js
```diff
@@ -68,7 +124,14 @@ app.GpxUploadController.prototype.importGpx_ = function(gpx) {
     featureProjection: 'EPSG:3857'
   });
   if (features.length > 0) {
-    this.scope_.$root.$emit('featuresUpload', features);
+    features = this.validatedFeatures_(features);
+    if (features) {
+      this.scope_.$root.$emit('featuresUpload', features);
+    } else {
+      if (goog.DEBUG) {
+        console.error('Fatal : failed to validate geometry');
```
asaunier: I am not very found of those ``console.*``. Do we really need them? I think they cause warning when building the JS code.

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +61,60 @@ app.GpxUploadController = function($scope) {
...
+ * @param {Array.<ol.Feature>} features - invalid features that have to be checked
+ * @returns {Array.<ol.Feature>} features - this geometry is valid
+ * @private
+ */
+app.GpxUploadController.prototype.validatedFeatures_ = function(features) {
+  for (var i = 0; i < features.length; i++) {
+    /**
+     *
+     * @type {ol.geom.Geometry}
+     */
```
dkocich: Yes, otherwise compilation won't complete (it does not know about used functions of undefined types). I was trying to use ` * @suppress {checkTypes}` but that is not a good solution too.

c2corg_ui/static/js/gpxupload.js
dkocich: It can be probably removed. I was thinking that it can be easier to debug later when some other problems with images will appear...

tsauerwein:
> Shouldn't we do those geometries filtering on the API side instead? What if some other app uses send poorly-formatted geometries to the API? By doing the processing in the API we could make sure the issue is fixed once for all

We could have done this on the API side, but the problem is that we can't use Shapely to modify geometries because it doesn't support 3D/4D geometries. Also it is related to the way ol3 parses GPX files. ol3 generates invalid geometries, so it seems reasonable to fix these geometries with ol3.

c2corg_ui/static/js/gpxupload.js
tsauerwein: You could also remove the type definition for `geom`. Because by doing the cast, the compiler knows the cast of the variable.

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +61,50 @@ app.GpxUploadController = function($scope) {
...
+    var geom = /**@type{ol.geom.Geometry}*/ (features[i].getGeometry());
+
+    if (geom instanceof ol.geom.MultiLineString) {
+      var multiLineString = /**@type{ol.geom.MultiLineString}*/ (geom);
+      /**
+       *
+       * @type {Array.<ol.geom.LineString>}
+       */
+      var lineStrings = multiLineString.getLineStrings();
```
tsauerwein: This type could also be removed.

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +61,50 @@ app.GpxUploadController = function($scope) {
...
+      }
+      lineStrings = lineStrings.filter(function(n) {
+        return n != undefined;
+      });
+
+      /**
+       *
+       * @type {ol.geom.MultiLineString}
+       */
+      var newMs = new ol.geom.MultiLineString([]);
```
tsauerwein: This one as well.

c2corg_ui/static/js/gpxupload.js
tsauerwein: In general, it is good practice to type function parameters and class members (`this.xyz`). But we usually don't do it for local variables because the compiler can figure out the type itself.

c2corg_ui/static/js/gpxupload.js
dkocich: No, because it checks only that [if geom is instance of LS or MultiLS](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/utils.js#L306-L313) and doesn't validate anything else with coordinates. This fix should be applied only for MultiLS so I should leave it like it is.

c2corg_ui/static/js/gpxupload.js
```diff
@@ -68,7 +101,10 @@ app.GpxUploadController.prototype.importGpx_ = function(gpx) {
     featureProjection: 'EPSG:3857'
   });
   if (features.length > 0) {
-    this.scope_.$root.$emit('featuresUpload', features);
+    features = this.validatedFeatures_(features);
```
asaunier: What about moving this line above ``if (features.length > 0) {``?
This would avoid the second ``if (features) {``.

By the way are you sure ``if (features) {`` is enough? An empty array evaluates to ``true``:
```
> [] ? 'true': 'false'
"true"
```
I then recommend:
```
features = this.validatedFeatures_(features);
if (features.length) {
  this.scope_.$root.$emit('featuresUpload', features);
}
```

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +60,38 @@ app.GpxUploadController = function($scope) {
...
 
 /**
+ *
+ * @param {Array.<ol.Feature>} features - invalid features that have to be checked
+ * @returns {Array.<ol.Feature>} features - this geometry is valid
+ * @private
+ */
+app.GpxUploadController.prototype.validatedFeatures_ = function(features) {
+  for (var i = 0; i < features.length; i++) {
+    var geom = /**@type{ol.geom.Geometry}*/ (features[i].getGeometry());
```
asaunier: Missing blank space after and before ``/**`` ``*/``

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +60,38 @@ app.GpxUploadController = function($scope) {
...
+
+    if (geom instanceof ol.geom.MultiLineString) {
+      var multiLineString = /**@type{ol.geom.MultiLineString}*/ (geom);
```
asaunier: Missing blank spaces

c2corg_ui/static/js/gpxupload.js
```diff
@@ -59,6 +60,38 @@ app.GpxUploadController = function($scope) {
...
+      var lineStrings = multiLineString.getLineStrings();
+
+      for (var j = 0; j < lineStrings.length; j++) {
+        if (lineStrings[j].getCoordinates().length === 1) {
+          delete lineStrings[j];
+        }
+      }
+      lineStrings = lineStrings.filter(function(n) {
+        return n != undefined;
+      });
```
asaunier: Isn't it possible to merge those 2 actions by directly removing the linestring from the list if it has only one set of coordinates?

asaunier:
Arf, my review came too late. Please at least address the comment about the final ``if (features)`` test, I think it's a bug.

asaunier:
By the way ``validatedFeatures_`` is not a correct method name because it's not an action name. It should rather be ``validateFeatures_``.

dkocich:
Yes. It is possible that empty `features` will be returned from validation function (so we should check `features.length`). Moving that line up should solve possible issues. `features` will always be an array and during validateFeatures, no loop will simply begin. I will create another PR for these changes you suggested.

Issue #1176, opened 2016-12-09
Images: UI must use date and time instead of date


brunobesson:
* On image page, date and time only displays date
* When editing, this is a date selector and not a date and time selector

fbunoz:
It is a big bug : the Exif time is not stored.

- [From the API](https://api.camptocamp.org/images/829497), the time is 2016-12-07T00:00:00+00:00
- But in the Exif ([in the Jpeg](https://media.camptocamp.org/c2corg_active/1481324429_523379614.jpg)), the time is : 2016-12-07 15:01:57

In an image doc from the V5, the time has been migrated correctly.
To correct the time, after fixed the bug, we need to reback only the image uplaoded in the V6.

loicperrin:
If time of images uploaded in V6 is not stored, this bug should be fixed in high prio.

fbunoz:
The bug is worst as described here : https://github.com/c2corg/v6_images/issues/34

brunobesson:
@stef74 you changed the priority from 6.1 to 6.4. IMHO, 6.4 is not a good choice, this is a pretty serious bug. The main impact being the ordering of pictures, which needs EXIF info.

stef74:
this issue is part 6.1 milestone : https://github.com/c2corg/v6_ui/milestone/10

brunobesson:
Grmpf. My bad, I'm sorry. It's the second time it happens to me.
The line in github is not clear enough :wink: 

>@stef74 modified the milestone: V6.1, V6.4 5 days ago

asaunier:
*[ ] Concerning the image datetime details, it seems an API issue, see https://github.com/c2corg/v6_api/issues/599#issuecomment-271585546

* [ ] Concerning the image editing, there is also a UI problem because only a date selector is provided, not a time selector.

tsauerwein:
https://github.com/c2corg/v6_ui/pull/1394 makes sure that the datetime is extracted correctly. What is left to do is to update the edit form so that also the time can be changed.

fbunoz:
Is there a script to update all images uploaded on the V6 until now ?
It's all images from this one : https://api.camptocamp.org/images/828237 (included), ~10000 images.

asaunier:
No

fbunoz:
Ok, I created a new issue : https://github.com/c2corg/v6_ui/issues/1419

asaunier:
[...]

PR #1394, opened 2017-01-19
Fix extracting location and date_time from EXIF data

tsauerwein:
* Do no try to send geometry for an image if the location in the EXIF data is invalid
* Correctly set `date_time` from EXIF data

Related to https://github.com/c2corg/v6_ui/issues/1176
Related to https://github.com/c2corg/v6_api/issues/600
Related to https://github.com/c2corg/v6_api/issues/599

### DIFF
--- a/c2corg_ui/externs/moment.js
+++ b/c2corg_ui/externs/moment.js
@@ -7,15 +7,20 @@
 
 /**
  * @param {Object | string} date
+ * @param {string=} opt_format
  */
-function moment(date) {};
-
+function moment(date, opt_format) {};
 
 /**
  * Formats the date in given format, for ex. 'YYYY-MM-DD'
- * @param {string} format
+ * @param {string=} opt_format
  */
-moment.prototype.format = function(format) {};
+moment.prototype.format = function(opt_format) {};
+
+/**
+ * @return {boolean}
+ */
+moment.prototype.isValid = function() {};
 
 /**
  * Difference in miliseconds between two dates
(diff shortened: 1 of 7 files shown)


asaunier:
The code changes look good to me!

The upload of an image with GPS data such as



works fine.

I have tested with images that were known to cause a crash but I still observe errors for some of them (different ones though).

For instance with https://github.com/c2corg/v6_ui/issues/929#issuecomment-273608912 the ``date_time`` attribute is set to ``Invalid date`` and it makes the API crashes because
```
  File "/home/alex/v6_api/c2corg_api/views/image.py", line 78, in validate_list_image_create
    for image in request.validated['images']:
KeyError: 'images'
```
capture du 2017-01-19 15-13-36

There is still an error with this image https://github.com/c2corg/v6_ui/issues/1341#issue-200101830
=> the coords are probably actually numbers but the computed geometry is weird:
capture du 2017-01-19 15-21-18
At least the API seems to refuse it instead of saving it as it did when this image was first uploaded (it caused super slow page loading because some script was having a hard time with those weird coordinates)
```
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/cornice/service.py", line 478, in wrapper
    validator(request, **args)
  File "/home/alex/v6_api/c2corg_api/views/image.py", line 78, in validate_list_image_create
    for image in request.validated['images']:
KeyError: 'images'
```
(by the way shouldn't the API return an actual error instead of crashing?)

On the other hand it's now OK for
* https://github.com/c2corg/v6_api/issues/600#issue-200099193 + the other one at this end of the comment
* https://github.com/c2corg/v6_ui/issues/1341#issuecomment-271899528

tsauerwein:
Thanks for the testing! I added checks to handle the two cases that did not work.

> (by the way shouldn't the API return an actual error instead of crashing?)

Indeed!

asaunier:
All tested images now pass, dates (when available) are now correctly saved with time, valid GPS coords are saved. Thanks, please merge when Travis is OK.

Issue #1321, opened 2017-01-06
Do not display virtual WP on the map of an associated route


fbunoz:
On the page of a route associated to a WP whose type = virtual, the virtual WP must not be displayed on the map.
Example where a virtual WP clutters the map without adding an information : https://www.camptocamp.org/routes/418850/fr/grand-tour-du-pic-d-asti-par-la-pointe-joanne-le-col-agnel-et-le-pic-segure

desnoes:
fixed on demo (FF)

PR #1397, opened 2017-01-19
Do not show virtual waypoints in route maps

brunobesson:
Related to #1321

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -414,7 +414,8 @@
   % for wp_category in ['waypoints', 'waypoint_children']:
     % if document.get('associations').get(wp_category):
       % for wp in document.get('associations').get(wp_category):
-        % if wp.get('geometry'):
+        % if wp.get('geometry') and \
+            not (document['doctype'] == 'routes' and wp['waypoint_type'] == 'virtual'):
           ,
           {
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/helpers/view.html
```diff
@@ -412,7 +412,9 @@ <h3 class="heading show-phone outings" ng-click="mainCtrl.animateHeaderIcon($eve
   % for wp_category in ['waypoints', 'waypoint_children']:
     % if document.get('associations').get(wp_category):
       % for wp in document.get('associations').get(wp_category):
-        % if wp.get('geometry'):
+        % if wp.get('geometry') and\
```
asaunier: Missing blank space before ``\``

c2corg_ui/templates/helpers/view.html
```diff
@@ -412,7 +412,9 @@ <h3 class="heading show-phone outings" ng-click="mainCtrl.animateHeaderIcon($eve
...
+            not (document['doctype'] == 'routes' and wp['waypoint_type'] == 'virtual'):
+            ## do not display virtual waypoints in route map #1321
```
asaunier: Not sure it is need to add a reference to #1321 => actually it's not that obvious it's a github issue and the issue only says the same thing than the comment above.

PR #1400, opened 2017-01-19
Fixe flex usage

### LINKED ISSUE #1248: Advanced search : Hide map on mobile landscape view
On mobile, with landscape view, the advanced search page show card list + map.
The map is too small and unusable.
It's better have the same behavior than vertical view : a button to show/hide the map on full screen.

stef74:
- remove mediaqueries on flex usage with map/no map
- increase phone detection to 700 to fix #1248 
- use same flex strategy to user selection (friend)
- reduction of font-size on mobile to show more content

Flex strategy : 
- card of 300pixel (flex-basis:300px)
- increase if space available (flex-grow:1)

Test localy see screenshot 

capture d ecran 2017-01-19 a 21 08 45
capture d ecran 2017-01-19 a 21 12 03
capture d ecran 2017-01-19 a 21 12 23
capture d ecran 2017-01-19 a 21 18 10
capture d ecran 2017-01-19 a 21 18 11

### DIFF
--- a/less/variables.less
+++ b/less/variables.less
@@ -1,8 +1,8 @@
 @big: ~"only screen and (min-width: 1400px) and (max-width: 3000px)";
 @desktop: ~"only screen and (min-width: 1100px) and (max-width: 1399px)";
-@tablet: ~"only screen and (min-width: 620px) and (max-width: 1099px)";
-@phone: ~"only screen and (max-width: 619px)";
-@phone-max: 620px;
+@tablet: ~"only screen and (min-width: 701px) and (max-width: 1099px)";
+@phone: ~"only screen and (max-width: 700px)";
+@phone-max: 700px;
 @tablet-max: 1100px;
 @large-width: 1800px;
 @lightgrey: #F0F0F0;
--- a/less/variables.less
+++ b/less/variables.less
@@ -1,8 +1,8 @@
 @big: ~"only screen and (min-width: 1400px) and (max-width: 3000px)";
 @desktop: ~"only screen and (min-width: 1100px) and (max-width: 1399px)";
-@tablet: ~"only screen and (min-width: 701px) and (max-width: 1099px)";
-@phone: ~"only screen and (max-width: 700px)";
-@phone-max: 700px;
+@tablet: ~"only screen and (min-width: 769px) and (max-width: 1099px)";
+@phone: ~"only screen and (max-width: 768px)";
+@phone-max: 768px;
 @tablet-max: 1100px;
 @large-width: 1800px;
 @lightgrey: #F0F0F0;
--- a/less/lists.less
+++ b/less/lists.less
@@ -18,7 +18,7 @@
   position: relative;
   cursor: pointer;
   background-color: white;
-  margin-top: 10px;
+  margin-top: 4px;
   outline: none;
   border-radius: 7px;
   min-height: 80px;
@@ -295,6 +295,7 @@
     margin-left: 3px;
     margin-right: 3px;
     flex-basis: 300px;	
+    max-width: 350px;
     .list-item-info {
       justify-content: center;
     }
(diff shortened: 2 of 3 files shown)


asaunier:
Do we really want to have cards whose size depends on the number of cards on the same line?

See the example below about the list of outings in a route page:
capture d ecran 2017-01-20 a 00 38 24

Or in the routes advanced search page:
capture d ecran 2017-01-20 a 00 40 22

Else the behaviour of this changes seems OK. But I have only tested on a recent Chrome and I have no clue if the CSS changes are OK :P

stef74:
I'ts due to flex-grow properties ... it offer possibility to enlarge the
box if space is available (good idea to adapt box to real space), but I
have now way to detect the number of card.

I have no real idea to fix that ....

2017-01-20 0:45 GMT+01:00 Alex Saunier <notifications@github.com>:

stef74:
Flex-grow is no used in friend selection ... so it's fixed sized (replace space available by margin not increase size of card).

asaunier:
Recently while ice climbing with a friend he told me has was a bit lost with the v6. I realized that pages were not very well rendered on his iphone (4, I think): the page header was a bit messy and it was impossible to use the simple search (so he used google...), outing cards in route pages were very narrow (titles were displayed as 1-character columns, vertically) so they were barely usable...

Perhaps that would be useful to make sure we don't use too recent CSS styles if we can also use better-supported alternate?

stef74:
Iphone 4 is very outdated ... it's big job to use it (probleme of width and support ...). 

Flex are supported on previsous version on Safari. We cannot ask user to upgrade her browser and ask to dev to support old device in specific case.

All flex properties are not supported but no major impact is advanced properties are not supported.

No body complain on flex usage (allready use in v6 previous this PR), only complain a slow scrolling on Ios but usable for me ( I use it every day ...)

I check tonight if can limit size of card with max_width, but not sure of rendering.

stef74:
I can limit size of card. I have force to max-width : 350px
If your are pixel perfect, the size on two last card are not the same with other (but not aligned ... so difficult to detect)
capture d ecran 2017-01-20 a 20 11 43

asaunier:
Don't worry, I'm not that pixel perfect :)

But I still have cards getting pretty wider on the last line (in the advanced search page or the associated outings of a route):

capture d ecran 2017-01-20 a 22 41 40
capture d ecran 2017-01-20 a 22 41 58

stef74:
Normal ... I have to do the commit 😌

stef74:
Done. I have also set same margin right/left and top/bottom.

stef74:
@asaunier @lbesson can we review ?

less/variables.less
```diff
@@ -1,8 +1,8 @@
 @big: ~"only screen and (min-width: 1400px) and (max-width: 3000px)";
 @desktop: ~"only screen and (min-width: 1100px) and (max-width: 1399px)";
-@tablet: ~"only screen and (min-width: 620px) and (max-width: 1099px)";
-@phone: ~"only screen and (max-width: 619px)";
-@phone-max: 620px;
+@tablet: ~"only screen and (min-width: 701px) and (max-width: 1099px)";
+@phone: ~"only screen and (max-width: 700px)";
+@phone-max: 700px;
```
brunobesson: How and why did you decide thoses limits? I understand this is based on #1248 but to me it seems like a rule of thumb rather than based on a global view.

stef74: Quick look on "standard" phone format.

brunobesson: I don't think it will ba a value like that, but rather a standard one, like 768px, 640px, etc. Don't you think?

brunobesson: Bootstrap, e.g. : https://v4-alpha.getbootstrap.com/layout/overview/#responsive-breakpoints

stef74: Ok tu use more standard value. I fix

Issue #942, opened 2016-11-19
WP creation 'Access': title errors


desnoes:
- [ ] 'Accès' shall be replaced by 'Accès routier'.
- [x] 'Période de gardiennage' shall be replaced by 'Période de déneigement ou d'ouverture'.

![access](https://cloud.githubusercontent.com/assets/19892089/20455395/6c23867a-ae5b-11e6-9a1f-058c47d61acc.jpg)

stef74:
Us only a translation issue ?

fbunoz:
No, because the string "Accès" is used for other WPs, and the string "Période de gardiennage" is used for the huts.

asaunier:
This issue has already been reported by @loicperrin at https://github.com/c2corg/v6_ui/issues/705#issuecomment-258557772

shortoland:
Indeed, "Accès" is used twice and should be duplicated.

access_period is used for accesses
opening_periods is used for huts
Translations were switched, I have corrected that.

access_period is now  Période de déneigement ou d'ouverture
 opening_periods is now Période de gardiennage

brunobesson:
`access_period` problem is already fixed.
#1401 attempts at solving the remaining translation issue.

asaunier:
@brunobesson Can this issue be closed since your PR has been merged?

brunobesson:
I.... think so. In prod, result seems fine to me.

Issue #1218, opened 2016-12-14
WP 'Access': the field "Accès" should be renamed "Accès TC" 


desnoes:
WP 'Acces': the field "Accès" should be renamed "Accès TC" (OK if we can do it in Transifex, but please check that it is not a common name used elsewhere)

brunobesson:
@desnoes I just committed #1401 assuming the french translation for `access` field should be "Accès routier". But here you're saying it should rather be "Accès en transport en commun". What's the correct choice?

fbunoz:
For an access WP, there are 2 text fields to describe the access : road access, an public transport access.
The V5 "road access" has been migrated in "description", and "PT access" in "access".
Be careful when you change the field name : "PT access" must be used only withe access WP.
Hut and climbing site WP use also this field, but as pedestrian access. And product WP uses it as "access" (road or pedestrian : last kilometers from the associated access or from an always accessible village).

brunobesson:
The art of obfuscating simple things in an attempt of optimization that may be useless :wink: I'll have a look later.

fbunoz:
I didn't say that each WP type must have its field name, but be careful with side effects :-)

brunobesson:
@fbunoz are you sure there is an access field for huts?

brunobesson:
@asaunier @fbunoz is the 'access' field still supposed to be available for huts? The API doesn't seem to return it.
I am working on #1401 to take into account more complicated cases:

* description field should be renamed 'road access' for waypoints
* access field may have different names depending on whether the waypoint is an access, a climbing_indoor, a climbing_outdoor, a local_product, or a *hut* (?)

I seem to handle it properly for all cases except huts: I can edit the field, but it is not returned by the API. Should I open a API issue?

(see also #942)

fbunoz:
> I can edit the field, but it is not returned by the API. Should I open a API issue?

Yes, open an issue.
The V5 hut access fields has been migrated in V6 WP access field : https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/waypoints/huts.py#L95
But the data don't appear in access field at the edition when changing WP type to product for instance.
I hope it's just an API issue and not a migration issue.

brunobesson:
I opended c2corg/v6_api#611

brunobesson:
I think this issue is solved, isn't it?

desnoes:
@brunobesson : yes it is solved

PR #1401, opened 2017-01-19
Use specific translation keys for waypoint access and description fields

brunobesson:
A waypoint of type 'access' has an `access` field which corresponds to the description of how to reach the parking by road.
The key `access` cannot be used because it is already used for access waypoint.
Thus, use another key `road access`.

References #942

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -461,12 +461,16 @@
   % endif
 </%def>
 
-<%def name="get_document_description(locale, doctype)">\
+<%def name="get_document_description(locale, doctype, title=None)">\
   % if locale.get('summary') or locale.get('description'):
     <div class="view-details-description col-xs-12  description">
       % if doctype not in ['article', 'profile']:
         <h3 class="heading show-phone">
-          <span translate>description</span>
+          % if title:
+            <span x-translate>${title}</span>
+          % else:
+            <span translate>description</span>
+          % endif
         </h3>
       % endif
       <span class="lead">
--- a/c2corg_ui/templates/waypoint/edit.html
+++ b/c2corg_ui/templates/waypoint/edit.html
@@ -741,7 +741,7 @@ from c2corg_common.attributes import default_langs, climbing_outdoor_types, publ
 
             ## ACCESS
             <div class="form-group data full" ng-if="['access', 'local_product', 'climbing_outdoor'].indexOf(type) > -1">
-              <label translate>access</label>
+              <label translate>road access</label>
               <textarea name="access" ng-model="waypoint.locales[0].access" class="form-control" placeholder="{{'Describe access' | translate}}"></textarea>
               <span class="glyphicon glyphicon-ok form-control-feedback" ng-show="waypoint.locales[0].access"></span>
             </div>
--- a/c2corg_ui/templates/waypoint/edit.html
+++ b/c2corg_ui/templates/waypoint/edit.html
@@ -735,14 +735,22 @@ from c2corg_common.attributes import default_langs, climbing_outdoor_types, publ
 
             ## DESCRIPTION
             <div id="description-group" class="form-group data full">
-              <label translate>description</label>
-              <textarea name="description" ng-model="waypoint.locales[0].description" class="form-control description" placeholder="{{'Describe here the waypoint' | translate}}"></textarea>
+              <label translate ng-if="type == 'access'">road access</label>
+              <label translate ng-if="type != 'access'">description</label>
+              <textarea name="description" ng-model="waypoint.locales[0].description" class="form-control description"
+                placeholder="{{ type =='access' ? mainCtrl.translate('Describe here the waypoint') : mainCtrl.translate('Describe road access') | translate}}"></textarea>
             </div>
 
             ## ACCESS
-            <div class="form-group data full" ng-if="['access', 'local_product', 'climbing_outdoor'].indexOf(type) > -1">
-              <label translate>road access</label>
-              <textarea name="access" ng-model="waypoint.locales[0].access" class="form-control" placeholder="{{'Describe access' | translate}}"></textarea>
+            <div class="form-group data full" ng-if="['access', 'local_product', 'climbing_outdoor', 'climbing_indoor', 'hut'].indexOf(type) > -1">
+              <label translate ng-if="type == 'access'">public transportation access</label>
+              <label translate ng-if="type == 'hut' || type == 'climbing_indoor' || type == 'climbing_outdoor'">pedestrian access</label>
+              <label translate ng-if="type == 'local_product'">road or pedestrian access</label>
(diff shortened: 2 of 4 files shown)


brunobesson:
See also https://github.com/c2corg/v6_ui/issues/1218#issuecomment-273909428
Let's wait for a clear answer before merging!

asaunier:
By "key" you mean "field title"?
I don't have a strong opinion on this question. If people at the association think it's better to rename it to ``road access``, please merge.

brunobesson:
I won't be merging now, @fbunoz explained complicated things about fields in waypoints and their meaning, seems more complicated than that...

brunobesson:
I renamed this PR and flagged it `[WIP]` as there are some issues to solve first.
See #942 #1218

c2corg_ui/templates/helpers/view.html
```diff
@@ -453,12 +453,12 @@ <h3 class="heading show-phone outings" ng-click="mainCtrl.animateHeaderIcon($eve
...
 </%def>
 
-<%def name="get_document_description(locale, doctype)">\
+<%def name="get_document_description(locale, doctype, title=None)">\
   % if locale.get('summary') or locale.get('description'):
     <div class="view-details-description col-xs-12  description">
       % if doctype not in ['article', 'profile']:
         <h3 class="heading show-phone">
-          <span translate>description</span>
+          <span x-translate>${title if title else 'description'}</span>
```
asaunier: OK because ``description`` is probably extracted somewhere else (?) but a "safer" syntax would have been:
```
% if title:
  <span x-translate>${title}</span>
% else:
  <span translate>description</span>
% endif
```

c2corg_ui/templates/waypoint/edit.html
```diff
@@ -735,14 +735,22 @@ <h2 class="heading show-phone" translate>description</h2>
...
+              <textarea name="description" ng-model="waypoint.locales[0].description" class="form-control description"
+                placeholder="{{ type =='access' ? mainCtrl.translate('Describe here the waypoint') : mainCtrl.translate('Describe road access') | translate}}"></textarea>
             </div>
 
             ## ACCESS
-            <div class="form-group data full" ng-if="['access', 'local_product', 'climbing_outdoor'].indexOf(type) > -1">
-              <label translate>access</label>
-              <textarea name="access" ng-model="waypoint.locales[0].access" class="form-control" placeholder="{{'Describe access' | translate}}"></textarea>
+            <div class="form-group data full" ng-if="['access', 'local_product', 'climbing_outdoor', 'climbing_indoor', 'hut'].indexOf(type) > -1">
+              <label translate ng-if="type == 'access'">pt access</label>
```
asaunier: pt access?
What does this mean? Are you sure translators will understand how to translate this in Transifex?

c2corg_ui/templates/waypoint/helpers/view.html
```diff
@@ -423,3 +423,27 @@
...
+
+<%def name="get_waypoint_access_field(locale, waypoint_type, open_tab=False)">\
+  % if locale.get('access'):
+    <%
+        if waypoint_type == 'access':
+            attribute_name = 'pt access'
+        elif waypoint_type in ['hut', 'climbing_indoor', 'climbing_outdoor']:
+            attribute_name = 'pedestrian access'
+        else:
+            attribute_name = 'road or pedestrian access'
```
asaunier: Again this is not a very "safe" way to switch among attribute titles. It's a bit more verbose but clearer (and at least you are sure the strings will be i18n-extracted) to use 
```
% if waypoint_type == 'access':
  <span x-translate>pt access</span>
...
```

c2corg_ui/templates/waypoint/edit.html
brunobesson: "pt" for public transport. But okay, I'll use a better key.

c2corg_ui/templates/waypoint/helpers/view.html
brunobesson: No problem. I tried to be consistent and used the same syntax already used for access_periods: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/helpers/view.html#L403

Should we change it here too?

brunobesson:
@asaunier there is an issue nonetheless with this PR, concerning huts. Looks like an API issue (access field is not returned), but I am invetigating it.
That's why I'll keep the WIP prefix for now.

brunobesson:
@asaunier c2corg/v6_api#611 being fixed, this PR is now in a correct state for merging. Still OK for this?

c2corg_ui/templates/waypoint/helpers/view.html
brunobesson: Before merging, this point should be clarified too

asaunier:
> c2corg/v6_api#611 being fixed, this PR is now in a correct state for merging. Still OK for this?

OK for me. Please merge when you have addressed the remaining point.

brunobesson:
>when you have addressed the remaining point

I am requesting some feedback here :wink:

c2corg_ui/templates/waypoint/helpers/view.html
asaunier: > Should we change it here too?

I guess so. At least it would be consistent to use the same approach for both functions. Make your choice :)

brunobesson:
@asaunier could you have a last quick check on the `translate` vs `x-translate` case? I think it's OK, but I have seen some strange things I don't understand sometimes (`x-translate` when `translate` seem appropriate to me), so I prefer a last review before merging.

c2corg_ui/templates/waypoint/helpers/view.html
```diff
@@ -423,3 +422,27 @@
     </div>
   % endif
 </%def>
+
+<%def name="get_waypoint_access_field(locale, waypoint_type, open_tab=False)">\
+  % if locale.get('access'):
+
```
asaunier: Empty line perhaps not necessary (OK minor I admit).

brunobesson:
>Is there a special line where you are unsure of your code?

No. It was only for a check anyway. I missed the empty line for example :wink: 

Some examples of cases I don't understand in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/helpers/view.html:

* https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/helpers/view.html#L357-L377 why are we using `x-translate`?
* https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/helpers/view.html#L332 why `x-translate` for `custodianship` (`translate` is used for the other 3 cases)?
* https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/helpers/view.html#L231 ditto

asaunier:
> Some examples of cases I don't understand

Those are most likely errors. No reason to use ``x-translate`` for fixed strings. Thanks for pointing the problems.
Feel free to switch them to ``translate`` instead.

Issue #988, opened 2016-11-24
not all float buttons have box shadow


brunobesson:
There are four float buttons: edit / add images / comments / plus (dropdown)
The plus button has no box-shadow (it is a rule for dropdowns - https://github.com/c2corg/v6_ui/blob/master/less/floatbuttons.less#L127), which I find disturbing. Is this a design choice?
(whether I would like to have the box-shadow is another question)

ginold:
I don't remember why, but it's a good catch! Maybe it's some very old code, along with `app-edit-button` seen above, which doesn't even exist anymore. You're right - all the buttons should have a shadow (or not).

brunobesson:
For now, maybe we could add the shadow, because all other buttons have it. I can propose a PR for this this evening if you like.
As whether we should have a shadow or not, I would rather not, but this should be discussed with more people, more globally<sup>*</sup> and is certainly not the priority.

(*) IMHO, we could sharpen the corners a little bit, remove some frames to lighten the layout, etc. But this is another subject!

brunobesson:
#1402 chooses to remove box-shadow from buttons.

desnoes:
I am unable to test, i did not catch the modification.

PR #1402, opened 2017-01-19
remove box-shadow for buttons

brunobesson:
For consistency, remove box-shadow from all buttons.
See #988

### DIFF
--- a/less/documentedit.less
+++ b/less/documentedit.less
@@ -27,14 +27,6 @@
     @media @phone {
       display: none;
     }
-    button {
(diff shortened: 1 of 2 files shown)


asaunier:
I wasn't disturbed by those shadows around the "document tool buttons" (except the "plus" one indeed) + "feed buttons" but OK for removing them if you think it's better/more consistent!

brunobesson:
I don't know if we need to keep them or remove them, but at least it should be consistent. I would rather remove them to lighten the L&F. I'll merge.

PR #1406, opened 2017-01-20
Add 24h time selector to Image edit page

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1176

So easiest was to add time picker from tools we already use and have time selector separately https://angular-ui.github.io/bootstrap/#/timepicker
What do you think - is this time picker ok?

Some images do not have date_time and in edit the form they have this date.
http://localhost:6558/images/edit/824766/fr
image

https://github.com/dalelotts/angular-bootstrap-datetimepicker#screen-shots
There is also this tool but I personally do not like when I want to select a time and I need to go through few steps with a selection of date also.

### DIFF
--- a/c2corg_ui/templates/image/edit.html
+++ b/c2corg_ui/templates/image/edit.html
@@ -133,7 +133,11 @@ from c2corg_common.attributes import default_langs, activities, image_types, ima
                 </span>
               </div>
 
-              <br>
+              ## TIME
+              <div class="input-group" uib-timepicker ng-model="image.date_time" hour-step="1" minute-step="1" second-step="1"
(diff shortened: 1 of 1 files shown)


asaunier:
> What do you think - is this time picker ok?

Yes, it looks great to me.

> Some images do not have date_time and in edit the form they have this date.

It's OK. Having the hour set to 0 would make more sense but it's OK as is (anyway the time is not the correct one since is is missing).

dkocich:
It probably just sets undefined as 0 in unix time https://en.wikipedia.org/wiki/Unix_time which is equal to this date. +1 hour is caused by our current timezone adjustment.

Issue #1408, opened 2017-01-20
Image editing map section


asaunier:
The map section could be more individualized in the image editing form, with a title (as for WP, routes or outings I think). In addition adding the lon/lat fields (as for WP) would make sense.

<img width="967" alt="capture d ecran 2017-01-20 a 22 03 20" src="https://cloud.githubusercontent.com/assets/1192331/22165586/e502c47a-df5c-11e6-83aa-91f9063b57d3.png">

desnoes:
ok on demo

PR #1413, opened 2017-01-24
Add lonlat fields to Image edit form

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1408
Looks more consistent now with other types.
I also added restriction to elevation because API crashed when I tried to add high value.

image


asaunier:
What about adding a section title above the map, as suggested in https://github.com/c2corg/v6_ui/issues/1408 and done for the other document types?

dkocich:
Now we have two styles -

- for Point (image, xreport, waypoint)
image

- for line (route, outing) - there are no XYZ fields so I did not add section just for the map
image

PR #1417, opened 2017-01-24
Rework article choice logic

brunobesson:
Don't know why, but I apparently made mistake when copying stuff from v5

### DIFF
--- a/c2corg_ui/templates/utils/__init__.py
+++ b/c2corg_ui/templates/utils/__init__.py
@@ -91,22 +91,52 @@ def get_route_gear_articles(route):
         if 'snowshoeing' in activities or 'skitouring' in activities:
             articles.append(('183333', 'skitouring gear'))
         if 'snow_ice_mixed' in activities and \
-                route.get('global_rating') in ['F', 'F+', 'PD', 'PD+']:
+                route.get('global_rating') in ['F', 'F+', 'PD-', 'PD', 'PD+']:
             articles.append(('185750', 'easy snow ice mixed gear'))
         if 'mountain_climbing' in activities and \
-                route.get('global_rating') in ['F', 'F+', 'PD', 'PD+', 'AD']:
-            articles.append(('185384', 'easy rock climbing gear'))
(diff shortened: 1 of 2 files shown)


c2corg_ui/templates/utils/__init__.py
```diff
@@ -95,18 +95,38 @@ def get_route_gear_articles(route):
...
+        if 'rock_climbing' in activities:
+            if route.get('equipment_rating') in ['P1', 'P1+']:
+                articles.append(('183332', 'bolted rock climbing gear'))
+            elif 'mountain_climbing' not in activities and \
+                    route.get('global_rating') in [
+                        'F',
+                        'F+',
+                        'PD',
+                        'PD+',
+                        'AD'
```
asaunier: What about PD- and AD-?

brunobesson: You're right. :sob:

fbunoz:
Instead of enumerating all items in a range, is it possible to define a range by > and < conditions ?
Because if we add some subdivision later, the enumeration will be incorrect, and will need an update.

brunobesson:
Feel free to propose something.
AFAIK there is no easy mapping between the possible values and numbers. Maybe in v6_common?

asaunier:
>  there is no easy mapping between the possible values and numbers.

Indeed. The attributes in DB are not numeric but are enumerations of the strings defined in https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py

For instance for ``ski_rating``: 
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/route.py#L78
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/enums.py#L65-L66
https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L314-L333

> Instead of enumerating all items in a range, is it possible to define a range by > and < conditions ?

I don't think it can be used in the UI because the UI has no access to the models (in the API).
In the API, especially for the search requests, it is possible to specify a range of ratings.

PR #1424, opened 2017-01-25
Translate 'persons' suffix in xreports edit

brunobesson:
certif30

### DIFF
--- a/c2corg_ui/templates/xreport/edit.html
+++ b/c2corg_ui/templates/xreport/edit.html
@@ -208,7 +208,7 @@ from c2corg_common.attributes import default_langs, activities, event_types, \
               <label translate>nb_participants</label>
               <div class="input-group">
                 <input class="form-control" type="number" min="0" max="1000" minlength="1" name="nb_participants" ng-model="xreport.nb_participants"/>
-                <span class="input-group-addon">persons</span>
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/xreport/edit.html
```diff
@@ -208,7 +208,7 @@ <h2 class="heading show-phone" translate>xreport general</h2>
               <label translate>nb_participants</label>
               <div class="input-group">
                 <input class="form-control" type="number" min="0" max="1000" minlength="1" name="nb_participants" ng-model="xreport.nb_participants"/>
-                <span class="input-group-addon">persons</span>
+                <span class="input-group-addon" x-translate>persons</span>
```
asaunier: You should use ``translate`` instead of ``x-translate`` here, because you want ``persons`` to be extracted by the translations extractor tool. Ditto for the other change.

``translate`` and ``x-translate`` are equivalent (the latter being an alias of the first directive), except that ``x-translate`` is not recognized by the extractor.

brunobesson: Oops. I was aware of this, simply mixed both.

brunobesson:
@asaunier OK for merging with latest modifications?

PR #1425, opened 2017-01-25
Use specific key for xreport condition field

brunobesson:
`conditions` is also used in other places. Thus use `conditions study` key instead.

See https://forum.camptocamp.org/t/nouvelle-version-v6-0-8/177838/7?u=b_b

### DIFF
--- a/c2corg_ui/templates/xreport/edit.html
+++ b/c2corg_ui/templates/xreport/edit.html
@@ -393,7 +393,7 @@ from c2corg_common.attributes import default_langs, activities, event_types, \
 
             ## conditions
             <div id="conditions-group" class="form-group full">
-              <label translate>conditions</label>
+              <label translate>conditions study</label>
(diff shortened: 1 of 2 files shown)


c2corg_ui/templates/xreport/edit.html
```diff
@@ -208,7 +208,7 @@ <h2 class="heading show-phone" translate>xreport general</h2>
               <label translate>nb_participants</label>
               <div class="input-group">
                 <input class="form-control" type="number" min="0" max="1000" minlength="1" name="nb_participants" ng-model="xreport.nb_participants"/>
-                <span class="input-group-addon">persons</span>
+                <span class="input-group-addon" x-translate>persons</span>
```
asaunier: ``translate`` instead of ``x-translate``, ditto for the following change

brunobesson: Plus, I mixed branch `translate-persons-suffix` commits in here :confused:

Issue #1404, opened 2017-01-20
Some articles generate Internal Server Error


fbunoz:
Ex : https://www.camptocamp.org/articles/465897/fr/concours-photo-sophie-2013
The particularity is this article is linked to ~500 images.

The API data seem correct, then it seem to be a bug in the UI.
I don't know if this bug exists only for article, or for any doc.

asaunier:
Here is the error message when loading this article in my instance:
```
  File "/home/alex/Desktop/v6_ui/c2corg_ui/templates/article/view.html", line 68, in render_body
    ${get_article_general(article)}
  File "/home/alex/Desktop/v6_ui/c2corg_ui/templates/article/helpers/view.html", line 34, in render_get_article_general
    % if article.get('author') and article['article_type'] != 'collab' and article['author']['user_id'] != article['associations']['users'][0]['document_id']:
IndexError: list index out of range
```

dkocich:
I just wanted to write that. So this is caused by the latest change PR
 https://github.com/c2corg/v6_ui/pull/1107 with showing/hiding of author / associated users for different article types.

tsauerwein:
There are more articles failing, but with a different error:

https://www.camptocamp.org/articles/106726/fr/140483
```
 File "/var/www/c2corg_ui/templates/base.html", line 57, in render_body ${self.body()} File "/var/www/c2corg_ui/templates/article/view.html", line 69, in render_body ${get_article_associated_users(article)} File "/var/www/c2corg_ui/templates/article/helpers/view.html", line 52, in render_get_article_associated_users % if article['associations']['users']: KeyError: 'associations'
```

desnoes:
Example of an article that generates the message 'Internal Server Error'

https://www.camptocamp.org/articles/844265/fr/test

desnoes:
It is possible to open personnal articles

PR #1426, opened 2017-01-25
Fix articles because of association key problems

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1404

All example articles (from related issue) should work
- https://www.camptocamp.org/articles/465897/fr/concours-photo-sophie-2013
- https://www.camptocamp.org/articles/106726/fr/140483
- https://www.camptocamp.org/articles/844265/fr/test (this one is new and not available in local DB to test)

Basically, there problems were with:
-  with `article['associations']['users'][0]['document_id']` when no associated users were available. I deleted it to make it work now.
-  in `get_article_associated_users` with `article['associations']['users']` - it was not available sometimes and causing an error (comparing versions in history)

### DIFF
--- a/c2corg_ui/templates/article/helpers/view.html
+++ b/c2corg_ui/templates/article/helpers/view.html
@@ -31,13 +31,13 @@
       <p><span class="value-title" translate>article_type</span>: <span class="value" x-translate>${article['article_type']}</span></p>
     % endif
 
-    % if article.get('author') and article['article_type'] != 'collab' and article['author']['user_id'] != article['associations']['users'][0]['document_id']:
+    % if article.get('author') and article['article_type'] == 'personal':
       <%
           user_id = article['author']['user_id']
           user_lang = article['locales'][0]['lang']
       %>
       <p><span class="value-title" translate>author</span>:
-        <a class="value" href="${request.route_path('profiles_view', id=user_id, lang=user_lang)}" >${article['author']['name']}</a>
+      <a class="value" href="${request.route_path('profiles_view', id=user_id, lang=user_lang)}" >${article['author']['name']}</a>
     % endif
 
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/article/helpers/view.html
```diff
@@ -49,30 +49,34 @@
...
-                    user_lang = user['locales'][0]['lang']
-                    is_last_associated_user = loop.last
-                %>
-                <a href="${request.route_path('profiles_view', id=user_id, lang=user_lang)}">${user['name']}</a>${'' if is_last_associated_user else ', '}
-              % endfor
-            </p>
-          % endif
+  % if article.get('associations'):
+    % if article['associations']['users']:
+      % if len(article['associations']['users']) > 1:
```
asaunier: Why not using ``and`` operators instead of embedded ``if`` blocks?

c2corg_ui/templates/article/helpers/view.html
```diff
@@ -49,30 +49,34 @@
...
+            <p translate>associated users</p>
+          </div>
+          <span class="detail-text accordion">
+            % if article.get('associations').get('users'):
+              <p>
+                % for user in article['associations']['users']:
+                  <%
+                      user_id = user['document_id']
+                      user_lang = user['locales'][0]['lang']
+                      is_last_associated_user = loop.last
```
asaunier: Do you really need this intermediary var? Using ``loop.last`` seems fairly understandable in the code below :)

c2corg_ui/templates/article/helpers/view.html
dkocich: Yes, that is better. I updated my code and it works with all previously not working articles. I can merge after your approval.

PR #1429, opened 2017-01-25
Fix for GPXupload features validation using DragNDrop.

dkocich:
related to gpx upload issue using drag and drop https://github.com/c2corg/v6_ui/issues/1143#issuecomment-275087238

I moved `validateFeatures` to the `app.utils` because it is now used in two places...

I tried it on localhost and this solution works without issue.

### DIFF
--- a/c2corg_ui/static/js/gpxupload.js
+++ b/c2corg_ui/static/js/gpxupload.js
@@ -5,8 +5,6 @@ goog.require('app');
 /** @suppress {extraRequire} */
 goog.require('ngeo.filereaderDirective');
 goog.require('ol.format.GPX');
-goog.require('ol.Feature');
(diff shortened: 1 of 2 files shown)


asaunier:
> I moved validateFeatures to the app.utils because it is now used in two places...

Is it really used in 2 places?
Actually both approaches trigger a call to ``handleFeaturesUpload_`` in the map directive:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/map.js#L641

I have the feeling it would make more sense to move ``validateFeatures`` to the map directive and to call it into ``handleFeaturesUpload_``.

dkocich:
Ok, I moved it to `map.js` to use it only in `handleFeaturesUpload_`

PR #1432, opened 2017-01-25
Support blocking user accounts

tsauerwein:
Moderators now have an entry "Block/unblock account" in the context menu on profile pages.

Requires https://github.com/c2corg/v6_api/pull/614

image

### DIFF
--- a/c2corg_ui/static/js/user.js
+++ b/c2corg_ui/static/js/user.js
@@ -120,4 +120,13 @@ app.UserController.prototype.hasEditRights = function(doctype, options) {
   return this.auth.hasEditRights(doctype, options);
 };
 
+
+/**
+ * @return {boolean}
+ * @export
+ */
+app.UserController.prototype.isModerator = function() {
+  return this.auth.isModerator();
+};
+
 app.module.controller('AppUserController', app.UserController);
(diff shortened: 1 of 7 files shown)


c2corg_ui/templates/profile/helpers/view.html
```diff
@@ -56,7 +56,7 @@
       <p class="float-button-text" translate>Add images</p>
     </div>
 
-    <div class="dropdown" ng-if="userCtrl.hasEditRights('profiles', ${options | n})">
+    <div class="dropdown" ng-if="userCtrl.hasEditRights('profiles', ${options | n}) || userCtrl.isModerator()">
```
asaunier: This is redundant because moderators already have editing rights on profiles:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/auth/authservice.js#L111-L117

c2corg_ui/static/js/user.js
```diff
@@ -120,4 +120,13 @@ app.UserController.prototype.hasEditRights = function(doctype, options) {
...
 };
 
+
+/**
+ * @return {boolean}
+ * @export
+ */
+app.UserController.prototype.isModerator = function() {
+  return this.auth.isModerator();
+};
```
asaunier: I don't have a strong opinion but is this alias really needed? You can use ``userCtrl.auth.isModerator()`` in the templates or partials. For instance https://github.com/c2corg/v6_ui/blob/1a6a6f27fcd1ed8b0a5421bcd3e167223112bda7/c2corg_ui/templates/article/edit.html#L82

tsauerwein: I like to keep the logic in templates to a minimum because the code in templates in not compiled and is harder to test.

PR #1436, opened 2017-01-25
Fix flex lists display

brunobesson:
To ensure a correct positioning of items in last row, we use the technique described here: http://stackoverflow.com/a/22018710, adding empty div elements at the end.

The base size for items is 300px, except for areas in preferences and users in following, which use a 200px basis. The sizes could be adjusted following your comments. The new behaviour is:

* the items have a min size of 300px (resp. 200px)
* they expand until there is place for another item in the row
* last row items are aligned to the left and have the same size as the others

Note: the less code is quite a mess on this matter (sorry), so I hope I didn't introduce bugs. Please test it carefully.

### DIFF
--- a/c2corg_ui/static/partials/advancedsearch.html
+++ b/c2corg_ui/static/partials/advancedsearch.html
@@ -1,8 +1,13 @@
-<div class="list" app-loading>
+<div class="list advanced-search" app-loading>
   <div id="card{{doc.document_id}}" class="list-item" ng-class="[searchCtrl.doctype, {'highlight': doc.document_id === searchCtrl.highlightId}]"
        ng-repeat="doc in searchCtrl.documents track by doc.document_id">
     <app-card app-card-doc="::doc" ng-mouseenter="searchCtrl.onMouseEnter(doc.document_id)"
               ng-mouseleave="searchCtrl.onMouseLeave(doc.document_id)"></app-card>
   </div>
+  <!--
+    add 5 empty items to fix last row item width
+    see http://stackoverflow.com/a/22018710
+  !-->
+  <div class="list-item" ng-repeat="n in [].constructor(5) track by $index"></div>
 </div>
 <p translate class="text-warning text-center no-results" ng-class="{'show': searchCtrl.noResults}">No results</p>
--- a/c2corg_ui/static/js/constants.js
+++ b/c2corg_ui/static/js/constants.js
@@ -11,7 +11,7 @@ app.module.constant('constants', app.constants);
 
 app.constants = {
   SCREEN: {
-    SMARTPHONE : 620,
+    SMARTPHONE : 768,
     TABLET : 1099,
     DEKTOP : 1400
   },
--- a/c2corg_ui/templates/following.html
+++ b/c2corg_ui/templates/following.html
@@ -2,7 +2,7 @@
 from c2corg_common.attributes import mailinglists
 %>
 <%inherit file="base.html"/>
-<%namespace file="helpers/common.html" import="show_title"/>
+<%namespace file="helpers/common.html" import="show_title, generate_empty_list_items"/>
 <%block name="pagetitle"><title ng-bind="mainCtrl.page_title('Followed users')">${show_title('Followed users')}</title></%block>
 <%block name="metarobots"><meta name="robots" content="noindex,nofollow"></%block>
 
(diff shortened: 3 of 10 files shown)


brunobesson:
The inital solution I was looking at, which has cleaner code (because we do not create dummy elements) was this one: https://www.fourkitchens.com/blog/article/responsive-multi-column-lists-flexbox

But managing media queries with all the cases we know have (different menu sizes, w(/o) map, etc) make it a very difficult task. Hence the other implementation I chose.

c2corg_ui/static/partials/advancedsearch.html
```diff
@@ -4,5 +4,6 @@
     <app-card app-card-doc="::doc" ng-mouseenter="searchCtrl.onMouseEnter(doc.document_id)"
               ng-mouseleave="searchCtrl.onMouseLeave(doc.document_id)"></app-card>
   </div>
+  <div class="list-item" ng-repeat="n in [].constructor(5) track by $index"></div>
```
asaunier: What about using ``Array(5)`` instead of ``[].constructor(5)``?

brunobesson: To be honest, I am not fluent with angular yet, so this is a solution I found on the web after several unsuccessful attempts. If it is working, I'll do it. Does it?

asaunier:
>  this is a solution I found on the web after several unsuccessful attempts

Indeed, for instance here http://stackoverflow.com/a/32518933

In my Chromium console ``Array(5)`` seems to be equivalent to ``[].constructor(5)`` but I don't know if it's OK for Angular and in other browsers. I was just curious if this alternate syntax (shorter and a bit clearer?) would work :)

brunobesson:
>I was just curious if this alternate syntax (shorter and a bit clearer?) would work

Indeed, and if it's the case, I would prefer this one.

stef74:
I have tested and just identifie one strange thing ... other seems good (map/no map/associated doc ...route/ outing / user)

Only associated route have strange behavior
capture d ecran 2017-01-26 a 20 21 17

asaunier:
I have tested the branch with Chrome, FF and Safari. It seems to work fine on various pages and screen widths. I see no change in the home feed though.

@brunobesson So the 5 empty divs are used to make sure cards are displayed with a consistent width on the last line, right? Perhaps you should comment this somewhere because it will be hard to figure this out when reading the code later.

Not sure to understand the problem reported by @stef74. Could you repost your screenshot please?

brunobesson:
>Not sure to understand the problem reported

Neither do I, there is no screenshot :wink: 

>Perhaps you should comment this somewhere because it will be hard to figure this out when reading the code later.

Last time I left a comment in the code you didn't like it :confused: OK, I admit it was a reference to a PR.

To explain the hack: we add 5 empty items (assuming we won't have a line with more than 6 items (otherwise we should add some more...)) to ensure that items on the last row have the same width as above. But those empty items have a 0 height so stay invisible.

I'll wait for @stef74 feedback and make adjustments then.

stef74:
I understand tha I have to wait the upload finish before publish a comment.

Look on Mont-Blanc waypoint on associated route the two end card iwth more resultat and add route are have not the same size than other. 

Sorry I have no material to reproduce until saturday evening.

brunobesson:
>Look on Mont-Blanc waypoint on associated route the two end card iwth more resultat and add route are have not the same size than other.

I don't observe anything specific (at least with at most 6 cards in a row). Unless you're speking of the height fo the cards, which is not the same for each row?

stef74:
Look here (is screenshot of prod ... so result is correct, but I remenber si not correct on you branch)
capture

the two specic card have not the same with than other ...

brunobesson:
I don't observe (or don't understand) what you're telling. Without a screenshot I can't go further.

stef74:
Yes but screenshoot is on my desktop ... cannot acces until saturday evening.

I suspect than the "AJOUTER *****" is too long in some case (318px)

brunobesson:
>I suspect than the "AJOUTER *****" is too long in some case (318px)

Let's way until next week then. As for the long button text, I think I made some adjustments in transifex (it's too long for mobile). Or maybe we could introduce line breaks, but I don't like it.

brunobesson:
>I see no change in the home feed though

@asaunier No I haven't changed anything there. What should be corrected?

Issue #1353, opened 2017-01-12
Map does not display on mobile


desnoes:
The map is always hidden even when i click on the "world" button.
Chrome version 51.0.2704.81
Android: 6.0.1

This bug appeared since the correction of issue #1312.

It seems that @stef74 also observed this bug. Could you confirm?

brunobesson:
There can be absolutely no relation between #1312 and the bug you are observing :wink: More likely other commits are the cause of this.
Still, I can't observe them.
Can you confirm it is working in prod but not on demo?

desnoes:
>Can you confirm it is working in prod but not on demo?

I confirm. In prod, i click on 'hide list' => a grey square appears on the screen => i click on the world button => the map appears at the same location of the grey square.
On the map, if i click on 'show list' => the map is hidden and the cards are displayed.

brunobesson:
>i click on 'hide list'

What? But there is (or should be) no hide list button on mobile! (at least on the demo). How can you do that? Do you have a screenshot?

desnoes:
> What? But there is (or should be) no hide list button on mobile! (at least on the demo). How can you do that? Do you have a screenshot?

My description concerned the usage in prod, not in demo.

Here is a screen shot of what i get on demo. When i click on the wordl button, the screen remains unchanged, the map is not displayed.
![screenshot_20170112-135651](https://cloud.githubusercontent.com/assets/19892089/21891093/53b09768-d8d1-11e6-856f-99efb0f5d282.png)

brunobesson:
OK then. That's strange, and I cannot help further, because all my devices work fine...

AlexisDemongeot:
The rounded earth-like button works fine on : 

Windows phone 10.0.14393.448
Microsoft Edge 38.14393.448.0

Android 7.1.1
Chrome 55.0.2883.91

Doesn't work on Android stock navigator (apprently not maintened) : 42.0.2311.1185 (747cb06)

stef74:
Not working on firefox desktop (I reduce the size of screen) vesrion ESR 45.6 (the last one une ESR branch)

desnoes:
I have updated Chrome and it works (sorry for the inconvenience, i thought my version was the last one but in fact it wasn't).
@stef74 : can i close the issue ?

stef74:
Nopt work on firefox ... work on chrome

AlexisDemongeot:
Apparently not work on stock browser too : 
[...]

PR #1438, opened 2017-01-26
Fix advanced search map display on mobile

brunobesson:
* now works on FF (and most likely on Safari and other failing browsers too, to be tested), see #1353
* correct dimensions (height wasn't computed properly)

### DIFF
--- a/less/c2corg_ui.less
+++ b/less/c2corg_ui.less
@@ -340,7 +340,7 @@ app-pagination.areas, app-pagination.articles app-pagination.books {
   transition: .3s;
 
   @media @phone {
-    margin-left: 0;
+    left: 0;
     bottom: 40px;
     position: fixed;
     opacity: 0;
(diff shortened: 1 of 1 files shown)


dkocich:
Good. I just spent some time looking for a solution. I will try your branch on my mobile FF which wasn't working.

asaunier:
> most likely on Safari (...), to be tested

OK, I'll test the branch on my macbook, hopefully tonight. Thanks :)

HSERGERAERT:
Still not working for me, I use IOS 10.2 (one of the last), I tested with safari and chrome, on both bowsers the map button becomes grey but no map at all happens...

brunobesson:
It hasn't been pushed on demo yet, it's ok

stef74:
Work onf FF ubuntu. I'm waiting push in demo to test on safari (desktop and mobile).

stef74:
@asaunier do you have install the ui in your mac directly or through a linux virtualised ?

asaunier:
I have just tested the branch in my instance: it works fine with Chrome, FF and Safari on my MacBook (with a narrow window simulating a mobile phone). Thanks also for fixing the map height!

Looks good to me then.

> do you have install the ui in your mac directly or through a linux virtualised ?

I have not installed the UI on my mac. I have a personal instance on the dev server.
Using a docker-based environment would make sense but it is not available yet.

PR #1442, opened 2017-01-27
Add article associations on add/edit page

dkocich:
I noticed that it is not possible to add associations to the currently created/edited article. This PR allows to:
- associate other document types (wrcb)
- associate users (should it be allowed?)

### DIFF
--- a/c2corg_ui/templates/article/edit.html
+++ b/c2corg_ui/templates/article/edit.html
@@ -6,7 +6,8 @@ from c2corg_common.attributes import default_langs, activities, article_categori
   updating_doc = article_id and article_lang
 %>
 <%namespace file="../helpers/common.html" import="show_title"/>
-<%namespace file="../helpers/edit.html" import="show_save_confirmation_modal, show_preview_container, show_editing_buttons"/>
+<%namespace file="../helpers/edit.html" import="show_save_confirmation_modal, show_preview_container, show_editing_buttons,
+  show_editing_associated_waypoints, show_editing_associated_routes, show_editing_associated_articles, show_editing_associated_books"/>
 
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/article/edit.html
```diff
@@ -109,6 +110,46 @@ <h2 class="heading show-phone" translate>general informations</h2>
...
+            <app-simple-search parent-id="article.document_id"
+              app-select="editCtrl.documentService.pushToAssociations(doc)" dataset="wrcb"></app-simple-search>
+            ${show_editing_associated_waypoints('article')}
+            ${show_editing_associated_routes('article')}
+            ${show_editing_associated_articles('article')}
+            ${show_editing_associated_books('article')}
+          </div>
+
+          <div id="participants-group" class="form-group  full">
+            <label translate>associated_users</label>
```
asaunier: That's probably not consistent to be able to associate users to a collaborative article. I think it should be possible only for personal articles.

dkocich: What if the user decides to change his personal article to collaborative - should all the user associations be deleted on API automatically in that moment? Maybe I should not allow it here if we aren't sure about this because it will just result in many user associations?

asaunier: That's a good question!
I have no answer. @stef74 @fbunoz ?

That's one of many reasons to disable associating users to articles :P

stef74: no real opinion ... but if it simple go ! This association on my point of view is not critical, but i'm probably not the right expert.

dkocich: I cherry-picked 2nd commit with users to another branch. We can merge If the first is ok.

Issue #1447, opened 2017-01-29
Card Layout


stef74:
Some optimization on card layout

- [ ] Remove some icon
- [ ] More compact 
- [ ] Area on dedicated line
- [ ] Rating and elevation information on same line

Other things possible (discussion in progress)

PR #1449, opened 2017-01-30
Card layout optim

stef74:
#1447


stef74:
capture d ecran 2017-01-29 a 22 02 45
capture d ecran 2017-01-29 a 22 02 05
capture d ecran 2017-01-29 a 22 00 52

asaunier:
@stef74 Are you sure you want to upload 4MB screenshots to this PR? :P 

On a Mac you can select a part to capture using the following combination: Cmd+Shift+4
https://support.apple.com/en-us/HT201361

I recommend to also resize the screenshots using for instance Gimp.

Last, please use an explicit title for the PR and commits, it helps understanding what they are for.

stef74:
capture d ecran 2017-02-05 a 14 42 19
capture d ecran 2017-02-05 a 14 41 24
capture d ecran 2017-02-05 a 14 40 55

Discussion is here : https://trello.com/b/TD2kEghs/design-cards

stef74:
I have two issues : 

1. Text hidden area

I want to hide area if is too long. 

It's work if all text are the same, like here : 
capture d ecran 2017-02-06 a 21 05 59

Not work if area are different (card size changing) : 
capture d ecran 2017-02-06 a 21 12 15

I have a bad idea to force the max-width at fix value ... it's work but not very "flexible"

2. Align area on left

Before text hidden work this point work, by use align-items space-between.

I need some help, @brunobesson an idea ?

stef74:
Ok 1st point is fixed by add min width at card level.

stef74:
capture d ecran 2017-02-06 a 22 32 01

Last issue fixed ... aera aligned at right and ellipsis

stef74:
@asaunier @brunobesson Ready for review. The work group have validated the design.

brunobesson:
I have not followed the decisions made, but I have some things that look strange or unlogical to me. If it's a choice, let's be it nonetheless.

* Is the orange surrounding the outings in feed voluntary?
* Placing the areas after the date in outings mixes both things together, so the area cannot be distinguished easily
* The date is bigger than the outing's name?
* The area on the opposite is positioned on the right hand side for e.g. routes. This is not consistent. Is it the only item aligned right?
* the (new) waypoints layout is too different from others (an icon alone on the left hand side, title moved to the right while for outings and route it starts on the left, etc)

I apologize if you have already answered all these points. I'll do a code review now.

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -1,30 +1,31 @@
...
 
   <div class="list-item-info">
-    <div class="list-item-title-lang margin-b-sm">
+    <div class="list-item-title-lang">
       <span class="list-item-title"><p>{{::locale.title}}</p></span>
       <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{::locale.lang}})</span>
     </div>
 
-    <div class="flex wrap-row margin-t-sm center-items-align">
+
```
brunobesson: 2 line returns is too much?

c2corg_ui/static/partials/cards/routes.html
```diff
@@ -30,32 +35,30 @@
...
-    </div>
-
-    <div class="full-rating flex" ng-init="fullRatings = cardCtrl.getFullRatings()">
-      <span class="icon-rating" ng-if="::fullRatings"></span>
-      <div ng-repeat="(attr, val) in ::fullRatings" uib-tooltip="{{attr | translate}}" class="rating" tooltip-append-to-body="true">
-        <span class="value">{{::val}}</span>
-      </div>
-    </div>
-
+
```
brunobesson: Ditto why 2 line returns?

less/feed.less
```diff
@@ -2,7 +2,7 @@
   display: block;
   padding-top: 10px;
   height: 100%;
-  background-color: rgba(229, 229, 229, 0.3);
+  background-color: #dcdcdc;
```
brunobesson: What's the purpose of this change? To avoid rgba codes? The previous value uses alpha chanel. Is this chnage necessary?

brunobesson:
In outings cards: there are some strange alignments:

* Area doesn't seem on the smae baseline?
* Completion icon should be moved too

toto

less/lists.less
```diff
@@ -403,54 +397,23 @@
...
-      flex-flow: wrap;
-      align-content: flex-start;
-      justify-content: center;
-    }
-    .author-name {
-      line-height: normal;
-      word-break: break-all;
-      word-break: break-word; // non-standard for webkit
-      text-align: center;
+      display: none;
```
brunobesson: Could this `display: none` be simply replaced by removing the content? Or is this something shown in the feed but not in lists (or vice versa)?

brunobesson:
Found strange rendering:

hum

less/feed.less
stef74: I have change the value .. the gray is more dark (more constrast).

less/lists.less
stef74: Yes ... in feed you don't show author; So for me it's correct.

stef74:
> 
>     Is the orange surrounding the outings in feed voluntary?

yes ... to increase visibility of outing vs other doc ( validate by work group).

>     Placing the areas after the date in outings mixes both things together, so the area cannot be distinguished easily

Yes ... but's wanted ... I add you comment in trello. probably we need to use not same color (more dark for date or area).

>     The date is bigger than the outing's name?

Yes i'ts wanted to increase visibility

>     The area on the opposite is positioned on the right hand side for e.g. routes. This is not consistent. Is it the only item aligned right?

It's aligned on irght because on outing it's on aligned on right. It's the behavior wanted by work group.

>     the (new) waypoints layout is too different from others (an icon alone on the left hand side, title moved to the right while for outings and route it starts on the left, etc)

Yes no work on it !

stef74:
If I resume change needed : 

- [ ] fix vertical alignement

- [ ] fix extra line 

- [ ] Fix quality align

stef74:
Issue have been fixed. I have also rebased

asaunier:
Thanks for the update/rebase/squashing.
Actually you could have used "fixup" instead of "squash" when grouping the commits => only the initial commit message is used instead of stacking the 10 commit messages (see "git log").

No opinion about the CSS here as well. I prefer that @brunobesson makes the review.
For the layout, WP looks fine but I am not very convinced by route and outing cards.
For me the content are too shifted to the left and the space usage is not very optimal with blank space in the middle. I don't understand why dates are bigger than the rest of the content. I would that it's not the correct way to highlight this data => should be placed to the left in its own column.

capture d ecran 2017-02-08 a 23 09 05


Placing the regions to the right sometimes creates weird rendering:

capture d ecran 2017-02-08 a 23 06 51

And it's not consistent with WP cards where regions are aligned left just after the elevation.

stef74:
@asaunier could you express your remark concerning layout on trello directly to share with other.

stef74:
I have try to create icon date with : https://www.sitepoint.com/create-calendar-icon-html5-css3/ ... not working. I have try a basic implementation (remove all content of card and just show a static date).

For split date in day month years ... it's easy with angular : https://docs.angularjs.org/api/ng/filter/date

stef74:
Create an icon containing the date see my 1st link.

brunobesson:
>Create an icon containing the date see my 1st link

My question was rather: for what purpose? And why not using font icons instead? E.g. glyphicons provide a calendar icon (https://getbootstrap.com/components/), so does font-awesome (I don't remember if we use it). We could also quite easily build our own.

brunobesson:
Sorry, maybe I misunderstood. You want to display the date in a nice way, within a calendar, is it so?

brunobesson:
OK. Then what? Are you asking for help? Feedback?

stef74:
I have issue to create style "calendar" (but tested in v6 context)
Now I have retry but out of v6 conrtext. And it's work ... but probably too big. We can give some space with use short month ( 'MMM' in angular).
[icon.zip](https://github.com/c2corg/v6_ui/files/766766/icon.zip)

I retry to implement it card this week-end.

stef74:
I have implement icon date ... 
capture d ecran 2017-02-19 a 14 45 38
capture d ecran 2017-02-19 a 14 41 09

But I have an issue with text-overflow. 
capture d ecran 2017-02-19 a 14 38 46

For me the issue is that the

stef74:
- [ ] add translation directive to icon-date

asaunier:
> I have implement icon date

Nice :)
Would it be possible that the icon shows the year instead of the weekday?
The weekday is not that important whereas the year can be critical, especially when you list outings in a route page or in the outing advanced search page.

stef74:
@asaunier yes possible and easy ... just change angular option

stef74:
Ok I have fix last issue : 
capture d ecran 2017-02-21 a 21 07 43

Ok for review (I squash after).

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -1,30 +1,36 @@
...
-        <span class="value-title glyphicon glyphicon-time"></span>
-        <span class="value">{{ ::doc['date_end'] | date: 'dd/MM/yyyy'}}</span>
-      </span>
-
+    <div class="flex center-items-align">
       <span class="elevation" ng-if="::doc.elevation_max" uib-tooltip="{{'elevation_max' | translate}}">
         <span class="value-title glyphicon glyphicon-sort-by-attributes rotate-180"></span>
         <span class="value">{{::doc['elevation_max']}}&nbsp;m</span>
-      </span>
+      </span>
```
dkocich: unnecessary trailing spaces

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -50,5 +55,5 @@
       <span class="quality-{{::doc['quality']}}"></span>
     </div>
   </div>
-
+</div>
```
dkocich: indentation inside `<a></a>` ?

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -1,30 +1,36 @@
...
 
-      <span class="location" ng-if="::doc['areas']">
-        <span class="glyphicon glyphicon-map-marker"></span>
+      <span class="flex row-reverse location" ng-if="::doc['areas']">
         <span class="value">{{::cardCtrl.showArea(doc['areas'])}}</span>
       </span>
+   </div>
+
+    <div class="flex wrap-row center-items-align">
+
```
dkocich: empty div?

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -1,30 +1,36 @@
...
         <span class="value">{{::doc['elevation_max']}}&nbsp;m</span>
-      </span>
+      </span>    
 
-      <span class="location" ng-if="::doc['areas']">
-        <span class="glyphicon glyphicon-map-marker"></span>
+      <span class="flex row-reverse location" ng-if="::doc['areas']">
         <span class="value">{{::cardCtrl.showArea(doc['areas'])}}</span>
       </span>
+   </div>
```
dkocich: indent

less/icon-date.less
```diff
@@ -0,0 +1,46 @@
...
+		background-color: #fd9f1b;
+		border-bottom: 1px dashed #f37302;
+		box-shadow: 0 2px 0 #fd9f1b;
+		font-size: 0.8em;
+	}
+	.year {
+		position: absolute;
+		bottom: 3px;
+		color: @C2C-orange;
+                font-size: 0.8em
```
dkocich: indent and missing semicolon

dkocich: + all less files use 2 spaces tab size

brunobesson:
Some aesthetics remarks:

* couldn't we simplify the calendar a little bit and remove the "pages indicator" in the bottom (I hope you understand what I mean)?
* could it be a little smaller? I think it catches the eye too much
* I don't understand how the completion indicator is handled: you seem to try to use narrowers margins, but  the indicator is a problem for you, so you add some more margin to the right of the author. This is strange when it appears alone on its line (e.g. Nouveau Chamechaude : Voie Normale test). I would rather ensure that activity icons and author name are sticked to the bottom whatever the card height.
* the completion icon itself could be move elsewhere, or included in the layout (because currently it has an absolute position independant of the rest of the card)
* the position of the activity icons is not optimal now that there is this quite huge calendar. Maybe have the activity icons placed under the calendar, with a smaller calendar?

dkocich:
connecting lines of the feed are not visible after the change of background colour (previously vs. now)
imageimage

stef74:
Do you remove also the little red line (under month) ?

For quality/completion I prefer to do on other PR. 

For move activity icon ... i'm not sure is and good idea. On simple outing that probably increase height of card. I propose to wait additionnal information in card to review again the layout (rating, gps, image etc ...)

I propose to fix : 
- [ ] simplify layout calendar
- [ ] reduce calendar size
- [ ] change connected line color

stef74:
capture d ecran 2017-02-22 a 21 51 50

Major issue are fixed.

For alignement of activities and author at bottom of card, I try to do in other PR (I have try but not found a solution).

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -1,54 +1,46 @@
...
 
-  <div class="list-item-info">
-    <div class="list-item-title-lang margin-b-sm">
-      <span class="list-item-title"><p>{{::locale.title}}</p></span>
-      <span class="list-item-lang" ng-if="cardCtrl.lang != locale.lang">({{::locale.lang}})</span>
+  <div class="flex wrap-row center-items-align">
+    <div class="icon-date">
+      <span class="year">{{ ::doc['date_end'] | date: 'yyyy'}}</span>
+      <span class="month">{{ ::doc['date_end'] | date: 'MMM' | translate}}</span>
+      <span class="number">{{ ::doc['date_end'] | date: 'dd' }}</span>
```
asaunier: What about using class "day" instead?

less/icon-date.less
```diff
@@ -0,0 +1,46 @@
+.icon-date {
```
asaunier: Indentation is incorrect in this file: please use 2-space indentation as in the other LESS files.

stef74:
@asaunier branch redy (conflict solve, squashed, rebased on new "master" ident icon-date fix). I have add more margin between card. other change needed ?

asaunier:
@stef74 @desnoes Let me know when I can publish this branch on the demo (this will not be a release).

desnoes:
For me it's ok, but i prefer that stef takes the decision.

stef74:
@asaunier you can !

asaunier:
I have just updated the demo and used the current branch card2.

stef74:
It's not correct ... I have issue of color

asaunier:
I have deployed the docker image "card2" available at https://hub.docker.com/r/c2corg/v6_ui/tags/
It is automatically built when the branch on github is updated.

> It's not correct ... I have issue of color

Actually the result seems similar to the branch I created last week based upon your commits. Are you sure all your changes are committed and pushed?

stef74:
Yes I have identifiend the issue ... ex : 
background

This color are not recognized ... I'm pretty sure on my ubuntu i'ts works !

stef74:
@asaunier I have fix color with new hexa code. You can redeploy.

asaunier:
Updated

stef74:
some issues 
img_1170
img_1171
img_1172
img_1173

stef74:
@asaunier I have done some fix. But impossible to test on safari without the demo updated (test only firefox/chrome possible). Could you reupdate the demo with this Branch ?

asaunier:
Updated

stef74:
Great, major bug are fixed on safari :+1:

stef74:
Final version (with condition) : 
capture d ecran 2017-03-28 a 09 18 03

@asaunier @desnoes ok for you ?

desnoes:
c'est bon pour moi, ça me plaît bien.

----- Mail original -----
De: "stef74" <notifications@github.com>
À: "c2corg/v6_ui" <v6_ui@noreply.github.com>
Cc: "desnoes" <mathieu.desnoes@free.fr>, "Mention" <mention@noreply.github.com>
Envoyé: Mardi 28 Mars 2017 09:22:23
Objet: Re: [c2corg/v6_ui] Card layout optim (#1449)



Final version (with condition) : 
capture d ecran 2017-03-28 a 09 18 03

@asaunier @desnoes ok for you ?

asaunier:
No objection!

asaunier:
Perhaps there is a small problem with the position of the condition rating icon with Chrome (tested with Chromium 56 - OK with FF 52):

capture du 2017-03-28 10-44-52

stef74:
@asaunier Fixed and tested on chromium & FF ( "end" are not recognized by chromium we have to use flex-end).

asaunier:
It's OK now, thanks.

stef74:
I have identified some align trouble on title on chromium. I fix it

asaunier:
I also notice duplicated ratings in the associated route cards when creating an outing (in Chromium and FF):

capture du 2017-03-28 11-12-09

asaunier:
To be honest, as I have already said it, I am not very enthousiastic with the changes about the gray background in the document pages or the black borders of the cards. But I don't really want to debate about those changes, especially if they have been accepted on the forum of the association.

asaunier:
Have you been able to fix the text color in the editing forms? (my instance is no longer set on your branch)

stef74:
Yes I have see for rating and also align trouble.

stef74:
I have fix alignement trouble. I have put on very grey for border, remove gray on list (not on homepage). Fix double rating issue

I propose to stop now ... and let react user in production directly !

stef74:
@asaunier all issue have been fixed include text color in edition.

asaunier:
OK for me

Issue #1451, opened 2017-01-30
Missing translation for licence type


fbunoz:
All translation ID are translated in transifex, but it misses some translations about licence type : 

- [ ] In article edition, the licence types are not translated in the selector : 'collab' and 'personal' even in french (https://www.demov6.camptocamp.org/articles/edit/828400/en)
- [ ] In a page of an image, or in the slideshow, the personal licence is not translated : 'personal' instead of 'personnelle' in FR (https://www.demov6.camptocamp.org/images/828416/fr/-).
- [ ] when editing an image, tranlastion is missing too
- [ ] In the image filter, the licence selector uses the article licence options instead of image licence options : 'collaboratif' instead of 'collaborative' (idem for 'personnel' (https://www.demov6.camptocamp.org/images)

brunobesson:
I have a branch ready for the first 3 points (I added the case of an image edition).

However, there is something confusing for the last one. This is rather a translation key clash.

The image_types are defined as follows:

    image_types = [
        'collaborative',
        'personal',
        'copyright'
    ]

And the articles_types as

    article_types = [
        'collab',
        'personal'
    ]

So there is an overlap in the translations. Currently, both `collab` and `collaborative` are translated as `collaboratif` in french. So there is no female translation currently.
@asaunier @fbunoz what should we do? Do we need to modify the keys in v6_api?

asaunier:
> Do we need to modify the keys in v6_api?

I recommend using the same gender for both ``image_types`` and ``article_types``. After all both are "types". Actually I would have recommended to have ``collaborative`` instead of ``collab`` in ``article_types``. I don't really understand why such a difference, especially since both attributes have ``personal`` as possible value.

Since changing those values would impact the database itself (those values are used as "enumerations" for ``image_types`` and ``article_types`` fields), it's not a trivial change (but not impossible).

brunobesson:
>I recommend using the same gender for both

I would agree with this advice. If there is no opposition, then we could create an issue in v6_api for using `collaborative` instead of `collab` for articles?

AFAIK #1454 is enough to solve this issue then?

asaunier:
> I would agree with this advice. If there is no opposition, then we could create an issue in v6_api for using collaborative instead of collab for articles?

Up to you. But if we can live with it, I would recommend to leave it as it. Else we need to create an Alembic script to update the values in the articles table. Not sure it's worth doing it!?

fbunoz:
[...]

PR #1454, opened 2017-01-31
Add translations for image / article license

brunobesson:
This PR solves the first 3 points of #1451.
However, there are pending question concerning the last point.

### DIFF
--- a/c2corg_ui/templates/article/edit.html
+++ b/c2corg_ui/templates/article/edit.html
@@ -76,21 +76,21 @@ from c2corg_common.attributes import default_langs, activities, article_categori
             </div>
 
             <div class="form-group half half-group">
-              <div ng-class="{ 'has-success': article.article_type}">
+              <div ng-class="{ 'has-success': article.article_type}" ng-init="article_types = ${article_types}">
                 <label><span translate>article_type</span> *</label>
                 % if updating_doc:
                   <select ng-if="userCtrl.auth.isModerator()" class="form-control" ng-model="article.article_type"
-                          ng-options="type as mainCtrl.translate(type) for type in ${article_types}" ><option value=""></option></select>
+                          ng-options="type as mainCtrl.translate(type) for type in article_types" ><option value=""></option></select>
                   <select ng-if="userCtrl.auth.isModerator() == false && article.article_type =='collab'"
                           class="form-control" ng-model="article.article_type"
-                          ng-options="type as mainCtrl.translate(type) for type in ${article_types} | filter: 'collab'" ><option value=""></option>
+                          ng-options="type as mainCtrl.translate(type) for type in article_types | filter: 'collab'" ><option value=""></option>
                   </select>
                   <select ng-if="userCtrl.auth.isModerator() == false && article.article_type =='personal'"
                           class="form-control" ng-model="article.article_type"
-                          ng-options="type as mainCtrl.translate(type) for type in ${article_types}" ><option value=""></option>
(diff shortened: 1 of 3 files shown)


c2corg_ui/templates/article/edit.html
```diff
@@ -80,17 +80,17 @@ <h2 class="heading show-phone" translate>general informations</h2>
                 <label><span translate>article_type</span> *</label>
                 % if updating_doc:
                   <select ng-if="userCtrl.auth.isModerator()" class="form-control" ng-model="article.article_type"
-                          ng-options="type as mainCtrl.translate(type) for type in ${article_types}" ><option value=""></option></select>
+                          ng-options="type as mainCtrl.translate(type) for type in ${article_types} | translate" ><option value=""></option></select>
```
asaunier: I don't think this correction (and the other similar ones below) is correct because the options are already translated, see ``type as mainCtrl.translate(type)`` => it does not make sense to make 2 calls to the ``translate`` directive.

I don't know exactly why the translations are not done as expected (are they?). For instance in other editing forms they seem to work as is:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/edit.html#L296

We already had this problem for some other attributes in the editing forms. I think the problem is related to the direct use of the mako variable ``${article_types}`` in ``ng-options``.
An alternate syntax that seems to work is to define an angular variable using ``ng-init`` and to use  it in the ``ng-options``. See for instance 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/outing/edit.html#L269-L272

brunobesson:
@asaunier your suggestion for using `ng-init` seems to be fine. I updated the PR accordingly.

c2corg_ui/templates/image/edit.html
```diff
@@ -141,8 +141,8 @@ <h2 class="heading show-phone" translate>general informations</h2>
 
               <div ng-class="{ 'has-success': image.image_type}">
                 <label><span translate>image_type</span> *</label>
-                <div class="input-group">
-                  <select class="form-control" ng-model="image.image_type" ng-options="rat as mainCtrl.translate(rat) for rat in ${image_types}"><option value=""></option></select>
+                <div class="input-group" ng-init="image_types = ${image_types}">
+                  <select class="form-control" ng-model="image.image_type" ng-options="rat as mainCtrl.translate(rat) for rat in image_types"><option value=""></option></select>
```
asaunier: Minor but I would recommend to use ``type`` instead of ``rat`` here because ``rat`` is probably a copy-paste from another part of the code, related to "ratings". ``rat`` is not really consistent here.

c2corg_ui/templates/image/edit.html
```diff
@@ -1,11 +1,11 @@
 <%!
 from c2corg_common.attributes import default_langs, activities, image_types, image_categories
 %>
-<%inherit file="../base.html"/>
+    <%inherit file="../base.html"/>
```
asaunier: Why adding an indentation here? Ditto at line 8?

brunobesson: Grmbl. Probably because I'm trying a new editor, vscode instead of the usual atom. I'll correct this (and I should have a better look at my commits to avoid stupid things like this, I'm sorry)

c2corg_ui/templates/image/edit.html
brunobesson: Good catch. I was focusing on my changes only, but this makes sense.

Issue #1441, opened 2017-01-26
In image edition, do not allow to change the licence to copyright


fbunoz:
The licence of an image should be changed to copyright only by a moderator.
But for now, everybody can change the licence of its personal images to copyright.
The copyright option must be limited to moderator only.
A simple member must be able to choose between personal and collaborative licence only.

There is no problem in the image upload tool, only in the image edit form when we modify an image.

asaunier:
It would make sense to appky a kind of filter in the image editing form at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/image/edit.html#L145
the same way than it's done in the image uploader:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/imageuploader.html#L83
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/imageuploader.js#L480-L489

asaunier:
> The licence of an image should be changed to copyright only by a moderator.

I have suggested a PR to fix the problem in the UI. #1459

But some checks might be needed in the API as well. Please open an API issue then.

desnoes:
the bug has been fixed

PR #1459, opened 2017-01-31
Fix image type field behaviour in image editing form

asaunier:
- Moderators can change to whatever type
- Non moderators can only change type from 'personal' to 'collaborative'. If type is 'collaborative' or 'copyright' the field is disabled.

Related to https://github.com/c2corg/v6_ui/issues/1441

### DIFF
--- a/c2corg_ui/static/js/editing/imageediting.js
+++ b/c2corg_ui/static/js/editing/imageediting.js
@@ -51,6 +51,11 @@ app.ImageEditingController = function($scope, $element, $attrs, $http,
    */
   this.exposureError = false;
 
+  /**
+   * @type {?string}
+   * @export
+   */
+  this.initialImageType = null;
 };
 goog.inherits(app.ImageEditingController, app.DocumentEditingController);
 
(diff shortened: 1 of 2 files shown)


c2corg_ui/static/js/editing/imageediting.js
```diff
@@ -51,6 +51,11 @@ app.ImageEditingController = function($scope, $element, $attrs, $http,
    */
   this.exposureError = false;
 
+  /**
+   * @type {?string}
+   * @export
+   */
+  this.initial_image_type = null;
```
tsauerwein: `this.initialImageType`?

asaunier: I wanted to be consistent with the attribute in the model which is ``image_type``.

tsauerwein: Elsewhere `imageTypes` is used.

asaunier: OK :)
I will change to ``this.initialImageType`` if you recommend to do so.

Issue #1457, opened 2017-01-31
Allow to export coordinates of a waypoint


brunobesson:
Similar to what has been done for routes and outings, some users express their desire to download coordinates of a waypoint.

asaunier:
It is probably possible to use quite the same code than for routes and outings:
https://github.com/c2corg/v6_ui/blob/2d099d64fffaf8e1cc1029791beafeb2cbe03c1b/c2corg_ui/templates/helpers/view.html#L834
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/trackdownload.js
except that in the current tools, tests are made to ignore points. Perhaps we'll have to pass the type of geometry to download.

stef74:
A good evolution if to offer possibility to download all waypoint associated to a route !

brunobesson:
@stef74 then it's another issue :wink:

desnoes:
works fine on desktop (could not test on mobile).

stef74:
Ok ... text is probably too long

PR #1460, opened 2017-02-01
Generalize the track downloader to point features as well

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1457

I have noticed that the feature title is not available in the exported GPX or KML (including for routes and outings). I guess it would make sense. For waypoints, the ``type`` is exported though.

But I have not figured out yet how to pass a list of attributes to be exported when writing the GPX or KML:
https://github.com/c2corg/v6_ui/blob/download-geom/c2corg_ui/static/js/geomdownload.js#L86-L88
http://openlayers.org/en/latest/apidoc/ol.format.GPX.html#writeFeatures


c2corg_ui/static/js/geomdownload.js
```diff
@@ -56,11 +66,21 @@ app.TrackDownloadController = function(ngeoDownload, mapFeatureCollection) {
...
   var geojson = new ol.format.GeoJSON();
   var features = geojson.readFeatures(this.featureCollection_);
-  features = features.filter(app.utils.isLineFeature);
+  switch (this.featureType) {
+    case 'Line':
+      features = features.filter(app.utils.isLineFeature);
+      break;
+    case 'Point':
+      // export only the waypoint itself, not its associated features
+      features = [features[0]];
```
tsauerwein: `features` also contains the geometries of associated documents? Is this different for routes and outing?

tsauerwein:
> But I have not figured out yet how to pass a list of attributes to be exported when writing the GPX or KML:

Setting the attributes on the feature before exporting should work:
http://openlayers.org/en/master/apidoc/ol.Feature.html#set
http://openlayers.org/en/master/apidoc/ol.Feature.html#getProperties

Maybe using the document slug as file name would also be a good idea.

c2corg_ui/static/js/geomdownload.js
asaunier: Yes, in detail view pages, ``features`` is set with ``mapFeatureCollection``:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/waypoint/view.html#L40-L68

Which is useful to display the associated WP on the map as well.

tsauerwein: So, maybe instead of using `features = features.filter(app.utils.isLineFeature);` for lines, it would be clearer to use `features = [features[0]];` as well?

asaunier: Good question. I don't know.

For a waypoint, it makes to get only the waypoint itself, not the associated ones. So only the first feature.

For routes or outings, we don't want the associated waypoints either, I think (?). But are we sure the route/track in the first (line) feature is enough? We could have a complex geometry with several line features, couldn't we?

tsauerwein: An outing or route has exactly one geometry (either of LineString or MultiLineString). So I guess there is also exactly one feature?

asaunier:
I have updated the PR to take @tsauerwein's suggestions into account:
* export the first feature only, whatever the geometry type (tested with various "weird" GPX files provided in  https://github.com/c2corg/v6_ui/issues/1143) => it also allowed to remove the type parameter I first introduced in the download directive (since no longer needed)
* added the feature name in the exported files

Using the feature slug in the filename would make sense but:
* no one has ever complained about the current naming convention (based upon the document id)
* it would probably require to factorize a bit the slug service config in https://github.com/c2corg/v6_ui/blob/08c60b977e6d20ee2c9a0b7b39774f18174882df/c2corg_ui/static/js/urlservice.js#L14-L22 to have a consistent behaviour in this download directive. But I don't have the motivation to spend some time doing so.

PR #1463, opened 2017-02-01
increase title size

stef74:
Title size are little bit to small in mobile.
Increase the size of title improve readibility vs other information

### DIFF
--- a/less/lists.less
+++ b/less/lists.less
@@ -88,6 +88,7 @@
     font-weight: bold;
     color: graytext;
     margin-right: 5px;
+    font-size: 1.2em;
     p {
       line-height: normal;
       display: block;


stef74:
desktop
mobile

fbunoz:
This PR doesn't change the card title color. Is it already done in another PR ?

stef74:
No we can do in same PR

fbunoz:
But your screenshots show a different color for the title, then I don't know if you forgot the color change or not.

stef74:
I have not changing anything on color ...

fbunoz:
In prod, there is the same color for all text in the card : graytext (~ #AAA).
I don't understand why there is a different color for the title in your screenshots.
Maybe another PR (not in prod) has changed the title color ?

stef74:
No I have tested directly on prod on my browser (commit not tested).

stef74:
I confirmed I have the same rendering in production ... probably due to weight, not color.

fbunoz:
The rendering in production on my screen (with FF 51), by changing only the font size : 

titre_card_1

And by changing the font size + the text color (changed to # 666) : 

titre_card_2

asaunier:
> probably due to weight, not color.

For sure!
The current PR is a one-line change about a ``font-size``: how could it change the color?!

In my instance, the card title are indeed bigger but the color has definitely not changed.

Prod:

capture d ecran 2017-02-02 a 00 13 02

This branch:

capture d ecran 2017-02-02 a 00 12 46

In a route page:

capture d ecran 2017-02-02 a 00 15 58

In my opinion the change is a bit weird (title are too big). For instance in the route page, the card titles are bigger than the "associated outings" section title or the attribute titles.

stef74:
Probably 1.1em is sufficiant. A global rework of font-size is here #1379

fbunoz:
@asaunier : how can you have a different color for the title and the area in prod ? In CSS, the color is the same for both !
Maybe your browser changes the color when the font weight is 700, but it's not already the case (I don't see that on my browsers as you can see on my screenshot).
I understand why you don't find a contrast problem on the card with this browser behavior,. But we must change the color to have the same rendering in all browsers.

> For instance in the route page, the card titles are bigger than the "associated outings" section title or the attribute titles.

On mobile the title is too small, moreover the color is ligt gray (# AAA, it isn't as on your screenshot).
The card title should have the same size as a topic title in a topic list on the forum.

fbunoz:
> how can you have a different color for the title and the area in prod ? In CSS, the color is the same for both !

Sorry, I'm wrong.
The colors are different : 

- title : graytext (~ # aaa on my browser, but it's different on your browsers !)
- date, altitude, area : # a9a9a9
- author : gray (~ # 888 on my browser)
- the default text color for the card is # 919191 (it appears nowhere in an outing card, I don't check for other cards).

We should avoid using named colors, and using only RGB colors, to avoid differences between browsers.

stef74:
Ok I manage in this PR #1379.

Could you confirm RGB color I have to use. I try to do this evening

dkocich:
> We should avoid using named colors, and using only RGB colors, to avoid differences between browsers.

http://htmlcolorcodes.com/color-names/ http://www.w3schools.com/colors/colors_names.asp I don't think that it is problem of named colours. Google said that this is Chromium issue with rendering of some colors.
https://www.sitepoint.com/community/t/colors-are-different-on-safari-v-chrome/103232/9
http://stackoverflow.com/questions/6237455/chrome-renders-colours-differently-from-safari-and-firefox
https://bugs.chromium.org/p/chromium/issues/detail?id=44872

fbunoz:
> I don't think that it is problem of named colours.

Ok, but the rendering on Chrome shows a dark gray, whereas it is a light gray on Firefox. The dev which use Chrome don"t show a problem of contrast and readability.
But thousands users using Firefox have a bad experience of V6 caused by this difference.

> Btw do you have the latest browser?

I use the latest viersion of Firefox, and I see "graytext" as # aaa.
This [page](https://bugs.chromium.org/p/chromium/issues/detail?id=44872) says that Firefox behavior is correct, and Chrome behavior is fail (but it should be fixed since 2010 ?).

fbunoz:
> I don't think that it is problem of named colours. They should standardized and same everywhere.

Your links show the list of standardized named colors.
We can see that "graytext" don't appear in these list, then it isn't a standardized color name.
[This page](http://help.dottoro.com/labpxqqe.php) shows that graytext is a Safari and Chrome specific color name. In fact, it can be a color defined by the OS.

Then, to avoid different rendering between browesr, we shouldn't use non standardized color names.

Issue #1343, opened 2017-01-11
Reduce icons size on map


stef74:
#1309 
reduction on icon on map

do something on : 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/map.js#L440

But not clear what exactly ....

brunobesson:
I started some work on branch map-cons-size: https://github.com/c2corg/v6_ui/tree/map-icons-size

I am not sure what your goal is. To me, the main problem is that highlighted icons are too big (scaled by a factor 2). Currently, I thus apply a factor of 1.2 only, which seems enough.

@stef74 Are there other things to do?

stef74:
I would also reduce the base icon itself

brunobesson:
Which ones? They don't have the same size. E.g. outings seem small enough for me?

stef74:
Hello,

For me all are too big

waypoint is very too big !
Route can be reduce a little to ease the search in some "dense" sector
Outings ... for me te too but ...  and the precision for outings is not
relevant.

The most important point is WP .... they hide some information in the map !



2017-01-17 9:59 GMT+01:00 Bruno Besson <notifications@github.com>:

brunobesson:
@stef74 I rework my PR to ensure all icons have the same size (because OL uses SVG, the width and height defined are relevant and were not all the same) and reduced it some more. Are you OK like this?

See #1465

brunobesson:
Working on demo

PR #1465, opened 2017-02-01
Reduce size of (highlighted) icons in maps

brunobesson:
* Ensure consistent (i.e. same) size for all items.
* Reduce highlight transform (there is one highlighted item per screnshot below)

Relates to #1343

Was (for waypoints, which were bigger than eg. profiles and outings - route were in an in-between state)

certif34

Now (same size for everything).
certif35


asaunier:
Please rebase your branch, you don't need commit 123709f which is already available in master.

Else the new size of the map icons looks good to me, except there is a shift/centering problem with outing icons:
capture d ecran 2017-02-02 a 00 29 38

brunobesson:
I rebased the branch (sorry for this, didn't check) and centered both icons that wren't (local products and outings).

Issue #1323, opened 2017-01-08
Modification follow


stef74:
We need a report of all modification in topoguide.

The first implementation can be only with API ( online exist to consert json in excel ... so moderator can easly do this conversion )

Probably more an API issues

asaunier:
@stef74 @desnoes A UI part is being developped, see #1466 
A UI URL will be avalaible: http://.../whatsnew
Should a link be made available somewhere or will the page only be accessible by typing the URL in the browser address bar?

fbunoz:
It would be useful having a link on the profile of an user, and on each doc type filter page (but I don't know where : at the bottom of the list ?).

desnoes:
@asaunier : This will only be used by moderators => typing URL in the address bar is fine for me.

brunobesson:
Sorry, I don't agree. IMHO it's a nonsense not to have a link to it. As far as I understand, there is only one page, not specific ones for different type of documents?
If so, adding the link to the user menu for moderators until we find a better place is OK for me.

asaunier:
I don't understand why this page would be limited to moderators. It's interesting for any users, including guest users, to see what changes are done. It also insists on the fact that c2corg is a collaborative website.
Anyway there is no access restriction applied to this page: anyone that knows the URL can access it.

Adding the link to the user menu does not make sense to me since it is not a user-related feature.

It could be added in the home page near other links such as the "see outings with my filters" button (see https://github.com/c2corg/v6_ui/issues/1342)

Please note that is is possible to filter the list on a given contributor to see their contributions. For instance ``whatsnew?u=12345``.
This link could be added to the profile page (similarly to the "see this user's outings" button at the beginning of the user's feed).

For the first version of the feature, we could say no explicit link is made available. But that's probably worth thinking how to make the tool public.

brunobesson:
>I don't understand why this page would be limited to moderators

You're right. I had thought likewise in the begining, then forgot.

desnoes:
works very fine on demo (see discussions and comment in https://github.com/c2corg/v6_ui/pull/1466)

PR #1466, opened 2017-02-02
Browse latest changes on the UI

dkocich:
Related to UI issue https://github.com/c2corg/v6_ui/issues/1323

Requires to merge/use API PR to work correctly https://github.com/c2corg/v6_api/pull/620

There is a new page `localhost:6558/whatsnew` which allows browsing of latest changes. ~~It uses [ngTable](http://ng-table.com/) and data are loaded in one request to API. It is possible to use filters, sort values, change page and its size.~~

image

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -428,6 +428,25 @@ app.Api.prototype.readFeed = function(token, lang, userId, isPersonal) {
 
 
 /**
+ * @param {undefined | string} token
+ * @param {?number} userId
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.readWhatsnewFeed = function(token, userId) {
+  var params = {};
+  if (token) params['token'] = token;
+  if (userId) params['u'] = userId;
+
+  var promise = this.getJson_('/documents/changes?' + $.param(params));
+  promise.catch(function(response) {
+    var msg = this.alerts_.gettext('Getting feed data failed:');
+    this.alerts_.addErrorWithMsg(msg, response);
+  }.bind(this));
+  return promise;
+};
+
+
+/**
  * @return {!angular.$q.Promise<!angular.$http.Response>}
  */
 app.Api.prototype.readAccount = function() {
--- a/c2corg_ui/static/js/feed.js
+++ b/c2corg_ui/static/js/feed.js
@@ -28,17 +28,18 @@ app.module.directive('appFeed', app.feedDirective);
  * @param {app.Api} appApi Api service.
  * @param {app.Lang} appLang Lang service.
  * @param {!string} imageUrl URL to the image backend.
+ * @param {ngeo.Location} ngeoLocation ngeo Location service.
  * @constructor
  * @ngInject
  * @struct
  */
-app.FeedController = function(appAuthentication, appApi, appLang, imageUrl) {
+app.FeedController = function(appAuthentication, appApi, appLang, imageUrl, ngeoLocation) {
 
   /**
    * @type {app.Api}
-   * @private
+   * @public
    */
-  this.api_ = appApi;
+  this.api = appApi;
 
   /**
    * @type {app.Authentication}
@@ -66,9 +67,9 @@ app.FeedController = function(appAuthentication, appApi, appLang, imageUrl) {
 
   /**
    * @type {string | undefined}
-   * @private
+   * @public
    */
-  this.nextToken_;
+  this.nextToken;
 
   /**
    * @type {boolean}
@@ -112,6 +113,12 @@ app.FeedController = function(appAuthentication, appApi, appLang, imageUrl) {
    */
   this.isPersonal = !this.userId;
 
+  /**
+   * @type {ngeo.Location}
+   * @public
+   */
+  this.ngeoLocation = ngeoLocation;
+
   this.getDocumentsFromFeed();
 };
 
@@ -123,25 +130,35 @@ app.FeedController = function(appAuthentication, appApi, appLang, imageUrl) {
  */
 app.FeedController.prototype.getDocumentsFromFeed = function() {
   this.busy = true;
-  this.api_.readFeed(this.nextToken_, this.lang_.getLang(), this.userId, this.isPersonal).then(function(response) {
-    this.error = false;
-    this.busy = false;
-    var data = response['data']['feed'];
-    var token = response['data']['pagination_token'];
-    this.nextToken_ = token;
-
-    for (var i = 0; i < data.length; i++) {
-      this.documents.push(data[i]);
-    }
-    if ((token && data.length === 0) || (!token && this.documents.length > 0)) {  // reached the end of the feed - disable scroll
-      this.feedEnd = true;
-    } else if (data.length === 0) { // first fetch with no feed returned.
-      this.noFeed = true;
-    }
+  this.api.readFeed(this.nextToken, this.lang_.getLang(), this.userId, this.isPersonal).then(function(response) {
+    this.handleFeed(response);
   }.bind(this), function() { // Error msg is shown in the api service
     this.busy = false;
     this.error = true;
   }.bind(this));
+};
+
+
+/**
+ * Handles feed processing for Feed.js and Whatsnew.js
+ * @param response
+ * @public
+ */
+app.FeedController.prototype.handleFeed = function(response) {
+  this.error = false;
+  this.busy = false;
+  var data = response['data']['feed'];
+  var token = response['data']['pagination_token'];
+  this.nextToken = token;
+  for (var i = 0; i < data.length; i++) {
+    this.documents.push(data[i]);
+  }
+  // reached the end of the feed - disable scroll
+  if ((token && data.length === 0) || (!token && this.documents.length > 0)) {
+    this.feedEnd = true;
+  } else if (data.length === 0) { // first fetch with no feed returned.
+    this.noFeed = true;
+  }
 };
 
 
@@ -167,7 +184,7 @@ app.FeedController.prototype.createActionLine = function(doc) {
     default:
       break;
   }
-  return line + this.documentType(doc['document']['type']);
+  return line + this.getDocumentType(doc['document']['type']);
 };
 
 
@@ -177,7 +194,7 @@ app.FeedController.prototype.createActionLine = function(doc) {
  */
 app.FeedController.prototype.toggleFilters = function() {
   this.isPersonal = !this.isPersonal;
-  this.nextToken_ = undefined;
+  this.nextToken = undefined;
   this.documents = [];
   this.getDocumentsFromFeed();
   window.scrollTo(0, 0);
@@ -200,7 +217,7 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
  * @export
  * @returns {string}
  */
-app.FeedController.prototype.documentType = function(type) {
+app.FeedController.prototype.getDocumentType = function(type) {
   return app.utils.getDoctype(type).slice(0, -1);
 };
 
--- a/c2corg_ui/static/js/whatsnewfeed.js
+++ b/c2corg_ui/static/js/whatsnewfeed.js
@@ -1,0 +1,83 @@
+goog.provide('app.WhatsnewFeedController');
+goog.provide('app.whatsnewFeedDirective');
+
+goog.require('app');
+goog.require('app.utils');
+
+/**
+ * @return {angular.Directive} The directive specs.
+ */
+app.whatsnewFeedDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'appWhatsnewFeedController as wfeedCtrl',
+    templateUrl: '/static/partials/feedtable.html'
+  };
+};
+app.module.directive('appWhatsnewFeed', app.whatsnewFeedDirective);
+
+
+/**
+ * @param {!angular.Scope} $scope Scope.
+ * @param {app.Authentication} appAuthentication
+ * @param {app.Api} appApi Api service.
+ * @param {app.Lang} appLang Lang service.
+ * @param {!string} imageUrl URL to the image backend.
+ * @param {ngeo.Location} ngeoLocation ngeo Location service.
+ * @constructor
+ * @extends {app.FeedController}
+ * @ngInject
+ */
+app.WhatsnewFeedController = function($scope, appAuthentication, appApi, appLang, imageUrl, ngeoLocation) {
+
+  goog.base(this, appAuthentication, appApi, appLang, imageUrl, ngeoLocation);
+
+  /**
+   * @type {number}
+   * @export
+   */
+  this.userId;
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.currentLang = appLang.getLang();
+};
+goog.inherits(app.WhatsnewFeedController, app.FeedController);
+
+/**
+ * Fills the feed with documents.
+ * Used by ng-infinite-scroll directive in the template.
+ * @export
+ * @override
+ */
+app.WhatsnewFeedController.prototype.getDocumentsFromFeed = function() {
+  this.busy = true;
+  if (this.ngeoLocation.hasFragmentParam('u')) {
+    this.userId = parseInt(this.ngeoLocation.getFragmentParam('u'), 10);
+  }
+  this.api.readWhatsnewFeed(this.nextToken, this.userId).then(function(response) {
+    this.handleFeed(response);
+  }.bind(this), function() { // Error msg is shown in the api service
+    this.busy = false;
+    this.error = true;
+  }.bind(this));
+};
+
+
+/**
+ * @param {appx.DocumentChange} doc
+ * @returns {string}
+ * @export
+ */
+app.WhatsnewFeedController.prototype.getVersionUrl = function(doc) {
+  return '/{module}/{id}/{lang}/{version}'
+    .replace('{module}', app.utils.getDoctype(doc.document.type))
+    .replace('{id}', String(doc.document.document_id))
+    .replace('{lang}', doc.lang)
+    .replace('{version}', String(doc.version_id));
+};
+
+
+app.module.controller('appWhatsnewFeedController', app.WhatsnewFeedController);
--- a/c2corg_ui/static/partials/feedtable.html
+++ b/c2corg_ui/static/partials/feedtable.html
@@ -1,0 +1,63 @@
+<div ng-show="wfeedCtrl.documents.length > 0"
+  infinite-scroll="wfeedCtrl.getDocumentsFromFeed()"
+  infinite-scroll-immediate-check="false"
+  infinite-scroll-distance="0.7"
+  infinite-scroll-disabled="wfeedCtrl.busy || wfeedCtrl.error || wfeedCtrl.feedEnd || wfeedCtrl.noFeed">
+
+  <p class="text-success text-center" ng-if="wfeedCtrl.userId">
+    <span translate>Feed for user:</span> {{::wfeedCtrl.userId}}
+  </p>
+
+  <div class="table-responsive">
+    <table class="text-center feedtable table table-condensed table-striped table-hover">
+      <tr>
+        <th class="text-center" translate>title</th>
+        <th class="text-center" translate>type</th>
+        <th class="text-center" translate>written_at</th>
+        <th class="text-center" translate>quality</th>
+        <th class="text-center" translate>author</th>
+        <th class="text-center" translate>comment</th>
+      </tr>
+
+      <tr class="feed-list" ng-repeat="doc in wfeedCtrl.documents track by doc.version_id"
+          ng-init="type = wfeedCtrl.getDocumentType(doc.document.type);">
+
+        <td class="text-left">
+          <a ng-href="{{wfeedCtrl.getVersionUrl(doc)}}">
+            <span ng-if="doc['document']['title']">{{::doc['document']['title']}}</span>
+            <span ng-if="!doc['document']['title']" translate>Untitled</span>
+          </a>
+          <span ng-if="doc['lang'] != wfeedCtrl.currentLang">({{::doc['lang']}})</span>
+        </td>
+
+        <td>
+          <label class="badge {{type}}">
+            <span class="glyphicon glyphicon-arrow-right"></span>&nbsp<span>{{type | translate}}</span>
+          </label>
+        </td>
+
+        <td data-original-title="999" data-container="body" tooltip-append-to-body="true" tooltip-placement="top" uib-tooltip="{{::doc['written_at']}}">
+          {{::doc['written_at'] | amUtc | amLocal | amDateFormat:'L LTS'}}
+        </td>
+
+        <td ng-if="::doc['document']['quality']" class="quality-{{::doc['document']['quality']}}" tooltip-append-to-body="true"
+            uib-tooltip="{{'quality'| translate}} : {{::doc['document']['quality'] | translate}}">
+        </td>
+
+        <td data-container="body" tooltip-append-to-body="true" tooltip-placement="top">
+          <a ng-href="/profiles/{{::doc['user']['user_id']}}/{{::doc['user']['lang']}}">{{::doc['user']['name']}}</a>
+        </td>
+
+        <td>{{::doc['comment']}}</td>
+
+      </tr>
+
+    </table>
+  </div>
+</div>
+
+  <h4 class="text-center text-danger" translate ng-show="wfeedCtrl.error">Sorry, there was an error while getting the feed.</h4>
+  <h4 class="text-success text-center" translate ng-show="wfeedCtrl.busy">Loading feed, please wait...</h4>
+  <h4 class="text-success text-center" translate ng-show="wfeedCtrl.feedEnd">No more feed to load.</h4>
+
+</div>
--- a/c2corg_ui/templates/whatsnew.html
+++ b/c2corg_ui/templates/whatsnew.html
@@ -1,0 +1,17 @@
+<%inherit file="base.html"/>
+<%namespace file="helpers/common.html" import="show_title"/>
+<%namespace file="helpers/view.html" import="show_badge"/>
+
+<%block name="pagetitle"><title>${show_title('Latest changes')}</title></%block>
+
+<section class="whatsnew">
+
+  <header>
+    <h1 class="text-center" translate>Latest changes</h1>
+  </header>
+
+  <div>
+    <app-whatsnew-feed app-loading></app-whatsnew-feed>
+  </div>
+
+</section>
--- a/c2corg_ui/views/index.py
+++ b/c2corg_ui/views/index.py
@@ -57,6 +57,11 @@ class Pages(object):
         return self._get_page(
             'following', 'c2corg_ui:templates/following.html')
 
+    @view_config(route_name='whatsnew')
+    def whatsnew(self):
+        return self._get_page(
+            'whatsnew', 'c2corg_ui:templates/whatsnew.html')
+
     def _get_page(self, page_key, template, no_etag=False):
         return get_or_create_page(
             page_key,
--- a/less/feed.less
+++ b/less/feed.less
@@ -1,3 +1,5 @@
+@import 'variables.less';
+
 #feed {
   display: block;
   padding-top: 10px;
@@ -191,5 +193,66 @@
     .timeline-bullet, .feed-list:before,   .feed-list-item:before {
       display: none;
     }
+  }
+}
+
+.whatsnew {
+  padding: 10px;
+  margin: 10px;
+
+  .tools {
+    padding: 10px;
+  }
+  .feed-list {
+    transition: .3s;
+
+    .badge {
+      font-size: 0.8em;
+      margin-right: 5px;
+      margin-left: 5px;
+      padding-bottom: 4px;
+      margin-top: 2px;
+
+      a {
+        color: white;
+        cursor: pointer;
+        text-decoration: none;
+      }
+
+      @media @phone {
+        display: none;
+      }
+
+      .glyphicon {
+        color: white;
+      }
+
+      &.article {
+        background-color: @darkcyan;
+      }
+      &.book {
+        background-color: @plum;
+      }
+      &.image {
+        background-color: @darkolivegreen;
+      }
+      &.route {
+        background-color: @brand-info;
+      }
+      &.xreport {
+        background-color: @tomato;
+      }
+      &.waypoint {
+        background-color: @bright-green;
+      }
+    }
+  }
+  .table > tbody > tr > td {
+     vertical-align: middle;
+  }
+  .feedtable {
+    overflow-y: hidden;
+    padding: 10px;
+    background-color: rgba(229, 229, 229, 0.5);
   }
 }
--- a/c2corg_ui/__init__.py
+++ b/c2corg_ui/__init__.py
@@ -192,6 +192,8 @@ def main(global_config, **settings):
     config.add_route('sitemap_index', '/sitemap.xml')
     config.add_route('sitemap', '/sitemaps/{doc_type:[a-z]{1}}/{i:\d+}.xml')
 
+    config.add_route('whatsnew', '/whatsnew')
+
     # health services
     config.add_route('health', '/health')
 
--- a/c2corg_ui/externs/appx.js
+++ b/c2corg_ui/externs/appx.js
@@ -113,6 +113,19 @@ appx.Document;
 
 /**
  * @typedef {{
+ *   version_id: number,
+ *   document: appx.Document,
+ *   written_at: (string | Date),
+ *   user: Object,
+ *   lang: string,
+ *   comment: string,
+ * }}
+ */
+appx.DocumentChange;
+
+
+/**
+ * @typedef {{
  *   associations: appx.DocumentAssociations,
  *   locales: Array.<Object>,
  *   type: string,
--- a/c2corg_ui/static/js/main.js
+++ b/c2corg_ui/static/js/main.js
@@ -31,6 +31,7 @@ goog.require('app.diffMapDirective');
 goog.require('app.doctypeSelectorDirective');
 goog.require('app.documentEditingDirective');
 goog.require('app.feedDirective');
+goog.require('app.whatsnewFeedDirective');
 goog.require('app.followDirective');
 goog.require('app.followingDirective');
 goog.require('app.geolocationDirective');
--- a/c2corg_ui/static/partials/feed.html
+++ b/c2corg_ui/static/partials/feed.html
@@ -30,7 +30,7 @@
   </article>
 
   <article class="feed-list-item" ng-repeat="doc in feedCtrl.documents track by doc.id"
-           ng-init="type = feedCtrl.documentType(doc.document.type);">
+           ng-init="type = feedCtrl.getDocumentType(doc.document.type);">
 
     <div class="timeline-bullet icon-{{::type}}s" uib-tooltip="{{::type | translate}}" tooltip-placement="left"></div>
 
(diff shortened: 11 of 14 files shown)


c2corg_ui/views/index.py
```diff
@@ -57,6 +58,28 @@ def following(self):
         return self._get_page(
             'following', 'c2corg_ui:templates/following.html')
 
+    @view_config(route_name='whatsnewdata', renderer='json')
+    def whatsnewdata(self):
```
tsauerwein: Do you need this view? The request could be made directly to the API (instead of making a request to the UI server-side which makes a request to the API).

c2corg_ui/static/js/whatsnew.js
```diff
@@ -0,0 +1,94 @@
...
+   */
+  this.alerts_ = appAlerts;
+
+  /**
+   *
+   * @type {Array}
+   * @export
+   */
+  this.columns = [
+    {field: 'version_id', title: 'version_id', show: true},
```
tsauerwein: To avoid that the Closure Compiler might rename the property names, you can either provide a definition in the externs file or use `'file': 'version_id'`.

c2corg_ui/static/js/whatsnew.js
```diff
@@ -0,0 +1,94 @@
...
+    {field: 'comment', title: 'comment', show: true},
+    {field: 'written_at', title: 'written_at', show: true}];
+
+  /**
+   * Used to store data from API
+   * @type {Array}
+   * @export
+   */
+
+  this.dataset = [{}];
```
tsauerwein: `this.dataset = [];`?

asaunier:
Does the API service really support cols sorting and pagination?
I see nothing about that in https://github.com/c2corg/v6_api/pull/620/files

dkocich:
> Does the API service really support cols sorting and pagination?

The higher amount of documents is loaded from API now and ordering/pagination/filtering is done with them on UI only. It can be useful to quickly search in the data without creating of additional requests to API. I started to work on pagination on API and I am not sure if there should be some/all filters implemented also. This is just first wip proposal so feel free to comment.

asaunier:
> The higher amount of documents is loaded from API now and ordering/pagination/filtering is done with them on UI only.

Is it really possible? There are about 850'000 documents in the database, probably millions of changes overall. I don't understand how we could handle this in the UI only?

asaunier:
By the way, are you sure sorting was really requested by the client? I rather think the goal is to be able to see the latest changes, ie. starting from the most recent down to older ones. Only a chronological sorting is needed.

Concerning the pagination, I would recommend to use something similar to the feed, with a direct infinite-scrolling request to the API (the UI is only used to display the page with the calling directive).

And if we could avoid relying on yet another library, it would save bytes to download :P

desnoes:
> By the way, are you sure sorting was really requested by the client? I rather think the goal is to be able to see the latest changes, ie. starting from the most recent down to older ones. Only a chronological sorting is needed.

I confirm, a chronological sorting is sufficient.

dkocich:
> Is it really possible? There are about 850'000 documents in the database, probably millions of changes overall. I don't understand how we could handle this in the UI only?

I tried to request 500-2500 documents and display them using ngTable. They were divided into pages based on the page size setting (bottom right in the previous picture). All the sorting, filtering, etc. worked only over one requested batch of documents and not over the complete table of course.

> @asaunier Concerning the pagination, I would recommend to use something similar to the feed, with a direct infinite-scrolling request to the API (the UI is only used to display the page with the calling directive).

I updated this PR. Changes are now listed as a feed into the table but JS to handle user filter is not yet implemented. If you would really like this approach more, then I can maybe try to add that user filter and improve some details to get this PR merged - what do you think? Thanks.

image

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,42 @@
...
+<%block name="moduleConstantsValues">
+  ## Set in the data request
+  module.value('mapFeatureCollection', null);
+
+</%block>
+
+<section id="whatsnew">
+
+  <header>
+    <h1 class="text-center" translate><span>Feed of the latest changes</span></h1>
```
asaunier: You don't need the span element here. Or at least the translate directive should apply to it, not to the parent, else the span tags are integrated to the string to translate.

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,42 @@
...
+##         <app-simple-search class="half form" dataset="u"></app-simple-search>
+##         <button type="button" class="btn btn-primary">Reload feed for user</button>
+##       </span>
+##
+##     </div>
+
+  </div>
+
+  <div>
+    <app-feed app-is-changes-feed="true" template-path="/static/partials/feedtable.html"></app-feed>
```
asaunier: Are you sure of this ``template-path`` attribute here? Shouldn't it be passed as ``templateUrl`` in the directive.
I'am afraid that using this attribute here will not make the partial available in the ``templatecache.js`` file.

dkocich: I tried `make clean`, `make build` and I see `$templateCache.put('/static/partials/feedtable.html'...`,  in my `static/build/templatecache.js`

asaunier: OK, perhaps that's because this file is built using all files in ``/static/partials``. But it doesn't mean this template cache will actually be used. It seems ``/static/partials/feedtable.html`` will always be called, including in prod mode.

c2corg_ui/views/index.py
```diff
@@ -57,6 +57,18 @@ def following(self):
         return self._get_page(
             'following', 'c2corg_ui:templates/following.html')
 
+    @view_config(route_name='whatsnew',
+                 renderer='c2corg_ui:templates/whatsnew.html')
+    def whatsnew(self):
+        resp, data = call_api(self.settings, 'document/changes')
```
asaunier: I don't understand why you call the API from within a UI server view instead of making a direct request from the browser, as done for the home/profile feeds.

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,42 @@
...
+<section id="whatsnew">
+
+  <header>
+    <h1 class="text-center" translate>Feed of the latest changes</h1>
+  </header>
+
+  <div class="tools">
+
+##     <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#filters">
+##       <span translate>Filters</span><span class="glyphicon glyphicon-menu-down"></span>
```
asaunier: I don't think we need to have filters.
A link in the user's profiles towards ``whatsnew?u=12345`` should be enough.

Perhaps the client will need advanced filtering on changes in the future. But it's not required for now.

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,42 @@
...
+
+<%block name="metarobots">
+  <meta name="robots" content="index,follow">
+</%block>
+
+<%block name="moduleConstantsValues">
+  ## Set in the data request
+  module.value('mapFeatureCollection', null);
+
+</%block>
```
asaunier: That's already the default value:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L153-L155
You don't need to override it.

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,69 @@
...
+  <div class="table-responsive">
+    <table class="feedtable table table-condensed table-bordered table-striped table-hover">
+
+      <p class="text-success text-center" ng-if="feedCtrl.userId" ng-init="user = feedCtrl.userId;" ng-model="amount = feedCtrl.feedLength;">
+        This feed for the user {{user}} contains {{amount}} changes.</p>
+      <p class="text-success text-center" ng-if="!feedCtrl.userId" ng-model="amount = feedCtrl.feedLength;">
+        This feed contains {{amount}} changes.</p>
+
+      <tr>
+        <th>type</th>
```
asaunier: translate

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,69 @@
...
+<div ng-show="feedCtrl.documents.length > 0"
+  infinite-scroll="feedCtrl.getDocumentsFromFeed()"
+  infinite-scroll-immediate-check="true"
+  infinite-scroll-distance="0.7"
+  infinite-scroll-disabled="feedCtrl.busy || feedCtrl.error || feedCtrl.feedEnd || feedCtrl.noFeed">
+
+  <div class="table-responsive">
+    <table class="feedtable table table-condensed table-bordered table-striped table-hover">
+
+      <p class="text-success text-center" ng-if="feedCtrl.userId" ng-init="user = feedCtrl.userId;" ng-model="amount = feedCtrl.feedLength;">
```
asaunier: Why ``ng-model="amount = feedCtrl.feedLength;"``?
It should be part of the ``ng-init`` I think.

c2corg_ui/static/js/feed.js
```diff
@@ -204,4 +235,31 @@ app.FeedController.prototype.documentType = function(type) {
...
+  return '/' + app.utils.getDoctype(doc.document.type) + '/'
+    + doc.document.document_id + '/' + doc.lang + '/' + doc.document.title;
+};
+
+/**
+ * Generate URL address to the changed document version
+ * @param doc {Object}
+ * @returns {string | boolean}
+ * @suppress {checkTypes}
+ * @constructor
```
asaunier: Constructor?

c2corg_ui/static/js/feed.js
```diff
@@ -204,4 +235,31 @@ app.FeedController.prototype.documentType = function(type) {
...
+ */
+app.FeedController.prototype.getDocHref = function(doc) {
+  return '/' + app.utils.getDoctype(doc.document.type) + '/'
+    + doc.document.document_id + '/' + doc.lang + '/' + doc.document.title;
+};
+
+/**
+ * Generate URL address to the changed document version
+ * @param doc {Object}
+ * @returns {string | boolean}
```
asaunier: It doesn't make sense that a function returns that unconsistent outputs. Make a choice :P

c2corg_ui/static/js/feed.js
```diff
@@ -204,4 +235,31 @@ app.FeedController.prototype.documentType = function(type) {
...
+ * @suppress {checkTypes}
```
asaunier: ??
Stashing problems under the carpet? :P

c2corg_ui/static/js/feed.js
```diff
@@ -204,4 +235,31 @@ app.FeedController.prototype.documentType = function(type) {
...
+ * @returns {string}
+ */
+app.FeedController.prototype.getDocHref = function(doc) {
+  return '/' + app.utils.getDoctype(doc.document.type) + '/'
+    + doc.document.document_id + '/' + doc.lang + '/' + doc.document.title;
+};
+
+/**
+ * Generate URL address to the changed document version
+ * @param doc {Object}
```
asaunier: Do you mean ``appx.Document``?
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js#L111

c2corg_ui/static/js/apiservice.js
```diff
@@ -399,17 +399,23 @@ app.Api.prototype.updatePreferredLanguage = function(lang) {
...
 /**
  * @param {undefined | string} token
- * @param {string} lang
+ * @param {undefined | string} lang
  * @param {?number} userId
  * @param {boolean} isPersonal
  * @return {!angular.$q.Promise<!angular.$http.Response>}
  */
-app.Api.prototype.readFeed = function(token, lang, userId, isPersonal) {
+app.Api.prototype.readFeed = function(token, lang, userId, isPersonal, isChangesFeed) {
```
asaunier: Since those are really different feeds, I would recommend to use dedicated functions.

c2corg_ui/templates/whatsnew.html
asaunier: OK I understand: you try to reuse the feed directive used for the home/profile feeds and would like to customize the partial to use...

I am not sure this is the right way. It makes things more complicated and finally all those feeds are different. It would make sense to use dedicated directives I think?!

To avoid duplicating the common code. You should create a parent controller class and use child classes for the specific parts. As we did for instance for the document editing.

asaunier:
I have set the [WIP] prefix back because it seems there is still a lot of work needed in this PR :P

asaunier:
I have run your branches in my instance. The tool works pretty well :)

capture du 2017-02-10 15-13-48

I see a few things that are not optimal and could be improved:
* ``document_id`` and ``version_id`` are not needed and could be removed
* make the title (or the whole line to make it easier to use on a mobile phone) clickable, pointing to the document archived version.
 * Perhaps pointing to the diff compared to the previous change would be even more relevant to figure the change out. But I guess we don't have the id of the previous version (perhaps a possible improvement in the future :P ).
* title should appear at the beginning of the line
* lang could be a suffix of the title, between parenthesis (as in cards)
* document types and qualities could be displayed using the icons as in cards.
* do we really need the columns titles? The content is quite straightforward, especially if we remove some cols and group some data
* the page is not usable on a mobile phone => the overflowing parts of the table are no longer visible => tables are not responsive :P 

Questions:
* where will this new "whatsnew" page be linked to? @stef74 @desnoes ?

brunobesson:
>tables are not responsive

There are several possibilities:
* bootstrap does it so that the tables can be scrolled horizontally (https://getbootstrap.com/css/#tables-responsive)
* change the layout like it has been done for snow conditions
* datatables hide some columns that can be expanded on purpose (https://datatables.net/extensions/responsive/), but it's a more complicated solution

asaunier:
> There are several possibilities:

Another possibilty: do we really need a table? That could be a line that would wrap if too long?

dkocich:
> Perhaps pointing to the diff compared to the previous change

I was also thinking about adding a link to the Diff between old and new docs... It would require changes to the API part to contain version's previous IDs...

> bootstrap does it so that the tables can be scrolled horizontally https://getbootstrap.com/css/#tables-responsive

[I did use it in this PR ](https://github.com/c2corg/v6_ui/pull/1466/files#diff-cf915b7eac979d3d23c4d6d4877eea28R11) but it probably don't work how it should :) ...

c2corg_ui/static/partials/feedtable.html
dkocich: It should not. `feedLength` is undefined on init and it contains a correct value after the response is received.

asaunier:
When the page first loads, the loading message is displayed twice. I guess only the message at the bottom of the feed is needed?!

@dkocich Thanks for the layout changes you have made. I find it pretty good now!
I would recommend to align the titles to the left, it's easier to read and parse than when they are centered.

c2corg_ui/static/js/apiservice.js
```diff
@@ -428,6 +428,28 @@ app.Api.prototype.readFeed = function(token, lang, userId, isPersonal) {
...
+ * @param {undefined | string} token
+ * @param {?number} userId
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.readWhatsnewFeed = function(token, userId) {
+  var params = {};
+  if (token) params['token'] = token;
+  if (userId) {
+    params['u'] = userId;
+  }
```
asaunier: The coding standard of those last 3 lines is not very consistent.

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+          ng-init="type = wfeedCtrl.documentType(doc.document.type); console.log(type);" >
+
+        <td class="text-left">
+          <a href="{{wfeedCtrl.getDocHref(doc)}}">{{::doc['document']['title']}}</a> ({{::doc['lang']}})
+          <span ng-if="doc['document']['version'] > 1 && (doc['document']['type'] !== 'i' && doc['document']['type'] !== 'o')">
+            (<a ng-href="{{wfeedCtrl.getVersionHref(doc)}}">previous version</a> | diff)
+          </span>
+        </td>
+
+        <td>{{::type}}</td>
```
asaunier: Just an idea: what about using the document type badge classes (with the color) as in the document page titles?
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/view.html#L824

It would add some color to the list and make it easier to identify the doc type at first sight.

c2corg_ui/static/js/whatsnewfeed.js
```diff
@@ -0,0 +1,186 @@
...
+ */
+app.WhatsnewFeedController.prototype.documentType = function(type) {
+  return app.utils.getDoctype(type).slice(0, -1);
+};
+
+
+/**
+ * Generate URL address of a document
+ * @param doc {appx.DocumentChange}
+ * @returns {string}
```
asaunier: Missing ``@export``

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+        <th class="text-center" translate>username</th>
+        <th class="text-center" translate>comment</th>
+      </tr>
+
+      <tr class="feed-list"
+              ng-repeat="doc in wfeedCtrl.documents track by doc.version_id"
+          ng-init="type = wfeedCtrl.documentType(doc.document.type); console.log(type);" >
+
+        <td class="text-left">
+          <a href="{{wfeedCtrl.getDocHref(doc)}}">{{::doc['document']['title']}}</a> ({{::doc['lang']}})
```
asaunier: ``ng-href``?
As for now the links are not working.

asaunier: I would show the lang only if different from the current interface language.
A bit like https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/cards/waypoints.html#L10

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+  infinite-scroll="wfeedCtrl.getDocumentsFromWhatsnewFeed()"
+  infinite-scroll-immediate-check="false"
+  infinite-scroll-distance="0.7"
+  infinite-scroll-disabled="wfeedCtrl.busy || wfeedCtrl.error || wfeedCtrl.feedEnd || wfeedCtrl.noFeed">
+
+  <div class="table-responsive">
+    <table class="text-center feedtable table table-condensed table-striped table-hover">
+
+      <p class="text-success text-center" ng-if="wfeedCtrl.userId"
+         ng-init="user = wfeedCtrl.userId;" ng-model="amount = wfeedCtrl.feedLength;">
```
asaunier: Missing translate directive

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+        This feed for the user {{user}} contains {{amount}} changes.
+      </p>
+      <p class="text-success text-center" ng-if="!wfeedCtrl.userId"
+         ng-model="amount = wfeedCtrl.feedLength;">
```
asaunier: Missing translate directive

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+      </tr>
+
+      <tr class="feed-list"
+              ng-repeat="doc in wfeedCtrl.documents track by doc.version_id"
+          ng-init="type = wfeedCtrl.documentType(doc.document.type); console.log(type);" >
+
+        <td class="text-left">
+          <a href="{{wfeedCtrl.getDocHref(doc)}}">{{::doc['document']['title']}}</a> ({{::doc['lang']}})
+          <span ng-if="doc['document']['version'] > 1 && (doc['document']['type'] !== 'i' && doc['document']['type'] !== 'o')">
+            (<a ng-href="{{wfeedCtrl.getVersionHref(doc)}}">previous version</a> | diff)
```
asaunier: Does the "previous version" link really work? The link seems not clickable in my instance and the cursor remains a text cursor, not a pointer.

Actually the link toward the archived version should be the main one (when clicking the title).

For instance if the feed advertises 5 changes of the same document in a row, it does not make sense that the main links redirect to the current version of the document, they should point to the actual archived version.
Perhaps we don't even need to make a link toward the last version of the document!?

What about the "diff"? Do you think it is fairly feasible? If not, please remove it.

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+         ng-model="amount = wfeedCtrl.feedLength;">
+        This feed contains {{amount}} changes.
+      </p>
+
+      <tr>
+        <th class="text-center" translate>title</th>
+        <th class="text-center" translate>type</th>
+        <th class="text-center" translate>written_at</th>
+        <th class="text-center" translate>quality</th>
+        <th class="text-center" translate>username</th>
```
asaunier: username => author

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+        <th class="text-center" translate>comment</th>
+      </tr>
+
+      <tr class="feed-list"
+              ng-repeat="doc in wfeedCtrl.documents track by doc.version_id"
+          ng-init="type = wfeedCtrl.documentType(doc.document.type); console.log(type);" >
+
+        <td class="text-left">
+          <a href="{{wfeedCtrl.getDocHref(doc)}}">{{::doc['document']['title']}}</a> ({{::doc['lang']}})
+          <span ng-if="doc['document']['version'] > 1 && (doc['document']['type'] !== 'i' && doc['document']['type'] !== 'o')">
```
asaunier: Why do you have such a restriction on outings and images?

c2corg_ui/static/partials/feedtable.html
asaunier: I see a problem with image with no title => it's not possible to click on the title because there is... no title :P 
We need a default (i18n) string when no title is available. Perhaps "Untitled image"?

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+        <td data-original-title="999" data-container="body" tooltip-append-to-body="true" tooltip-placement="top" uib-tooltip="{{::doc['written_at']}}">
+          {{::doc['written_at'] | amUtc | amLocal | amDateFormat:'L LTS'}}
+        </td>
+
+        <td ng-if="::doc['document']['quality']" class="quality-{{::doc['document']['quality']}}" tooltip-append-to-body="true"
+            uib-tooltip="{{'quality'| translate}} : {{::doc['document']['quality']}}">
+        </td>
+
+        <td data-original-title="999" data-container="body" tooltip-append-to-body="true" tooltip-placement="top"
+            uib-tooltip="user_id: {{::doc['user']['user_id']}} username: {{::doc['user']['username']}} ">
```
asaunier: We don't need this (user id and username). All need info are provided with ``doc['user']['name']``.

c2corg_ui/static/partials/feedtable.html
```diff
@@ -0,0 +1,66 @@
...
+
+        <td data-original-title="999" data-container="body" tooltip-append-to-body="true" tooltip-placement="top" uib-tooltip="{{::doc['written_at']}}">
+          {{::doc['written_at'] | amUtc | amLocal | amDateFormat:'L LTS'}}
+        </td>
+
+        <td ng-if="::doc['document']['quality']" class="quality-{{::doc['document']['quality']}}" tooltip-append-to-body="true"
+            uib-tooltip="{{'quality'| translate}} : {{::doc['document']['quality']}}">
+        </td>
+
+        <td data-original-title="999" data-container="body" tooltip-append-to-body="true" tooltip-placement="top"
```
asaunier: 999?

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,18 @@
...
+<%namespace file="helpers/common.html" import="show_title"/>
+
+<%block name="metarobots">
+  <meta name="robots" content="index,follow">
+</%block>
+
+<section id="whatsnew">
+
+  <header>
+    <h1 class="text-center" translate>Feed of the latest changes</h1>
```
asaunier: What about simply "Latest changes"?

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,18 @@
+<%inherit file="base.html"/>
+<%namespace file="helpers/common.html" import="show_title"/>
+
+<%block name="metarobots">
+  <meta name="robots" content="index,follow">
```
asaunier: I don't think this page should be indexed. But no big deal.
Anyway this is already the default value: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L29
=> you can omit this block if you keep it as is.

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,18 @@
+<%inherit file="base.html"/>
+<%namespace file="helpers/common.html" import="show_title"/>
```
asaunier: Where is ``show_title`` used?

asaunier: On the other hand the HTML page title is not correct (as for now is is "home page") => please update the page title.

less/feed.less
```diff
@@ -193,3 +193,71 @@
...
+        &:before {
+          font-family: 'Glyphicons Halflings';
+          content: "\e107"; // warning-sign
+          font-size: 21px;
+          margin-left: 1px;
+        }
+      }
+
+    }
+
```
asaunier: Useless blank lines

c2corg_ui/templates/whatsnew.html
```diff
@@ -0,0 +1,18 @@
...
+
+<%block name="metarobots">
+  <meta name="robots" content="index,follow">
+</%block>
+
+<section id="whatsnew">
```
asaunier: What about using a class instead?

less/feed.less
```diff
@@ -193,3 +193,71 @@
...
+  margin: 10px;
+
+  .tools {
+    padding: 10px;
+  }
+
+  .feed-list {
+    transition: .3s;
+
+    .timeline-bullet {
```
asaunier: Do you really use those styles? There is no timeline in the page, only a table.

c2corg_ui/static/partials/feedtable.html
dkocich: I wasn't sure how we store old versions of these documents because many created URL addresses of these document types weren't accessible. I simply hide them but it is possible that it was caused by some other problem. For example 1st version of this image can't be opened (probably created as private and later changed to collaborative?) https://www.camptocamp.org/images/history/851612/fr

asaunier: > For example 1st version of this image can't be opened

That's weird. Ditto for any archive version of https://www.camptocamp.org/images/history/208579/fr
There is probably an error somewhere. I have created an issue at https://github.com/c2corg/v6_ui/issues/1503

I would recommend to drop those special cases: image archives should be accessible and outings will probably no longer be listed in the feed. https://github.com/c2corg/v6_api/issues/625

c2corg_ui/static/partials/feedtable.html
dkocich: That badges look good. I added some colours to the types which had none selected.
image

asaunier: Nice! Let me know when the PR is ready for review.

asaunier: > I added some colours to the types which had none selected.

Have you change the badge colors in the actual document pages too?
That would make sense, to have the same colors for badge in the change feed in the document detail views.

c2corg_ui/static/partials/feedtable.html
dkocich: Prev. version works.

I think that it is better to have those links (to document + to the previous version) separately. In a case of documents which have just one version, it is better to point to that document otherwise an user has no chance to find it.
It can be confusing to sometimes have a title's `href` pointing to the previous version and sometimes for latest document (if it was just created)...

Now when there are those document badges I have an idea if it is better to have link to the document on the badge and if the document has any previous version -> point to it with the title (and delete "previous version" text). What do you think?

image

dkocich:
I have no issue but travis fails :( https://travis-ci.org/c2corg/v6_ui/builds/202559775#L692-L708

```npm install
npm ERR! Linux 4.4.0-51-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v6.9.5/bin/node" "/home/travis/.nvm/versions/node/v6.9.5/bin/npm" "install"
npm ERR! node v6.9.5
npm ERR! npm  v3.10.10
npm ERR! shasum check failed for /tmp/npm-4869-28477fd8/registry.npmjs.org/moment/-/moment-2.15.2.tgz
npm ERR! Expected: 1bfdedf6a6e345f322fe956d5df5bd08a8ce84dc
npm ERR! Actual:   41b4e7ef38051e320e9b7b8f6eae35003f57e44c
npm ERR! From:     https://registry.npmjs.org/moment/-/moment-2.15.2.tgz
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
npm ERR! Please include the following file with any support request:
npm ERR!     /home/travis/build/c2corg/v6_ui/npm-debug.log```

File: c2corg_ui/static/partials/feedtable.html (hunk 22 above)
Comment (asaunier) [inline c2corg_ui/static/partials/feedtable.html]: > I think that it is better to have those links (to document + to the previous version)

I am not sure to understand what you mean. Actually why do we need links towards several versions?

What do you call "previous version"? It seems your "previous version" links point to the archived version concerned by the change, doesn't it? This is actually the main link we want to see.

If you have several changes in a row of the same document, it's no use that the main links (on titles) point to the same URL...

The goal is to report that user X has created/edited a document at time T and to show what changes they have made. It should work a bit like the history pages (no need to compare several versions though) => point to the concerned version of the document and (if possible - not sure the previous version id is provided by the API response, is it?) to the diff of the changes that have been made (no diff anyway if this is a creation).
No need to point towards the last version of the document (or as a secondary link, could be named "current version").
No need to point towards the previous version of the document.

>  if it is better to have link to the document on the badge and if the document has any previous version

I don't know. In detail view pages, the badge points to the advanced search page. It would not be very consistent to make it point here to the document itself.
On the other hand, the badge name uses the singular form, not the plural form, so it's not really the same badge...

**Comment by asaunier** [comment]
I have just tested the last version of your branch, I see problems anyway:

* [ ] Your links towards the document detail views are not correct, their slug is defective.
For instance "http://c2corgv6-demo.gis.internal:6555/outings/842111/fr/Mont%20de%20Grange%20:%20Couloir%20de%20Pertuis"
* [ ] Title of routes are not correct: the ``title_prefix`` is missing.
* [ ] The mouse cursor on badges become a pointer (as for links) where there is no link.

As I said in my previous comment, I don't think using the standard document URL is relevant as the main link.

**Comment by asaunier** [comment]
I have just tested the last version of the branch.

Why don't you put the link on the document title instead of using a "current version" link?
It makes more sense (and it's more handy) to put the link on the title.
In addition "current version" is a confusing term. For me it means "the most recent version", not the version of the link. 
Finally I see many entries with no link at all (not "current version" link)!? Is it because my API is not up-to-date?

What about putting the link on the whole line as I suggested it earlier?

What I recommend:
* drop "current version"
* make sure a title is provided (at least a default "Untitled" => should be translated then)
* badges should be translated
* it seems quality tooltips are not translated => "quality" is translated but not the attribute value
* lang data => only if different from the interface language (as in cards)

**Comment by asaunier** [comment]
@dkocich With commit f1fe8ee I think I have fixed/simplified the links and main titles.

**Comment by asaunier** [comment]
I see other minor problems I think. Perhaps I will suggest another commit.

File: c2corg_ui/static/partials/feedtable.html [hunk 33]
```diff
@@ -24,12 +24,11 @@
...
         <td class="text-left">
-          <span ng-if="doc['document']['version'] > 1 && doc['document']['type'] !== 'i'">
-            <a ng-href="{{wfeedCtrl.getVersionHref(doc)}}">{{::doc['document']['title']}}</a>
-          </span>
-          <span ng-if="doc['document']['version'] == 1 || doc['document']['type'] == 'i'">{{::doc['document']['title']}}</span>
-          <span ng-if="doc['document']['title'].length == 0" translate>Untitled</span>
-          <span ng-if="doc['lang'] != locale.lang">({{::doc['lang']}})</span>
+          <a ng-href="{{wfeedCtrl.getVersionUrl(doc)}}">
+            <span ng-if="doc['document']['title']">{{::doc['document']['title']}}</span>
+            <span ng-if="!doc['document']['title']" translate>Untitled</span>
```
dkocich: I checked and there is again that problem with not-accessible image versions.

asaunier: > I checked and there is again that problem with not-accessible image versions.

That's not our business here. It's an API issue: https://github.com/c2corg/v6_ui/issues/1503
Hopefully it will be fixed and the links will work here as well. But it does not make sense to make workarounds here, it's better to fix the problem at its source.

asaunier:
Argh, there is a problem with your changes (I have tried to remove mine with no more success): the home page feed is broken, only the initial list of items is loaded but when scrolling the feed is not refreshed.

c2corg_ui/static/js/feed.js
```diff
@@ -126,6 +126,7 @@ app.FeedController = function(appAuthentication, appApi, appLang, imageUrl, ngeo
 /**
  * Fills the feed with documents.
  * Used by ng-infinite-scroll directive in the template.
+ * @export
  * @public
```
asaunier: Are you sure the ``@public`` keyword are needed?

dkocich: it works without it just with `export`

asaunier:
Please do no more changes to this branch I was about to commit small problems fixes...

asaunier:
With commit 741501d I have made small improvements.

For me the PR is OK now, except that commits should be squashed (including mines, no problem).

dkocich:
@asaunier This pr is prepared to be merged.

desnoes:
@dkocich : nice job ! 
I have a question: what are the types of comments that should be displayed in the table ? I only observe 'creation'. Is it the expected behavior ?
I found a very minor bug: there is a 1 hour delay between the timestamp written in the table and the timestamp written in the tooltip. I would suggest to remove the tooltip on this field.

bug whats new

asaunier:
> I found a very minor bug: there is a 1 hour delay between the timestamp written in the table and the timestamp written in the tooltip.

I think the tooltip shows the GMT time.

> I would suggest to remove the tooltip on this field.

+1

desnoes:
I have removed the geometry in this outing and the modification is not reported.
http://www.demov6.camptocamp.org/outings/828402/en/tete-de-vauclave-rrr

dkocich:
> I have a question: what are the types of comments that should be displayed in the table? I only observe 'creation'. Is it the expected behaviour?

What would you like to see? When someone creates the document then it is "creation" automatically. When someone edits document then he is asked to write the short comment about the new version and to change quality.
@asaunier maybe we should add `translate` directive which would replace this word with something like "This document was just created." in all languages? It should be changed on doc. history pages too... 

image

>I would suggest to remove the tooltip on this field.

Ok.

> I have removed the geometry in this outing and the modification is not reported. http://www.demov6.camptocamp.org/outings/828402/en/tete-de-vauclave-rrr

I think that it is not possible to see any outing change - in this PR we removed User profile creations and Outing changes from the feed https://github.com/c2corg/v6_api/issues/625#issuecomment-280233276

desnoes:
I understood my mistake. The comment column report the text written in the field 'Comment about the changes'. Sorry for the inconvenience.

I consider the result is very fine. I close the issue and let you remove the tooltip on the timestamp.

dkocich:
No problem. :) Thanks!

fbunoz:
With this PR, is it possible to filter on the doc type, the lang, the linked areas, the activities (routes and outings only) ?

dkocich:
@fbunoz it is not possible right now

Issue #1469, opened 2017-02-03
Internal server error for some routes


dkocich:
Some routes are crashing (prod/demo/master br.) - f.e.
https://www.camptocamp.org/routes/191005/fr/el-yelmo-kindelan
https://www.camptocamp.org/routes/300430/fr/pointe-de-la-fourche-facette-ne
https://www.camptocamp.org/routes/299313/fr/trelod-arete-n

```
pyramid_mako.MakoRenderingException: 

Traceback (most recent call last):
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid_mako/__init__.py", line 148, in __call__
    result = template.render_unicode(**system)
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/template.py", line 471, in render_unicode
    as_unicode=True)
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/runtime.py", line 838, in _render
    **_kwargs_for_callable(callable_, data))
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/runtime.py", line 873, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/runtime.py", line 899, in _exec_template
    callable_(context, *args, **kwargs)
  File "/home/dkocich/projects/c2corg/v6_ui/c2corg_ui/templates/base.html", line 57, in render_body
    ${self.body()}
  File "/home/dkocich/projects/c2corg/v6_ui/c2corg_ui/templates/route/view.html", line 133, in render_body
    ${get_route_gear(route, locale)}
  File "/home/dkocich/projects/c2corg/v6_ui/c2corg_ui/templates/route/helpers/view.html", line 264, in render_get_route_gear
    <%
  File "/home/dkocich/projects/c2corg/v6_ui/c2corg_ui/templates/utils/__init__.py", line 127, in get_route_gear_articles
    articles.append('185384', 'easy mountain climbing gear')
TypeError: append() takes exactly one argument (2 given)
```

asaunier:
Good catch, thanks :)
I have suggested a fix (untested) at #1471

dkocich:
[...]

PR #1476, opened 2017-02-07
Check properly for missing route attributes to prevent server error

dkocich:
related to https://github.com/c2corg/v6_ui/issues/1469#issuecomment-277222520

It was not possible to browse some archived versions of Routes. For some reason (missing attributes?), there are unknown keys in historic documents. They created internal server error because of Mako templates rendering.

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -111,7 +111,8 @@
       title = ''
       if 'title_prefix' in locale and locale['title_prefix']:
           title = locale['title_prefix'] + ' : '
-      title += locale['title']
+      if 'title' in locale and locale['title']:
+          title += locale['title']
   %>\
   ${title}${' - Camptocamp.org' if is_tab_title else ''}\
 </%def>
--- a/c2corg_ui/templates/route/helpers/view.html
+++ b/c2corg_ui/templates/route/helpers/view.html
@@ -264,7 +264,7 @@ from c2corg_ui.templates.utils import get_route_gear_articles
   <%
     gear_articles = get_route_gear_articles(route)
   %>
-  % if locale['gear'] or gear_articles:
+  % if ('gear' in locale and locale['gear']) or gear_articles:
     <div class="view-details-info col-xs-12">
       <h3 class="heading show-phone"
           ng-click="mainCtrl.animateHeaderIcon($event)" data-toggle="collapse" data-target="#gear">
--- a/c2corg_ui/templates/route/view.html
+++ b/c2corg_ui/templates/route/view.html
@@ -40,7 +40,7 @@ other_langs, missing_langs = get_lang_lists(route, lang)
 
 <%block name="moduleConstantsValues">
   <%
-      if route['geometry']:
+      if 'geometry' in route and route['geometry']:
           geometry = route['geometry']['geom_detail'] \
               if route['geometry']['geom_detail'] else route['geometry']['geom']
       else:


c2corg_ui/templates/route/helpers/view.html
```diff
@@ -264,7 +264,7 @@
   <%
     gear_articles = get_route_gear_articles(route)
   %>
-  % if locale['gear'] or gear_articles:
+  % if ('gear' in locale and locale['gear']) or gear_articles:
```
asaunier:
```
% if locale.get('gear') or gear_articles:
```
should do the job I think.

c2corg_ui/templates/route/view.html
```diff
@@ -40,7 +40,7 @@
 
 <%block name="moduleConstantsValues">
   <%
-      if route['geometry']:
+      if 'geometry' in route and route['geometry']:
```
asaunier: Perhaps
```
if route.get('geometry'):
```

c2corg_ui/templates/helpers/view.html
```diff
@@ -111,7 +111,8 @@
       title = ''
       if 'title_prefix' in locale and locale['title_prefix']:
           title = locale['title_prefix'] + ' : '
-      title += locale['title']
+      if 'title' in locale and locale['title']:
+          title += locale['title']
```
asaunier: I don't think this change is satisfaying because you stash the problem instead of solving it.

Isn't the problem here that the API does not return a ``title``? That does not look correct. Perhaps there's a bug in the API.
The API request for http://localhost:6555/routes/111829/fr/291699 (see the examples at https://github.com/c2corg/v6_ui/issues/1469#issuecomment-277222520) returns a ``locale`` with many lacking attribute:

capture du 2017-02-08 13-12-12

The previous and next versions in http://localhost:6555/routes/history/111829/fr return all attributes as expected.

Why is the behaviour different for this version?
API issue?

tsauerwein: The problem is that the route does not have any activities in that version. Without activity, no fields (as defined in the [common project](https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py#L50)) are returned. Activities are now a required field, but apparently there are old versions without. I will work on a fix.

dkocich: note - for that wrong version Title field is strange (----) and description/remarks/gear fields are empty http://localhost:6548/routes/111829/fr/291699
``` sql
SELECT * 
FROM 
  guidebook.documents_versions, guidebook.documents, guidebook.documents_archives, 
  guidebook.documents_geometries_archives, guidebook.documents_locales_archives, 
  guidebook.history_metadata, guidebook.routes_locales_archives
WHERE 
  documents_versions.document_id = documents.document_id AND
  documents_versions.document_archive_id = documents_archives.id AND
  documents_versions.document_geometry_archive_id = documents_geometries_archives.id AND
  documents_versions.history_metadata_id = history_metadata.id AND
  documents_locales_archives.id = documents_versions.document_locales_archive_id AND
  routes_locales_archives.id = documents_locales_archives.id AND
  guidebook.documents_versions.document_id = 111829;
```

asaunier: I think that change was made by an upset user that wanted to sabotage the route...

tsauerwein: PR is here: https://github.com/c2corg/v6_api/pull/623

dkocich:
Problematic routes are displaying correctly with that API PR so we can probably close this one.

fbunoz:
> The problem is that the route does not have any activities in that version. 

Before march 2012, it was possible to save a route with "empty" activity (coded as 0 in DB, and displayed as an empty option in the selector), which was different of unset activity. "empty" activity was understand as an activity by the validator. Before 2012, the validator check if the activity was set, but it didn't check if the activity wasn't "empty".
Then some users selected "empty" option, then a moderator corrected the activity, but there are some version with "empty" activity (which has been migrated as unset ?).

https://github.com/c2corg/camptocamp.org/commit/94dc24a73666c037386272b9f6d8b1500bc54400#diff-2ae46812b70e77389563f9c3f8921a6b

Issue #1455, opened 2017-01-31
Use npm package references for some packages


asaunier:
* [x] angular-recaptcha https://github.com/c2corg/v6_ui/blob/master/package.json#L17 See https://www.npmjs.com/package/angular-recaptcha => perhaps we could upgrade to 4.0.1 instead of 3.0.1
* [ ] angular-debounce https://github.com/c2corg/v6_ui/blob/master/package.json#L34 See https://www.npmjs.com/package/angular-debounce but not sure it's the same package!

dkocich:
`angular-debounce`
It isn't.

- Our - latest 0.1.7 https://github.com/shahata/angular-debounce/releases
- The one from NPM don't have this version... https://github.com/rubenv/angular-debounce/releases

asaunier:
> It isn't same.

OK than please close the issue when your PR #1479 is merged. Thanks!

PR #1479, opened 2017-02-08
Update angular-recaptcha to v4.0.3

dkocich:
Related to https://github.com/c2corg/v6_ui/issues/1455

The need to include the Google script was removed here https://github.com/VividCortex/angular-recaptcha/commit/82d33e6adabdd81a0341b6e68866f28b0f760522
I don't know how to test recaptcha - should work (?)...

### DIFF
--- a/c2corg_ui/templates/auth.html
+++ b/c2corg_ui/templates/auth.html
@@ -167,4 +167,3 @@ from pyramid.settings import asbool
 
   </section>
 </div>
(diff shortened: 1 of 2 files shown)


asaunier:
> I don't know how to test recaptcha - should work (?)...

The branch is built with no error and I can actually check the "I am not a robot" checkbox in the signup page:

capture du 2017-02-08 19-44-30

Issue #1472, opened 2017-02-03
Right of xreport edition


fbunoz:
The right to edit a xreport must be the same as for outings : an user associated to a xreport must be able to edit this xreport, even if he isn't the creator of this xreport.

For now, an user associated to a xreport but who isn't the creator, can't edit this xreport.

PR #1480, opened 2017-02-08
Allow Xrep. edition  to Xrep. associated users

dkocich:
related to https://github.com/c2corg/v6_ui/issues/1472

It should be possible to edit Xreport when logged user is associated to it. It requires change of API also.


c2corg_ui/static/js/auth/authservice.js
```diff
@@ -189,8 +189,18 @@ app.Authentication.prototype.hasEditRightsArticle_ = function(articleType, autho
  * @return {boolean}
  * @private
  */
-app.Authentication.prototype.hasEditRightsXreport_ = function(authorId) {
-  return this.userData.id === parseInt(authorId, 10);
+app.Authentication.prototype.hasEditRightsXreport_ = function(authorId, users) {
```
asaunier: Since outings and xreports editing rights seem to be the same, what about using ``hasEditRightsOuting_`` (could be renamed) instead?

Issue #1473, opened 2017-02-05
Missing translation ID in xreport


fbunoz:
- [x] In the xreport read and edit page, the title of the section "personal" must have a specific translation ID to avoid conflict with other uses of "personal". Because "personal" is the translation ID for "personal licence", using for image and article. The xreport section title should be "personal infos" ("Infos personnelles" in FR).
- [x] ~~~In the outing read and edit page, the field title for the checkbox "soft mobility outing" must have a specific translation ID.
For now, the ID is "public_transport", which is also use in access WP.
It should be "soft mobility outing", to allow the outings with bicycle access.~~~

fbunoz:
> For now, the ID is "public_transport", which is also use in access WP.

Wrong, in access WP it's public_transportation_rating

brunobesson:
Translation is OK when editing xreport, but the box doesn't appear in view. Is there some reason it is not shown? I am checking a report I made with my own account.

desnoes:
@mgerbaux @fbunoz : the box is displayed in the preview but not in view. Is-it the expected result ?

fbunoz:
In V5, only the field "author" was visible in view for simple membre.
For moderator, all field of "personal" box was visible.
In the preview, it's better to show the same field as the view, the only the "author" field should be shown.

desnoes:
@fbunoz : so the bug is fixed in view but a minor bug remain in preview. I am right ?

PR #1483, opened 2017-02-08
Use 'personal informations' title instead of 'personal' for xreports

brunobesson:
Related to #1473 
I re-use `personal informations` key, which is similar and used also in outings creation wizard. @asaunier is it best to use different keys, or is it OK like this to limit the number of translation items?

### DIFF
--- a/c2corg_ui/templates/xreport/edit.html
+++ b/c2corg_ui/templates/xreport/edit.html
@@ -267,7 +267,7 @@ from c2corg_common.attributes import default_langs, activities, event_types, \
 
         ## PERSONAL
         <section class="section title" ng-show="userCtrl.auth.isAuthenticated()">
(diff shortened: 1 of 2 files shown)


asaunier:
>  is it best to use different keys, or is it OK like this to limit the number of translation items?

For me it is better to reuse strings if the translations are the same from a document type to another and avoid adding new strings unless really needed.

The PR looks good to me. @dkocich OK for you?

dkocich:
Ok for me (except one note in relation to PR that I made some time ago :rofl:  https://github.com/c2corg/v6_ui/pull/734)

brunobesson:
>except one note in relation to PR

You mean we have to take care to "redo" the translation after merging?

dkocich:
leave it like it is if that key is already translated from the previous version...

Issue #1487, opened 2017-02-09
Advanced search API returns an error if offset is above 9970


brunobesson:
Maybe this could be an API issue too, but is it really pertinent to show a "Go to last" button if there are so many results?
Hiding the button until there are fewer results would make sense too.

brunobesson:
E.g.

https://api.demov6.camptocamp.org//outings?bbox=587706,5526776,660474,5661916&pl=fr&offset=9971
and
https://api.demov6.camptocamp.org//outings?bbox=587706,5526776,660474,5661916&pl=fr&offset=9970

dkocich:
duplicate of https://github.com/c2corg/v6_api/issues/578

dkocich:
Hiding that button on UI is a good idea (oops - I wanted to close API issue).

brunobesson:
@dkocich I now of the c2corg/v6_api#578 issue. But is it really a duplicate if it concerns another project? IMHO we should close the issue in v6-api instead. Moreover, I have a PR for this issue.

brunobesson:
>oops - I wanted to close API issue

OK, we're on the same line then. Thanks.

brunobesson:
Working on demo.

PR #1488, opened 2017-02-09
Do not show "Got to last" button if there are too many results

brunobesson:
If offset is above 9970, the API will return an error.

See #1487

### DIFF
--- a/c2corg_ui/static/js/search/pagination.js
+++ b/c2corg_ui/static/js/search/pagination.js
@@ -73,6 +73,11 @@ app.PaginationController = function($scope, ngeoLocation) {
     this.handleSearchChange_.bind(this));
 };
 
+/**
+ * @const
+ * @type {number}
+ */
+app.PaginationController.MAX_RESULT_OFFSET = 10000;
 
 /**
  * @param {Object} event
@@ -85,11 +90,10 @@ app.PaginationController.prototype.handleSearchChange_ = function(event,
     features, total, recenter) {
   this.total = total;
   this.offset = this.location_.getFragmentParamAsInt('offset') || 0;
-  // don't show the "Go to last page" button when doing a full-text-search
-  // (the full result set has to be iterated in ElasticSearch when doing
-  // requests with large offsets)
-  this.showGoToLastPage = !(this.location_.hasFragmentParam('q') &&
-    this.location_.getFragmentParam('q') !== '');
+  // don't show the "Go to last page" button if the offset is above
+  // maxResultOffset_, or the API will return an error
+  this.showGoToLastPage =
+    this.total <= app.PaginationController.MAX_RESULT_OFFSET;
 };
 
 


asaunier:
Wasn't https://github.com/c2corg/v6_ui/pull/1200 enough?

brunobesson:
>Wasn't #1200 enough?

Nope, it was focused on elastic search only. Even without full text search, there is a limit to the offset allowed by API. See c2corg/v6_api#578

tsauerwein:
> Wasn't #1200 enough?

I was wrong with my [assumption](https://github.com/c2corg/v6_ui/pull/1200#issuecomment-268485119) ("In fact this is only an issue when doing FTS queries"). No matter what kind of query, ElasticSearch has this kind of limitation.

brunobesson:
>I was wrong
Could the PR be simplified then by removing https://github.com/c2corg/v6_ui/blob/a0281b6dd55df1bcd7b2f0014873190fd699eeef/c2corg_ui/static/js/search/pagination.js#L94-L98?

c2corg_ui/static/js/search/pagination.js
```diff
@@ -47,6 +47,12 @@ app.PaginationController = function($scope, ngeoLocation) {
 
   /**
    * @type {number}
+   * @private
+   */
+  this.maxResultOffset_ = 9970;
```
tsauerwein: The default limit is 10000, see: https://www.elastic.co/guide/en/elasticsearch/reference/master/search-request-from-size.html

tsauerwein:
> Could the PR be simplified then by removing https://github.com/c2corg/v6_ui/blob/a0281b6dd55df1bcd7b2f0014873190fd699eeef/c2corg_ui/static/js/search/pagination.js#L94-L98?

Yes!

c2corg_ui/static/js/search/pagination.js
brunobesson: Sure, be you have to take the result size limit into account too (usually 30). But maybe I should check the computation logic again

brunobesson:
@tsauerwein @dkocich I updated the PR to remove first condition on FTS and make the offset limit more clear.

c2corg_ui/static/js/search/pagination.js
```diff
@@ -47,6 +47,12 @@ app.PaginationController = function($scope, ngeoLocation) {
 
   /**
    * @type {number}
+   * @private
+   */
+  this.maxResultOffset_ = 10000;
```
asaunier: I don't have a strong opinion about this but since this is a constant, it could be a static property with a ``@const`` keyword as for instance in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/map/map.js#L348-L352

c2corg_ui/static/js/search/pagination.js
```diff
@@ -85,11 +91,10 @@ app.PaginationController.prototype.handleSearchChange_ = function(event,
...
   this.offset = this.location_.getFragmentParamAsInt('offset') || 0;
-  // don't show the "Go to last page" button when doing a full-text-search
-  // (the full result set has to be iterated in ElasticSearch when doing
-  // requests with large offsets)
-  this.showGoToLastPage = !(this.location_.hasFragmentParam('q') &&
-    this.location_.getFragmentParam('q') !== '');
+  // don't show the "Go to last page" button if the offset is above
+  // maxResultOffset_, or the API will return an error
+  var lastOffset = this.total - (this.total % this.limit);
+  this.showGoToLastPage = lastOffset <= this.maxResultOffset_ - this.limit;
```
tsauerwein: Wouldn't `this.showGoToLastPage = this.total < this.maxResultOffset_` be enough?

brunobesson: You're right :sob:

brunobesson:
Everything should be ready for a last review now :wink:

asaunier:
For me it looks OK but I have not tested and not really followed the discussions about this part.
@tsauerwein is the PR OK for you?

tsauerwein:
Yes, this looks good. Thanks, @brunobesson!

Issue #1486, opened 2017-02-08
Scroll up with pagination buttons


asaunier:
In the advanced search, it would be handy to go back to the top of the list after clicking on the pagination buttons (especially the ones in the bottom toolbar).

I have tried to add commands like
``window.scrollTo(0, 0);``
or
``$('.list').scrollTop = 0;`` 
to the ``goTo`` actions in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/pagination.js#L99-L136 (or to type the commands in my console) but with no success yet. An idea anyone? :)

A few tips using angular and ``scrollTop``:
http://stackoverflow.com/questions/23736647/how-to-retain-scroll-position-of-ng-repeat-in-angularjs
http://stackoverflow.com/questions/24040985/scroll-to-top-of-div-in-angularjs

tsauerwein:
`$('.documents-list-section').scrollTop(0)` does the trick.

brunobesson:
Working on demo.

PR #1489, opened 2017-02-09
Scroll to top when changing page in advanced search

brunobesson:
Related to #1486 

(let's try to avoid using jquery when not necessary)


c2corg_ui/static/js/search/pagination.js
```diff
@@ -133,6 +136,14 @@ app.PaginationController.prototype.goToLast = function() {
   var nextOffset = this.total - (this.total % this.limit);
   this.location_.updateFragmentParams({'offset': nextOffset});
   this.scope_.$root.$emit('searchFilterChange');
+  this.scrollToTop();
+};
+
+/**
+ * @private
+ */
+app.PaginationController.prototype.scrollToTop = function() {
```
asaunier: Since it is a private method, the naming convention is to have a trailing underscore: ``scrollToTop_``.

PR #1490, opened 2017-02-10
Add some more values in manifest.json

brunobesson:
These are better if user wants to add a shortcut to camptocamp.org on his mobile phone.


asaunier:
Seems OK to me but I don't know much about this part.

brunobesson:
This was basically some feedback from lighthouse extension on chrome. It only adds some basic values.

* name, short_name are apparently used for the name of the shortcut
* background_color is used when loading the "application", so that the color is consistent with our them (I reused the lightgray)

Issue #1491, opened 2017-02-10
Error on some outing's versions page because of missing geometry


dkocich:
~~Related to https://github.com/c2corg/v6_ui/pull/1479 ?~~

~~Maybe we should remove/adjust this also https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L149-L151 ?~~

On this page http://localhost:6558/outings/808860/fr/1273493 I get error:

```
pyramid_mako.MakoRenderingException: 
Traceback (most recent call last):
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid_mako/__init__.py", line 148, in __call__
    result = template.render_unicode(**system)
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/template.py", line 471, in render_unicode
    as_unicode=True)
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/runtime.py", line 838, in _render
    **_kwargs_for_callable(callable_, data))
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/runtime.py", line 873, in _render_context
    _exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
  File "/home/dkocich/projects/c2corg/v6_ui/.build/venv/lib/python3.5/site-packages/mako/runtime.py", line 899, in _exec_template
    callable_(context, *args, **kwargs)
  File "/home/dkocich/projects/c2corg/v6_ui/c2corg_ui/templates/base.html", line 150, in render_body
    vcRecaptchaServiceProvider.setSiteKey('${settings["recaptcha_site_key"]}');
  File "/home/dkocich/projects/c2corg/v6_ui/c2corg_ui/templates/outing/view.html", line 38, in render_moduleConstantsValues
    <%
TypeError: 'NoneType' object is not subscriptable
```

asaunier:
I don't think it has a connection with #1479 since the latter PR made no change in ``settings["recaptcha_site_key"]``.

Are you sure your ``recaptcha_site_key`` is set in your ``common.ini``?

tsauerwein:
We just looked at it together, it's unrelated to recaptcha. The problem is here:
[...]

PR #1492, opened 2017-02-10
Fix error in old outing versions without any geometry

dkocich:
related to https://github.com/c2corg/v6_ui/issues/1491

there is `geometry: null` in this version of an outing which was causing an error ...
http://localhost:6548/outings/815298/fr/1284552

I removed map too when there is no geometry.

### DIFF
--- a/c2corg_ui/templates/outing/view.html
+++ b/c2corg_ui/templates/outing/view.html
@@ -42,26 +42,30 @@ other_langs, missing_langs = get_lang_lists(outing, lang)
       else:
           geometry = None
   %>
-  module.value('mapFeatureCollection', {
-    "type": "FeatureCollection",
-    "properties": {},
-    "features": [{
-      "type": "Feature",
-      % if geometry:
-        "geometry": ${geometry | n},
-      % endif
-      "properties": {
-        "title": ${dumps(locale['title']) | n},
-        "lang": "${locale['lang']}",
-        "documentId": ${outing['document_id']},
-        "module": "outings"
-      }
-    }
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/outing/view.html
```diff
@@ -36,15 +36,20 @@
...
+      else:
+          geometry = None
   %>
   module.value('mapFeatureCollection', {
     "type": "FeatureCollection",
     "properties": {},
     "features": [{
       "type": "Feature",
-      "geometry": ${geometry | n},
+      % if geometry:
```
asaunier: What's the point of adding a feature with no geometry?
If no geometry is available, the feature should be skipped from the featurecollection.

c2corg_ui/templates/outing/view.html
```diff
@@ -36,15 +36,20 @@
...
   <%
-      geometry = outing['geometry']['geom_detail'] \
-          if outing['geometry']['geom_detail'] else outing['geometry']['geom']
+      if outing.get('geometry'):
+          geometry = outing['geometry']['geom_detail'] \
+              if outing['geometry']['geom_detail'] else outing['geometry']['geom']
+      else:
+          geometry = None
   %>
   module.value('mapFeatureCollection', {
```
tsauerwein: If there is no geometry and you are not showing the map, you don't have to create the GeoJSON structure. You could simply use `module.value('mapFeatureCollection', null);` in that case.

asaunier:
Please squash the commits.

Issue #1322, opened 2017-01-08
Moderator tools


stef74:
- [x] Merging doc. See also API issue https://github.com/c2corg/v6_api/issues/386
- [x] Deleting doc. See also API issue https://github.com/c2corg/v6_api/issues/465
- [x] Reset geometry for a route/outing. See also https://github.com/c2corg/v6_ui/issues/1120.

loicperrin:
See §9 (p50) of RFP. I think there is also another issue on this subject.

loicperrin:
- [ ] Tool for reporting a problem -> see #306 and #317

asaunier:
> Tool for reporting a problem -> see #306 and #317

@loicperrin That looks not that prioritary. I recommend to keep the issues distinct.

desnoes:
History of association (usefull tool for moderators) is already handled by #193

desnoes:
I tested the merge between 2 documents: the associated documents (route and WP) of the source are correctly associated with the target document.

PR #1494, opened 2017-02-13
Add dialog to merge documents

tsauerwein:
Related to https://github.com/c2corg/v6_ui/issues/1322

image

image

### DIFF
--- a/c2corg_ui/static/partials/mergedocuments.html
+++ b/c2corg_ui/static/partials/mergedocuments.html
@@ -1,0 +1,31 @@
+<div class="alert alert-warning">
+  <p class="text-center"><strong translate>Warning: This action cannot be undone!</strong></p>
+</div>
+
+<p translate>Merging a source document with a target document transfers all
+associations of the source document to the target document, and sets up a
+redirection from the source to the target document.</p>
+<p><strong translate>Note that comments have to be transferred manually in Discourse before merging.</strong></p>
+
+<div class="document-to-merge">
+  <p>
+    <label>Source:</label> {{ mergeCtrl.sourceDocument.title }}
+    (<a ng-href="/{{ mergeCtrl.module }}/{{ mergeCtrl.sourceDocument.document_id }}"
+        target="_blank">{{ mergeCtrl.sourceDocument.document_id }}</a>)
+  </p>
+  <p>
+    <label>Target:</label>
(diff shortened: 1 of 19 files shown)


c2corg_ui/static/partials/mergedocuments.html
```diff
@@ -0,0 +1,31 @@
...
+
+<p translate>Merging a source document with a target document transfers all
+associations of the source document to the target document, and sets up a
+redirection from the source to the target document.</p>
+<p><strong translate>Note that comments have to be transferred manually in Discourse before merging.</strong></p>
+
+<div class="document-to-merge">
+  <p>
+    <label>Source:</label> {{ mergeCtrl.sourceDocument.title }}
+    (<a href="/{{ mergeCtrl.module }}/{{ mergeCtrl.sourceDocument.document_id }}"
```
dkocich: `ng-href`([doc](https://docs.angularjs.org/api/ng/directive/ngHref)) instead of href is better maybe?

tsauerwein: > Using AngularJS markup like {{hash}} in an href attribute will make the link go to the wrong URL if the user clicks it before AngularJS has a chance to replace the {{hash}} markup with its value.

This should not be a problem because the links are shown in dialog. When the dialog is shown, Angular should have replaced the ``{{hash}}` markup. But I replaced it anyway.. :)

asaunier:
The code looks good to me.

A few comments about the usability:

in the modal box:
* it would probably make sense to explicitly indicate the document type of the documents (WP, routes, outings). For outings would it be possible/interesting to add the date to the outing title?
* if there are too many results in the search results list, currently a "more results" link is displayed => it doesn't make sense. It would be more relevant to show "too many results, refine your search" as when searching for associated routes during an outing editing. See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/simplesearch.js#L268-L278
* perhaps a "cancel" button would be handy?

After confirming "merge documents", it would make sense to refresh the current page (merged!) and redirect to the target document.

Else the tool seems to work as expected (old URL redirects to target doc, associations are transfered).

tsauerwein:
Thanks for your feedback, Alex!

The dialog now looks like this:

image

asaunier:
Have you solve the "too many results" problem?

tsauerwein:
> Have you solve the "too many results" problem?

Yes.

Issue #1120, opened 2016-12-07
Geometry for an outing/route cannot be reset


brunobesson:
There is currently no way to reset the geometry of a route or outing.
When doing this, the default centroid (or whatever strategy is used currently) should be applied again.

In V5, this action is reserved to moderators. Should we open this to all users? IMHO, yes.

See also #1056

desnoes:
Fixed on demo

PR #1501, opened 2017-02-15
Make it possible to remove geometry of document

tsauerwein:
Related to https://github.com/c2corg/v6_ui/issues/1120

### DIFF
--- a/c2corg_ui/static/js/editing/documentediting.js
+++ b/c2corg_ui/static/js/editing/documentediting.js
@@ -193,8 +193,12 @@ app.DocumentEditingController = function($scope, $element, $attrs, $http,
     return;
   }
 
-  this.scope.$root.$on('mapFeaturesChange', function(event, features, isReset) {
-    this.handleMapFeaturesChange_(features, isReset || false);
+  this.scope.$root.$on('mapFeaturesChange', function(event, features) {
+    this.handleMapFeaturesChange_(features);
+  }.bind(this));
+
+  this.scope.$root.$on('mapFeaturesReset', function(event, initialGeometry) {
+    this.handleMapFeaturesReset_(initialGeometry);
   }.bind(this));
 };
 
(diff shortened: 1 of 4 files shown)


c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -216,38 +220,52 @@ app.DocumentEditingController.prototype.filterData = function(data) {
...
+
+/**
+ * @param {string} str
+ * @private
+ */
+app.DocumentEditingController.prototype.toCoordinates_ = function(str) {
+  var point = /** @type {ol.geom.Point} */
+      (this.geojsonFormat_.readGeometry(str));
+  return this.getCoordinatesFromPoint_(point);
+}
```
asaunier: The linter reports a missing semi-colon after the closing curly brace.

c2corg_ui/static/partials/map/map.html
```diff
@@ -13,7 +13,12 @@
     </div>
   </div>
 
-  <button type="button" ng-show="mapCtrl.edit" ng-click="mapCtrl.resetFeature()"
-          class="btn btn-primary btn-xs map-reset-feature" translate>Reset feature</button>
+  <button type="button" ng-show="mapCtrl.canReset()" ng-click="mapCtrl.resetFeature()"
+          class="btn btn-primary btn-xs map-reset-feature"
+          title="{{ 'Reset geometry' | translate }}" translate>Reset geometry</button>
```
asaunier: Why not reusing the existing string "Reset feature"?

asaunier:
I have tested the branch in my instance:
* advanced search maps: OK
* view page maps: OK
* editing page maps: the reset/delete buttons works pretty well. I have noticed a weird behaviour though when dragging an existing node: the old node remains in place (with the segments coming/going from/to the other nodes) AND a new node+segments are created. See:
capture du 2017-02-15 17-48-47

But wait: I can reproduce the same behaviour in http://www.demov6.camptocamp.org/routes/45168/fr/mont-de-grange-couloir-de-pertuis (without the proposed changes) and I have no problem with another GPS track. Perhaps there is something weird with the track above?

Then except the minor comments in the code review, it looks good to me.

dkocich:
> Perhaps there is something weird with the track above?

I think that this track is somehow damaged. After displaying height profile from KML there are two tracks which are exactly same (the user didn't go twice because it would be slightly different). Then you can add nodes (to both lines at once) or move nodes (one by one).


It is visible from gpx too (two points at one place).

asaunier:
Thanks @dkocich for investigating! Then it's only a data problem with this track and the tool behaviour is correct.

c2corg_ui/static/partials/map/map.html
tsauerwein: I was expecting a comment about this. ;) *Feature* is a technical term that you can only understand if you have a background in GIS. I know that you want to keep the changes to the translation files to a minimum. But as a new translation has to be given for "Delete geometry" anyway, I thought it would be a good opportunity to get rid of this confusing string.

tsauerwein:
Thanks for testing, @asaunier. And thanks for debugging the problem, @dkocich!

c2corg_ui/static/partials/map/map.html
asaunier: "Geometry" is also a GIS-related term. I don't think normal users know more about "geometries" than about "features".
If you think "feature" is not common enough (for me it is not a problem because the translation can be translated in some other way by translators), it would make more sense to use "track" instead.

I have not tested with point documents. Does your tool apply to such features?

tsauerwein: > I don't think normal users know more about "geometries" than about "features".

I think everybody knows at least the word *geometry*.

> I have not tested with point documents. Does your tool apply to such features?

Yes, you can also delete point geometries.

asaunier: > I think everybody knows at least the word geometry.

Everybody knows the word feature :P 
Nervermind, OK for me to change as you suggest.

> Yes, you can also delete point geometries.

I have just tested it. It works. Of course deleting the geometry for a WP prevent to save the document changes until a new geometry is created, so it's not that relevant to be able to delete a WP geometry. But at least it is consistent with other document types and we have point doc types (images!) where we should be able to remove the geometry.

Issue #1340, opened 2017-01-11
Images uploader should upload images sequentially


asaunier:
Forked from https://github.com/c2corg/v6_ui/issues/929#issuecomment-269907514

A problem with the image uploader is that **it tries to upload all submitted images at the same time**. Which is a non-sense: if too many images are uploaded at the same time, the bandwidth might be not enough to send all of them in a reasonable amount of time, thus causing timeout errors.

**It would make more sense that only 1 or 2 images are uploaded at the same time and that the uploader starts uploading additional images when the latter uploads are done.**

By the way I have realized that when many images are uploaded in the same batch, the API request to save the associated data (not the files!) takes a while... Perhaps there are things to optimize in the API to reduce the delay needed to save those data?

tsauerwein:
> By the way I have realized that when many images are uploaded in the same batch, the API request to save the associated data (not the files!) takes a while... 

See also: https://github.com/c2corg/v6_api/issues/569

arnaud-morvan:
@asaunier : Do you remember why you put a ``this.scope_.$apply()`` executed every second in the ``upload_`` method here : 
https://github.com/c2corg/v6_ui/blob/ecb06aa78ac6bea80d56a74b2f9d96a4d4254553/c2corg_ui/static/js/imageuploader.js#L190
Because it seems to work perfectly without that.

asaunier:
I don't remember at all.
Feel free to remove it if you think it is useless.

stef74:
![img_0229](https://cloud.githubusercontent.com/assets/3962997/23184860/a86723b0-f881-11e6-8fc7-a65ddca6b42d.PNG)
I havé Try to upload 10 photos ... but error message on browser

arnaud-morvan:
This shoud be related with SSL error on demo : #637

desnoes:
Upload of several images works fine. It is impossible to save the images.

arnaud-morvan:
Note that we have also added Retry/Cancel buttons in case of failure.
It is possible to test it by enabling throttling in browser development tools.

desnoes:
> enabling throttling in browser development tools

What do you mean exactly ? I am not an expert in those things. My browser cannot be configure that much at work.

arnaud-morvan:
If you are using firefox or chrome, You can use Ctrl + Shift + i to open the dev tools
In chrome you should go to the network tab, and there is a throttling option in the top toolbar.
In Firefox I don't find it.

[...]

PR #1502, opened 2017-02-16
Upload images sequentially

arnaud-morvan:
Relate #1340

### DIFF
--- a/c2corg_ui/static/js/imageuploader.js
+++ b/c2corg_ui/static/js/imageuploader.js
@@ -167,7 +167,7 @@ app.ImageUploaderController = function($scope, $uibModal, $compile, $q,
     return this.files;
   }.bind(this), function() {
     if (this.files.length) {
-      this.upload_();
+      this.processFiles_();
     }
   }.bind(this));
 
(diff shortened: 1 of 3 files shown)


arnaud-morvan:
@asaunier @tsauerwein : this is ready for review.

asaunier:
@arnaud-morvan That's awesome! :)
The code is OK for me.

I have tested the branch in my instance and it works really well (images uploaded 1 by 1 when drag-dropping for instance 10 images at the same time).

However I have often observed (almost every time) that at least an image (once 2) fails uploading because of a timeout error. Sometimes thrown only a few seconds after drag-dropping my set of images. It's not always the same image that fails when uploading the same 7 or so images.

arnaud-morvan:
When using GPRS throttling, with big images (several Mo), I get normal timeout after 120s (the timeout of by images backend).

When we get an error, the corresponding image immediately disappear from the list, and the user cannot retry the upload, and do not knows which image have disappear.
IMHO we should keep and show informations about failed images, maybe add a retry button.

Apart from this, everything works fine for me. I do not reproduce your error. But if you get a timeout after a few second, this may come from your images backend, or it is not a real timeout.
https://github.com/c2corg/v6_ui/blob/3ed29c2fbfaaebf1c02659ae7771208fb27dd4e6/c2corg_ui/static/js/imageuploader.js#L259:L261

I suppose you use the default value for the images backend:
```
export image_backend_url = http://images.demov6.camptocamp.org
```
I will give it a try.

asaunier:
> IMHO we should keep and show informations about failed images, maybe add a retry button.

+1
If the failed image card is removed, it's hard to figure out what image has not been uploaded.

> I suppose you use the default value for the images backend:

Indeed: https://github.com/c2corg/v6_ui/blob/master/config/alex#L7

If you think this is a problem not related with your changes, we can merge your PR, have the demo tested and if the problem persists, create a new issue about this.

arnaud-morvan:
I've tested with the demov6 images backend that use async io http workers, and I get no timeout, everything seems ok with slow connection and big images.

arnaud-morvan:
Now I show the retry/cancel button instead of removing the image from list.
User has to manually retry/cancel the image before he get the save button.

asaunier:
> Now I show the retry/cancel button instead of removing the image from list.

I have just tested it: nice.
For me it's OK as is even though a bit of styling of the error message would be interesting to make it even more obvious (red font or red background?).

asaunier:
Thanks for updating the error message styles :)
It looks really great now!

One more thing (sorry!): I have noticed the "Error while uploading the image" is not translated in my french interface whereas the string is available and translated in Transifex and in the i18n JSON file.

It seems using ``this.alerts_.gettext()`` is not enough to have the string actually translated:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/alertservice.js#L35-L37

arnaud-morvan:
Translation done using the translate attribute in the partial template.

asaunier:
I have just tested it, it works fine now. Please merge when Travis is OK.

PR #1504, opened 2017-02-16
Document deleting tool

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1322
Requires https://github.com/c2corg/v6_api/pull/610

I have also made some changes for code and layout consistency.
For instance small layout changes in the documents merging modal to make it more consistent with the association removal tool and eventually the document deleting tool.

capture du 2017-02-16 20-33-03

capture du 2017-02-16 20-33-34

capture du 2017-02-16 20-33-16

### DIFF
--- a/c2corg_ui/static/js/deletedocument.js
+++ b/c2corg_ui/static/js/deletedocument.js
@@ -1,0 +1,98 @@
+goog.provide('app.deleteDocumentDirective');
+goog.provide('app.DeleteDocumentController');
+
+goog.require('app');
+
+/**
+ * This directive is used to manage the dialog to delete a document.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.deleteDocumentDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppDeleteDocumentController',
+    controllerAs: 'deldocCtrl',
+    templateUrl: '/static/partials/deletedocument.html',
+    bindToController: {
+      'module': '<'
+    }
+  };
+};
+
+app.module.directive('appDeleteDocument', app.deleteDocumentDirective);
+
+
+/**
+ * @param {appx.Document} documentData Data set as module value in the HTML.
+ * @param {app.Api} appApi appApi.
+ * @param {app.Alerts} appAlerts
+ * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
+ * @param {ui.bootstrap.$modalStack} $uibModalStack $uibModalStack.
+ * @constructor
+ * @ngInject
+ */
+app.DeleteDocumentController = function(documentData, appApi, appAlerts, gettextCatalog, $uibModalStack) {
+
+  /**
+   * @type {appx.Document}
+   * @export
+   */
+  this.documentData = documentData;
+
+  /**
+   * @type {app.Api}
+   * @private
+   */
+  this.appApi_ = appApi;
+
+  /**
+   * @type {app.Alerts}
+   * @private
+   */
+  this.appAlerts_ = appAlerts;
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.gettextCatalog_ = gettextCatalog;
+
+  /**
+   * @type {ui.bootstrap.$modalStack}
+   * @private
+   */
+  this.$uibModalStack_ = $uibModalStack;
+
+  /**
+   * @export
+   * @type {string}
+   */
+  this.module;
+};
+
+
+/**
+ * @export
+ */
+app.DeleteDocumentController.prototype.deleteDocument = function() {
+  this.appApi_.deleteDocument(this.documentData.document_id).then(
+      function(response) {
+        this.closeDialog();
+        this.appAlerts_.addSuccess(this.gettextCatalog_.getString(
+            'Document successfully deleted'
+        ));
+      }.bind(this)
+  );
+};
+
+
+/**
+ * @export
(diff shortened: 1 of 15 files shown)


asaunier:
Perhaps a problem to deal with:

If an error occurs (eg. because trying to delete a main waypoint), the API actually returns the error
capture du 2017-02-16 21-18-59

but in the API service, ``response.data`` is ``null`` and the alert only reports an "unknown error".
https://github.com/c2corg/v6_ui/blob/b938c3e26cf5f1a7d37f6841bf341b915b32b65d/c2corg_ui/static/js/apiservice.js#L706

Else playing with WP/routes/outings the deletion works as expected.

asaunier:
I have tried to delete images: the document is actually deleted but I still can load the image file from the s3 storage (or maybe it's a cache issue?).

I have also encountered errors when deleting an outing with archives (OK if the outing has only 1 version), see https://github.com/c2corg/v6_api/pull/610#issuecomment-280454074

asaunier:
The PR is ready to be reviewed.

There is still a problem with the error message that are not reported (see comment above at https://github.com/c2corg/v6_ui/pull/1504#issuecomment-280452196) but I have no clue how to fix this. It seems the response is not correctly parsed by the ``$http`` service? Or perhaps there is something missing in the response itself? As anyone a tip?

arnaud-morvan:
Concernant le ``delete`` sur un main waypoint, je me tape un ``400`` sans le header ``Access-Control-Allow-Origin:*``
est-ce normal ???

asaunier:
Error 400 is expected because a main waypoint cannot be deleted.
No clue about the header.

arnaud-morvan:
Dans d'autres validations, je vois : 
```
request.errors.add('body',
                           '{}/{}'.format(document_id, lang),
                           'Document not found')
```
je pense que ce sont ces erreurs qui sont bien affichées dans la partie client

arnaud-morvan:
I've tried to replace ``raise HTTPBadRequest`` by ``request.errors.add``, the server return 400 with cors header and the message is correctly displayed on client side.

c2corg_ui/static/js/deletedocument.js
```diff
@@ -0,0 +1,98 @@
...
+  this.module;
+};
+
+
+/**
+ * @export
+ */
+app.DeleteDocumentController.prototype.deleteDocument = function() {
+  this.appApi_.deleteDocument(this.documentData.document_id).then(
+      function(response) {
```
arnaud-morvan: Don't you think we should redirect th user once the document is deleted.

c2corg_ui/templates/helpers/view.html
```diff
@@ -120,20 +120,17 @@ <h3 class="text-center"><span translate>Merge documents</span> (<span x-translat
   </div>
 </%def>
 
-<%def name="show_delete_document_dialog()">\
+<%def name="show_delete_document_dialog(module)">\
+  <%
+      # get the module without the 's' at the end ('routes' -> 'route')
+      # to reuse existing translation strings
+      document_type = module[:-1]
```
arnaud-morvan: Not sure it is useful to add put an ending ``s`` to remove it afterward.
Should be simpler to pass the ``document_type`` string to the ``show_delete_document_dialog``

tsauerwein:
> Concernant le delete sur un main waypoint, je me tape un 400 sans le header Access-Control-Allow-Origin:*
> est-ce normal ?

The CORS headers seem not to be set if the response is generated by the error views: https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/__init__.py#L31

It would probably make sense to set the CORS headers manually in that case. Similar to here: https://github.com/camptocamp/c2cwsgiutils/blob/64d55f24b58ee2d92ae822563a51bedf4aa97bdc/c2cwsgiutils/errors.py

c2corg_ui/templates/helpers/view.html
asaunier: Perhaps. I have copied-paster what @tsauerwein has done for document merging :package:
The plural form is used (at least in the document merging) for instance to generate URL https://github.com/c2corg/v6_ui/blob/4ae76094a25c5b92e8d30a7950c8ea0e2854c7c1/c2corg_ui/static/partials/mergedocuments.html#L17

I have preferred to use a consistent approach for the deletion.

c2corg_ui/static/js/deletedocument.js
asaunier: We have discussed this with @tsauerwein about the documents merging (that does not redirect either).

It would make sense to redirect. both after merging or deleting, yes. On the other hand, as for now we have no way to display a confirmation message ("the document has successfully been merged or deleted") after a redirection. And keeping the document displayed could be useful if the user finally realizes they have forgotten to save some data. Finally it's a moderator-only feature and can be improved later if the moderators think it's better to redirect.

arnaud-morvan:
It seems ok for me, but it's a huge modification, and I am not really aware with everything here.

asaunier:
The huge change was on the API side (and is merged now).
Changes in this PR are consistent with what has been done for the documents merging.

I'm testing the branch with your recent changes in the API (CORS etc.) and will merge if it works as expected and if no one objects in the meantime.

Thanks for the review!

asaunier:
OK the changes seem OK, I actually get an explicit error message (with the description of what has prevented to delete the document). Perhaps the ``: document_id`` mention at the end of the message is a bit weird but it's not critical, we'll see later if needed.

sans titre

PR #1506, opened 2017-02-20
[WIP] iPhone-Safari map buttons - styles fix

dkocich:
Selector for Safari 9+ should work fine. https://browserstrangeness.bitbucket.io/css_hacks.html?#webkit I don't have issues with desktop and my Android, but I am afraid that iOS Chrome will be broken now because there is no selector only for Safari which wouldn't work in iOS version of Chrome too.

If someone with iPhone would confirm that the iOS Chrome is broken now, then probably the only option is to restyle the map only for Safari using JS (somewhere in `map.js`)???
http://browserhacks.com/#hack-462504c4ab517b400419d1b3d73d943a

### DIFF
--- a/config/dkocich
+++ b/config/dkocich
@@ -1,6 +1,6 @@
 include config/dev
 
-export debug_port = 6558
+export debug_port = 3000
 export instanceid = dkocich
 export base_url = /${instanceid}
 export api_url = http://localhost:6548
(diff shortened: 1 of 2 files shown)


stef74:
I can test only if on demo site.

asaunier:
Replaced by https://github.com/c2corg/v6_ui/pull/1755

Issue #1295, opened 2017-01-04
Reloading the outings search page with date filter causes an error


pvalsecc:
How to reproduce:

- From the home page, click "outings"
- Click all filters
- Set a start date and an end date
- Click "Show results"
- Click on one of the results
- Click the back button of your browser

You'll see that:
![4tableseditor](https://cloud.githubusercontent.com/assets/1180608/21644971/6faaee24-d290-11e6-8ab6-6d6495fa5a80.png)

asaunier:
@pvalsecc Thanks for the report!

It seems the problem only occurs with dates (I have no problem restoring a search page with other filters).
When there is a date filter in the outing search, reloading the page generates the following error message:

> ﻿angular.js:13920 TypeError: Cannot read property 'push' of undefined
    at app.OutingFiltersController.<anonymous> (http://localhost:6555/static/113592d/js/search/outingfilters.js:66:17)

See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/outingfilters.js#L66

asaunier:
It seems to be an inheritance problem:

``this`` in the overridden method ``getFilterFromPermalink`` (see https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/outingfilters.js#L66) seems to refer to the parent class and expose only the properties of the parent class, not the properties of the child class (among them ``dates``, defined at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/outingfilters.js#L26).

![sans titre](https://cloud.githubusercontent.com/assets/1192331/21692336/8973a840-d37c-11e6-9da7-695a915a8d77.png)

Not sure where this behaviour comes from and how to make sure ``this`` refers to the actual object with the child properties.

desnoes:
I was unable to reproduce the bug.

asaunier:
When following the initial list of steps I indeed get the error (with Chrome).

fbunoz:
> I was unable to reproduce the bug (FF).

Test simlply with this link : https://www.camptocamp.org/outings#date=2017-01-16,2017-01-21

PR #1511, opened 2017-02-21
Fix activation of date filter from URL for outings

dkocich:
related to https://github.com/c2corg/v6_ui/issues/1295

`this.dates` was undefined in the outing's filter controller (even if it should be defined [here](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/outingfilters.js#L21-L25)) and it caused an error ([here](https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/outingfilters.js#L65)) while the date was pushed into it.

I moved `this.dates` to the parent controller and it works fine for both document types.

It is a bit strange because for Xreport this works fine and both controllers are very similar (if not same).

Test -
* Outings http://localhost:6558/outings?debug#date=2016-11-01%252C2017-02-20
* Xreports http://localhost:6558/xreports?debug#date=2016-11-29%252C2017-02-20

### DIFF
--- a/c2corg_ui/static/js/search/outingfilters.js
+++ b/c2corg_ui/static/js/search/outingfilters.js
@@ -18,11 +18,6 @@ app.OutingFiltersController = function($scope, ngeoLocation, ngeoDebounce,
 
   goog.base(this, $scope, ngeoLocation, ngeoDebounce, advancedSearchFilters);
 
-  /**
-   * @type {Array.<Date>}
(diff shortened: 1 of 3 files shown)


asaunier:
> I moved this.dates to the parent controller and it works fine for both document types.

But the goal of those child classes was actually to avoid having doctype-specific properties in the generic class...

PR #1512, opened 2017-02-21
Prefer EXIF field DateTimeOriginal to get image date

tsauerwein:
Some photos have more than one EXIF field related to the date time. So far we were using `DateTime`, which is the last time the photo was modified. It is better to use `DateTimeOriginal` (if available) which is the time the photo was taken.

Related to https://github.com/c2corg/v6_api/issues/627


tsauerwein:
Thanks for the review!

Issue #1342, opened 2017-01-11
Enable to load "prefs" filters in advanced search filters


asaunier:
Users would like to easily apply their preferences (currently used only for the homepage feed) to advanced search pages.

If applicable, activities filter would use the preferred activities and the map would be recentered on the preferred areas. 2 options for the areas:
* use the overall bbox of preferred areas
* actually filter on the preferred areas by adding a ``a=...`` parameter to the URL. For instance https://www.camptocamp.org/outings#a=14427,14421

Questions:
* It should possible to toggle on/off those preferred filters: on/off button in the interface?
* Should those filters be applied by default? (the initial URL must be consistent then and make it obvious preferred filters are applied)
* Should we use a special URL parameter to enable the preferred filters (eg. ``#pref=true``) or retrieve the user's preferences (via an API call? or from the localstorage?) such as ``#a=123,456&act=skitouring,ice_climbing``?

stef74:
Probably need to open discussion on trello/forum ... but I prefer discuss with possibility technically validated (and have macro cost).

But my opinion : 
- for activities is pertinent ( a toogle can be good idea ... reuse the same as in a feed)
- for aera the interact with map can be very confusion ... 

Probably we need to have : 
- map centering on the "union" of aera in pref user search by default ( and not reset it if change filter ... but is probably an other issue)
- offer possibility to filter on area show on the map ( as in airbnb we can filter on "quartier" ... but not search on it directly) You can check by defaut the area of user preference ... but we need to reuse the filtering aera on advanced search for area out of preference.

Concerning url ... they can be consistent, if you need to show on the url the filter added by preference. 
![capture](https://cloud.githubusercontent.com/assets/3962997/21854352/b487d05e-d81a-11e6-913a-142182b2706a.PNG)

desnoes:
> It should possible to toggle on/off those preferred filters: on/off button in the interface?

yes. If you are doing research in a region which was not set in your preference, you need to toggle off the preferred filters. Otherwise, if for instance you are searching for routes and then huts, the preferred filters will apply when you will switch between the 2 search and it will be very annoying.

>Should those filters be applied by default?

yes

asaunier:
Here are my suggestions about making the personal preferences (areas and activities) available in the various advanced search pages.

Concerned document types would be:
* outings, routes, images => areas + activities
* waypoints => areas
* areas => no prefs
* books => activities (but is it really useful?)
* articles => activities (but is it really useful?)
[...]

PR #1520, opened 2017-02-22
Add tool to load prefs in advanced search

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1342

Still some testing and fixes to do:
* [x] make sure the map is recentered on areas when loading areas prefs
* [x] make sure the activities selector is updated and consistent with the loaded prefs
* [x] when filters are cancelled (removing area prefs), make sure the map extent is reset, or at least consistent with the returned results

As for now I have not implemented any storage of the preferences as considered in https://github.com/c2corg/v6_ui/issues/1342#issuecomment-280308466 => prefs are loaded on demand when pushing the "apply my prefs" buttons, in my instance the delay is acceptable. Don't know yet if it could be OK in demo/prod as well.

capture du 2017-02-22 18-20-16
capture du 2017-02-22 18-20-42

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,6 @@ LIBS_JS_FILES += \
     c2corg_ui/static/lib/angular-bootstrap/ui-bootstrap-custom-1.3.2.min.js \
     c2corg_ui/static/lib/angular-bootstrap/ui-bootstrap-custom-tpls-1.3.2.min.js \
     node_modules/angular-gettext/dist/angular-gettext.min.js \
-    node_modules/angular-debounce/angular-debounce.js \
     node_modules/angular-messages/angular-messages.min.js \
     node_modules/angular-cookies/angular-cookies.min.js \
     node_modules/corejs-typeahead/dist/typeahead.bundle.min.js \
--- a/build.json
+++ b/build.json
@@ -22,7 +22,6 @@
       "c2corg_ui/externs/angular-1.5.js",
       "c2corg_ui/externs/angular-1.5-q_templated.js",
       "c2corg_ui/externs/angular-1.5-http-promise_templated.js",
-      "c2corg_ui/externs/debounce.js",
       "node_modules/openlayers/externs/bingmaps.js",
       "node_modules/openlayers/externs/closure-compiler.js",
       "node_modules/openlayers/externs/esrijson.js",
--- a/c2corg_ui/externs/debounce.js
+++ b/c2corg_ui/externs/debounce.js
@@ -1,7 +1,0 @@
-
-/**
- * @typedef {
- *  function(function(), number)
- * }
- */
-var debounce;
--- a/c2corg_ui/static/js/appmodule.js
+++ b/c2corg_ui/static/js/appmodule.js
@@ -26,6 +26,5 @@ app.module = angular.module('app', [
   'ngFileUpload',
   'slug',
   'vcRecaptcha',
-  'infinite-scroll',
-  'debounce'
+  'infinite-scroll'
 ]);
--- a/c2corg_ui/static/js/loadpreferences.js
+++ b/c2corg_ui/static/js/loadpreferences.js
@@ -113,7 +113,7 @@ app.LoadPreferencesController.prototype.loadPreferences_ = function() {
       this.location_.deleteFragmentParam('bbox');
     }
     this.location_.updateFragmentParams(params);
-    this.scope_.$root.$emit('searchFilterChange');
+    this.scope_.$root.$emit('searchFilterChange', true);
   }
 };
 
--- a/c2corg_ui/static/js/loadpreferences.js
+++ b/c2corg_ui/static/js/loadpreferences.js
@@ -1,0 +1,174 @@
+goog.provide('app.LoadPreferencesController');
+goog.provide('app.loadPreferencesDirective');
+
+goog.require('app');
+
+
+/**
+ * @return {angular.Directive}
+ */
+app.loadPreferencesDirective = function() {
+  return {
+    restrict: 'A',
+    controller: 'AppLoadPreferencesController',
+    controllerAs: 'lpCtrl',
+    bindToController: {
+      'module': '@appLoadPreferences',
+      'url': '@appLoadPreferencesUrl'
+    },
+    link: function(scope, el, attr, ctrl) {
+      el.click(function() {
+        ctrl.applyPreferences();
+      });
+    }
+  };
+};
+
+app.module.directive('appLoadPreferences', app.loadPreferencesDirective);
+
+
+/**
+ * @param {angular.Scope} $scope Directive scope.
+ * @param {ngeo.Location} ngeoLocation ngeo Location service.
+ * @param {app.Api} appApi Api service.
+ * @constructor
+ * @ngInject
+ * @struct
+ */
+app.LoadPreferencesController = function($scope, ngeoLocation, appApi) {
+
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
+
+  /**
+   * @type {ngeo.Location}
+   * @private
+   */
+  this.location_ = ngeoLocation;
+
+  /**
+   * @type {app.Api}
+   * @private
+   */
+  this.api_ = appApi;
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.module;
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.url;
+
+  /**
+   * @type {?appx.UserPreferences}
+   * @private
+   */
+  this.preferences_;
+
+  /**
+   * @type {Object}
+   * @private
+   */
+  this.params_ = {};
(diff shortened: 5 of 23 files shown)


asaunier:

To summarize the changes:
* I have removed ``angular-debounce`` since a debouncing tool is already available in ngeo (used in most part of the code). I have then replaced ``angular-debounce`` by ``ngeo.Debounce``.
* A ``loadPreferences`` directive has been created. It is used as an attribute of whatever button (works a bit like the existing ``protected-url-btn`` directive https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/protectedurlbtn.js)
* The directive can either be used to open a new page if an additional directive attribute is provided (eg. ``app-load-preferences-url="/outings"``) or to update the fragment params (eg. ``#act=skitouring&a=12345,12346``) of the current page (used in the advanced search pages).
* Prefs are retrieved from the API the first time the button is clicked, but the prefs are not saved between pages (I think the delay is acceptable and it saves some complexity - to be confirmed).
* Depending on the document type, ``act`` (activities) and/or ``a`` (areas) params are applied.
* When area prefs are loaded the map is recentered in the extent of those areas.
* When clicking the "clear filters" button, area prefs are removed but the map extent remains unchanged.
* To be able to recenter on the area prefs, I had to skip the checkboxes panels debouncing (if you click several checkboxes in a row, the API request is made only after a slight delay, not after every single click) because I needed to had a param to the ``searchFilterChange`` event (which seemed not possible with the debounce). Perhaps it was acceptable as is. But I have however added a ``checkboxes`` buffer that is synchronised with the filters through a debounce tool to recreate the former behaviour.

Screenshots above are still valid.

arnaud-morvan:
Unwanted behavior with activities checkboxes.
In /outings page, load my preferences, remove the filters, open activities dropdown and check one, "preferred activities" checboxes state change in background.

arnaud-morvan:
After the preceding workflow, re-check exactly the prefered activities. The result count is not the same because of the filter on the prefered regions has been removes. The prefered regions filter is not materialized on the page, and common user do not take care of the url. This is not obvious.

arnaud-morvan:
IMHO we should display something for the prefered areas filter in the seach form.

arnaud-morvan:
A "Prefered areas" check box or "Areas" dropdown.

arnaud-morvan:
Not sure the new button "See outins according to my preferences" in activity feed is at the right place. It takes a lot of place (with corresponding card / margins) for such button. May be possible move it inside the same card as "Add an outing".

asaunier:
> IMHO we should display something for the prefered areas filter in the seach form.
> A "Prefered areas" check box or "Areas" dropdown.

This is out of scope and would require much more work.

There are 679 available areas of 3 different types: https://www.camptocamp.org/areas
Using a simple list of checkboxes or a dropdown is not possible.
The topic of an area selector has already been discussed concerning the selection of areas in the preferences page. See https://github.com/c2corg/v6_ui/issues/598 
As for now a simplified selector based upon the search tool has been implemented.

There are discussions about adding an area filter section in the advanced search page but:
* is it really needed? It's already possible to do most of the geographical filtering using the map.
* we don't really know how to render this area selection tools. It's not that simple.
* applying prefs with this PR already does most of the job, it's better to split tasks to avoid spending weeks on a PR.

I really prefer not dealing with an area selector in this PR!

> Not sure the new button "See outins according to my preferences" in activity feed is at the right place. It takes a lot of place (with corresponding card / margins) for such button. May be possible move it inside the same card as "Add an outing".

Maybe you're right. I can try moving the button to the same card than the "add an outing" button.

arnaud-morvan:
Ok but it is really not obvious when the checkbox "Limit to visible bbox" is checked that the prefered areas are used instead. Maybe two radio buttons with new option "Limit to prefered areas".

arnaud-morvan:
Sorry the two apply the same time so it's an "and", not radio buttons.

asaunier:
> Ok but it is really not obvious when the checkbox "Limit to visible bbox" is checked that the prefered areas are used instead. Maybe two radio buttons with new option "Limit to prefered areas".

That sounds a bit complicated. Both in the interface (having too many inputs makes it harder to use) and in the code which is already not simple at all.

We have too much complexity in this projet. I prefer to keep things as simple as possible.

Clicking the "load my prefs" button is only a shortcut that "checks" activities and areas (perhaps langs in the future) for you depending on your prefs. But then you can change the activities (not the areas but perhaps later). So it's not very relevant to have inputs mentioning "preferd areas".
In addition having a "Limit to prefered areas" would be a duplicate of the "load my prefs" button which also limit to the prefered areas.

asaunier:
> In /outings page, load my preferences, remove the filters, open activities dropdown and check one, "preferred activities" checboxes state change in background.

Indeed. Thanks for catching this!

arnaud-morvan:
According that It seems really difficult to materialize the prefered areas filter in the UI (value is a list of areas, not a boolean). After you fix the clear filter behavior, it is ok for me to merge.

asaunier:
I have updated the PR to fix the error reported by @arnaud-morvan when clearing filters.

Concerning the position of the button in the home feed:

> May be possible move it inside the same card as "Add an outing".

this definitely makes sense! By the way adding a "See outings" button instead of "See outings according to my preferences" when not authenticated would make sense too (perhaps the "Outings" entry could be removed from the menu then?).

Unfortunately I have not been able to change the styles and have the buttons rendered nicely in the same card!

Here are the additional changes I did in the feed partial https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/feed.html
```
-  <article class="feed-list-item" protected-url-btn url="/outings/add" ng-if="!feedCtrl.userId">
+  <article class="feed-list-item" ng-if="!feedCtrl.userId">
     <div class="timeline-bullet icon-outings" uib-tooltip="{{'outing'| translate}}" tooltip-placement="left"></div>
     <div class="list-item vertical-align">
-      <button class="orange-btn btn" translate>add an outing</button>
+      <button class="orange-btn btn" protected-url-btn url="/outings/add" translate>add an outing</button>
+      <button class="green-btn btn" ng-if="userCtrl.auth.isAuthenticated()"
+              app-loading app-load-preferences="outings" app-load-preferences-url="/outings"
+              translate>See outings according to my preferences</button>
+      <a href="/outings" ng-if="!userCtrl.auth.isAuthenticated()">
+        <button class="green-btn btn" translate>See outings</button>
+      </a>
     </div>
   </article>
 
@@ -29,14 +35,6 @@
     </div>
   </article>
 
-  <article class="feed-list-item" ng-if="userCtrl.auth.isAuthenticated() && !feedCtrl.userId">
-    <div class="timeline-bullet icon-outings" uib-tooltip="{{'outing'| translate}}" tooltip-placement="left"></div>
-    <div class="list-item vertical-align">
-      <button class="green-btn btn" app-loading app-load-preferences="outings" app-load-preferences-url="/outings"
-              translate>See outings according to my preferences</button>
-    </div>
-  </article>
-
```
which more or less results in:

capture du 2017-03-02 17-02-14

capture du 2017-03-02 17-02-48

In addition the layout must be adapted when using a mobile screen.

I guess that on a mobile screen both buttons should take the full width of the card and be stacked above each other.
On wider screens we could have 2 side-by-side buttons (as the "Topoguide" and "Forum" buttons when not authenticated) with same width.

But I don't feel skilled enough to do so...

arnaud-morvan:
I do not find annoying to have the buttons stacked even on wide screen.
Note that having buttons with text as long as "See outings according to my preferences" is not usual and seems little weird.

asaunier:
> I do not find annoying to have the buttons stacked even on wide screen.

Shame on me but I am struggling to do it as well :P There are a bunch of styles on those buttons...

> Note that having buttons with text as long as "See outings according to my preferences" is not usual and seems little weird.

I understand but what text could we use instead?
It's probably useful to make people know that the button opens the outings page WITH the prefs, especially if the "Outings" entry is kept in the menu. Else people might not understand what's the use of this button?

Or perhaps we could:
* remove the "Outings" entry in the sidemenu
* have this button named "Last outings" both for authenticated users (with prefs) and for guests (no pref in that case).
?

asaunier:
I have updated the PR to group buttons "Add an outing" and "See outings" or "See outings according to my preferences" in the same card:

capture d ecran 2017-03-03 a 17 43 07

capture d ecran 2017-03-03 a 17 43 44

When the screen gets narrower:

capture d ecran 2017-03-03 a 17 44 01
capture d ecran 2017-03-03 a 17 44 13
capture d ecran 2017-03-03 a 17 44 24

Always having the buttons the same height would make sense (see the last screenshot, for a mobilephone screen). Don't know yet how to do it.

asaunier:
Any comment/review about this PR?
Is it OK for merging?

arnaud-morvan:
For me it's OK, it seems that it's the better thing we can do now.

asaunier:
@arnaud-morvan Thanks for reviewing!

PR #1522, opened 2017-02-23
Handle vimeo, youtube and dailymotion videos in topoguide documents

brunobesson:
Pretty similar (but simplified, too) to what was done previously. Accepts [video] http://xxx [/video] URLs.

(for reference: https://github.com/c2corg/camptocamp.org/blob/72e777075f2e6260b1d2ef563ab0f90c383038ba/web/forums/include/parser.php)

Is there a nicer way to iterate over the regex comparisons?

### DIFF
--- a/c2corg_ui/format/video.py
+++ b/c2corg_ui/format/video.py
@@ -1,0 +1,77 @@
+'''
+c2corg video Extension for Python-Markdown
+==============================================
+
+Converts video tags to advanced HTML video tags.
+'''
+
+from markdown.extensions import Extension
+from markdown.inlinepatterns import Pattern
+from markdown.util import etree
+import re
+
+VIDEO_RE = r'\[video\](.*?)\[/video\]'
+
+
+class C2CVideoExtension(Extension):
+
+    def extendMarkdown(self, md, md_globals):  # noqa
+        self.md = md
+
+        pattern = C2CVideo(VIDEO_RE)
+        pattern.md = md
+        # append to end of inline patterns
+        md.inlinePatterns.add('c2cvideo', pattern, "<not_strong")
+
+
+class C2CVideo(Pattern):
+
+    def handleMatch(self, m):  # noqa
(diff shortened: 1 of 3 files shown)


c2corg_ui/format/video.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class C2CVideo(Pattern):
+
+
+    def handleMatch(self, m):  # noqa
+        link = m.group(2).strip()
+
+        # youtube http://www.youtube.com/watch?v=3xMk3RNSbcc(&something)
+        match = re.search(r"https?:\/\/(?:www\.)?youtube\.com/watch\?(?:[=&\w]+&)?v=([-\w]+)(?:&.+)?(?:\#.*)?", link)
```
asaunier: The linter expects lines to be smaller than 80 characters. In cases where it's not possible to do so, you can tell the linter to ignore the line by adding ``# noqa`` at the end of the line. See the function name for instance (in that case the problem was the camelCase used instead of function_name but here we override a third-party lib class function).

asaunier:
Your code looks nice to me, except the line length issues.

> Is there a nicer way to iterate over the regex comparisons?

I don't know. I would say it is OK.

Don't forget to create an issue, so that we can list this change in the next release change log. Thanks!

brunobesson:
@asaunier should I care about snyk? It doesn't seem related to this PR.

asaunier:
> should I care about snyk? It doesn't seem related to this PR.

You can ignore it.

brunobesson:
But now because of snyk, CI is failing too :cry: The commit I added to limit video size on large screens is not related to it.

Unless you don't agree, I'll merge this PR tomorrow.

asaunier:
> But now because of snyk, CI is failing too

I have restarted Travis (there's a "restart build" button at the top of the Travis report). It passes now.

brunobesson:
Related to #1523

PR #1529, opened 2017-02-24
limit card size in feed

stef74:
To limit image distortion et too wide card.
It's not enough to improve globaly but simple in 1st approach.

### DIFF
--- a/less/feed.less
+++ b/less/feed.less
@@ -81,6 +81,7 @@
     background-color: #fbfbfb;
     border-radius: 10px;
     width: 40%;
+    max-width: 500px;
     transition: .3s;
 
     &:first-of-type .list-item {


brunobesson:
Wouldn't that accentuate the furstration of those having a wide screen which will now have even more green empty space?

stef74:
Not because you have blank space in card ... information density not change
... only image are more correct (currently in very large screen image are
not usable in feed).
Is no an optimal solution ... but it's better than existing solution.

2017-02-24 14:10 GMT+01:00 Bruno Besson <notifications@github.com>:

asaunier:
Does it impact the profile feed too?
I have noticed the the profiles in the home and profile pages had not the same width, which is not consistent. Will this PR fix this inconsistency?

stef74:
@asaunier not is font size issue. I have try to manage in https://github.com/c2corg/v6_ui/pull/1379

stef74:
An other way is to limit only image (that is the problem in fact). 
current screen width : 2500px


new one 
capture d ecran 2017-02-26 a 17 54 38


An evolution can be add more photo ... but it's an other job !

Issue #1382, opened 2017-01-18
Add a language filter to preferences


asaunier:
In the feed preferences page https://www.camptocamp.org/preferences

See the related API issue at https://github.com/c2corg/v6_api/issues/608

desnoes:
For information, I put here a description of how it worked in V5.

Sur la V5, y compris pour les invités, la langue de l'interface était présélectionnée par défaut en fonction de la langue du navigateur. On pouvait changer ensuite bien sûr.
La personnalisation de la langue des sorties/forum par défaut (y compris pour les invités) dépendaient de cette langue d'interface :

Si la langue de l'interface est FR, EN, DE ou EU : pas de personnalisation de langue, tout est affiché.
Si la langue de l'interface est ES ou CA : personnalisation en ES, CA, EN.
Si la langue de l'interface est IT : personnalisation en IT.
Ceci afin d'éliminer le FR dans les langues minoritaires mais un peu actives, tout en évitant d'avoir un contenu qui ne bouge pas pour les langues très peu actives.

L'algo a changé plusieurs fois en fonction du nb de sorties par mois dans chaque langue.
Il se trouve là : https://github.com/c2corg/camptocamp.org/blob/master/apps/frontend/lib/c2cPersonalization.class.php#L30-L73

asaunier:
Tag "high prio" and milestone "6.3" are not very consistent :P

asaunier:
@desnoes I have the feeling a language filter is a bit a non sense, bringing extra work just to remove content that could interest people.

In the home page, the forum block will anyway not be managed by preferences filters, but by the prefs set directly in the forum.

In the activity feed, you can already filter on regions (and activities), which already strips out most of the content that don't interest you.

True, you might still get reports written in other languages for outings done in your selected areas. But why hidding them? It's still useful data about the conditions, etc. : 
* Pictures are not language-dependant and are definitely interesting.
* Figures data (overall condition ratings, conditions per level - especially the amount of snow at various elevations - , crowding, hut status, elevation of the snow, etc.) are not or not much language-dependent.

Of course the text comments might be difficult to read if you don't understand the language but browsers often detect "foreign language" contents and propose a translation. I guess it will generalize progressively. Or we can add a automatic translating service (as in v5) with a button in documents whose language does not match the current interface language.

To summarize, I strongly recommend not to implement a language filter but rather to make sure "foreign" contents can be autotranslated on demand.

desnoes:
[...]

PR #1530, opened 2017-02-24
[WIP] Add a language preferences to the feed - UI part

dkocich:
related to issue https://github.com/c2corg/v6_ui/issues/1382

requires API part https://github.com/c2corg/v6_api/pull/642 to test 

STILL WIP!

Few notes to look at later -
- since language preference is an `ArrayOfEnum`, it makes sense to enable some kind of "sorting" on UI preferences setting page (I have IT+ES enabled and I want to see IT if both locales are available). Right now there are only simple checkboxes and language preference is chosen on API automatically

### DIFF
--- a/c2corg_ui/static/js/preferences.js
+++ b/c2corg_ui/static/js/preferences.js
@@ -54,7 +54,7 @@ app.PreferencesController = function($scope, $cookies, appAuthentication, appApi
 
   /**
    * @type {Array.<string>}
-   * @public
+   * @export
    */
   this.lang_preferences = [];
 
@@ -84,7 +84,7 @@ app.PreferencesController = function($scope, $cookies, appAuthentication, appApi
       this.followed_only = data.followed_only;
       this.lang_preferences = data.lang_preferences;
 
-      console.log('data jsou po otevr. stranky ...', data);
+      console.log('data after page load ...', data);
 
       this.scope_.$watch(function() {
         return this.followed_only;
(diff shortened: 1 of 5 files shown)


c2corg_ui/templates/preferences.html
```diff
@@ -37,6 +37,22 @@ <h3 translate class="green-text">Areas</h3>
         </div>
       </div>
 
+      <div class="preferences-langs">
+        <h3 translate class="green-text">langs</h3>
```
asaunier: Please use real strings. For instance ``Languages``.

c2corg_ui/templates/preferences.html
```diff
@@ -37,6 +37,22 @@ <h3 translate class="green-text">Areas</h3>
...
+        <p translate ng-show="prefCtrl.lang_preferences.length == 0">Filter the document languages of your interest by selecting them.</p>
+        <div class="form-group full full-group">
+          <ul class="types">
+            <li ng-repeat="lang in ${default_langs}" ng-click="prefCtrl.updateLanguages(lang)">
+              <div class="checkbox">
+                <input type="checkbox" ng-checked="prefCtrl.lang_preferences.indexOf(lang) > -1">
+                <span>{{lang | translate}} {{lang}}</span>
```
asaunier: Why do you need to display both the translated string and the untranslated one?

Instead of checkboxes, it would make sense to use buttons that would switch colors (grey to orange?) when activated (like for the activities).

c2corg_ui/static/js/preferences.js
```diff
@@ -101,6 +117,33 @@ app.PreferencesController.prototype.updateActivities = function(activity) {
...
+      return item !== language;
+    });
+  } else {
+    this.lang_preferences.push(language);
+  }
+
+  console.log('inserting cookie', this.lang_preferences.join(','));
+  var d = new Date();
+  d.setFullYear(d.getFullYear() + 1); // today + 1 year
+  this.cookies_.put('preferred_lang', this.lang_preferences.join(','), {
```
asaunier: Why do you need cookies?

asaunier:
Still todo:
* [ ] remove debug code
* [ ] remove langs cookies introduced by this PR (still don't understand why we need them)
* [ ] fix/improve the languages selector in the preferences page

stef74:
@asaunier @fjacon what is the status of this PR ?

PR #1531, opened 2017-02-24
use another (shorter) translation key for float button help on mobile

brunobesson:
Should be e.g. 'Images' instead.
Currently it is - in french - 'Ajouter des images' so only 'Ajouter' is displayed.

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -563,7 +563,7 @@
       <div class="float-button float-favorites" tooltip-placement="left" uib-tooltip="{{'Add images' | translate}}"
            ng-click="uplCtrl.openModal()" ng-if="userCtrl.hasEditRights('${doctype}', ${options})">
         <span class="glyphicon glyphicon-picture"></span>
-        <p class="float-button-text" translate>Add images</p>
+        <p class="float-button-text" translate>Images</p>
       </div>
     % endif
 


asaunier:
> Currently it is - in french - 'Ajouter des images' so only 'Ajouter' is displayed.

Why not using "Add" then?

Using "Add images (short)" does not really make sense to me. Those strings are not translation keys. They are the base translation strings in english (actually it is not needed to translate those strings in Transifex if the original translation fits).

brunobesson:
>Why not using "Add" then?

Because maybe the translation we want is not "Add", but more specific? I am not the sole translator.

>Those strings are not translation keys

Well, I don't see it the same way you do. If so, we should have corrected many many string whose base translation is incorrect. E.g. because it misses a capital letter. IMHO, we should see american english as translation keys, and english (UK ?) as the default translation to be used. But whatever, I can deal with your point of view.

asaunier:
> Well, I don't see it the same way you do.

I assure you it's the way it has been designed in v6. That's the reason why you have long texts as original translation strings in the templates or partials instead of short keys such as "preferences_intro". For instance https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/preferences.html#L13-L14

There are several reasons for this:
* those strings actually appear in the rendered HTML pages, thus are read as is by robots and search engines.
* translators have access to the full text to be translated instead of confusing short keys with no context at all.
* ???

Then I would recommend to be consistent and keep using the same approach.

> because it misses a capital letter

I know. It's a pity the case is not correct for several strings.

brunobesson:
I reworked the PR to use `Images` key, which already exist in the project (but for another purpose, akthough translation matches)

asaunier:
@brunobesson Feel free to merge the PR, thanks.

Issue #1535, opened 2017-03-03
Image infos


stef74:
When you use this syntaxte : [img=857571] test [/img] the information show are not correct ... 

![image](https://cloud.githubusercontent.com/assets/3962997/23547893/446f2d86-0005-11e7-8075-48b8dd40ad53.PNG)

stef74:
A probably link issue ... that sometime we can show exif infos (correctly display) with this step to reproduce : 
    Cliquer sur une image dans les commentaires
    Cliquer sur le bouton infos ⓘ
    Fermer l'apperçu Ｘ

Chaque image dans les commentaires à alors toutes les infos exif affichées de la photo précédemment sélectionnée.

stef74:
An other thing if I can see image toolbar 

![capture](https://cloud.githubusercontent.com/assets/3962997/23551618/e6300264-0017-11e7-9174-4c67ce88b16e.PNG)

asaunier:
Can be tested in this outing: https://www.camptocamp.org/outings/857542/fr/breitwangflue-flying-circus

> When you use this syntaxte : [img=857571] test [/img] the information show are not correct ...

It seems the "infos" title that is automatically added is the "infos" section that is displayed in the slideshow when clicking the "i" button.

When the page is first loaded, an "infos" title is shown below the image:

<img width="476" alt="capture d ecran 2017-03-03 a 14 47 52" src="https://cloud.githubusercontent.com/assets/1192331/23553309/87c0a644-0020-11e7-915a-b07fb1c6a434.png">

If I click on the image, the slideshow is shown. If I click on the "i" button, metadata are shown in the slideshow. After closing the slideshow, those metadata are shown in the text of the outing:

<img width="452" alt="capture d ecran 2017-03-03 a 14 45 39" src="https://cloud.githubusercontent.com/assets/1192331/23553393/dd04f164-0020-11e7-8542-9446d01d5600.png">

> An other thing if I can see image toolbar

Indeed.

In addition there is no border around the embedded images.

On the other hand all those weird behaviours cannot be observed in a route document. For instance https://www.camptocamp.org/routes/857378/fr/breitwangflue-flying-circus

<img width="388" alt="capture d ecran 2017-03-03 a 14 54 55" src="https://cloud.githubusercontent.com/assets/1192331/23553520/67849a4c-0021-11e7-9bb8-27d8d6458fcf.png">

=> borders and legend are shown as expected 
=> no "infos" title is shown under the image
[...]

PR #1536, opened 2017-03-03
Fix images embedded in outing description

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1535


asaunier:
Thanks for reviewing!

PR #1537, opened 2017-03-03
Image position in the feed

stef74:
Move image in the feed to right (same "line") if enought space.

By the way image are now under the card.

This PR was tested with older image layout (but probably no conflict to merge after). 

capture d ecran 2017-03-03 a 18 23 14
capture d ecran 2017-03-03 a 18 16 31

PR #1538, opened 2017-03-04
collapsed menu on medium screen

stef74:
Many user conplain that menu take place in desktop (on small and medium screen).

A very simple way to fix is to considere that small screen are a tablet (collapsed menu by defaut).

An other gain that decrease menu size is to give more space for photo (when the PR #1537 is merged)

### DIFF
--- a/less/variables.less
+++ b/less/variables.less
@@ -1,9 +1,9 @@
-@big: ~"only screen and (min-width: 1400px)";
-@desktop: ~"only screen and (min-width: 1100px) and (max-width: 1399px)";
-@tablet: ~"only screen and (min-width: 769px) and (max-width: 1099px)";
+@big: ~"only screen and (min-width: 1900px)";
+@desktop: ~"only screen and (min-width: 1365px) and (max-width: 1899px)";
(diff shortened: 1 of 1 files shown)


stef74:
previous : 
capture d ecran 2017-03-04 a 22 18 35

brunobesson:
IMHO it is not a good idea until we handle the ads another way (i.e. have small horizontal ad on small screens + need to expand menu to select an entry when collapsed)

stef74:
You have probably right ... a small evoluation can be have the feed more on the left to offer more space for the photo

asaunier:
> until we handle the ads another way

Do we already have an issue about the menu and ads?

Please note that with PR https://github.com/c2corg/v6_ui/pull/1520 an "outings" button is added in the first card of the home feed => perhaps we could remove the "outings" entry in the left menu?

desnoes:
> Please note that with PR #1520 an "outings" button is added in the first card of the home feed => perhaps we could remove the "outings" entry in the left menu?
I think we need to find a solution suitable for all screen sizes. 
cf. https://forum.camptocamp.org/t/affichage-des-bannieres-de-pub-et-resolution-de-lecran/181263/26

PR #1545, opened 2017-03-10
Fix collapse with flex in document views

arnaud-morvan:


asaunier:
Could you provide before/after screenshots or precise what part of document views should be tested?
Is there a related issue?

asaunier:
I don't really see the difference with associated docs.
In the "information" section, it seems the titles are moved upward, not sure it is correct.

Before:

capture d ecran 2017-03-13 a 16 44 56

After:

capture d ecran 2017-03-13 a 16 45 09

arnaud-morvan:
This shoud not be due to my changes, try to collape the associated route, by clicking in the section header, the body height is set to 0 but as there is a display: flex that override the (collapse) display: none, the cards appear outside the section.
This PR apply the display: flex only on elements that do not have the (collapse) "in" class.

arnaud-morvan:
It seems I shoud add a rule to apply the flex when there is no collapse class.

asaunier:
Indeed the associated documents sections are incorrectly displayed in prod when folded (OK with your branch).

For instance https://www.camptocamp.org/routes/736816/fr/sainte-victoire-le-signal-l-innominata

capture du 2017-03-14 18-28-12

Issue #239, opened 2016-03-18
Handle protected documents


tsauerwein:
API issue: https://github.com/c2corg/v6_api/issues/217

With https://github.com/c2corg/v6_api/pull/191 the `protected` field of documents will be included. Protected documents can only be edited by moderators.

In the UI maybe we can disable the edit button and show a hover message ("Document is protected, ...") when a user is not moderator. The edit page should also show a "not authorized" message and the form should not be shown.

Example for a protected document in v5:
http://www.camptocamp.org/summits/213785/it/piramide-ovest-di-somalbosco

asaunier:
The protection status of a document is set in a document attribute, right?
Wich means we could have a "protected" attribute in the editing forms only available to the moderators?

tsauerwein:
I was thinking about providing a separate web-service for changing the protected status (similar to merging a document). The protected checkbox might be a bit lost in the normal editing form, so maybe it's better to have an explicit command to lock/unlock a document (as part of the moderator tools).

asaunier:
In v5, the lock/unlock function is indeed a separated function.

Wouldn't it save time to have a checkbox in the editing forms (only for moderators)? We would not have to create a dedicated webservice, it's quite similar to the "quality" attribute (~metadata), and locking/unlocking creates a new version in the history anyway (doesn't it?).

Since it is a special tool for advanced users, it's not that a big deal if it is a bit lost. As I said it could be in a metadata section placed near the "save" button, containing also the quality attribute and the "comments about the change" attribute.

tsauerwein:
> Wouldn't it save time to have a checkbox in the editing forms (only for moderators)? We would not have to create a dedicated webservice, it's quite similar to the "quality" attribute (~metadata), and locking/unlocking creates a new version in the history anyway (doesn't it?).

For me it would actually be easier to have a separate web-service.

asaunier:
OK then. I have created an issue about that in the API: https://github.com/c2corg/v6_api/issues/217

I guess it would be a function available in the "More" menu button.

asaunier:
@fjacon This task cannot be done until the API webservice for it is available: https://github.com/c2corg/v6_api/issues/217

asaunier:
TODO:

* [ ] add a "protect" (resp. "unprotect") action in the "plus" button (after the "merge" and "delete" actions) - only for moderators
* [ ] this new action should call the protect/unprotect API service, see https://github.com/c2corg/v6_api/issues/217
[...]

PR #1553, opened 2017-03-15
Add protect/unprotect document tool

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/239

TODO:

* [x] add a "protect" (resp. "unprotect") action in the "plus" button (after the "merge" and "delete" actions) - only for moderators
* [x] this new action should call the protect/unprotect API service, see https://github.com/c2corg/v6_api/issues/217
* ~~display a warning in the document if it is protected? (a bit like the disclaimer in archived versions of a document? https://www.camptocamp.org/waypoints/37355/fr/1305265)~~
* [x] hide the "edit" button for non moderators
* [x] disable the editing form (or display a warning) for non moderators

### DIFF
--- a/c2corg_ui/static/js/protectdocument.js
+++ b/c2corg_ui/static/js/protectdocument.js
@@ -1,0 +1,98 @@
+goog.provide('app.protectDocumentDirective');
+goog.provide('app.ProtectDocumentController');
+
+goog.require('app');
+goog.require('app.Alerts');
+goog.require('app.Api');
+goog.require('app.Authentication');
+
+
+/**
+ * @return {angular.Directive} The directive specs.
+ */
+app.protectDocumentDirective = function() {
+  return {
+    restrict: 'A',
+    controller: 'appProtectDocumentController',
+    controllerAs: 'protectCtrl',
+    templateUrl: '/static/partials/protectdocument.html'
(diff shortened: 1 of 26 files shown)


asaunier:
This PR is ready for review (as well as the API counterpart at https://github.com/c2corg/v6_api/pull/644).

capture du 2017-03-16 16-03-37

capture du 2017-03-16 14-53-51

c2corg_ui/static/js/protectdocument.js
```diff
@@ -0,0 +1,104 @@
...
+};
+
+
+/**
+ * @export
+ */
+app.ProtectDocumentController.prototype.protect = function() {
+  if (this.auth_.isModerator()) {
+    this.api_.protectDocument(this.documentId_).then(function(response) {
+      this.protected = true;
```
asaunier: That would make sense indeed. I will make the change, thanks!

asaunier:
I have updated the PR to directly update ``this.documentData`` when changing the ``protected`` attribute.

> I also note that it is possible (even if the edit button is not there) for a non moderator user to get the edit page for a protected document using url.

Indeed. But a warning message is shown saying the document is not editable (see screenshot above).

PR #1560, opened 2017-03-20
feed position more left on desktop

stef74:
Use same feed position in desktop than in tablet. 

It's give more space to put photo en right. 

**Before** 
capture d ecran 2017-03-20 a 21 53 59

**After**
capture d ecran 2017-03-20 a 21 50 26

### DIFF
--- a/less/feed.less
+++ b/less/feed.less
@@ -100,7 +100,7 @@
     border-radius: 10px;
     width: 80%;
     @media @big {
-      max-width: 50%;    
+      width: 60%;    
     }
     transition: .3s;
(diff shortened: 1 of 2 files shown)


stef74:
last rendering on medium and big screen : 

capture d ecran 2017-03-23 a 21 20 36
capture d ecran 2017-03-23 a 21 20 15

@asaunier @desnoes ok for merging ?

asaunier:
OK for me. What it looks like in my instance (images are broken, but that's not related with the proposed PR).

capture du 2017-03-24 12-11-54

PR #1566, opened 2017-03-22
Apply description styles to other text attributes in detail view

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1535#issuecomment-287640045

### DIFF
--- a/less/viewdetails.less
+++ b/less/viewdetails.less
@@ -728,7 +728,7 @@ app-add-association {
     }
   }
 }
-.view-details-description {
+.view-details-section {
   img.image {
     max-width: 100%;
   }


asaunier:
@brunobesson @arnaud-morvan Any objections? I have tested for various document types and noticed no side effects (extensive testing in the demo would be useful though).

brunobesson:
I haven't been able to follow all the developments that have been made recently, sorry. I assume you are simply trying to expand the styling to other sections?
That is fine to me!

asaunier:
> I assume you are simply trying to expand the styling to other sections?

Exactly. An alternative solution would be to add a ``view-details-description`` class to the other text attributes that could contain images. But that would not really make sense to use a "description" keyword except for the ``description`` attribute itself. So I preferred generalizing the style to the parent element ("section") and it seems to work.

PR #1567, opened 2017-03-23
Add mardown editor

arnaud-morvan:
Relate #305 

* Create new directive ``appMarkdownEditor``
* Apply ``app-markdown-editor`` on every ``textarea`` tag

Note that tooltips and all markdown editor translations are completely handled using angular-gettext and transifex and fully updated using the language selector without reloading the page.

### DIFF
--- a/c2corg_ui/templates/base.html
+++ b/c2corg_ui/templates/base.html
@@ -34,6 +34,7 @@ closure_library_path = settings.get('closure_library_path')
     <link rel="stylesheet" href="${request.static_path('%s/photoswipe/dist/photoswipe.css' % node_modules_path)}">
     <link rel="stylesheet" href="${request.static_path('%s/photoswipe/dist/default-skin/default-skin.css' % node_modules_path)}">
     <link rel="stylesheet" href="${request.static_path('%s/slick-carousel/slick/slick-theme.css' % node_modules_path)}">
+    <link rel="stylesheet" href="${request.static_path('%s/bootstrap-markdown/css/bootstrap-markdown.min.css' % node_modules_path)}">
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/build.css')}">
     <link rel="stylesheet" href="${request.static_path('c2corg_ui:static/build/build-print.css')}" media="print">
 % else:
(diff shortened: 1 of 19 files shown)


arnaud-morvan:
@asaunier : Ready for review.

package.json
```diff
@@ -13,12 +13,14 @@
   },
   "license": "AGPL-3.0",
   "dependencies": {
+    "angular-sanitize": "1.5.8",
```
asaunier: What is this lib used for?

asaunier:
What's the use of the "4 arrow" icon placed at the end of the toolbar and after the textarea?

capture du 2017-03-23 18-36-06

Could you make the heading tool creates ``<h3>`` tags (ie. ``##`` markup) instead of ``<h4>`` tags (``###``)? I think in v5-migrated documents, h3 are more commonly used. For instance https://www.camptocamp.org/routes/53781/fr/mont-blanc-arete-des-bosses

Most likely we won't need the code and quote tools but will have to add custom tools for wikilinks, images etc.

package.json
arnaud-morvan: Good catch, it was an angular-markdown-editor dependency, but I do not use it. will be removed ...

arnaud-morvan:
@asaunier : This button put the editor in fullscreen mode.
Note that you snapshot seems to miss bootstrap-markdown.min.css.
You may have to use make clean or delete the deps.css file manually to get it refreshed.

asaunier:
> This button put the editor in fullscreen mode.

OK, it works fine after a "make clean":

capture du 2017-03-24 10-28-27

Would it be possible to have a tooltip on this full-screen tool?

Inactive textareas now have a gray background (a bit like in github by the way). Using a lighter gray would be nicer I think. As well as using the same font than in other inputs.

arnaud-morvan:
Help button with link page describing syntax and custom tags seems mandatory, does this page exists somewhere ?

asaunier:
See https://www.camptocamp.org/articles/151910/fr/aide-mise-en-forme-du-texte

arnaud-morvan:
I've updated my PR, for me this is ready for merging.
Custom buttons (wikilinks, images etc.) should be added in later pull requests.
Here is a snapshot : 
markdown-editor

c2corg_ui/templates/base.html
```diff
@@ -141,6 +143,7 @@
          module.constant('imageUrl', '${image_url}');
          module.constant('discourseUrl', '${discourse_url}');
          module.constant('authUrl', '${request.route_path('auth')}');
+         module.constant('textFormatingUrl', '${request.route_path('articles_view_id', id=151910)}');
```
asaunier: Isn't it possible to add this constant only for editing forms?

arnaud-morvan: I do not want to add it in each edit template, here is the place that we will seach for it. Note that ``route_path`` is really cheap.

asaunier: OK

asaunier:
@arnaud-morvan Feel free to merge your PR.

PR #1573, opened 2017-03-30
fix alignement condition rating

stef74:


stef74:
overflow with ellipsis is name is too big : 

capture d ecran 2017-03-30 a 20 23 05

max 3 activities : 
capture d ecran 2017-03-30 a 20 43 22

@asaunier ok for merge ?

PR #1578, opened 2017-03-30
Outing card

stef74:


asaunier:
Delete branch after merging a PR :P

stef74:
Yes ... but not just after merging ... on one branch I have forgot a commit ... so it's more simple if branch allready exist.

asaunier:
I don't think so. If the branch has been merged, you can really reuse it to add a new commit, you have to create a new one.
Then it's better to always delete the branch after merging the PR.

Issue #1581, opened 2017-04-03
Add print button


brunobesson:
Although users could simply press Ctrl+P, many seem to still be looking for a print button. We could add one more option in the plus menu.

desnoes:
ok on demo

PR #1582, opened 2017-04-03
Print button

brunobesson:
* Adds print button to "plus" menu
* Adds icons for dropdown actions
* Changes "plus" menu icon

movement5

### DIFF
--- a/c2corg_ui/static/js/viewdetails.js
+++ b/c2corg_ui/static/js/viewdetails.js
@@ -339,7 +339,7 @@ app.ViewDetailsController.prototype.createTopic = function() {
   }.bind(this), function(resp) {
     if (resp.status == 400) {
       var topic_id = resp['data']['errors'][0]['topic_id'];
-      if (topic_id != undefined) {
+      if (topic_id !== undefined) {
         this.documentService.document['topic_id'] = topic_id;
         this.getComments();
       }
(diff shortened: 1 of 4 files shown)


brunobesson:
See #1581

brunobesson:
Or maybe, assuming there is need for this button only on desktops and tablets, we could add the print button as a top level button between comments and plus, hidden on mobile.
What do you think?

asaunier:
> What do you think?

I don't have a strong opinion. I feel a bit amazed that we need to add such a print button: people don't even know the basic features of a browser/website? no surprise then that they sometimes have a hard time with c2corg which offers a lot of advanced features.
I am a bit concerned that adding another button will make the other buttons harder to figure out (the fewer buttons, the better?).

brunobesson:
Let's stick to this proposal then, we can easily switch to another solution if needed. @desnoes @mgerbaux do you have a non-dev point of view about this subject?

mgerbaux:
If I understand Alex's point of view, I think that adding a dedicated printing button make people more confident than using the default browser 's one : they would think that lead to a good printing quality rather than a screen print, many of them do not know very well the behavior of their browser.

brunobesson:
>I think that adding a dedicated printing button make people more confident

OK, that's why I added this button in the plus button dropdown. The question being: is that enough, or do we need a "top leve" print button?

mgerbaux:
I think we can leave it in the "plus" sub-menu

desnoes:
> I think we can leave it in the "plus" sub-menu

I agree with this solution

Issue #1555, opened 2017-03-16
Add a tool to remove a linguistic version of a document


asaunier:
See API issue at https://github.com/c2corg/v6_api/issues/645

TODO:
* add a new entry in the "plus" dropdown next to the "delete document" entry (moderators only) https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/view.html#L594-L595
* extend the ``app-delete-document`` directive to add an optional lang parameter https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/deletedocument.js
* update the existing deletion modal window to show that only the locale will be deleted (if no the only one available)
* pass the optional lang parameter to ``deleteDocument`` method in the ``apiservice`` and call the API URL accordingly https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/apiservice.js#L719-L730

desnoes:
I cannot delete this version of a route : http://www.demov6.camptocamp.org/routes/828490/en/chamechaude-essai-en. Same thing with the italian version : http://www.demov6.camptocamp.org/routes/828490/it/essi-it

WP: same pb with http://www.demov6.camptocamp.org/waypoints/828492/en/essai-en

I did the following experiment:

- i've created a WP (http://www.demov6.camptocamp.org/waypoints/828493/fr/toto-sommet) in french.
- i've created an english version of this document and then deleted it successfully.
- i've created an italian version, and then tried to delete. It was not possible (error message).

It seems it is not possible to delete more than one linguistic version of a document.

arnaud-morvan:
This may be due to the ESDeletedLocale primary key, this table is not purged.

asaunier:
@desnoes @arnaud-morvan Arf, right, the primary key is incorrect.
I have created an issue about this https://github.com/c2corg/v6_api/issues/653 and will make a PR soon.

Thanks for testing :)

asaunier:
PR is available here https://github.com/c2corg/v6_api/pull/654

asaunier:
The PR is now finalized and ready for review.
In my instance it is now possible to delete several locales (linguistic versions) of the same document, following the steps described by @desnoes at https://github.com/c2corg/v6_ui/issues/1555#issuecomment-293480695

desnoes:
ok on demo

PR #1587, opened 2017-04-04
Add a locale deletion tool

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1555

API part: https://github.com/c2corg/v6_api/pull/648

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -760,4 +760,20 @@ app.Api.prototype.deleteDocument = function(documentId) {
 };
 
 
+/**
+ * @param {number} documentId
+ * @param {string} lang
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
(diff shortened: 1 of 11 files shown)


arnaud-morvan:
Note that, if I load some document (fr), change the locale using the locale dowpdown selector to (de), fr is still in the url, but I see the german version. If I delete the current locale, it deletes the french version.

asaunier:
> change the locale using the locale dowpdown selector to (de),

It does not change the document locale. Only the language of the interface.
If you want to see other document locales you have to use the "Plus" button > "View in other language".
Actually it's the same if you want to edit the document: if you load the FR locale, switch the interface to DE and press the editing button, you'll still edit the FR locale.

Perhaps the distinction between the interface language and the document locales language is confusing but at least it is not related to the this feature.

> there is still some finishes to be done on the corresponding API pull request.

The API PR is now ready :)

asaunier:
@arnaud-morvan OK for merging?

Issue #1516, opened 2017-02-22
Improve access to the SERAC DB


desnoes:
The access to the SERAC DB is not user friendly. Solution: create a dedicated home page for SERAC (similar to the topoguide HP: https://www.camptocamp.org/topoguide)

asaunier:
@desnoes @mgerbaux We need more details about the content of this page. Could you precise the needs?

(partial) TODO:
* [ ] Add a new route/view/template for this static page
* [ ] Design template according to the desired content
* [ ] Update the URL of the xreport entry in the side menu

mgerbaux:
I propose this type of page
![1412112537_1597118560 jpg](https://cloud.githubusercontent.com/assets/22116605/24581859/b4dfdd06-1723-11e7-93ed-fe0b11b96058.png)

it should be homogenized in design with https://www.camptocamp.org/topoguide (if needed, we can use a darker picture)

asaunier:
@mgerbaux Thanks! Looks good to me.

asaunier:
A very first and unfinished version is available here https://github.com/c2corg/v6_ui/pull/1591

desnoes:
I let @mgerbaux validate this issue. I found the result very nice!

PR #1591, opened 2017-04-06
Add a home page for SERAC

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1516

TODO:
* [x] make the background image bigger and more centered
* [x] margins around the intro block (with title and text)
* [x] factorize CSS with the styles of the topoguide home?
* [x] better responsive layout

(tips are welcome because CSS is not my cup of tea)

The URL of this new page will be https://www.camptocamp.org/serac => for search engines optimization should we use and other URL with more explicit keywords? For instance https://www.camptocamp.org/serac-incidents-accidents ? Ditto for the page title and description meta?


desnoes:
@asaunier : it's nice! Unfortunately i cannot help you on css.

asaunier:
PR is ready to be reviewed.

A few screenshots for several screen sizes:

capture du 2017-04-10 15-32-27

capture du 2017-04-10 15-32-52

capture du 2017-04-10 15-33-30

What about my question:
> The URL of this new page will be https://www.camptocamp.org/serac => for search engines optimization should we use and other URL with more explicit keywords? For instance https://www.camptocamp.org/serac-incidents-accidents ? Ditto for the page title and description meta?

asaunier:
@desnoes @stef74 Any idea about this question?

> The URL of this new page will be https://www.camptocamp.org/serac => for search engines optimization should we use and other URL with more explicit keywords? For instance https://www.camptocamp.org/serac-incidents-accidents ? Ditto for the page title and description meta?

mgerbaux:
>  The URL of this new page will be https://www.camptocamp.org/serac 

the shorter the better. Even if it could be better for search engines to add key words in url, I think we shoud keep an easy memorizing or typing url.

desnoes:
@asaunier : i have no idea. I follow the Martin's proposition.

I guess we will have to translate some sentences. If yes, do you have the list off key words used by transifex ?

asaunier:
> I guess we will have to translate some sentences. If yes, do you have the list off key words used by transifex ?

New untranslated strings will automatically be available in Transifex when this pull request is acccepted and merged. Those new strings can be seen in the screenshots above.

stef74:
explicit is better for google ! 
An other simple way is to use explict url by defaut and create a shorter url that redirect to explicit url

On my personnal opinion nobody type this url directly

asaunier:
> use explict url by defaut and create a shorter url that redirect to explicit url

For me it is a clear no.
The rule is "keep things as simple as possible" :P

stef74:
Alex ... is not complex you cannot refuse to deal with google !

For me best is simple si only use explicit url ... short url is not needed

asaunier:
> New untranslated strings will automatically be available in Transifex when this pull request is acccepted and merged.

Actually it seems there are already available in Transifex (that's weird because I thought only strings from the "master" branch were pushed to Transifex!?).
https://www.transifex.com/camptocamp-association/c2corg_ui/translate/#fr/v6/111346875?translated=no

> Alex ... is not complex you cannot refuse to deal with google !
For me best is simple si only use explicit url ... short url is not needed

What I mean is we don't want to have several URLs for the same page. Please stick to one. Either a short one or a long one. Let me know.

asaunier:
@arnaud-morvan Thanks for reviewing!

I merge the PR as is. We'll do a another PR if the URL must be changed.

PR #1598, opened 2017-04-25
orientation overlap

stef74:


stef74:
capture d ecran 2017-04-25 a 12 38 11

Now orientation is only show 2 and after "..."

stef74:
ready for review

brunobesson:
A note aside: couldn't we imagine using SVG fonts similar to the wind rose for this? Would it be readable enough?

asaunier:
>  couldn't we imagine using SVG fonts similar to the wind rose for this?

That looks like a good idea :)

Issue #1096, opened 2016-12-04
Contextual Help


stef74:
TDB

desnoes:
Requirements:

- A question marking is displayed when the mousse hover over a field
- When clicking on a field, a window is opened. It contains a text which gives information about this field (ex: difficulty)
- The window contains a close button.

example: http://www.pre-prod.dev.camptocamp.org/outings/527014/fr/gouffre-d-enfer-cascade-du-gouffre-d-enfer

stef74:
I have more simple approach allready use ... if you over with a mouse or click in mobile you can show black pop-up (as in card)

![tooltip](https://cloud.githubusercontent.com/assets/3962997/23164623/4f76c06a-f838-11e6-9b8e-b5c83e2c7dea.png)

No need question mark  or close buttun.

In term of code is simple add : 
uib-tooltip="{{'height_diff_difficulties'| translate}}" tooltip-append-to-body="true"

You have only to defined a key for translation. 

I can do the job but need : 
- define key : You can simple use field_name_description 
- which field need contextual help and where (not necessary all).

brunobesson:
I don't agree. Tooltips aren't appropriate for this kind of contextual help which can be pretty long (it is often several paragraphs).

stef74:
We can do multine in tooltips ? 

For mobile on chrome it's work not on safari. But for me contextual help in mobile is not necessary (mobile cannot have the same feature than desktop). 

I have not the knowledge to propose a better solution ... just capacity to copy an existing one.

arnaud-morvan:
It is possible to open modal dialog, sometime on the element.onClick, but for links, we should add a "i" button near the link. The modal is closed by clicking (touching) outside.
Here is a snapshot : 
![modal](https://cloud.githubusercontent.com/assets/6095494/23998280/de88c0e0-0a54-11e7-905d-1fdca2608179.png)

fbunoz:
We must have contextual help on tablet and mobile.

asaunier:
Using a modal window is indeed probably a good idea, especially if the help content is a bit long.
In v5 you could have a small window listing for instance the possible values of a rating attribute with explanations about those values.

fjacon:
@desnoes we need your OK and one example of help text that we could add on demo.
thanks.

desnoes:
Here is an example of a help text for the selection of the activity.

**Activités**
[...]

PR #1601, opened 2017-04-26
Add context help on activities label

arnaud-morvan:
Relate #1096

Add new directive ``app-context-help`` for openning contextual help in modal dialog.

For adding more contextual help, you shoud :
 
* Add the ``app-context-help`` directive on the label, example: 
```
<label app-context-help
                context-help-title="{{'activities' | translate}}"
                context-help-content="<b>Ceci est un test</b>">
                <span translate>activities</span> *
              </label>
```

For longer help texts, it possible to use a partial/template that will be placed in the $templateCache (this will reduce the network traffic):
```
<label app-context-help
                context-help-content-url="/static/partials/contexthelp/activities.html"
                context-help-title="{{'activities' | translate}}">
                <span translate>activities</span> *
              </label>
```
And the corresponding template ``c2corg_ui/static/partials/contexthelp/activities.html``: 
```
<p translate>Type d'activités : dans une liste ou un document, passez la souris sur les pictogrammes pour connaître leur signification.</p>
<p translate>Un document peut être associé à une ou plusieurs activités :</p>
<ul>
  <li translate>ski, surf : randonnée à ski et/ou surf</li>
  <li translate>raquette : randonnée en raquettes</li>
...
```

I've added an indicator that there is a context help available:
image

Here is the modal dialog on desktop:
image

On mobile:
image

### DIFF
--- a/c2corg_ui/static/js/appmodule.js
+++ b/c2corg_ui/static/js/appmodule.js
@@ -26,5 +26,6 @@ app.module = angular.module('app', [
   'ngFileUpload',
   'slug',
   'vcRecaptcha',
-  'infinite-scroll'
+  'infinite-scroll',
+  'ngSanitize'
 ]);
--- a/c2corg_ui/static/js/contexthelp.js
+++ b/c2corg_ui/static/js/contexthelp.js
@@ -1,0 +1,82 @@
+goog.provide('app.contextHelpDirective');
+goog.provide('app.ContextHelpModalController');
+
+goog.require('app');
+
+/**
+ * This directive is used to display a contextual help modal dialog.
+ * @param {ui.bootstrap.$modal} $uibModal modal from angular bootstrap.
+ * @param {angular.$templateCache} $templateCache service.
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.contextHelpDirective = function($uibModal, $templateCache) {
+  return {
+    restrict: 'A',
+    link: function(scope, element, attrs) {
+      element.append(' <span class="glyphicon glyphicon-info-sign context-help-sign">');
+      element.on('click', function() {
+        $uibModal.open({
+          controller: 'AppContextHelpModalController',
+          controllerAs: 'contextHelpModalCtrl',
+          templateUrl: '/static/partials/contexthelpmodal.html',
+          'windowClass': 'context-help-modal',
+          resolve: {
+            content: function() {
+              var templateUrl = attrs['contextHelpContentUrl'];
+              if (templateUrl) {
+                return app.utils.getTemplate(templateUrl, $templateCache);
+              } else {
+                return attrs['contextHelpContent'];
+              }
+            },
+            title: function() {
+              return attrs['contextHelpTitle'];
+            }
+          }
+        });
+      });
+    }
+  };
+};
+
+
+app.module.directive('appContextHelp', app.contextHelpDirective);
+
+
+/**
+ * We have to use a secondary controller for the modal so that we can inject
+ * uibModalInstance which is not available from the first level controller.
+ * @param {ui.bootstrap.modalInstance} $uibModalInstance modal from angular bootstrap
+ * @param {string} content
+ * @param {string} title
+ * @constructor
+ * @ngInject
+ * @returns {app.ContextHelpModalController}
+ */
+app.ContextHelpModalController = function($uibModalInstance, content, title) {
+  /**
+   * @type {string}
+   * @export
+   */
+  this.content = content;
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.title = title;
+
+  /**
+   * @type {ui.bootstrap.modalInstance} $uibModalInstance angular bootstrap
+   * @private
+   */
+  this.modalInstance_ = $uibModalInstance;
+};
+
+app.ContextHelpModalController.prototype.close = function() {
+  this.modalInstance_.close();
+};
+
+
+app.module.controller('AppContextHelpModalController', app.ContextHelpModalController);
--- a/c2corg_ui/templates/outing/edit.html
+++ b/c2corg_ui/templates/outing/edit.html
@@ -90,7 +90,10 @@ updating_doc = outing_id and outing_lang
 
             ## ACTIVITIES
             <div class="form-group full" id="outing-activities" ng-init="activities = ${activities}">
-              <label><span translate>activities</span> *</label>
+              <label app-context-help
+                context-help-title="{{'activities' | translate}}"
+                context-help-content-url="/static/partials/contexthelp/activities.html"
+                ><span translate>activities</span> *</label>
               <div class="route-activities">
                 <div ng-repeat="activity in activities" class="activity">
                   <div>
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ LIBS_JS_FILES += \
     node_modules/angular-gettext/dist/angular-gettext.min.js \
     node_modules/angular-messages/angular-messages.min.js \
     node_modules/angular-cookies/angular-cookies.min.js \
+    node_modules/angular-sanitize/angular-sanitize.min.js \
     node_modules/corejs-typeahead/dist/typeahead.bundle.min.js \
     node_modules/moment/min/moment.min.js \
     node_modules/moment-timezone/builds/moment-timezone-with-data.min.js \
--- a/c2corg_ui/static/js/main.js
+++ b/c2corg_ui/static/js/main.js
@@ -26,6 +26,7 @@ goog.require('app.baselayerSelectorDirective');
 goog.require('app.blockAccountDirective');
 goog.require('app.cardDirective');
 goog.require('app.constants');
+goog.require('app.contextHelpDirective');
 goog.require('app.deleteAssociationDirective');
 goog.require('app.deleteDocumentDirective');
 goog.require('app.diffMapDirective');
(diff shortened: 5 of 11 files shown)


arnaud-morvan:
Note that for testing purpose, I've modified .build/locales/ files manually.

c2corg_ui/static/js/contexthelp.js
```diff
@@ -0,0 +1,76 @@
+goog.provide('app.contextHelpDirective');
+goog.provide('app.ContextHelpModalController');
+
+goog.require('app');
+
+/**
+ * This directive is used to display a contextual help modal dialog.
+ *
+ * @return {angular.Directive} The directive specs.
```
asaunier: I think the ``@param`` info are missing.

c2corg_ui/static/js/contexthelp.js
```diff
@@ -0,0 +1,76 @@
...
+  return {
+    restrict: 'A',
+    link: function(scope, element, attrs) {
+      element.append('<span class="glyphicon glyphicon-info-sign context-help-sign">');
+      element.on('click', function() {
+        $uibModal.open({
+          controller: 'AppContextHelpModalController',
+          controllerAs: '$ctrl',
+          templateUrl: '/static/partials/contexthelpmodal.html',
+          'windowClass': 'context-help-modal',
```
asaunier: Just curious: why do you need quotes around ``windowClass``?

c2corg_ui/static/js/contexthelp.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+app.contextHelpDirective = function($uibModal, gettextCatalog) {
+  return {
+    restrict: 'A',
+    link: function(scope, element, attrs) {
+      element.append('<span class="glyphicon glyphicon-info-sign context-help-sign">');
+      element.on('click', function() {
+        $uibModal.open({
+          controller: 'AppContextHelpModalController',
+          controllerAs: '$ctrl',
```
asaunier: Are you sure ``$ctrl`` is a relevant controller name? It looks like a variable whereas it is just a string.
At least it is not very consistent with other directives whose controller name is something like ``fooCtrl``.

c2corg_ui/templates/outing/edit.html
```diff
@@ -90,7 +90,9 @@ <h2 class="heading show-phone" translate>general informations</h2>
 
             ## ACTIVITIES
             <div class="form-group full" id="outing-activities" ng-init="activities = ${activities}">
-              <label><span translate>activities</span> *</label>
+              <label app-context-help ="context_help_activities" context-help-title="activities">
```
asaunier: * Extra blank space before the = sign.
* Can't you use a translate filter instead of adding the string to ``c2corg_ui/templates/i18n.html``? ``app-context-help="{{'context_help_activities' | translate}}"``

c2corg_ui/static/js/appmodule.js
```diff
@@ -26,5 +26,6 @@ app.module = angular.module('app', [
   'ngFileUpload',
   'slug',
   'vcRecaptcha',
-  'infinite-scroll'
+  'infinite-scroll',
+  'ngSanitize'
```
asaunier: Is this really used?

c2corg_ui/static/js/contexthelp.js
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * @type {Object} $uibModalInstance angular bootstrap
+   * @private
+   */
+  this.modalInstance_ = $uibModalInstance;
+};
+
+app.ContextHelpModalController.prototype.close = function() {
+  this.modalInstance_.close();
+}
```
asaunier: Missing semi-colon => makes the build fail

c2corg_ui/static/js/contexthelp.js
arnaud-morvan: I do not really know why. I've searched in ui.bootstrap externs, ``open`` function is declared as:
```
/**
 * @param {!Object} modalOptions
 * @return {!ui.bootstrap.modalInstance}
 */
ui.bootstrap.$modal.open = function(modalOptions) {};
```
Note that I am not really conformtable with closure.

c2corg_ui/static/js/appmodule.js
arnaud-morvan: This is need to insert HTML using ng-bind-html. I have a "no safe content ..." error without ngSanitize.
https://docs.angularjs.org/api/ng/directive/ngBindHtml

c2corg_ui/templates/outing/edit.html
arnaud-morvan: Thanks, I was not aware that this form was catched.

arnaud-morvan:
Thanks @asaunier for review, I've update my PR with you comments.
There is still the ``'windowClass': 'context-help-modal',`` but I really do not understand, not why I have to give this property as string, but why the others are not renamed by closure.

arnaud-morvan:
Also note that translators will have to enter the full HTML help corresponding to the context help key in transifex.

asaunier:
In my instance the contextual button is a "i" not a "?" :P Nevermind, both are explicit enough to me.
Perhaps the button is a bit too small to be handy on a mobile touch screen?

@stef74 @desnoes When in the demo, this (only) example will be available near the activities section of the outing editing form.

> There is still the 'windowClass': 'context-help-modal', but I really do not understand, not why I have to give this property as string

Perhaps it is not available in some externs file as you suspect it. It's OK for me if you leave it as is.

> Also note that translators will have to enter the full HTML help corresponding to the context help key in transifex.

Mmmm I guess this will be a problem. Actually using a translation key as ``context_help_activities`` will be confusing for translators. In Transifex they will only see ``context_help_activities`` and won't know what to translate it to. Obviously they won't know they can use HTML code (and most likely most translators don't even know how to write HTML).

arnaud-morvan:
> In my instance the contextual button is a "i" not a "?" :P Nevermind, both are explicit enough to me.

I've hesitated between ``?`` and ``i``, the important thing is to use the same everywhere. Here it is set in the directive, it is easy to change it globally.

> Perhaps the button is a bit too small to be handy on a mobile touch screen?

Note that click event is bind on the label element, so clicking (or touching) on the label text also works.

> Mmmm I guess this will be a problem. Actually using a translation key as context_help_activities will be confusing for translators. In Transifex they will only see context_help_activities and won't know what to translate it to. Obviously they won't know they can use HTML code (and most likely most translators don't even know how to write HTML).

It is also possible to use partials/templates for the help texts, with ``translate`` on all elements. for example using a ``context_help_activities.html`` file with :
```
<p translate>Type d'activités : dans une liste ou un document, passez la souris sur les pictogrammes pour connaître leur signification.</p>
<p translate>Un document peut être associé à une ou plusieurs activités :</p>
<ul>
  <li translate>ski, surf : randonnée à ski et/ou surf</li>
  <li translate>raquette : randonnée en raquettes</li>
...
```
But I do not want to add the full text in the edit template, to avoid passing it to the browser in each edit request.

arnaud-morvan:
Note the example here is in French and should be translated in English

asaunier:
Thanks for updating the doc in the initial comment above with the 2 possibilities (inline short contextual help messages or partials for longest ones)!

> Note the example here is in French and should be translated in English

I have opened a dedicated issue https://github.com/c2corg/v6_ui/issues/1605
@stef74 @desnoes That would be better the FR strings are quickly converted to EN strings, because they will appear in Transifex as the EN source strings (whereas they are in french). If translations are done in Transifex about those strings, they will be lost when the expected EN strings will be used instead.

desnoes:
@arnaud-morvan @asaunier : can you tell me the file you have modified for the example (activities in an outing form)? I would like to compare the procedure describe above with a practical example. I need to explain the procedure to Sylvain.

arnaud-morvan:
@desnoes : If I correctly understand you are searching for the changes applied to outing/edit template : https://github.com/c2corg/v6_ui/pull/1601/files#diff-abecb3d7ee99e295aeab6e796b94f92a

Note that here I use `context-help-content-url` attribute, but for shorter texts, it possible to put the text inline with `context-help-content` attribute.

asaunier:
> can you tell me the file you have modified for the example (activities in an outing form)?

@desnoes I guess you mean this one https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/contexthelp/activities.html

desnoes:
@asaunier @arnaud-morvan : i would like to see based on a practical example, which files i will have to modify and how. The 'how' hase been described above by Arnaud, i simply would like to have the list of files that have been modified in order to insert the contextual help. Did you only modify this file:  https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/outing/edit.html ?

asaunier:
> Did you only modify this file: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/outing/edit.html ?

Yes. Here is the change: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/outing/edit.html#L93-L95
See also the documentation in the first comment of this issue.
Please note that ``context-help-content-url="/static/partials/contexthelp/activities.html"`` points to  https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/contexthelp/activities.html

This means that someone that wants to add some contextual help to such a field must:
* edit the HTML template (some editing form for instance) and add the attributes above
* in case a simple help string is enough, the content attribute to use is like ``context-help-content="{{'Some text' | translate}}"``
* in case a longer formatted content is required, one should create a "partial" (some HTML file in https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials/contexthelp) and reference it like ``context-help-content-url="/static/partials/contexthelp/activities.html"``.
* translations must be done in Transifex as usual

desnoes:
@asaunier : ok, thanks for the explanation. I will have look this week-end.

Issue #1610, opened 2017-05-02
Contextual help partials are not translated


asaunier:
See https://github.com/c2corg/v6_ui/issues/1096#issuecomment-297995825
> I think there is a problem left with PR #1601: it's OK with simple strings passed as
context-help-content="{{'Some text' | translate}}"
but it seems the translation does not work for complex strings passed as a partial (HTML file)
context-help-content-url="/static/partials/contexthelp/activities.html".
![capture du 2017-05-04 10-07-40](https://cloud.githubusercontent.com/assets/1192331/25694990/9355fa2e-30b1-11e7-879f-2902ab4157e9.png)

PR #1617, opened 2017-05-04
Compile contexthelp partials for translations

arnaud-morvan:
Relate #1610


asaunier:
I had similar changes in mind when I had a look last week but my approach was not the good one :)
Yours makes sense except that it causes a build error :P

PR #1620, opened 2017-05-04
[WIP] add local storage

stef74:


asaunier:
I think it's better to use the approach of PR https://github.com/c2corg/v6_ui/pull/1621

PR #1621, opened 2017-05-04
Add listSwitch directive

asaunier:
@stef74 I don't think your current approach (adding list switching buttons by hand in all document types list pages) is the good one: it is better to use an AngularJS directive (an HTML template with a controller where you can define actions) once and to reuse it in all appropriate templates.

This is what this PR does.
I'm also working on a way to save the user's choice in the local/sessionStorage as/in the ``userData`` (object passing some user attribute to all pages) from the directive controller, but the code is not ready yet.

### DIFF
--- a/c2corg_ui/static/js/listswitch.js
+++ b/c2corg_ui/static/js/listswitch.js
@@ -1,0 +1,42 @@
+goog.provide('app.ListSwitchController');
+goog.provide('app.listSwitchDirective');
+
+goog.require('app');
+
+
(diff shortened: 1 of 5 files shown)


stef74:
thank @asaunier I'm waiting to merge your PR, After I rebase my 2 PR.

asaunier:
@stef74 OK thanks. I merge the PR as is and will open a dedicated PR for storing the user choice.

PR #1615, opened 2017-05-03
Add list in document

stef74:
#1614

PR #1628, opened 2017-05-06
Fix indentations

asaunier:


stef74:
Please not merge ... because that create conflict #1615 , I prefer to fix in this one. (I'm away form 2 days)

Issue #1618, opened 2017-05-04
Store list/card preference in localstorage


stef74:
List are now available but need manual switch.

A simple way is to store the value in cookie and reuse it. 

The defaut value of list or card is here : https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/outing/index.html#L37

- [ ] update ng-init="rowList = xxxx"  where xxx is the value of cookies

- [ ] update the cookies when you change list/card

I don't know how to manipulate a cookies in angular. So I need some help.

stef74:
use https://docs.angularjs.org/api/ngCookies/service/$cookies ? ngcookies seem to be allready use. 

so something like this, can be work :  

```
<button class="btn orange-btn btn-default no-list-btn hidden-xs card-list" ng-init="rowList = **$cookies.get('list'**);" 
		ng-click="rowList = !rowList **;   $cookies.put('list', rowList)** " >
	  <span class="glyphicon glyphicon-globe"></span>
	  <span translate ng-show="!rowList">List Format</span>
	  <span translate ng-show="rowList">Card Format</span>
	</button>
```

Two question : 
- I'm not very sure that my ng-click can work ... 
- in example of documentation they have : ".controller('ExampleController', ['$cookies', function($cookies)" i don't know if i need to do something ...

asaunier:
I am not sure cookies are the right way to store that kind of information.
Perhaps it could be saved to the ``userData`` (along with id, username, name, role) that stores info in the localStorage/sessionStorage:
https://github.com/c2corg/v6_ui/blob/7e1c83f764f2766c61f443c7074f9c6c6be3e306/c2corg_ui/static/js/auth/authservice.js#L204-L228

> so something like this, can be work :

Putting too much JS code in the HTML attributes is not a good practise (as putting CSS styles directly in the tags). It's better to write or extend angular directives https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/js

PR #1629, opened 2017-05-06
Save list switch option

asaunier:
@stef74 This PR makes the list/card switch pref persistant.

Related to https://github.com/c2corg/v6_ui/issues/1618


stef74:
@asaunier not tested but the approach seem good.

Issue #1458, opened 2017-01-31
Use release tag instead of commit hash as UI version id


asaunier:
See https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/sidemenu.html#L70-L74

asaunier:
Could be achieved by modifying (or extending) https://github.com/c2corg/v6_ui/blob/master/config/default#L26
using 
```
git describe --tags `git rev-parse HEAD`
```
instead of 
```
git rev-parse --short HEAD
```

For instance:
```
$ git describe --tags `git rev-parse HEAD`
v6.1.5
```

That might impact the health service https://github.com/c2corg/v6_ui/blob/bcb6106395398029498bc77cad98349d98b748b7/c2corg_ui/views/health.py#L29 so perhaps it is safer to add a ``tag`` data next to the ``version`` data.

Template https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/sidemenu.html#L71-L72 should be updated as well as some view file I guess.

asaunier:
It seems it does not work as expected in the demo:
<img width="245" alt="capture d ecran 2017-05-14 a 10 16 53" src="https://cloud.githubusercontent.com/assets/1192331/26032226/7da04772-388e-11e7-9c8c-ecdeeb06d099.png">

desnoes:
I cannot observe the string 'version' in the home page. Where should it be visible ?

![bug release tag](https://cloud.githubusercontent.com/assets/19892089/26035576/22cd2108-38cf-11e7-9dfb-0e2d2ffb27a5.png)

asaunier:
Your screen is too small. The version info is not shown when the screen height is not enough. See my screenshot above (but the version number seems to be not correct in the demo). See also https://github.com/c2corg/v6_ui/pull/1630#issuecomment-300601307

asaunier:
I prefer reopening the issue because the version info is not correct in demo (probably not in prod either). See https://github.com/c2corg/v6_ui/issues/1458#issuecomment-301297791

asaunier:
In prod ``version: 0`` is shown as well, instead of the actual release number.

The ``common.ini`` file in the docker containers is indeed ``release = 0``:
https://github.com/c2corg/v6_ui/blob/master/common.ini.in#L40

It seems
```
https://github.com/c2corg/v6_ui/blob/master/config/default#L27
```
[...]

PR #1630, opened 2017-05-06
Show release tag instead of commit hash

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1458


asaunier:
A screenshot example
capture d ecran 2017-05-10 a 22 19 48

See the version number in the footer. For an actual release, only the release tag (ie "v6.1.7") is shown.

stef74:
ok for me

asaunier:
Thanks for reviewing @stef74

PR #1624, opened 2017-05-05
Homedesign

Courgetteandratatouille:
I guess it's ready for a test and maybe we can improve the performance of the column controller.

PR #1632, opened 2017-05-07
Homedesign

Courgetteandratatouille:
In the next commit i work on user profile page, it's need a little redesign too after the modification of the feed.


asaunier:
Does this PR replace https://github.com/c2corg/v6_ui/pull/1624?

Courgetteandratatouille:
Yes, with a little update , i'am going to close the old PR

c2corg_ui/static/js/feed.js
```diff
@@ -140,27 +228,134 @@ app.FeedController.prototype.getDocumentsFromFeed = function() {
...
+  }
+};
+
+/**
+ * @param {Object} doc
+ * @return {string | undefined}
+ * @export
+ */
+app.FeedController.prototype.createURL = function(doc) {
+
```
asaunier: The current feed use the card directive defined at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js which indeed has a ``createURL`` function https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js#L163

c2corg_ui/static/js/feed.js
```diff
@@ -211,6 +408,157 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
...
+    }
+
+  }
+  return null;
+};
+
+
+/**
+ * Show title with the good language:
+ * 1) range 2) admin limits 3) country
```
asaunier: ?

c2corg_ui/static/js/feed.js
```diff
@@ -211,6 +408,157 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
...
+ * @param {?Array<Object>} locales
+ * @return {string | null}
+ * @export
+ */
+app.FeedController.prototype.showTitle = function(locales) {
+
```
asaunier: ...

c2corg_ui/static/js/feed.js
```diff
@@ -211,6 +408,157 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
...
+  for (var i = 0; i < locales.length; i++) {
+    if (locales.lang == this.lang_) {
+      if (!locales[i].title_prefix) {
+        return locales[i].title;
+      } else {
+        return locales[i].title_prefix  + ' : ' + locales[i].title;
+      }
```
asaunier: OK but you could save a line using the following syntax:
```
if (...) {
  return ...;
}
return ...;
```

c2corg_ui/static/js/feed.js
```diff
@@ -211,6 +408,157 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
...
+
+
+/**
+ * Show title with the good language:
+ * 1) range 2) admin limits 3) country
+ * @param {?Array<Object>} locales
+ * @return {string | null}
+ * @export
+ */
+app.FeedController.prototype.showTitle = function(locales) {
```
asaunier: Why do you need this? Couldn't you use the existing way to show titles in the feed cards?

c2corg_ui/static/js/feed.js
```diff
@@ -211,6 +408,157 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
...
+    return locales[0].title;
+  } else {
+    return locales[0].title_prefix  + ' : ' + locales[0].title;
+  }
+};
+
+
+/**
+ * Show summaru with the good language:
+ * 1) range 2) admin limits 3) country
```
asaunier: ?

c2corg_ui/static/js/feed.js
```diff
@@ -211,6 +408,157 @@ app.FeedController.prototype.createImageUrl = function(filename, suffix) {
...
+ * @param {?Array<Object>} locales
+ * @return {string | null}
+ * @export
+ */
+app.FeedController.prototype.showSummary = function(locales) {
+
+  for (var i = 0; i < locales.length; i++) {
+    if (locales.lang == this.lang_) {
```
asaunier: Are you sure this works?
Perhaps you mean ``if (locales[i].lang == this.lang_)``?

Anyway I am a bit surprised you have to do this. Is there that kind of code in the cards showing localized contents?

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
-          ng-if="::doc.image3">
-          <span class="image" style="background-image: url({{feedCtrl.createImageUrl(doc.image3.filename, 'MI')}})"></span>
+  <div class="feed-list" ng-show="feedCtrl.documents.length > 0"
+       infinite-scroll="feedCtrl.getDocumentsFromFeed()"
+       infinite-scroll-immediate-check="false"
+       infinite-scroll-distance="0.7"
+       infinite-scroll-disabled="feedCtrl.busy || feedCtrl.error || feedCtrl.feedEnd || feedCtrl.noFeed">
+
+
+
```
asaunier: Extra blank lines

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+                {{::feedCtrl.showArea(doc['document']['areas'])}}
+              </div>
+              <div class="grow">
+              </div>
+            </div></a>
+        </div>
+        <div class="parent-infop">
+          <div class="infop-bar">
+
+
```
asaunier: Extra blank line

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+                <div class="bubble">
+                  <span class="route-activity icon-{{::activity}}"></span>
+                </div>
+              </a>
+            </div>
+
+            <div class="grow">
+            </div>
+
+
```
asaunier: Extra blank lines

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+            <div class="parent-bubble">
+              <a href="#" class="link share-facebook" ng-if="::doc['document']['condition_rating']" uib-tooltip="{{'condition_rating' | translate}} : {{doc['document']['condition_rating'] | translate}}" tooltip-append-to-body="true">
```
asaunier: Just curious: what's the use of the ``share-facebook`` class? Why do you need to use a ``<a>`` here and for other data around?

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+    <div style="clear:both">&nbsp;</div>
+
+    <h4 class="text-center text-danger" translate ng-show="feedCtrl.error">Sorry, there was an error while getting the feed.</h4>
+    <h4 class="text-success text-center" translate ng-show="feedCtrl.busy">Loading feed, please wait...</h4>
+    <h4 class="text-success text-center" translate ng-show="feedCtrl.feedEnd">No more feed to load.</h4>
+  </div>
+</div>
+<div class="feed-cr">
+  <div class="info">
+    <h3>Échos du forum</h3>
```
asaunier: In english please, with a ``translate`` directive

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+    <h4 class="text-danger"translate ng-show="feedCtrl.errorForum">Sorry, there was an error while getting the forum data.</h4>
```
asaunier: Missing blank space before ``translate``

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+    <div class="fpost" ng-repeat="post in feedCtrl.forums track by post.id" ng-if="::feedCtrl.forums.length > 0">
+      <a href="https://forum.camptocamp.org/t/{{::post['slug']}}/{{::post['id']}}" title="{{::post['title']}}">
```
asaunier: Can't you use the ``discourseUrl`` instead of hardcoding the prod forum base URL?

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+              <a href="#" class="link share-facebook" uib-tooltip="{{doc['document']['quality'] | translate}}" tooltip-append-to-body="true">
+
+                <span class="quality-{{::doc['document']['quality']}}"></span>
+              </a></div>
+            </div>
+          </div>
+        </div>
+      </article>
+    </div> 
+    <div style="clear:both">&nbsp;</div>
```
asaunier: Inline styling is not recommended

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+
+            <div class="parent-bubble">
+              <a href="#" class="link share-facebook" ng-if="::doc['document']['condition_rating']" uib-tooltip="{{'condition_rating' | translate}} : {{doc['document']['condition_rating'] | translate}}" tooltip-append-to-body="true">
+                <div class="condition-rating" ng-class="::doc['document']['condition_rating']" ></div>
+              </a>
+            </div>
+
+            <div class="parent-bubble"><div class="bubble bquality">
+              <a href="#" class="link share-facebook" uib-tooltip="{{doc['document']['quality'] | translate}}" tooltip-append-to-body="true">
+
```
asaunier: extra blank line

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
```
asaunier: extra blank line

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+                <span class="quality-{{::doc['document']['quality']}}"></span>
+              </a></div>
```
asaunier: This would be a little more readable to close tags consistently with the opening tags.
```
<div class="parent-bubble"><div class="bubble bquality">
  <a href="#" class="link share-facebook" >
  </a>
</div></div>
```
instead of 
```
<div class="parent-bubble"><div class="bubble bquality">
  <a href="#" class="link share-facebook" >
  </a></div>
</div>
```
Or to have tags on dedicated lines.

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+                {{::feedCtrl.showArea(doc['document']['areas'])}}
+              </div>
+              <div class="grow">
+              </div>
+            </div></a>
+        </div>
+        <div class="parent-infop">
+          <div class="infop-bar">
+
+
```
asaunier: blank line

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+                <div class="bubble">
+                  <span class="route-activity icon-{{::activity}}"></span>
+                </div>
+              </a>
+            </div>
+
+            <div class="grow">
+            </div>
+
+
```
asaunier: blank line

c2corg_ui/static/partials/feed.html
```diff
@@ -1,76 +1,269 @@
...
+            <div class="fportrait">
+              <img ng-src="https://forum.camptocamp.org/{{::post['avatar_template']}}">
+            </div>
+            <div class="ftxt">
+              <div class="fauthor">{{::post['last_poster_username']}}</div>
+              <div class="ftitle"><span class="title">{{::post['title']}}</span> - {{::post['last_posted_at'] | amUtc | amLocal | amTimeAgo}}</div>
+            </div>
+          </div>
+
+        </div></a>
```
asaunier: Blank line + unconsistent tags grouping

less/c2corg_ui.less
```diff
@@ -508,8 +508,22 @@ app-simple-search {
   border-radius: 6px;
   padding-left: 10px;
   margin-right: 10px;
-  background: @C2C-orange;
+  background: rgba(255,255,255,0.25);
 
+  .form-control {
+        box-shadow: none;
+    }
```
asaunier: All those lines are not correctly indented (indentation = 2 cols + selectors of the same level should be aligned)

less/c2corg_ui.less
```diff
@@ -539,9 +553,18 @@ app-simple-search {
...
   }
 
-  label {
-    margin-bottom: 0;
-  }
+   .btn {
+        border: none;
+    }
+
+    .search {
```
asaunier: Indentation problems around here

less/c2corg_ui.less
```diff
@@ -579,16 +602,23 @@ app-simple-search {
     padding: 5px;
     &:focus, &:hover {
       width: 420px;
+      /*
```
asaunier: What about simply removing those styles instead of commenting them out?

less/c2corg_ui.less
```diff
@@ -579,16 +602,23 @@ app-simple-search {
...
+      */
+      box-shadow: none;
+      outline: none;
+      border: none;
       @media @tablet {
         width: 150px;
       }
       @media @phone {
-        .calc(width, "100vw - 45px");
+                background: rgba(255,255,255,0.25) !important;
```
asaunier: Indentations :P Make sure your editor convert tabs to blank spaces

less/feed.less
```diff
@@ -1,5 +1,6 @@
 @import 'variables.less';
 
+/*
```
asaunier: Styles to be removed?

less/newfeed.less
```diff
@@ -0,0 +1,694 @@
+#feed {
```
asaunier: Indentations of this file are incorrect (2 blank spaces)

c2corg_ui/static/js/feed.js
```diff
@@ -120,6 +160,52 @@ app.FeedController = function(appAuthentication, appApi, appLang, imageUrl, ngeo
...
+   */
+
+};
+
+
+/**
+* refresh the feed column according the width
+* @export
+*/
+app.FeedController.prototype.feedColumnController = function() {
```
Courgetteandratatouille: Ok for the rename, but this function can be have a lot of change in the future, (maybe complety remove with a css solution), i guess it's more readable to separe this code from the rest because the goal it's not to get the doc, only manage the look.

asaunier:
@Courgetteandratatouille A HUGE THANKS for your HUGE WORK :)
This looks like an excellent job. I have made many comments though.

The home page (tested in my dev instance) is quite fine in my browsers (FF and Chrome), whatever the screen size.

The orange header is nice. I just wonder if it does not make a bit too much orange in the pages? Perhaps we could use another color for the page sections headers? Anyway this could be done in some other PR.

The forum feed fails displaying in my instance. Perhaps because some CORS settings are missing in the demo forum? Have you tried with the prod forum? Do you know if there are still a problem with https://github.com/c2corg/v6_forum/issues/93 ?

I have also noticed that a call to the forum is also done in the profile feed whereas it should not.
I think you should add a ``if (this.userId)`` (set only when in a profile feed) test on 
https://github.com/c2corg/v6_ui/blob/343e0bfe862f21592b36ea0c4e5d053f036207b7/c2corg_ui/static/js/feed.js#L163

You have removed the "Add an outing" and "See outings" button at the beginning of the homepage feed. Is it on purpose?

Last but not least I have noticed that you have duplicated a lot of code from the card directive 
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js
and of the card partials https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials/cards

To be honest I feel quite concerned about those duplicates:
* duplicating code always mean there is something wrong
* wasn't it possible to keep using the card directive as it is the case in the current feed?

Courgetteandratatouille:
Thanks you @asaunier :) , it's not perfect, but i try to do better the next time.
I made the little correction on the spaaaace x) and other thing in the next commit.

`
The orange header is nice. I just wonder if it does not make a bit too much orange in the pages? Perhaps we could use another color for the page sections headers? Anyway this could be done in some other PR.`

You're right, i planned to work on the other page for increase the readability.

`The forum feed fails displaying in my instance. Perhaps because some CORS settings are missing in the demo forum? Have you tried with the prod forum? Do you know if there are still a problem with c2corg/v6_forum#93 ?`

Yes the problem is the cors, it's the same with the prod, but maybe the prod is working with www.camptocamp.org. Actually the forum is down, so i cannot get the topics where we talk about that. 

```
I have also noticed that a call to the forum is also done in the profile feed whereas it should not.
I think you should add a if (this.userId) (set only when in a profile feed) test on
https://github.com/c2corg/v6_ui/blob/343e0bfe862f21592b36ea0c4e5d053f036207b7/c2corg_ui/static/js/feed.js#L163
```
Indeed, i work it out later !

`You have removed the "Add an outing" and "See outings" button at the beginning of the homepage feed. Is it on purpose?`

Yes, the page have already button to do this and we need some space for the new feed, so i delete it.

```
Last but not least I have noticed that you have duplicated a lot of code from the card directive
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js
and of the card partials https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials/cards
```
I reply a little later.

c2corg_ui/static/js/feed.js
```diff
@@ -140,27 +228,134 @@ app.FeedController.prototype.getDocumentsFromFeed = function() {
...
   var token = response['data']['pagination_token'];
   this.nextToken = token;
-  for (var i = 0; i < data.length; i++) {
-    this.documents.push(data[i]);
+  if (window.innerWidth >= 1360) {
+    this.nbcolumn = 2;
+
+    for (var i = 0; i < data.length / 2; i++) {
+      this.documentsL.push(data[i]);
+      this.documents.push(data[i]);
```
Courgetteandratatouille: Not easily, i keep it in mind for later

c2corg_ui/static/partials/feed.html
Courgetteandratatouille: it's a forget :)

Courgetteandratatouille:
@asaunier 

About the copy/paste from cards.js, i'am not very familiar with angular1. I dont know to access properly to the cards method, so i had prefer put the code into feed.js.

Second point, actually the cards element is use into feed and in the pages with list, but with this update, we are not in this case, so i had  prefer separe the two thing and i guess  later with an harmonisation, we can reuse the code into cards.js.

asaunier:
> Yes, the page have already button to do this and we need some space for the new feed, so i delete it.

That's true there is already an "add an outing" button in the header (but I often wonder if it is really useful? perhaps the button in the homepage is enough?).
The "see outings according to my preferences" has no equivalent and it's a nice shortcut to opening the outings page + pressing the "apply my prefs" button, isn't it?
I was wondering if a third button "What's new?" opening https://www.camptocamp.org/whatsnew would not make sense as well? There is no link at all pointing to that page whereas it is useful figuring out the editing activity. Placing those 3 buttons in a card at the beginning of the feed does not take that much place.

> actually the cards element is use into feed and in the pages with list, but with this update, we are not in this case, so i had prefer separe the two thing and i guess later with an harmonisation, we can reuse the code into cards.js.

I don't understand what you mean. With the current feed we already used the card directive to build the feed item cards, why wouldn't it be possible to do so with this new version of the feed?
You could define dedicated partials for those new card types:
https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials/cards
See also the example of the advanced search directive that itself uses card directives to display cards: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/advancedsearch.html#L4

I really do think it's a problem reinventing the wheel, duplicating code and behaviours from the card directive.
If you need help about this part, I can try to have a look and suggest changes to your PR.

c2corg_ui/static/js/feed.js
```diff
@@ -67,9 +68,9 @@ app.FeedController = function(appAuthentication, appApi, appLang, appUrl,  image
 
   /**
    * @type {number}
-   * @export
+   * @private
    */
-  this.nbcolumn = 0;
+  this.nbCols = 0;
```
asaunier: Private properties must have a trailing ``_`` (so that it is obvious they are private). Eg. ``this.nbCols_``

c2corg_ui/static/js/feed.js
```diff
@@ -228,11 +218,10 @@ app.FeedController.prototype.getDocumentsFromFeed = function() {
...
 
 /**
- * get lastest topics
- * @export
+ * get latest topics
+ * @private
  */
-app.FeedController.prototype.getDocumentsFromForum = function() {
-
+app.FeedController.prototype.getTopics = function() {
```
asaunier: Private method must also have a trailing ``_``: ``getTopics_``.

stef74:
Concerning the card, agree with @asaunier ... probably better thing to reuse the existing html. I don't identifying a blocking point to have your design with this html. Flex offer many posibility. List layout confirm that. 
If you want to implement a multi colum with none alignement card (https://material.angularjs.org/latest/demo/card). But we need to manage the number of column dynamicly.
A good idea to really create a none alignement is to have an action small card on first (as in G+ ... you have search feature). 
I prefer to have 3 image than one ...

Courgetteandratatouille:
@stef74 
@asaunier talk about the javascript and not the html, css flex cannot helps us in this case, the vertical align is bad. Your example work like my code.  But if you have a solution to align verticaly the card in column  with css, i'am open to change :).

We can have 3 pictures in the card, no problem with that, but before doing a lot a change, it's better to have a test version in the demo to avoid useless work if c2c's members doesnt accept the new design.

stef74:
Ok the number of image, you agree start simple.

Concerning you problem of vertical align is not clear for me. Coudl you provide a screenshoot (I have not your branch available with me).

Courgetteandratatouille:
Look the bad vertical align with flex.  Or,  the size oh cards is not ok and the card take all the blank space. That why you need column.

flexalign

asaunier:
I think this PR is too big, doing too many things at the same time. I would recommend to split some parts (ie. make dedicated PRs) that could then be pushed to production without waiting that every thing is ready.

For instance the following parts could be independent:
* header layout
* feed cards content and styling
* multi-column layout, including the forum column

Concerning the feed cards content and styling, I think the card directive should be used instead of duplicating some of its code. I will think at the changes it requires.

Concerning the multi-column layout:
* the approach used by @Courgetteandratatouille (dispatching cards alternatively to the left and right columns) is quite simple but I see several flaws. For instance depending on the height of every single card, both columns do not grow at the same speed. In addition the chronology of the feed is not respected. Finally relying on 2 distinct pieces of code for managing 1 or 2 columns (depending on the window width) is a bit cumbersome.
Perhaps we could use a CSS (or JS?) technic that would automatically display the feed as 1 or 2 (or more?) columns depending on the available width. It's worth investigatimg.

Concerning the forum block: please note that as for now it's not shown at all in a mobile screen despite the request to Discourse API is triggered anyway.

stef74:
If you have look a G+ code concerning multi colomun layout ... it managed column by colonum.
If you want to distribute left-right column , probably no other way to manage it column in JS.

@brunobesson other idea ?

stef74:
agree with Alex to split in multiple PR. Theay can probaly push and merge some element more quickly than other.

stef74:
have look on https://masonry.desandro.com/

Courgetteandratatouille:
I know mansory but some people complainting about the speed of loading of the website on mobile device, we need to keep a lightweight approch in my opinion.

Courgetteandratatouille:
close for spliting in other PR.

Issue #1613, opened 2017-05-03
Modification of collaborative article


desnoes:
It is possible to select a blank line when we try to modify the article type. This gives an error message, but it is not clear. For a collaborative article, this option should be avoided.

![1150f92c02c56f9b6edea256f2f28f8a8e77aa50](https://cloud.githubusercontent.com/assets/19892089/25649118/076ad9fa-2fd3-11e7-919a-adf69b7fd5e5.png)

PR #1633, opened 2017-05-07
Fix article type field when editing an article

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1613

Use a similar approach to the image type selector in image editing form:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/image/edit.html#L152-L156
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/editing/imageediting.js#L73

What it does:
* create an editing controller extension for article in order to store the initial article type (at form loading time)
* remove empty options for article type when editing an article (it is no longer "possible" to set the type to no value)
* when the initial type was "collaborative", display a fixed (disabled) text input with the translated collaborative type
* when the initial type was not "collaborative" (ie. "personal"), both collaborative and personal types are available until the form is submitted.
* moderators can change to whatever type at any time

### DIFF
--- a/c2corg_ui/static/js/editing/articleediting.js
+++ b/c2corg_ui/static/js/editing/articleediting.js
@@ -1,0 +1,56 @@
+goog.provide('app.ArticleEditingController');
+
+goog.require('app');
+goog.require('app.DocumentEditingController');
+goog.require('app.Document');
+
+
+/**
+ * @param {!angular.Scope} $scope Scope.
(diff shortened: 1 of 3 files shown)


asaunier:
The moderator case has actually not been modified in this PR, only the standard user case :P 
Since you have no objection about the code, I merge the PR and we will have the asso test if the behaviour is as expected.
Thanks!

Issue #1638, opened 2017-05-10
add support of list on tablet


stef74:

desnoes:
@stef74 : i let you test

PR #1639, opened 2017-05-10
add list for tablet

stef74:
#1638

### DIFF
--- a/less/lists.less
+++ b/less/lists.less
@@ -610,6 +610,9 @@
   
 .card-list {
   display: none ;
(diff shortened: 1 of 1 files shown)


asaunier:
Tested in my instance: OK in Chrome, in FF it's quite OK as well except that some data are overlapping on each other:

capture d ecran 2017-05-10 a 21 53 31

stef74:
Open a specific issue for that I have look for netx release. But not enough space difficult to have good rendering

PR #1637, opened 2017-05-10
fix bad rendering in mobile

stef74:
#1623

PR #1647, opened 2017-05-11
Fix home topoguide rendering

stef74:
#1637


stef74:
@asaunier ready for review (test is ok)

Issue #523, opened 2016-07-23
Add filter in doc list


loicperrin:
- [ ] In the page of a WP : associated routes should be displayed:
  1) grouped by activity
  2) in alphabetic order for a same activity
- [ ] In the page of a route: associated WP should be diplayed:
  1) grouped by WP type (first summits, second: parkings, third: huts,...)
  2) in alphabetic order for a same WP type

asaunier:
> In the page of a WP : associated routes should be displayed:
1) grouped by activity

How should card be grouped when routes have several activities?

fjacon:
Proposition : 
- order by name
- Always keep a link to advanced search (even if less than 30 results
@tsauerwein any opinion ?

loicperrin:
In the page of a WP : associated routes should be displayed (as in V5):
1) grouped by activity, with a sub-title for each activity containing routes
2) in alphabetic order (route name) for a same activity
3) when routes have several activities, the cards are duplicated in each activity

asaunier:
For me this solution is not satisfaying because duplicating multiactivity-route cards will take even more space (scroooooolling) and will confuse people (as already the case in v5 - remember Ro's recent email...) if they don't realize that's the same routes.

tsauerwein:
We could display all routes of a waypoint in one group (in alphabetic order) and provide a filter on the activities at the top of the group. Something like this:

![c2c-activities-filter](https://cloud.githubusercontent.com/assets/266474/17204751/9164f76a-54a7-11e6-9e6f-5a198f19271d.png)

stef74:
@tsauerwein great idea !

fjacon:
Indeed, it is also additional work : about 1 day. 
@stef74  Is that OK for you ?

loicperrin:
> @tsauerwein great idea ! 

+1

stef74:
Ok we validated the solution of @tsauerwein . For filtering use the icon instead a text. @fjacon please send the extention to hervé.

asaunier:
Solutions discussed with @tsauerwein to implement this filtering:

> I guess we could add an "ng-show" condition to the app-card element, a bit like this:
```
<div class="list-item" ng-repeat="doc in ${associations}">
  <app-card app-card-doc="doc" ng-show="doc.activities.indexOf(filteredActivity) !== -1"></app-card>
</div>
```

> You can also use a filter with "ng-repeat":
[...]

PR #1651, opened 2017-05-12
Add an activity filter in associated routes list

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/523

The PR adds a set of activity buttons (depending on the activities of the associated routes) shown above the list of associated routes in a WP page. When a button is clicked a filter on the activity is applied to the list, showing only the routes with the given activity.

See https://github.com/c2corg/v6_ui/issues/523#issuecomment-237206666
https://docs.angularjs.org/api/ng/filter/filter

I have still a problem with this PR: the ``this`` context of the controller's filter function is set to ``Window`` instead of the controller instance itself, thus the selected activity property is not accessible in https://github.com/c2corg/v6_ui/blob/06e7954c6ecfaf21e58bbe745cb336ea745a9143/c2corg_ui/static/js/activityfilter.js#L64

Binding a context as in ``ng-repeat="doc in ${associations} | filter: afilterCtrl.filter.bind(afilterCtrl)"`` causes an angular error ("bind is disallowed here").

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -318,6 +318,11 @@
     routes = document['associations']['all_routes']['documents'] if type == 'all_routes' else document['associations']['routes']
     associations = 'detailsCtrl.documentService.document.associations.'
     associations += 'all_routes.documents' if type == 'all_routes' else 'routes'
+    activities = []
+    for route in routes:
+        for activity in route['activities']:
+            if activity not in activities:
+                activities.append(activity)
   %>
     ${add_seo_association_links(routes, 'routes')}
     <section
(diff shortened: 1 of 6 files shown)


asaunier:
> I have still a problem with this PR

I have used another approach with fa1c0cb by directly filtering the source array of the ``ng-repeat`` instead of applying a ``filter`` on the items. The route cards are actually refreshed when switching from an activity button to another.

See also http://stackoverflow.com/a/35559482 or

arnaud-morvan:
@asaunier : Note that I have already done a prototype for this feature here : https://github.com/c2corg/v6_ui/tree/amorvan_associations_switches

asaunier:
> Note that I have already done a prototype for this feature here

@arnaud-morvan Gloops! Please always make a PR or at least mention your branch in the issue so that people know there is work in progress.

I will have a look at your branch, but isn't your code more complicated than the current PR?

arnaud-morvan:
Yes, mine is more complex, because it is possible to activate more that one activity at a time.

asaunier:

Main features:
* when the current document as more than 3 associated routes, with more than 1 activity, an activity filter is displayed:
capture du 2017-05-30 18-12-47
* activity icons are used the same way than in outing/route editing forms or in the preferences page, selected activities are made orange
* it is possible to select several activities. To unselect an activity click on it again
* the filter shows only activity for which routes are actually associated

c2corg_ui/templates/helpers/view.html
```diff
@@ -329,37 +334,40 @@ <h3 class="heading show-phone" ng-click="mainCtrl.animateHeaderIcon($event)" dat
...
-          % if type == 'routes':
-            <app-delete-association child-doctype="routes" parent-id="${document['document_id']}" child-id="doc.document_id"
-              ng-if="${test_association_rights(document)}
-              % if document['type'] == 'o':
-                  && ${associations}.length > 1
-              % endif
-            "></app-delete-association>
+      <div class="collapse in" id="associated-routes">
+        <app-activity-filter activities="${activities}" documents="${associations}"></app-activity-filter>
+        <div class="associated-documents">
```
asaunier: The change in this file is mainly that CSS classes ``collapse in`` have been moved to a new parent container so that the ``app-activity-filter`` directive is shown  inside the associated routes section but is not impacted by the card's flex layout.
Most of the changed lines have actually been reindented.

PR #1656, opened 2017-05-13
Fix escaping in urls

tsauerwein:
URLs like `[url=https://httpbin.org/get?value1=1&value2=2]link[/url]` are converted to `<a href="https://httpbin.org/get?value1=1&amp;amp;value2=2">link</a>` (note the `&amp;amp;`), while it should be `<a href="https://httpbin.org/get?value1=1&amp;value2=2">link</a>`.

This PR prevents that the BBCode parser escapes characters again, that the Markdown parser already has escaped.

### DIFF
--- a/c2corg_ui/format/__init__.py
+++ b/c2corg_ui/format/__init__.py
@@ -43,6 +43,9 @@ def _get_markdown_parser():
 def _get_bbcode_parser():
     global _bbcode_parser
     if not _bbcode_parser:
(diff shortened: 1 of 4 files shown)


tsauerwein:
> Shall we rename _get_bbcode_parser to get_bbcode_parser since it may be called outside of c2corg_ui/format/__init__.py?

Right now, `_get_bbcode_parser` is only called in a test outside. And for me it's ok to call a private functions in a test. I would leave it as it is.

Issue #1576, opened 2017-03-30
Text formating button: insertion of a link


desnoes:
I made some test in this document: http://www.demov6.camptocamp.org/outings/828446/en/tete-de-vauclave-toto-fait-du-ski

- [ ]  The insertion of a link does not work.

- [ ]  ~~(There is a problem when using ordered and unordered lists successively, only the first type of list is selected. If there is another balise in between, it works fine.)~~

asaunier:
> There is a problem when using ordered and unordered lists successively, only the first type of list is selected.

Does this use case really happen?
Who need to have ordered and unordered lists successively?

asaunier:
> The insertion of a link does not work.

Could you precise what does not work?

I have just tried to use the link button and indeed observed a strange result:
* the button create some code similar to 
```
[enter link description here](http://example.com)
```
(which is a regular mardown link notation)
* but then when previewing the document or saving it a link is created with the following URL:
```
http://www.demov6.camptocamp.org/outings/edit/828446/a%20href=%22http://example.com%22%3Ehttp://example.com%3C/a
```
or
```
http://www.demov6.camptocamp.org/outings/828446/en/a%20href=%22http://example.com%22%3Ehttp://example.com%3C/a
```

=> I suspect there is a conflict between the bbcode/markdown parsers. It seems the first one replaces the raw URL by ``<a href="http://example.com">http://example.com</a>`` which creates the following code
```
[enter link description here](<a href="http://example.com">http://example.com</a>)
```
Then the markdown parser converts this syntax to 
```
<a href="<a href="http://example.com">http://example.com</a>">enter link description here</a>
```
=> the link is converted twice!

We might have to remove the auto converting of URLs to links.
What about disabling the bbcode parser once for all?

desnoes:
> Does this use case really happen?
Who need to have ordered and unordered lists successively?

I agree this use case is very unlikely to happen. But the bug is still there.

> Could you precise what does not work?

When i click on the link, i have the message 'Page non trouvée'.

> What about disabling the bbcode parser once for all?

What is the impact on legacy documents ?

asaunier:
> I agree this use case is very unlikely to happen. But the bug is still there.

But what's the point of spending time on something "very unlikely to happen"?

In addition this behaviour is not related to the work done for the project, it's specific to the (external) markdown parser used to interpret the markdown code. So if this is really an issue, it should rather be reported to https://github.com/waylan/Python-Markdown/issues

[...]

Issue #1660, opened 2017-05-15
Code interpration, bbcode, markdown


cbeauchesne:
Regarding this 

https://www.camptocamp.org/articles/838875/en/api-c2c-v6

I think I've understood what's happen : I think that bbcode parser sanitize markdown output, which is already sanitized, and that's why all & and </> are replaced by html characters. If it's the case, there is two possible solutions : 

1. Modify both parser sanitazitation, and add one after all
2. Recode BBcode parsing => this one is not so hard, bbcode is very simple. FYI, it's what I've done, and it was very simple as it's most of the case a simple replacement. 

Another suggestion : as most of bbcode can be done by markdown, do a migration to remove all bbcode and stop supporting it. It would simplify a lot the parsing :smile: The only part to keep should be images...

tsauerwein:
> Another suggestion : as most of bbcode can be done by markdown, do a migration to remove all bbcode and stop supporting it. It would simplify a lot the parsing

+1 I think we could avoid quite some problems by only supporting Markdown.

cbeauchesne:
@tsauerwein,  To be honest, it's what I would do if I was the only person who decides. it's quite a big  effort to do, but, once it's done, life will be far more easier for both dev and users.

Let me know if you want to follow that way, I will be glad to give a hand on that subject.

brunobesson:
@cbeauchesne no need to remind us of your opinion on decision-making...

Almost everyone, (except maybe @fbunoz?) agrees that removing old bbcode parser would be best.
We already discussed this.
Still, the database migration script is not that easy to write, plus that would mean communicating appropriately with users. But everyone should be delighted if you have the time and will to solve this issue.

fbunoz:
We'll be able to remove bbcode when we'll be able to do the same thing whith markdown, without side effect.
It implies to debug or configure markdown syntax.

cbeauchesne:
Ok guys, I can take this point if you want.

I think this subject is far to complicated to be handled in one big migration/PR. I propose a step-by-step approch that will help us to limit side effects : 

1. build a script that runs on entire DB and outputs list of documents ids that contains a certain pattern
  * this script will be very usefull for plenty of subject
2. build a bot editor that use API to read/write documents
  * this bot will be also  very usefull
3. list all patterns that can be migrated
4. write markdown extension to implement part of BBcode that can't be handled by markdown (I see video and images)
5. pattern by pattern follow this process : 
  1. Run script to output ids
[...]

PR #1661, opened 2017-05-15
Fix markdown links

tsauerwein:
Markdown links (`[Link](http://www.link.com)`) were not replaced properly. After the Markdown parser had inserted the link (`<a href="...">...</a>`), the BBCode parser created again a link for the part inside `href="..."`.

This PR proposes not to replace links without tag with the BBCode parser, instead to do it with a Markdown extension.

See also: https://github.com/c2corg/v6_ui/issues/1576#issuecomment-290510655

### DIFF
--- a/c2corg_ui/format/__init__.py
+++ b/c2corg_ui/format/__init__.py
@@ -2,6 +2,7 @@ import bbcode
 import markdown
 import html
 
+from c2corg_ui.format.autolink import AutoLinkExtension
 from c2corg_ui.format.wikilinks import C2CWikiLinkExtension
 from c2corg_ui.format.img import C2CImageExtension
 from c2corg_ui.format.video import C2CVideoExtension
@@ -34,6 +35,7 @@ def _get_markdown_parser():
             C2CWarningExtension(),
             Nl2BrExtension(),
             TocExtension(marker='[toc]', baselevel=2),
+            AutoLinkExtension()
         ]
         _markdown_parser = markdown.Markdown(output_format='xhtml5',
                                              extensions=extensions)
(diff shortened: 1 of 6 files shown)


c2corg_ui/format/__init__.py
```diff
@@ -34,6 +35,7 @@ def _get_markdown_parser():
             C2CWarningExtension(),
             Nl2BrExtension(),
             TocExtension(marker='[toc]', baselevel=2),
+            AutoLinkExtension()
```
tsauerwein: @asaunier, do you remember why not all of these extensions are enabled for the unit tests?

cbeauchesne:
Guys, You should try parsing bbcode before markdown. Markdown parser if far more advanced, and wont mess up with html tag produced by bbcode parser.

c2corg_ui/format/__init__.py
asaunier: No special reason I think, probably the unit tests have been forgotten when adding those extensions.

asaunier:
> You should try parsing bbcode before markdown

@cbeauchesne I don't remember the details but we have chosen to run the bbcode parser before the markdown parser because other conflicts were occuring.
As you and @tsauerwein told it in https://github.com/c2corg/v6_ui/issues/1660 I would recommend to get rid of the bbcode parser.

cbeauchesne:
@asaunier, my bad, I've probably read the code too quiqly, thought bbcode was run after.

tsauerwein:
In the tests, the Markdown parser was run before the BBCode parser. But for rendering the pages, it was the opposite. :O 

I cleaned this up.

c2corg_ui/format/__init__.py
tsauerwein: I am now using the same function with all extensions as for rendering the pages.

asaunier:
Great, thanks!
Looks good to me, please merge when Travis is OK.

Issue #1622, opened 2017-05-05
Advanced search display on mobile


desnoes:
The filter area is floating and moves to the left when i try to move the right par of a slider.
Chrome (last version)
![screenshot_20170504-190753](https://cloud.githubusercontent.com/assets/19892089/25736897/0937d58c-3175-11e7-9bdc-a2ca55b4329e.png)

asaunier:
I indeed observe this behaviour on my android phone with Chrome (not tested with another problem).
At least the page remains correctly usable: it is still possible to scroll down and this behaviour occurs only when the filters section is open while panning to the right.

I suspect this is caused by the buttons "More filters" / "Clear filters" / "Load my preferences" that are shown on the same line whereas there is not enough width to show them all:
![capture du 2017-05-18 14-37-05](https://cloud.githubusercontent.com/assets/1192331/26202404/a0923fdc-3bd7-11e7-83bd-58f2dc3d3145.png)

When clicking on "More filters", the buttons are dispatched on several lines, eg.
![capture du 2017-05-18 14-39-28](https://cloud.githubusercontent.com/assets/1192331/26202453/d1abb5bc-3bd7-11e7-88b5-522b8b622f99.png)

PR #1666, opened 2017-05-19
Fix incorrect positioning of filters buttons

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1622

See the explanations and screenshots at https://github.com/c2corg/v6_ui/issues/1622#issuecomment-302391613

Currently this PR makes the buttons in the primary filters section wrap to a new line if all 3 are shown and the screen width is too small (instead of overflowing and causing a horizontal scroll bar).

Side-effect to handle: when scrolling down in a mobile screen, the button are made "sticky" at the top of the screen but remain on 2 lines:
capture du 2017-05-19 19-23-31

Should I "simply" make the gray background bigger so that both lines are covered (but then the screen height available to show the cards is reduced)? Or perhaps we could use reduced button titles to make sure the buttons are rendered on a single line, at least when "sticky"?

### DIFF
--- a/less/filters.less
+++ b/less/filters.less
@@ -292,6 +292,7 @@ form[app-search-filters] {
 .more-filters-btn-container {
   .flex();
   justify-content: center;
+  flex-flow: wrap row;
   padding-top: 5px;
   transition: .2s;
   @media @phone {
--- a/less/stickyfilters.less
+++ b/less/stickyfilters.less
@@ -5,7 +5,6 @@
   font-size: 1.4em;
   z-index: 29;
   top: @page-header-height + @page-main-title-height;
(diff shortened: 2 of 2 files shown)


asaunier:

> Side-effect to handle: when scrolling down in a mobile screen, the button are made "sticky" at the top of the screen but remain on 2 lines:

I have simply made sure the gray background is displayed behind the third button if displayed on a second line:
capture du 2017-05-30 18-49-04

If there is only 2 buttons (for instance when no filter is applied => the "Clear filters" button is not shown) or if the screen width is big enough, only one line is shown with the adapted background:

capture du 2017-05-30 18-50-56

asaunier:
Thanks @brunobesson !

PR #1669, opened 2017-06-02
Add document change reverting tool

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1068
Alternate approach to https://github.com/c2corg/v6_ui/pull/1101

API part: https://github.com/c2corg/v6_api/pull/671
capture du 2017-06-08 16-25-52

The "restore this version" is only shown to moderators when viewing an archived version of a document.
capture du 2017-06-08 16-38-43


asaunier:
A use case that could be improved (in the API revert service): if you revert a route to a version whose main waypoint is no longer associated to the route, the name of this waypoint will be shown as route's title prefix even though the waypoint remains unassociated to the route. Which is inconsistent.

We could add a check and reassociate the WP if needed in the API revert service.
But at least:
* it causes no crash
* it is still possible to associate the WP by hand
* it's probably not that common

Issue #1672, opened 2017-06-06
Confusing month translations in the outing cards


asaunier:
In outing cards, only the first 3 letters of the month names are shown:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/partials/cards/outings.html#L6

which may be confusing. For instance in French june and july have the same first 3 letters:
![capture du 2017-06-06 09-35-16](https://cloud.githubusercontent.com/assets/1192331/26818625/b5d6e76c-4a9b-11e7-8610-feaa068e0884.png)

desnoes:
Proposition: JUN for June (latin: Junius, french: Juin) et JUL pour July (Julius/Juillet)

asaunier:
@desnoes I think I have found a simple and efficient solution (#1675)

PR #1675, opened 2017-06-07
Use angular moment to format month names in outing cards

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1672

capture du 2017-06-07 15-20-36
capture du 2017-06-07 15-20-05


asaunier:
Anyone can review this easy PR? @stef74 I think you are the better placed for this job since you have worked on the dates in outing cards. The change is that we use angular-moment formatting for months as done for instance for the date of the outing (or xreports or images) view pages: https://github.com/c2corg/v6_ui/blob/b24a07d6287fcde1beb0d44f1898ce912b9147ea/c2corg_ui/templates/outing/helpers/view.html#L10

asaunier:
Thanks @brunobesson :)

PR #1681, opened 2017-06-09
Add MW contest link

brunobesson:
when public_transport checkbox is checked in outing edit form, add link to participate to MW changerdapproche contest

### DIFF
--- a/c2corg_ui/templates/outing/edit.html
+++ b/c2corg_ui/templates/outing/edit.html
@@ -397,6 +397,7 @@ updating_doc = outing_id and outing_lang
               <label translate>public_transport</label>
               <input type="checkbox" ng-model="outing.public_transport" ng-checked="outing.public_transport == true">
               <span ng-click="editCtrl.pushToArray(outing, 'public_transport', !outing.public_transport, $event)"></span>
+              <a href="http://changerdapproche.org/le-concours/presentation.html" target="_blank" ng-if="outing.public_transport" translate>Participate to MW contest</a>
             </div>
 
             ## FREQUENTATION


c2corg_ui/templates/outing/edit.html
```diff
@@ -397,6 +397,7 @@ <h2 class="heading show-phone">
               <label translate>public_transport</label>
               <input type="checkbox" ng-model="outing.public_transport" ng-checked="outing.public_transport == true">
               <span ng-click="editCtrl.pushToArray(outing, 'public_transport', !outing.public_transport, $event)"></span>
+              <a href="http://changerdapproche.org/le-concours/presentation.html" target="_blank" ng-if="outing.public_transport == true" translate>Participate to MW contest</a>
```
asaunier: You can probably simply write
```
ng-if="outing.public_transport"
```
?

brunobesson: How stupid :confounded: 
Fast copy and paste (have a look at line 398... :wink:)

asaunier: I wonder if line 398 is correct: do we really need the ``ng-checked`` when we already have an ``ng-model`` based on the same property.
Have you tried to remove it?

asaunier:
I had not noticed (?) Travis reported some errors with this commit.
Please always make sure tests pass before merging :P

c2corg_ui/templates/outing/edit.html
brunobesson: Nope, I was focusing on adding the link only (I should have done it a long time ago already). Not much time to participate currently :sob: 
I'll have a look at it sometimes.

brunobesson:
The tests failed only because I did amend my commit that was already pushed to github, or so I understand. I don't think adding this line could cause any test to fail, does it?

asaunier:
I don't know what has happened with the test. Here is the output of Travis build related to this branch:
capture du 2017-06-09 10-31-47

At least tests pass in my instance with the HEAD of the master branch so I guess it's OK.

PR #1683, opened 2017-06-09
remove useless '== true'

brunobesson:


asaunier:
Thanks! Perhaps try to remove the ``ng-checked``? I have the feeling that a checkbox whose ``ng-model`` is true will already be checked (not sure).

brunobesson:
I am more fluent with angular 2+, but your assertion seems OK to me. I'll have a look sometimes, when I find the time. I'll add the correction there if applies, before merging.

asaunier:
By the way the doc https://docs.angularjs.org/api/ng/directive/ngChecked says
> Note that this directive should not be used together with ngModel, as this can lead to unexpected behavior.

:P

brunobesson:
> if you have tested and make sure the behaviour is as expected.

Of course I did :wink:

PR #1686, opened 2017-06-09
Add a spinner when the profile page data are loading

asaunier:
I have noticed that when loading a profile page, if the API response is a bit long, nothing tells the visitor that something is going on => this PR simply adds a loading spinner when loading profile data until the content is displayed.


asaunier:
Minor change and tested in my instance => merging

Issue #1691, opened 2017-06-14
Feed layout: add margin between document details and images


brunobesson:
Currently, there is no space between document description and images:
![chrome3](https://user-images.githubusercontent.com/2234024/27134425-cf4619d6-5115-11e7-9fcc-cde010375948.png)

PR #1692, opened 2017-06-14
Ensure consistent margin with feed images

brunobesson:
Related to #1691.

Was:
chrome3

Now:
chrome4

Off course, margin is consistent when images are positioned below:
chrome5


asaunier:
This looks good to me, thanks.
@stef74 OK for you?

PR #1698, opened 2017-06-19
Use black and white coloring for printed wind rose

brunobesson:
c2c1

* no green and grey coloring, use black & white instead
* remove text
* reduce size a little


brunobesson:
Related to #1699

PR #1701, opened 2017-06-19
Use consistent trash button positioning

brunobesson:
Ensures that trash button in cards is shown homogeneously.

There is a remaining problem for moderators displaying results as lists. But at least the current result is not worse than before.

### DIFF
--- a/less/c2corg_ui.less
+++ b/less/c2corg_ui.less
@@ -942,12 +942,13 @@ ul {
   }
 }
 
-.glyphicon-trash:not(app-delete-association) {
+.glyphicon-trash {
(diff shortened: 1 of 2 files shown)


brunobesson:
Was
c2c4
Now
c2c5
Although there are still problems with list view (but not worse than before anyway)
c2c6

asaunier:
@brunobesson Perhaps this change had a side-effect with the trash icons in the outing editing forms (conditions):
capture du 2017-06-21 19-45-17

brunobesson:
Nice catch, thanks. I'll commit a fix.

PR #1702, opened 2017-06-19
Remove useless mixins

brunobesson:
flex(), shadow() and calc() are supported natively by the browsers we aim

### DIFF
--- a/less/404.less
+++ b/less/404.less
@@ -81,11 +81,11 @@
     margin-bottom: 5px;
   }
 
-  .flex();
+  display: flex;
   flex-flow: wrap row;
   a {
     color: graytext;
-    .flex();
+    display: flex;
     flex-direction: column-reverse;
     justify-content: center;
     font-weight: bold;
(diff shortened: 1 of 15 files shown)


asaunier:
I agree it is better to use standard CSS, no clue if there was a reason for using flex(), shadow() and calc().
I don't know much about CSS and have not much opinion about the changes :)
By the way a conflict should be fixed anyway.

brunobesson:
I rebased the branch.

To see browser support, have a look at 

* https://caniuse.com/#search=flex
* https://caniuse.com/#search=box-shadow
* https://caniuse.com/#search=calc

There is no use for vendor prefixes.

asaunier:
I don't understand what your links mean: they seem to show that all recent browsers support natively those styles but what about (a bit) older browsers? For instance don't we support Safari 9? (even though the GPX download does not work as expected with Safari 9 :P )

brunobesson:
>I don't understand what your links mean

This website indeed show support of (JS, HTML5...) features in web browsers. You can switch the view for usage relative or date relative presentation.
For the 3 elements that concern us, they are supported correctly for Safari 9+, IE 11 (even for flex in our case), Edge and not too old versions of FF and Chrome. So there is really no need for vendor prefixes, trust me!

Issue #1054, opened 2016-12-01
Images search date filter do not work


dkocich:
* [ ] Image count is not updated and list remains same after selection of any date.

![image](https://cloud.githubusercontent.com/assets/6165660/20787894/7fa8bb14-b7ae-11e6-9c61-4d2e034a21bf.png)

asaunier:
In addition
* [ ] The date picker is cut:
![sans titre](https://cloud.githubusercontent.com/assets/1192331/20788358/c62904e8-b7b0-11e6-8cb1-c0c2d4e27a10.png)

ginold:
^ the same problem as stated in #655 -> it only happens when you scroll down and open a menu, not happening if you are on the top of the list.

asaunier:
> Image count is not updated and list remains same after selection of any date.

Selecting a date does not even update the filter parameters in the URL, thus not trigger the search either.

That's because the date filter in the images search page:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/image/filters.html#L41
use a ``filtersCtrl.setDate()`` function that is only available for outings:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/outingfilters.js#L89
not in the standard search controller used for images:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/search/searchfilters.js

To add the date support we should create a extended fitler controller for images (as for outings) with a ``setDate`` function (or factorize it to the parent controller but it's not really a common function needed by all document types).

Question: do we really need a date filter for images? If not, I would recommend to simply remove this filter. Has anyone already complained about it?

asaunier:
> Question: do we really need a date filter for images? If not, I would recommend to simply remove this filter. Has anyone already complained about it?

@desnoes Has the association given some feedbacks about this question? Is the issue still relevant?

desnoes:
According to the discussion in the forum, this functionality is not really necessary. You can remove the filter.

PR #1708, opened 2017-06-20
Remove not working date selector from images filters

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1054

### DIFF
--- a/c2corg_ui/templates/image/index.html
+++ b/c2corg_ui/templates/image/index.html
@@ -7,9 +7,9 @@
 <%block name="moduleConstantsValues">
   module.value('mapFeatureCollection', null);
   module.value('advancedSearchFilters', {
-    'date': {'type': 'date'},
     'act': {'type': 'list'},
-    'oalt': {'type': 'range'}
+    'cat': {'type': 'list'},
+    'ityp': {'type': 'list'}
   });
 </%block>
 
--- a/c2corg_ui/templates/image/filters.html
+++ b/c2corg_ui/templates/image/filters.html
@@ -35,17 +35,6 @@ from c2corg_common.attributes import default_langs, activities, image_categories
         ${add_multiselect('ityp', image_types)}
       </div>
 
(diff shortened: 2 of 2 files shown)


c2corg_ui/templates/image/index.html
```diff
@@ -7,9 +7,9 @@
 <%block name="moduleConstantsValues">
   module.value('mapFeatureCollection', null);
   module.value('advancedSearchFilters', {
-    'date': {'type': 'date'},
     'act': {'type': 'list'},
-    'oalt': {'type': 'range'}
+    'cat': {'type': 'list'},
```
brunobesson: In this PR, you remove filter, but also add some new values. Were they missing?

asaunier: There is no ``oalt`` filter for images, it's probably an incorrect copy-paste of the outings filters.
Here are the available filters for images: https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/image_mapping.py#L11-L19

I have added some filters that were not correctly defined. Those definitions are used in the search filters directive, for instance to parse the filtering criterias in the URL: https://github.com/c2corg/v6_ui/blob/a86893997e2b92976d79dab55b15c467eedad7ac/c2corg_ui/static/js/search/searchfilters.js#L163-L193

Currently when loading https://www.camptocamp.org/images#bbox=773697%252C5610223%252C778464%252C5617799&cat=action%252Ctrack&act=snow_ice_mixed%252Cmountain_climbing
* activities are correctly restored (because ``act`` is defined in the list above)
* categories are not (see for instance the number of selected items in the categories list)
capture du 2017-06-21 13-10-30

asaunier:
@brunobesson OK for merging?
@desnoes has confirmed this date filter is not really necessary and can be removed. See https://github.com/c2corg/v6_ui/issues/1054#issuecomment-309758164

brunobesson:
Sure. Easy review otherwise :slightly_smiling_face:

Issue #1515, opened 2017-02-22
Block comments on xreports and outings


desnoes:
An user should be able to authorize or ban comments from other users on its personnal reports:

- xreport: mandatory
- outing: likely

asaunier:
@desnoes Could you precise the need:
* could you confirm what is asked is only to allow/disallow comments in an outing/xreport, not to give moderation rights on comments to the outing's associated users?
* would the ban be on a per-document basis, using for instance a checkbox in the editing form?
* would it be possible to change the policy by editing the document and (un)checking the checkbox? What would happen if a user disallow comments after comments have already been posted? Would comments be deleted? hidden? Would it be no longer possible to disallow comments when there are existing comments?

> xreport: mandatory

Does this mean that comments would anyway be disabled for xreports? (the author would then not even have the choice to allow comments)

desnoes:
> could you confirm what is asked is only to allow/disallow comments in an outing/xreport, not to give moderation rights on comments to the outing's associated users?

Could you explain? I am not sure i understood the difference between the 2 options.

> would the ban be on a per-document basis, using for instance a checkbox in the editing form?

From my personal point of view, the answer is yes.

> would it be possible to change the policy by editing the document and (un)checking the checkbox? What would happen if a user disallow comments after comments have already been posted? Would comments be deleted? hidden? Would it be no longer possible to disallow comments when there are existing comments?

I would select the simpliest option: no modification of policy by editing the document. If a problem arises it is still possible to delete comments manually.

> Does this mean that comments would anyway be disabled for xreports? (the author would then not even have the choice to allow comments)

No, i meant the requirement (possibility to allow/disallow comment) is mandatory for xreport, while it is "a nice to have" functionnality for outing. If it is simpler to implement the same procedure for xreport and outing, this is great. If additional complexity is implied for outing, then the requirement can be dropped for outings.

@mgerbaux : what is your opinion ?

asaunier:
> I am not sure i understood the difference between the 2 options.

* option 1: the user decide whether or not comments are allowed in the outing
* option 2: the user can remove (edit?) some of the comments posted in the outing => kind of moderation rights as sometimes suggested in the association forum. But I think it is really complicated and would not recommend this option.

>  i meant the requirement (possibility to allow/disallow comment) is mandatory for xreport, while it is "a nice to have" functionnality for outing.

[...]

PR #1712, opened 2017-06-21
Allow to disable comments in outings or xreports

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1515
Requires https://github.com/c2corg/v6_common/pull/57 and https://github.com/c2corg/v6_api/pull/673

[for outings and xreports only]

The PR adds a ``disable_comments`` attribute that can be checked/unchecked in the outings/xreports editing forms:
capture du 2017-06-21 16-04-29

If checked, the comments section (even though comments have been posted in the meantime) is replaced by "Comments are disabled.":
capture du 2017-06-21 16-05-09

If not checked comments are shown as usual.

### DIFF
--- a/c2corg_ui/templates/outing/edit.html
+++ b/c2corg_ui/templates/outing/edit.html
@@ -528,7 +528,7 @@ updating_doc = outing_id and outing_lang
             </div>
 
             ## PARTICIPANTS
-            <div id="participants-group" class="form-group  full" ng-class="{'has-success': outing.locales[0]['participants']}">
+            <div id="participants-group" class="form-group full" ng-class="{'has-success': outing.locales[0]['participants']}">
               <label translate>participants</label>
               <div class="input-group">
                 <input type="text"  ng-model="outing.locales[0]['participants']" class="form-control" />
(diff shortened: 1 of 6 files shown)


brunobesson:
How does this handle the case when comments have already been posted on the subject?

asaunier:
They are simply not shown. Until the checkbox is unchecked again. See the "specs" here: https://github.com/c2corg/v6_ui/issues/1515#issuecomment-289016578

brunobesson:
OK, sorry I didn't follow. This does not look like the perfect choice to me, but in the vast majority of cases, this should be OK :slightly_smiling_face: 
Not sure I should make this review though.

asaunier:
> This does not look like the perfect choice to me, but in the vast majority of cases, this should be OK. Not sure I should make this review though.

It's indeed a simplified version but I do think we should do simple first and then see if more complicated approaches are needed.

@arnaud-morvan is in vacation.
@brunobesson or @stef74 could you make the review please so that I can add this feature to the new release (today hopefuly since I am in vacation next week :P )?
The changes are quite simple, both in the UI and the API and the review is quite easy I think.
The API PR is here: https://github.com/c2corg/v6_api/pull/673
Thanks!

c2corg_ui/templates/outing/edit.html
```diff
@@ -555,11 +555,18 @@ <h5 translate ng-show="outing.associations.users.length == 0">You can look up fo
...
+            <div class="full form-group">
               <label translate>personal comments</label>
               <textarea app-markdown-editor class="form-control description" placeholder="{{'write your comments' | translate}}" ng-model="outing.locales[0]['description']"></textarea>
             </div>
 
+            ## COMMENTS MANAGEMENT
+            <div class="form-group half">
+              <label translate>disable_comments</label>
+              <input type="checkbox" ng-model="outing.disable_comments">
+              <span ng-click="editCtrl.pushToArray(outing, 'disable_comments', !outing.disable_comments, $event)" translate>If checked, comments cannot be posted or read.</span>
```
brunobesson: I guess you checked, but just to be sure. Is it really `!outing.disable_comments` (negated)?

brunobesson:
Done on all 3 projects.

PR #1713, opened 2017-06-21
Use different list switch button for associations in views than for lists

brunobesson:
* better visuals (IMHO)
* prepare for a possible actions toolbar
* some ng-show -> ng-if optimisations

c2c9
c2c11


c2corg_ui/static/js/listswitchaction.js
```diff
@@ -0,0 +1,48 @@
...
+  /**
+   * @type {app.ListSwitch}
+   */
+  this.listSwitchService = appListSwitch;
+
+  /**
+   * @type {boolean}
+   * @export
+   */
+  this.showList = this.listSwitchService.isList();
```
asaunier: ``isList()`` does not return anything!?

c2corg_ui/static/js/listswitchbutton.js
```diff
@@ -0,0 +1,48 @@
+goog.provide('app.ListSwitchButtonController');
```
asaunier: These directive/controller seem the same than the ones of the action. Perhaps you could avoid duplicating code by switching the template depending on some parameter? See for instance https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/card.js#L24-L25

Or you could group both templates into one and use ``ng-if`` to figure out which layout to use?

c2corg_ui/static/js/listswitchaction.js
brunobesson: isList() returns whether the display is configured for lists or cards: https://github.com/c2corg/v6_ui/pull/1713/files#diff-1cd444cb2344de44c8f6791257e43f67R37

Or maybe I didn't understand your question?

c2corg_ui/static/js/listswitchbutton.js
brunobesson: Indeed I didn't know how to optimize this redundancy. If so, maybe the service itself is useless then, and I could make a much simpler PR? What do you think in terms of modeling?

c2corg_ui/static/js/listswitchaction.js
asaunier: Forget my comment, I think I have misread something :P

Issue #1689, opened 2017-06-12
Error 500 for outings containing a [video] tag


desnoes:
A message error is returned when opening these outings: https://www.camptocamp.org/outings/883813/fr/verdon-l-escales-l-offre
https://www.camptocamp.org/outings/894458/fr/les-tampes-col-du-freu-et-couloir-s

According to https://forum.camptocamp.org/t/v6-vos-questions-les-reponses/173468/491?u=miko, the bug could come from the link to youtube. This sounds weird. To be clarified. The 2 outings add a link to youtube for a video.

asaunier:
The 2 example outings contain a ``[video]`` tag such as ``[video]https://www.youtube.com/watch?v=rdgyTDAoQiY[/video]``. This tag seems to make the document viewing and previewing crash. Here is the error message found in the logs:
```
Traceback (most recent call last):
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid_debugtoolbar/toolbar.py", line 192, in toolbar_tween
    response = _handler(request)
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid_debugtoolbar/panels/performance.py", line 57, in resource_timer_handler
    result = handler(request)
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid/tweens.py", line 62, in excview_tween
    reraise(*attrs['exc_info'])
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid/compat.py", line 148, in reraise
    raise value
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid/router.py", line 158, in handle_request
    view_name
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/home/alex/Desktop/v6_ui/.build/venv/lib/python3.5/site-packages/pyramid/viewderivers.py", line 465, in rendered_view
    request, result, view_inst, context)
[...]

PR #1716, opened 2017-06-22
Video extension should come before autolink extension

brunobesson:
@fbunoz was right: the autolink extension was eating the video tag.
This PR fixes it by applying the video extension before the autolink.

@asaunier I don't know how to run the autolink tests. Could you check it is still working appropriately?

### DIFF
--- a/c2corg_ui/format/__init__.py
+++ b/c2corg_ui/format/__init__.py
@@ -30,12 +30,12 @@ def _get_markdown_parser():
         extensions = [
             C2CWikiLinkExtension(),
             C2CImageExtension(api_url=_parsers_settings['api_url']),
-            C2CVideoExtension(),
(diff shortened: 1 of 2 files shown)


brunobesson:
See #1689

asaunier:
> I don't know how to run the autolink tests. Could you check it is still working appropriately?

I have tested the branch in my instance: inserting a video tag now works as expected.
The tests pass as well (so does Travis). To run the test in the UI, since there are much less tests in the UI than in the API you can simply run them like this:
```
make -f config/... check
```

To run a special test, see https://github.com/c2corg/v6_ui#run-the-tests eg.
```
.build/venv/bin/nosetests c2corg_ui/tests/format/test_format.py
```

PR #1717, opened 2017-06-22
Rework sidemenu

brunobesson:
* to ensure ad is displayed, no more intermediate status for tablet
* when opened on mobile or tablet, content is grayed, clicking anywhere closes the menu
* buttons to open or close the menu are positioned identically
* rework layout

next stuff to (possibly) handle:

* smaller ad when there is not enough place for the big one? e.g. on mobile, 160 * 60?
* allow to fold menu on desktop (and save preferences)

### DIFF
--- a/c2corg_ui/static/js/style/sidemenu.js
+++ b/c2corg_ui/static/js/style/sidemenu.js
@@ -6,17 +6,27 @@ goog.require('app');
  * @export
  */
 app.sidemenu = function() {
-  $('.menu-open-close').click(function() {
-    if ($('#sidemenu').hasClass('menu-open')) {
-      $('#sidemenu').removeClass('menu-open');
-      $('#sidemenu').addClass('menu-closed');
+  /* handle sidebar hiding */
+  var body = $('body');
+  var content = $('.page-content');
+
+  $('.menu-open-close').on('click', function() {
+    body.toggleClass('menu-toggled');
+    if (body.hasClass('menu-toggled')) {
+      content.prepend('<div id="content-toggled">');
     } else {
-      $('#sidemenu').removeClass('menu-closed');
-      $('#sidemenu').addClass('menu-open');
+      $('#content-toggled').remove();
     }
   });
 
-  $('.menu-open-close').click(function() {
-    $('.menu-open-close.menu').toggleClass('is-active');
+  $(window).resize(function() {
+    if (window.innerWidth >= 1100 && body.hasClass('menu-toggled')) {
+      body.toggleClass('menu-toggled', false);
+      $('#content-toggled').remove();
+    }
+  });
+  content.on('click', '#content-toggled', function() {
+    body.toggleClass('menu-toggled', false);
+    $('#content-toggled').remove();
   });
 };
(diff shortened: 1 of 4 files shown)


brunobesson:
c2c14
c2c15
c2c16

brunobesson:
And of course, there will be some work to adapt it for discourse... But we need to validate it first.

asaunier:
Looks good to me, thanks!

> to ensure ad is displayed, no more intermediate status for tablet

OK for me but perhaps it could be validated by more people :)
@stef74 @desnoes maybe?

> when opened on mobile or tablet, content is grayed, clicking anywhere closes the menu

👍 

> buttons to open or close the menu are positioned identically

👍 

By the way I have noticed the logo is now centered in the desktop sidemenu, instead of being left-aligned. Why not even though we perhaps loose the consistency with menu items that are left-aligned too. Is it related with the need to have space for the close menu button to the left?

> allow to fold menu on desktop (and save preferences)

+1

brunobesson:
>Is it related with the need to have space for the close menu button to the left?

@asaunier yes it is, plus I do think it is nicer also on desktop when centered. But I could change that depending on feedback of course.

Concerning folding on desktop, how would you see that? For the same reason I think we need to get rid of the previous tablet menu (half hidden, no ad), I don't think having a reduced menu like this is OK. Thus, if we wish to store in preferences the fact that the menu should be hidden by default, we cannot simply use the switch hamburger for this.

asaunier:
@brunobesson I totally agree with you. For me it's OK to merge the PR.

brunobesson:
OK, thanks. Let's wait for some feedback from @desnoes or @stef74 then.

desnoes:
Nice job! Ok for the merge.

brunobesson:
OK, I'll merge this PR.
I won't be working on the other two points unless I get some feedback:

* how could we handle the option to hide the menu by default on desktop?
* do we want to be able to display smaller ads for screens that do not have enough height? or could we try a scrollable menu? (@mgerbaux this subject interests you I think)

asaunier:
> how could we handle the option to hide the menu by default on desktop?

The same way prefs about list/cards or maps fold/unfold are managed (saved in the localStorage when changed)? Perhaps that would be worth to develop a service that would store such prefs in a consistent (factorized) way instead of managing them in dedicated controllers.

brunobesson:
>The same way (...)

Sorry, my question was rather focusing on the UI.

brunobesson:
>Perhaps the same button than for mobile except that the menu would be open by default on desktop on the contrary of mobile screens?

Then there would be no way to store preferences:

* I click on the hamburger to hide the menu. Yay! (preferences are save: menu: hidden)
* I want to go to another page. I click on the hamburger to open the menu, then go to [whatever] (menu: shown)
* I land on the new page. Menu is open because (menu: shown) in preferences. But I would like it to be hidden by default now!

brunobesson:
>Perhaps that would be worth to develop a service that would store such prefs in a consistent (factorized) way instead of managing them in dedicated controllers.

:+1: for this anyway. We could already factorize list display and map disaply options. Do you think it is worth doing it already? Maybe wait until we converged on this subject?

asaunier:
> Then there would be no way to store preferences:

Indeed, I had not thought about that :P 

>  We could already factorize list display and map disaply options. Do you think it is worth doing it already? Maybe wait until we converge on this subject?

We can wait we have a solution for the sidemenu.

Issue #1507, opened 2017-02-21
Anonymize xreport


desnoes:
A user should be able to choose to hide his name when the report is publicly displayed.
A check box should be inserted in the creation form to offer this choice.

asaunier:
Is it OK if this is handled on the UI side only (the template does not show the associated users section if the box is checked)? Or should we make sure the API does not return at all the list of associated users?

Option 1 is much easier. Is it enough?

Anyway please note that making the report anonymous will most likely make it no longer possible to edit.
Because editing credentials are related to associated users. Unless the associated users are still provided to the client code but not displayed (which is a quite superficial anonymization).

Depending on the requirements of this demand, it can become quite difficult to implement.

stef74:
Yes for me UI approach is sufficient. If user want really anonymised contribution ... it use an other account.

desnoes:
@asaunier: for me it's ok, We shall add a clear statement about this constraint when a user creates the report.
@mgerbaux: what is your opinion ?

mgerbaux:
In my point of view, It could be only a short term patch to do this only in UI ; It is not the way to make a confidante relationship with the users.
Maybe the time to make some work on API to pass author as private argument ?

> If user want really anonymised contribution ... it use an other account.
It is not the way to encourage and facilitate contributions asked by our partner

Note that anonymization should also be done in the versions data (maybe an other issue related to all personnal documents).

asaunier:
> Maybe the time to make some work on API to pass author as private argument ?

You mean it would be passed like the personal data?

> Note that anonymization should also be done in the versions data (maybe an other issue related to all personnal documents).

Associations are not integrated in archived versions and are independent (ditto for all document types).

asaunier:
## Approach 1, allowing to make an existing or new report anonymous 

TODO:

API side:
* [ ] add an "anonymous" boolean attribute to the data model https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/xreport.py
* [ ] remove attributes ``author`` and ``users`` from the the "detail" data if ``anonymous`` is true (already OK for the "list" request: no user info is provided in these data)
* [ ] disable the history request (for all xreports)? => problem: there is only one service for all document types https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_history.py
* [ ] disable the old versions service: https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/xreport.py#L103
[...]

PR #1720, opened 2017-06-23
Add support of anonymous xreports

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1507 (approach 2)
Requires https://github.com/c2corg/v6_common/pull/58
API part: https://github.com/c2corg/v6_api/pull/674

### DIFF
--- a/c2corg_ui/static/js/editing/xreportediting.js
+++ b/c2corg_ui/static/js/editing/xreportediting.js
@@ -64,13 +64,26 @@ goog.inherits(app.XreportEditingController, app.DocumentEditingController);
 
 
 /**
-* @param {appx.Document} data
-* @return {appx.Document}
-* @override
-* @public
-*/
+ * @param {appx.Document} data
+ * @return {appx.Document}
+ * @override
+ * @public
+ */
 app.XreportEditingController.prototype.filterData = function(data) {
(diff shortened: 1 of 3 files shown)


asaunier:

What it does:
* add an ``anonymous`` attribute right above the associated users section in the xreport editing form (only when creating a report - existing reports cannot be made anonymous)
capture du 2017-07-05 12-44-26

* if checked, the associated users section is hidden (shown again if unchecked)
capture du 2017-07-05 12-44-38

* when submitting the form, if ``anonymous=true``, the associated users list is reset
* the API side assigns the report to some special account (which means the report can no longer be edited by the original author, only by moderators or by the special account owners)
capture du 2017-07-05 12-43-36
capture du 2017-07-05 12-44-01

desnoes:
Nice !
@mgerbaux: which account should be associated as the author of the report ? We should have a special account for the SERAC DB moderators ?

mgerbaux:
> which account should be associated as the author of the report ?

https://www.camptocamp.org/profiles/271737/fr

asaunier:
@mgerbaux Thanks!
I have added this setting to the ``docker-compose.yml`` file of the demo and prod, to have them use the "Base accidents" account instead of the default "Camptocamp Association" account.

PR #1723, opened 2017-06-26
Outings CSV download

brunobesson:
Adds an icon right to the "My outings" menu item that allows to retrieve one's outings in CSV format (with very simple information currently).

I'll appreciate some feedback on this matter.

c2c17

### DIFF
--- a/c2corg_ui/static/js/outingscsvdownload.js
+++ b/c2corg_ui/static/js/outingscsvdownload.js
@@ -1,0 +1,61 @@
+goog.provide('app.OutingsCsvDownloadController');
+goog.provide('app.outingsCsvDownloadDirective');
+
+goog.require('app');
+
+
+/**
+ * This directive is used to display a download button for outings in CSV format.
+ *
+ * @return {angular.Directive} The directive specs.
+ * @ngInject
+ */
+app.outingsCsvDownloadDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'AppOutingsCsvDownloadController',
+    controllerAs: 'csvCtrl',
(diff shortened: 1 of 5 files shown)


c2corg_ui/static/js/outingscsvdownload.js
```diff
@@ -0,0 +1,150 @@
...
+  this.getOutings_().then(function(outings) {
+    var csvFile = '';
+    outings.forEach(function(outing) {
+      var cells = [
+        outing.date_start,
+        outing.locales[0].title,
+        outing.activities.map(function(activity) {
+          return langService.translate(activity);
+        }).join(';'),
+        'https://www.camptocamp.org/outings/' + outing.document_id
```
asaunier: There is probably a way to avoid hard-coded URL here :)

asaunier: On the other hand it's just to add the outing URLs in a CSV file, then no big deal to have some static content here. OK for me then.

c2corg_ui/static/js/outingscsvdownload.js
```diff
@@ -0,0 +1,150 @@
...
+      ];
+      csvFile += '"' + cells.map(function(cell) {
+        if (cell.indexOf('"') > -1) {
+          return cells.replace(/"/g,'""');
```
asaunier: There is probably a typo here: ``cell`` instead of ``cells``.
In the demo the CSV export fails with error:
```
TypeError: cells.replace is not a function
    at http://localhost:6555/static/0dd7605/js/outingscsvdownload.js:84:24
```

PR #1724, opened 2017-06-27
enable pswp history + code cleanup

brunobesson:
when gallery is open, navigating back in browser simply only gallery


asaunier:
> when gallery is open, navigating back in browser simply only gallery

I don't understand what you mean and what this PR does?!

brunobesson:
Try on prod: open the gallery, then click back in your browser. You are directed to the previous page.
With this commit, clicking back button when gallery is open will only close the gallery and stay on the same page.

PR #1725, opened 2017-06-27
Hide action buttons on mobile when editiing a field

brunobesson:
On mobile, action buttons use too much space on screen when editing.
This PR hides them when an input field gains focus.


asaunier:
Interesting! But aren't you afraid that people might get confused when editing a page because the submit buttons are no longer available, in case they don't figure out that buttons are shown back when taping outside of the inputs?

The code is OK for me.

stef74:
if we not decide to continue on this way can we close the PR ?

brunobesson:
I have to think first a little bit. Because I was trying to solve something I find really annoying. But maybe I just did it to eagerly.

PR #1728, opened 2017-06-28
Add elevation profile for outings

brunobesson:
(Note: we use d3 3.5.17, which ngeo depends on)

c2c19
c2c17
c2c18

### DIFF
--- a/c2corg_ui/static/js/elevationprofile.js
+++ b/c2corg_ui/static/js/elevationprofile.js
@@ -1,0 +1,443 @@
+goog.provide('app.ElevationProfileController');
+goog.provide('app.elevationProfileDirective');
+
+goog.require('app');
+goog.require('app.Lang');
+goog.require('ngeo.Debounce');
+
+/**
+ * @return {angular.Directive} The directive specs.
+ */
+app.elevationProfileDirective = function() {
+  return {
+    restrict: 'E',
+    controller: 'appElevationProfileController',
+    controllerAs: 'elevationProfileCtrl',
+    templateUrl: '/static/partials/elevationprofile.html'
+  };
+};
+app.module.directive('appElevationProfile', app.elevationProfileDirective);
+
+/**
+ * @param {angular.Scope} $scope Scope.
+ * @param {?GeoJSONFeatureCollection} mapFeatureCollection FeatureCollection of
+ *    features shown on the map.
+ * @param {app.Lang} appLang Lang service.
+ * @param {ngeo.Debounce} ngeoDebounce ngeo Debounce service.
+ * @constructor
+ * @ngInject
+ */
+app.ElevationProfileController = function(
+  $scope,
+  mapFeatureCollection,
+  appLang,
+  ngeoDebounce
+) {
+  /**
+   * @type {angular.Scope}
+   * @private
+   */
+  this.scope_ = $scope;
+
+  /**
+   * @type {?GeoJSONFeatureCollection}
+   * @private
+   */
+  this.featureCollection_ = mapFeatureCollection;
+
+  /**
+   * @private
+   * @type {ngeo.Debounce}
+   */
+  this.ngeoDebounce = ngeoDebounce;
+
+  /**
+   * @export
+   */
+  this.mode = 'distance';
+
+  this.i18n = {
+    elevation: appLang.translate('Elevation'),
+    elevation_legend: appLang.translate('Elevation (m)'),
+    meters: appLang.translate('meters'),
+    distance: appLang.translate('Distance'),
+    distance_legend: appLang.translate('Distance (km)'),
+    km: appLang.translate('kilometers'),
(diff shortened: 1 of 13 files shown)


c2corg_ui/templates/outing/helpers/view.html
```diff
@@ -304,3 +304,17 @@
     </tbody>
   </table>
 </%def>
+
+<%def name="get_outing_elevation_profile(geometry)">\
+  <div class="view-details-info col-xs-12">
+    <h3 class="heading closed show-phone" ng-click="mainCtrl.animateHeaderIcon($event)"
+        data-toggle="collapse" data-target="#elevation-profile">
+      <span x-translate>Elevation profile</span><span class="glyphicon glyphicon-menu-right"></span>
```
asaunier: You should use ``translate`` instead of ``x-translate``. The latter directive is an alias that should only be used when the content of the element must not be extracted and pushed to Transifex. Which is not what you want here, is it?

c2corg_ui/static/js/elevationprofile.js
```diff
@@ -0,0 +1,443 @@
...
+      return coord.ele > 0;
+    })
+  ) {
+    $('#elevation-profile').closest('.view-details-info').remove();
+    return;
+  }
+  this.createChart();
+};
+
+app.ElevationProfileController.prototype.createChart = function() {
```
asaunier: Perhaps you could add the usual "export" keyword (ditto for the other functions):
```
/**
 * @export
 */
```

asaunier:
Really cool :)
But in my instance the tool does not work and I see the following error in my debugger:
> TypeError: Cannot read property 'geometry' of undefined
    at new app.ElevationProfileController (elevationprofile.js:73)

c2corg_ui/templates/i18n.html
```diff
@@ -535,3 +535,14 @@
...
+## elevation profile
+<span translate translate-comment="Elevation title for mouseover in elevation profile">Elevation</span>
+<span translate translate-comment="Legend in elevation profile">Elevation (m)</span>
+<span translate translate-comment="Elevation unit for mouseover in elevation profile">meters</span>
+<span translate translate-comment="Distance title for mouseover in elevation profile">Distance</span>
+<span translate translate-comment="Legend in elevation profile">Distance (km)</span>
+<span translate translate-comment="Distance unit for mouseover in elevation profile">kilometers</span>
+<span translate translate-comment="Time title for mouseover in elevation profile">Time</span>
+<span translate translate-comment="Duration title for mouseover in elevation profile">Duration</span>
+<span translate translate-comment="Alternate legend in elevation profile">Duration (hrs)</span>
```
asaunier: Perhaps we don't need those "fake" translate strings since they are already made available in the JS directive file (I think the extractor extracts strings in ``translate()`` functions). Not sure it is possible to have "translate-comments" in the JS code but do we really need them here? The strings are quite straightforward.

By the way, do you really need to have almost duplicate strings such as "Elevation" / "Elevation (m)"? Isn't it possible to have the same string for both locations? You could also check in Transifex that some similar strings do not already exist. For instance "elevation" is already available.

asaunier:
> But in my instance the tool does not work and I see the following error in my debugger:

I have rebuilt the branch on my instance and made further testing.
It seems to work perfectly with outings actually having a GPS track (``geom_detail``).
But I get the error above (`` Cannot read property 'geometry' of undefined``) for outings with no track.

There is probably a check missing to show the profile only for outings with an actual ``geom_detail`` attribute set.

asaunier:
I have rebased the PR and add commit 673894a that:
* makes sure the directive is only called for outings with a GPS track
* removes the hard-coded translation strings in template ``i18n.html`` (the strings are marked as to be translated and extracted directly in the JS file using a ``gettext()`` function)
* fixes some coding standard errors, especially with some private properties and functions (I have not renamed all private props, that should have a trailing ``_`` but that would have made sense - nevermind that's quite OK like this now)
* extended the support to ``MultiLineString``s and not only ``LineString``s (my GPX files are all multilinestrings I think).

asaunier:
@brunobesson OK to merge the PR with my additions? The errors in the API that made the tests fail are now fixed.

asaunier:
@brunobesson Any objection against merging the PR as is?

brunobesson:
I haven't got any time recently to have a look at anything, sorry. I hope
i'll find some next week. In the meantime feel free to do as you think
best, I trust you ;-)

Le 24 août 2017 11:42 AM, "Alex Saunier" <notifications@github.com> a
écrit :

asaunier:
@brunobesson I recommend to go forward and merge the PR. My changes are only fixes of what you have done, not a major redesign. It's not reasonable to have pending PRs for too long, especially with all the layout changes that are coming.

> feel free to do as you think best, I trust you

Thanks :) I will merge a bit later unless you want to do it yourself :)

Issue #1072, opened 2016-12-02
Date picker layout


brunobesson:
Buttons are not placed properly, bottom margin is incorrect

![toto](https://cloud.githubusercontent.com/assets/2234024/20828053/e519c96c-b875-11e6-8237-26dd471063e6.png)

brunobesson:
Also, why use 3 different colors for buttons? And these colors do not seem to respect the color palette?

ginold:
Those are the default values and color. But you are completely right there should be a margin and a different button style, it annoys me too :P

brunobesson:
See also #1135

PR #1734, opened 2017-06-29
enhance datepicker CSS

brunobesson:
also cleanup some duplicated or incorrect stuff for button

was
toto

now
c2c22


brunobesson:
See #1072

Issue #971, opened 2016-11-23
Default value of quality dropdown in editing forms


asaunier:
As for now the quality dropdown is set to a blank value when creating a new document (the API converts it to "draft" I think).

It would make more sense that the default quality is directly displayed in the dropdown ("draft" I think).
Dropping this line should do the job:
https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/helpers/edit.html#L56

desnoes:
@asaunier @brunobesson : i did not understand the objective of this PR. I cannot test it.

desnoes:
@asaunier: thanks for the explanation. It works fine.

PR #1741, opened 2017-06-30
use default draft value when creating new document

brunobesson:
See #971

@asaunier do you think we should also create an alembic script to set a default value for all documents that don't have any yet?

### DIFF
--- a/c2corg_ui/static/js/editing/documentediting.js
+++ b/c2corg_ui/static/js/editing/documentediting.js
@@ -666,7 +666,7 @@ app.ConfirmSaveController = function($uibModalInstance, appDocument, appLang) {
    * @type {string}
    * @export
    */
-  this.quality = appDocument.document.quality;
+  this.quality = appDocument.document.quality || 'draft';
 
 };
 
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -99,9 +99,7 @@
       ## QUALITY
       <div class="full comment" ng-class="{'has-success': saveCtrl.quality}">
         <label translate>quality</label>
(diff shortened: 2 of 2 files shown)


asaunier:
> do you think we should also create an alembic script to set a default value for all documents that don't have any yet?

I think ``quality`` is set for all existing documents, default being ``draft``.

brunobesson:
>I think quality is set for all existing documents, default being draft

I am quite sceptical about this, because I have seen cards in the feed with no quality set.

asaunier:
> I have seen cards in the feed with no quality set in prod.

Do you have examples?

``quality`` actually defaults to ``draft`` according to https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L60-L61

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -666,7 +666,7 @@ app.ConfirmSaveController = function($uibModalInstance, appDocument, appLang) {
    * @type {string}
    * @export
    */
-  this.quality = appDocument.document.quality;
+  this.quality = appDocument.document.quality || 'draft';
```
asaunier: A small downside of having an hard-coded default quality will prevent to have it updated if the list of qualities changes in https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L601-L607

A cleaner solution would have been to retrieve this value from the config file above. But I admit it would have been quite awkward because it would probably imply to inject this value through a module constant passed in the mako template as in https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L137-L151

It's probably unrequired complexity, then OK for me.

asaunier:
@brunobesson Feel free to merge

Issue #1528, opened 2017-02-24
when filling non textual fields, typing "enter" should not open "completude" popup


mgerbaux:
It is the case for outings or routes, but not for articles.
It is especially destabilizing for auto-completion fields on which newbies are tempted tempted to validate their keyboarding

brunobesson:
Could you more explicit? I understand the global meaning of the issue, but could you point to both a working and a non-working examples?

mgerbaux:
> Could you more explicit?
hardly…
Try it ? 
![screenshot at 2017-02-24 13 02 57](https://cloud.githubusercontent.com/assets/22116605/23303001/be67711c-fa91-11e6-83c6-066ba0030742.png)
(I have typed the whole summit name then enter before autocompletion propose anything)

asaunier:
Perhaps related to this issue https://github.com/c2corg/v6_ui/issues/1293 ?

brunobesson:
That hitting enter submits the form in some cases is not really a problem, this is a standard web feature. What I don't understand, is why the completion modal shows up while some fields are invalid. Can you submit the form when clicking "Save" in the dialog then? I guess not.

Thus, @asaunier guess might be the correct one.

asaunier:
> It is the case for outings or routes, but not for articles

Indeed the behaviour is observed with all document types except articles and books, which are the only documents with no map. I have tried to remove the maps in the editing forms of waypoints and outings: the problem no longer occurs.

Then I am pretty sure this issue is a duplicate of https://github.com/c2corg/v6_ui/issues/1293 and could be closed. I have ported the behaviour described here to https://github.com/c2corg/v6_ui/issues/1293#issuecomment-302407447

brunobesson:
>Then I am pretty sure this issue is a duplicate 

I agree. Closing

Issue #1293, opened 2017-01-03
Edit page : conflict between map and modal box


fbunoz:
On the edit page of a WP, route or outing, when we click on the "Center on my position" button on the map, it open the summary modal box (with quality choice, save button, etc).

brunobesson:
I assume this is because the action would like to rewrite the location (adding coordinates) and thus submits the form?

asaunier:
> when we click on the "Center on my position" button on the map, it open the summary modal box

Same behaviour when pressing the ENTER key in a text input (except textareas) as described in the (probably related) issue https://github.com/c2corg/v6_ui/issues/1528 : 
* the map is recentered on the user's current position (as if the "center on my position" button was clicked)
* the quality modal box is shown

asaunier:
See #1743 for investigations

PR #1743, opened 2017-06-30
prevent showing quality dialog if form is invalid

brunobesson:
There might be some work to cleanup code for this area + handle #1528

### DIFF
--- a/c2corg_ui/static/js/editing/documentediting.js
+++ b/c2corg_ui/static/js/editing/documentediting.js
@@ -544,7 +544,7 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
       }
     }
   }
-  if (missing['data']['errors'].length > 0) {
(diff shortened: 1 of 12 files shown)


asaunier:
There might be something wrong: if I try to edit an existing outing, the "Save" button is enabled (no missing property) but clicking on it always causes the "Invalid form" alert to be shown (not possible to save the changes).

Concerning issues https://github.com/c2corg/v6_ui/issues/1528 (quality popup shown when pressing Enter while the focus in a text input) and https://github.com/c2corg/v6_ui/issues/1293 (quality popup shown when clicking on the map "recenter on my location" button), the same behaviour is observed with this PR: "invalid form" alert shown in both cases.

stef74:
@brunobesson what is the status of this PR ?

brunobesson:
I need to have a look again on it. It's quite old, and @asaunier tests showed I didn't solve the issue properly.

PR #1754, opened 2017-07-19
Fix conflicts between map geolocation and editing forms

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1293 and https://github.com/c2corg/v6_ui/issues/1528
See also pending PR https://github.com/c2corg/v6_ui/pull/1743

What I have observed:
* pressing the ENTER key in a text input triggers the editing form ``ng-submit`` (as a result it shows the quality modal box)
* clicking on the geolocation button also triggers the editing form ``ng-submit`` etc.

If we replace the ``ng-submit`` by an ``ng-click`` on the "Save" button, pressing ENTER no longer submit the form (as asked by https://github.com/c2corg/v6_ui/issues/1528). It also prevents the form from being submitted when clicking the recenter button.

Adding an explicit ``type="button"`` attribute to the geolocation button ("recenter on my current position") was necessary to prevent the button handler to be triggered when pressing the ENTER key while in an input field of the form.

### DIFF
--- a/c2corg_ui/static/js/editing/documentediting.js
+++ b/c2corg_ui/static/js/editing/documentediting.js
@@ -544,7 +544,7 @@ app.DocumentEditingController.prototype.hasMissingProps = function(doc, showErro
       }
     }
   }
-  if (missing['data']['errors'].length > 0) {
+  if (missing['data']['errors'].length) {
     this.alerts.addError(missing);
   }
   return hasError;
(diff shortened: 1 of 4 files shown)


asaunier:
Anyone willing to review this PR? :P

brunobesson:
>Adding an explicit type="button" attribute to the geolocation button ("recenter on my current position") was necessary

Yes, default type for a button is 'submit'

PR #1760, opened 2017-07-24
Fix incorrect route history placeholder

asaunier:
The string in Transifex has no trailing whitespace. Which probably explains why this string is not translated in the UI despite it "exists" and is translated in Transifex.

### DIFF
--- a/c2corg_ui/templates/route/edit.html
+++ b/c2corg_ui/templates/route/edit.html
@@ -809,7 +809,7 @@ updating_doc = route_id and route_lang
               <textarea app-markdown-editor name="route_history"
                         ng-model="route.locales[0].route_history"
                         class="form-control"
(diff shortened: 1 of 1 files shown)


c2corg_ui/templates/route/edit.html
```diff
@@ -809,7 +809,7 @@ <h2 class="heading show-phone" translate>Comments</h2>
               <textarea app-markdown-editor name="route_history"
                         ng-model="route.locales[0].route_history"
                         class="form-control"
-                        placeholder="{{'Describes here historical information about the route (date, names..) ' | translate}}"></textarea>
+                        placeholder="{{'Describe here historical information about the route (date, names..)' | translate}}"></textarea>
```
tsauerwein: For me "Describe historical information about the route (date, names..) here" would sound more natural.

asaunier: Right, updated!

Issue #1745, opened 2017-07-05
The button 'Hide list' is inactive


desnoes:
When clicking on the button 'Hide list', nothing happends. This is observable in the outing and topoguide home pages.

PR #1762, opened 2017-07-25
Revert change 14cc7f4 because it breaks map/list hidding switches

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1745

Replacing ``ng-show`` by ``ng-if`` in 14cc7f4 seems to break the map/list hidding switches.
Perhaps the explanation is here https://stackoverflow.com/a/21870119
> ng-if will remove elements from DOM. This means that all your handlers or anything else attached to those elements will be lost. For example, if you bound a click handler to one of child elements, when ng-if evaluates to false, that element will be removed from DOM and your click handler will not work any more, even after ng-if later evaluates to true and displays the element. You will need to reattach the handler.

I have not reverted the whole commit 14cc7f4 because the buttons seem to work as expected with the suggested change. More testing might be needed.


brunobesson:
Sorry for the bug introduced...

PR #1767, opened 2017-07-30
Spread the disable comments option over the full editing form width

asaunier:
Before:

capture d ecran 2017-07-30 a 14 59 21

After:

capture d ecran 2017-07-30 a 15 03 06

This is already the case for in xreports editing forms.


asaunier:
A review anyone? :)

PR #1773, opened 2017-08-10
Cotometer menu changes

asaunier:
The cotometer is a nice tool but I feel a bit "unhappy" because:

* the cotometer button icon is not very intuitive in the route editing form:
capture d ecran 2017-08-10 a 17 01 38
=> I doubt that users will easily understand they can open the cotometer by clicking on this icon. I suggest to rather use the question mark icon:
capture d ecran 2017-08-10 a 17 02 47

=> commit 67879b6

* there is a "cotometer" entry in the "Plus" menu of any document (whatever its type, including WP or images for instance - by the way clicking on the "Cotometer" menu in those pages does nothing) => that does not look very consistent :P 
In addition I don't think this entry is relevant here because in the "Plus" menu other entries are document-related tools (history, merging, deleting, other languages, translations), they does not deal with the content itself. In addition, as a user, I don't find it intuitive to look at a document "Plus" menu to get an assistant about skitouring route ratings => OK when I have to provide the rating (editing form), but not really in the viewing page. Or maybe in a dedicated page, without having to open whatever document to find it :P 

As a result I suggest to remove this entry from the Plus menu: commit fd92529

### DIFF
--- a/c2corg_ui/templates/helpers/view.html
+++ b/c2corg_ui/templates/helpers/view.html
@@ -621,7 +621,6 @@
       </div>
       <ul class="dropdown-menu dm-icon" role="menu">
         <li><a ng-click="detailsCtrl.printPage()"><span class="glyphicon glyphicon-print"></span> <translate>Print page</translate></a></li>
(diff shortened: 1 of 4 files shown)


Courgetteandratatouille:
good idea :) i'am agree

Issue #1680, opened 2017-06-08
Improve visibility of document translations


asaunier:
As for now other linguistic versions of a document are available via the "Plus" menu in document view pages. The visibility of this action is pretty poor though.

Perhaps we could add links to the other linguistic version at the top of the document?

Should probably be considered along with the document view page organisation https://github.com/c2corg/v6_ui/issues/1196

asaunier:
The current position of other lang versions is definitely not OK, people not noticing this possibility, rather trying to switch the language using the interface language selector in the top right corner.

Here are feedbacks a dutch user just sent to the dev mailing list:
> * When I google for routes, I only find the french variant of camptocamp (previously this was not the case). It seems that the english variations just don't exist any more?
> * When I change FR to EN settings on the page  (right on top) of a route itself it is not directing me to the EN site.
> * All friends of mine have the same problem. I am wondered that you didn't know about this..

desnoes:
works correctly with routes, waypoints and articles

PR #1774, opened 2017-08-10
Show available langs directly in the document pages

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1680

The layout of the lang links are still a bit raw but I have not much inspiration to make them nicer (patch/help welcome :P ). The available langs entry in the Plus menu has been removed.

@Courgetteandratatouille this might impact your work about the page layout redesign.

capture d ecran 2017-08-10 a 22 16 37

### DIFF
--- a/less/viewdetails.less
+++ b/less/viewdetails.less
@@ -606,6 +606,10 @@
       color: @C2C-orange
     }
   }
+
+  .other-langs {
+    clear: both;
+  }
 } /*.view-details-section*/
 
 .view-details-map {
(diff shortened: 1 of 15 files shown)


asaunier:
@brunobesson @Courgetteandratatouille Looks good to you?

less/viewdetails.less
```diff
@@ -606,6 +606,10 @@
       color: @C2C-orange
     }
   }
+
```
stef74: Why a blank line ?

asaunier: To be consistent with the other blocks? Except maybe ``.area-associations`` just above, there are blank lines before selector blocks.

asaunier:
@stef74 Thanks for the review. Merging :)

PR #1775, opened 2017-08-11
Improve loading time

asaunier:

As for now, an ``ng-cloak`` directive is set on the main ``<html>`` tag, to make sure the page is shown only when all angular related parts are ready. The downside is that it delays the rendering of the pages, making them appearing slower.

This PR finetunes a bit the use of ``ng-cloak`` to make the page being displayed even if some secondary parts are not ready yet, making the page loading faster.

I have also removed the white shade that is generally shown when the loading directive is enabled (including when the page was loading): I don't think this white shade is really useful and it makes the pages flashing.

An improvement I would like to add as well (but have not succeeded in doing so yet) would be to use a loading bar at the top of the page, as in github pages, eg. https://github.com/c2corg/v6_ui
See https://github.com/chieffancypants/angular-loading-bar

An incomplete PR already exists, see https://github.com/c2corg/v6_ui/pull/963 I have tried to port it to the current PR, but at the moment with no effect (perhaps because of the ``ng-cloak``). See https://github.com/c2corg/v6_ui/pull/1776

### DIFF
--- a/c2corg_ui/static/js/loading.js
+++ b/c2corg_ui/static/js/loading.js
@@ -24,10 +24,8 @@ app.loadingDirective = function($http) {
        }, function() {
          if ($http.pendingRequests.length > 0) {
            $('.loading-gif').show();
-           el.addClass('loading');
          } else {
            $('.loading-gif').hide();
-           el.removeClass('loading');
          }
        });
      }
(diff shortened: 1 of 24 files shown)


asaunier:
The PR is ready for review, the loading bar could be added later if relevant.

asaunier:
I will merge the PR today if no one objects or ask for a delay.

PR #1787, opened 2017-09-12
Contextual help for edition of routes

desnoes:


asaunier:
@desnoes I will have a look a but later but I already notice something wrong: the fiest 3 commits of this PR have already been merged to master: https://github.com/c2corg/v6_ui/commit/6f952611c5000d0073846f2893b0b54563dedb14
Only the last commit is relevant, right?

I think the problem is that you have reused your old "contextual_help" branch. Which is not correct.
You have to start from an up-to-date copy of the master branch:
```
git checkout master
git pull
```
and then you create a new branch (using another branch name is probably better):
```
git checkout -b routes_contextual_help
```
before adding/committing your changes and pushing the branch to github.

I can fix the PR if you want.

2 other comments:

It's not critical but it would make sense to add intermediary ``route/``, ``xreport/``, ``outings/``, etc. directories in ``https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/static/partials/contexthelp`` instead of having a bunch of ``route-....html``, ``outing-....html`` files directly at the root of the directory. A bit like for the templates: https://github.com/c2corg/v6_ui/tree/master/c2corg_ui/templates
Eg. ``c2corg_ui/static/partials/contexthelp/route-equipment_rating.html`` would be ``c2corg_ui/static/partials/contexthelp/route/equipment_rating.html``

I am a bit concerned by the number of contextual help partials. It's probably not that critical either but we should try to have as less partials (and as small) as possible because all those files are concatenated altogether along with the JS files into a "big" JS file that users download when they first open the site (even though they never open the contextual help contents) => if this built JS file grows to big, the site will become longer to load and browser memory use will increase.
If there are too many contextual help partials, I would recommend to make them served on demand only by a UI: it would mean that when opening the contextual help modal box, a call to the UI would ask for the HTML content to display.

asaunier:
@desnoes I have removed the extra commits and rebased (updated) the branch.
I have also noticed errors in your change and added a fixup commit 72ee7f9
For instance:
* try reusing strings as much as possible, this way we reduce the number of strings to be translated
* avoid adding HTML in the translations. It's OK but this might make translations harder to do for translators with no HTML skills.
* you don't need to use HTML escaping (eg. ``&deg;`` instead of ``°``) since we use HTML5 which does not require an XML syntax (as does XHTML - which we don't use).

If you want to test the updated branch in your computer:
```
git fetch
git reset --hard origin/contextual_help
```
Be careful: the last command makes a "hard reset", which means it fully replaces your current branch by a copy of the "origin/contextual_help" which has been updated from github by the "git fetch" command.

Feel free to merge the PR if my changes are OK for you.

desnoes:
@asaunier : many thanks for your help. I will propose a commit to reorganize files in folders.

asaunier:
@desnoes There are now a bunch of not translated strings in Transifex (230 for French) :P 
By the way I see strings that should probably not be translated. For instance

capture du 2017-09-15 10-38-18

It would make sense that you had a look at the new strings in Transifex and check if there are no duplicates or strings that don't really need to be translated (changes in the contextual help partials are then needed).

desnoes:
@asaunier : you are right!
For the translation, i have stored in an excel file all the traductions retrieved from i18n in V5. I guess this work will be cumbersome but hope it could be done quite fast.
I will check in files if there are unnecessary strings that don't need to be translated.

Issue #1137, opened 2016-12-08
Missing recentering tools in mobile maps


stef74:
On mobile we don't have recentring tools on the map.

ginold:
you sure you clicked on the map button on the bottom of the page? or is it another page you are speaking of?
![screenshot from 2016-12-08 12-22-20](https://cloud.githubusercontent.com/assets/7814311/21008213/03fd7aac-bd41-11e6-807d-d51b3d25d486.png)

stef74:
My screen (iphone5s) 
![img_0867](https://cloud.githubusercontent.com/assets/3962997/21008543/757fb3a6-bd42-11e6-846e-013fa3264ceb.PNG)

stef74:
@ginold it s Word on m'y iPad mini not on m'y iPhone , size screen issue ?

ginold:
I don't understand 

>  it s Word on m'y iPad mini 

You are browsing with Word (microsoft) ? :confused:  i think not. 
All in all, it's surely not a screen size issue.

brunobesson:
I think I saw someone reporting something similar: apparently on some devices buttons are placed on an upper position and hidden by the header.

brunobesson:
E.g. on my phone, I get the following result: 
![screenshot_2016-12-24-08-39-30](https://cloud.githubusercontent.com/assets/2234024/21465867/ba48d0fc-c9b4-11e6-9805-9d24833e286a.png)

stef74:
Word = works. It s probably an issue of size if screen ...

stef74:
I can show it in chrome on IOS not on safari.

stef74:
Same issue on iphone 7s (screen are more big than 5)

fjacon:
So it is just a problem on safari, right ?

dkocich:
> So it is just a problem on safari, right ?

Yes, I see this problem only on iPhone (portrait view) with BrowserStack.

A problem is probably with the use of calc in CSS http://caniuse.com/#feat=calc and VH http://caniuse.com/#feat=viewport-units (known issues for Safari). Some styles are not applied the same way and then there are some differences on some places in different browsers and versions. I am trying to solve this using media queries to target only iPhone's safari and apply different styles but I am not sure if it is a good way.
http://stephen.io/mediaqueries/
https://css-tricks.com/snippets/css/media-queries-for-standard-devices/

asaunier:
Tested in browserstack.com (emulator):
[...]

PR #1803, opened 2017-11-02
fix 1137

julsbreakdown:
#1137 
Works on android chrome, I haven't tested it on safari
for testing i had to use my one phone since classic debug tools do not render the google search tab cf image
capture du 2017-11-02 16-46-04

### DIFF
--- a/less/c2corg_ui.less
+++ b/less/c2corg_ui.less
@@ -380,6 +380,7 @@ div.help-block {
     opacity: 0;
     pointer-events: none;
     width: 100%;
+    height: ~"-webkit-calc(100% - 120px)";
   }
   .map {
     @media @phone {


less/c2corg_ui.less
```diff
@@ -380,6 +380,7 @@ div.help-block {
     opacity: 0;
     pointer-events: none;
     width: 100%;
+    height: ~"-webkit-calc(100% - 120px)";
```
asaunier: Safari is based upon Webkit, right? Then it's probably concerned by this change.
Chrome is no longer using it since 2013 (OK it uses a fork I think), then will this ``-webkit-calc`` really impacts Chrome?

julsbreakdown:
I do not think that browserstack enables you to add the search tool bar.

less/c2corg_ui.less
julsbreakdown: I honestly don't know much about Safari
As I said i've tried it using the lastest chrome version on my android.
https://developers.google.com/web/tools/chrome-devtools/remote-debugging/

Adding this in the CSS results in fixing the bug

PR #1816, opened 2017-11-13
Add view sso_login

arnaud-morvan:
As authentication is stored as JWT token in local storage, we need to pass through UI authCtrl to call API sso_login.

Needs https://github.com/c2corg/v6_api/pull/706

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -391,6 +391,20 @@ app.Api.prototype.login = function(data) {
 
 
 /**
+ * @param {Object} data
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.ssoLogin = function(data) {
+  var promise = this.postJson_('/sso_login', data);
+  promise.catch(function(response) {
+    var msg = this.alerts_.gettext('SSO login failed:');
+    this.alerts_.addErrorWithMsg(msg, response);
+  }.bind(this));
+  return promise;
+};
+
+
+/**
  * @param {string} lang
  * @return {!angular.$q.Promise<!angular.$http.Response>}
  */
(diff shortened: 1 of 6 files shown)


c2corg_ui/templates/sso_login.html
```diff
@@ -0,0 +1,6 @@
+<%inherit file="./base.html"/>
```
asaunier: Since this is is a dummy page, you should probably make sure it is not indexed by adding:
```
<%block name="metarobots"><meta name="robots" content="noindex,follow"></%block>
```
See the default value at https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L29

c2corg_ui/templates/sso_login.html
```diff
@@ -0,0 +1,6 @@
...
+
+<div app-auth class="auth-form" ng-init>
```
asaunier: Is this ``ng-init`` required?

c2corg_ui/static/js/apiservice.js
```diff
@@ -391,6 +391,20 @@ app.Api.prototype.login = function(data) {
 
 
 /**
+ * @param {Object} data
+ * @return {!angular.$q.Promise<!angular.$http.Response>}
+ */
+app.Api.prototype.ssoLogin = function(data) {
+  var promise = this.postJson_('/sso_login', data);
+  promise.catch(function(response) {
+    var msg = this.alerts_.gettext('Login failed:');
```
asaunier: Perhaps you could use a more specific error message (``SSO login failed:``?) to make it more obvious if something wrong happens?

c2corg_ui/templates/sso_login.html
arnaud-morvan: Done

arnaud-morvan:
@asaunier : Thanks for review, note that I do not have written tests that includes javascript part, only tested manually.

PR #1821, opened 2017-11-17
better rendering on css print, fix some useless space and indentation

Courgetteandratatouille:
It's not perfect but I guess it's ok for route waypoint and outing.

We need more work for books areas  etc... later :)


asaunier:
Shouldn't you have added the new styles to the less files dedicated to the printing: https://github.com/c2corg/v6_ui/tree/master/less-print ?

Perhaps we could also clean this less-print files from the old styles deprecated by the new layout?

config/demo
```diff
@@ -2,13 +2,13 @@ include Makefile
 
 export instanceid = ui
 export base_url = /
-export api_url = https://api.camptocamp.org
-export api_url_internal = https://api.camptocamp.org
-export api_url_host = api.camptocamp.org
+export api_url = http://api.demov6.camptocamp.org
```
asaunier: Why switching to http? The demo should use https URLs. Ditto for the discourse URL below.
Anyway this file no longer used when updating the demo, everything is done using env vars set in the ``docker-compose.yml`` file.

asaunier:
Please rebase your PR when fixing commits, see https://github.com/c2corg/v6_ui/wiki/Git-Cookbook#managing-commits

Courgetteandratatouille:
It's a fail, i made the correction soon and i rebase the commit. After this it's ok ?

config/demo
```diff
@@ -6,7 +6,7 @@ export api_url = http://api.demov6.camptocamp.org
 export api_url_internal = http://api
 export api_url_host = api
 
-export discourse_url = https://forum.demov6.camptocamp.org
+export discourse_url = http://forum.demov6.camptocamp.org
```
asaunier: Why do you need this? Perhaps you have forgotten to undo this change?

Courgetteandratatouille: indeed it's a forget, all the change in this file are an error.

asaunier: OK, please undo that change as well and I think it's OK.

PR #1827, opened 2017-11-23
unshow image header

stef74:
unshow img on header


stef74:
ready for review (easy)

asaunier:
My **personal** opinion: I would keep the image headers, at least a few weeks, to figure out if they really cause problems, how they could be solved and the feedbacks of users in the real website.

Issue #1831, opened 2017-11-27
title emphasis


cbeauchesne:
Handle this pattern

    ### title # emphasis

stef74:
ready for testing = on the demo server , it is the case ?

cbeauchesne:
Oups, no sorry, I've missunderstood the meaning of this label.

asaunier:
I have added the "ready for testing" label because it will soon be available on the demo (anticipating a bit).
When a PR has been merged, the related issue should be tagged "ready for demo" (this way we know that it should become "ready for testing" when the demo is updated).

cbeauchesne:
ok

PR #1832, opened 2017-11-27
Header emphasis

cbeauchesne:
Feature #1831


cbeauchesne:
I've done a squash&merge, is it a good solution?

asaunier:
> I've done a squash&merge, is it a good solution?

It's OK. But I think it is better to have well organised PR before reviewing and merging.
Next time don't forget to use "git commit --amend" or "git rebase -i" etc. as described in https://github.com/c2corg/v6_ui/wiki/Git-Cookbook#managing-commits Thanks :)

PR #1839, opened 2017-11-30
Fix design new design issue

stef74:
- [x] fix area size

- [x] fix activity size (for more than 2 act)

- [x] highligh card when select an object on map (search)

### DIFF
--- a/less/cards.less
+++ b/less/cards.less
@@ -1,6 +1,16 @@
 .list.advanced-search {
   margin-top: 10px;
 }
+
+.hovered-card() {
+  background-color: #EFEFEF;
+  transition: 0.1s ease-in-out;
+}
+
+.highlight .post {
+  .hovered-card;
+}
+
 .post {
   -webkit-box-shadow: 0 1px 4px 0 rgba(0,0,0,0.14);
   box-shadow: 0 1px 4px 0 rgba(0,0,0,0.14);
@@ -16,6 +26,10 @@
   margin-bottom: 15px;
   outline: none;
   cursor:pointer;
+  
+  &:hover  {
+    .hovered-card;
+  }
 
   &.card-outing {
     .infoa {
@@ -242,7 +256,7 @@
 
       .parent-activity {
         display: flex;
-        width: 80px;
+        width: 90px;
         overflow: hidden;
         height: 30px;
       }
@@ -443,7 +457,7 @@
           .parent-infop {
             width: 20%;
             &.parent-infop-first {
-              width: 70px;
+              width: 50%;
             }
           }
         }
@@ -590,8 +604,6 @@
             width: 35%;
           }
         }
-
-
       }
       &.card-outing {
         .inlist {
@@ -624,16 +636,15 @@
         .outing-icon-infos {
           width:  150px;
         }
-        .area  {
-          width:  10%;
-          max-width: 120px;
+        .area {
+          width:  20%;
+          max-width: 500px;
         }
         
         .large {
           width:  25%;
-          max-width: 250px;
+          max-width: 500px;
         }
-        
       }
       &.card-waypoint {
         .bquality{
@@ -656,7 +667,7 @@
         .parent-infop {
           width: 20%;
           &.parent-infop-first {
-            width: 70px;
+            width: 50%;
           }
         }
       }
@@ -792,19 +803,12 @@
         display: none;
       }
     }
-  }
-}
-
-@media @big {
-  .area {
-    width:  20%;
-    max-width: 600px;
-  }
-}
-@media @verybig {
-  .large {
-    width:  20%;
-    max-width: 600px;
+    .post.card-outing {
+      .area {
+        width:  10%;
+        max-width: 120px;
+      }
+    }       
   }
 }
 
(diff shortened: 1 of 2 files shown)


stef74:
@asaunier I'm blcoking on hilight the card when have focus on map object ... m'y undetrsand is use hilight on .post class ... but not working

brunobesson:
@stef74 could you more explicit on what you are trying to solve for someone that couldn't follow the discussions and needs an update?
I might be able to help you

stef74:
I try to highlight thé card When hover élément on map ( same behavior thaï previous design)

brunobesson:
The `highlight` class is applied to the div class wrapping the card itself.

Instead of

    &:hover, &.highlight {
      background-color: #EFEFEF;
      transition: 0.1s ease-in-out;
      cursor: pointer;
    }

you should use only

    &:hover {
      background-color: #EFEFEF;
      transition: 0.1s ease-in-out;
      cursor: pointer;
    }

and put something link the following aside

    .highlight .post {
      background-color: #EFEFEF;
      transition: 0.1s ease-in-out;
      cursor: pointer;
    }

Maybe some mixin could be used to factorize the code.

brunobesson:
@stef74 is it OK for you with the explanations I provided?

stef74:
I have not found time to try ... lot of work and travel ... and lot of snow
!

2017-12-09 18:14 GMT+01:00 Bruno Besson <notifications@github.com>:

stef74:
@brunobesson it's work. Ready for review

less/cards.less
```diff
@@ -443,7 +462,7 @@
           .parent-infop {
             width: 20%;
             &.parent-infop-first {
-              width: 70px;
```
brunobesson: Could you explain where this applies? I couldn't find.

less/cards.less
```diff
@@ -784,27 +802,20 @@
     }
   }
 }
-
+
```
brunobesson: useless spaces, please remove

less/cards.less
```diff
@@ -625,15 +644,14 @@
           width:  150px;
         }
         .area  {
```
brunobesson: why not removing the duplicated space before the bracket in the meantime? :wink:

less/cards.less
```diff
@@ -625,15 +644,14 @@
...
-          width:  10%;
-          max-width: 120px;
+          width:  20%;
```
brunobesson: two spaces between value

brunobesson: Concerning the area modifications, I wasn't able to understand where your modifications apply either, sorry (still this mess in the CSS :sob:). Could you provide more details or some screenshots of before/after?

less/cards.less
```diff
@@ -16,6 +23,18 @@
   margin-bottom: 15px;
   outline: none;
   cursor:pointer;
+  
+  &:active {
```
brunobesson: What is this for? I see no difference when clicking the item, because there is already no outline for `.post` on line 17/24

less/cards.less
```diff
@@ -16,6 +23,18 @@
...
+    -webkit-transition: .2s ease-in-out;
+    transition: 0.2s ease-in-out;
+    outline: none;
+  }
+  
+  &:hover  {
+    background-color: #EFEFEF;
```
brunobesson: Didn't like the mixin suggestion?
I think it would improve maintenance and readability of the code. Something like

    .hovered-card() {
      background-color: #EFEFEF;
      transition: 0.1s ease-in-out;
    }

    &:hover {
      .hovered-card;
    }
    
    .highlght .post {
      .hovered-card;
    }

Also, why is the cursor set to pointer hen hovering and highlight class is applied? Either set it only when hovered, or always.

less/cards.less
stef74: Copy past methology ... I have removed this block

less/cards.less
stef74: fix

less/cards.less
stef74: in list format they permit to show 2 activity  and the number of other activity

less/cards.less
stef74: I don't know I can do this ;-)

less/cards.less
stef74: It's for list to equilibrate the content. i provide some screenshot

stef74:
capture d ecran 2017-12-10 a 12 04 21
capture d ecran 2017-12-10 a 12 04 18

stef74:
@brunobesson better now ?

less/cards.less
brunobesson: Isn't it the goal of line 245/259?

less/cards.less
```diff
@@ -1,6 +1,16 @@
 .list.advanced-search {
   margin-top: 10px;
 }
+
+.hovered-card() {
+  background-color: #EFEFEF;
+  transition: 0.1s ease-in-out;
+}
+
+.highlght .post {
```
brunobesson: there seem to be a typo: `.highlght` > `.highlight`.
Have you checked your modifications?

brunobesson:
Mmh. OK, maybe I haven't understood everything also because I haven't tested the list view! It would explain the parts I was missing.

When your PR is ready, don't forget to rebase before merging, thanks :stuck_out_tongue:

PR #1846, opened 2017-12-01
Hide inner scrollbar on mobile

### LINKED ISSUE #1845: Viewdoc content should hide scrollbars

brunobesson:
(and by the way use less variable for color)


asaunier:
Could you use more explicit PR and commit titles?

brunobesson:
There you go.
As this is an really simple change, thought a link to the corresponding ticket would be enough.

PR #1852, opened 2017-12-04
Fix user cards in outings and xreports edits

### LINKED ISSUE #1848: JS error when edit xreport
To be investigated

snapshot2

brunobesson:
* add missing information in user cards (lang)
* add auth user only when creating new outing
* rapid fix in order to have user cards - some more love needed though

fixes #1848

### DIFF
--- a/c2corg_ui/static/js/editing/outingediting.js
+++ b/c2corg_ui/static/js/editing/outingediting.js
@@ -81,11 +81,18 @@ app.OutingEditingController = function($scope, $element, $attrs, $http,
       );
     }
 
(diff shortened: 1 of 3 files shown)


stef74:
capture d ecran 2017-12-14 a 20 24 51


capture d ecran 2017-12-14 a 20 22 41

brunobesson:
>(lby adding empty card)

True. In another PR I think I have solved it, but I'll do it directly in this one then.

As for the design, yes, I explained in the commit that it needs some love. But it's always better than the current state!

Also, to test this PR, you should have the dev console open and check that there are no dubious JS errors occuring...

Issue #1790, opened 2017-09-15
Only the first segment of GPX tracks are imported


asaunier:
When uploading a GPX track with several segments (for instance if the GPS has been turned off while recording), all segments are shown on the document editing form map but only the first segment is actually sent to the API.

Most likely because of https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/static/js/editing/documentediting.js#L428

It would make sense to concatenate segments instead so the the whole track is sent to the API and saved to the document.

Simplified example of multi-segment GPX file:
```
<?xml version="1.0" encoding="UTF-8"?>
<gpx xmlns="http://www.topografix.com/GPX/1/1" version="1.1" creator="MyGPSFiles" xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <trk>
    <name>Bellegarde - St Cergue</name>
    <desc>Total: 63.6km</desc>
    <trkseg>
      <trkpt lat="46.11004142090678" lon="5.827384488657117">
        <ele>382.79998779296875</ele>
        <time>2017-09-10T03:11:34Z</time>
      </trkpt>
      <trkpt lat="46.110052736476064" lon="5.82740711979568">
        <ele>382.79998779296875</ele>
        <time>2017-09-10T03:11:36Z</time>
      </trkpt>
    </trkseg>
  </trk>
  <trk>
    <name>Bellegarde - St Cergue section 2</name>
    <trkseg>
      <trkpt lat="46.36913954280317" lon="6.017430983483791">
        <ele>1330.199951171875</ele>
        <time>2017-09-10T12:09:55Z</time>
      </trkpt>
      <trkpt lat="46.369090508669615" lon="6.017514215782285">
        <ele>1330</ele>
        <time>2017-09-10T12:10:02Z</time>
      </trkpt>
    </trkseg>
  </trk>
</gpx>
```

pixelfr:
This issue seems similar to that one:
[https://gis.stackexchange.com/questions/240697/how-to-merge-two-or-more-features-in-openlayers]()

I can start to implement the solution with JSTS.

pixelfr:
I made the following commit:
[https://github.com/pixelfr/v6_ui/commit/cc8c1cb2b3fa97cb6c68658d2dc3306b57522235]()

[...]

PR #1855, opened 2017-12-07
[WIP] Gps merge tracks

pixelfr:
I am trying to use JSTS to merge multiple tracks within one input `.gpx` file.

This PR is trying to solve issue #1790

At the moment:
* I created an externs file to use JSTS within closure. The externs file seems functional.
* I added some code in the scripts loading the `.gpx` tracks but it generates errors (see comments).


c2corg_ui/static/js/gpxupload.js
```diff
@@ -66,6 +66,20 @@ app.GpxUploadController.prototype.importGpx_ = function(gpx) {
...
     featureProjection: 'EPSG:3857'
   });
+  if (features.length > 1) {
+    var parser = new jsts.io.OL3Parser();
+    var gf = new jsts.geom.GeometryFactory();
+
+    var jstsFeatures = [];
+    for (var i = 0; i < features.length; i++) {
+      var geofeat = features[i].getGeometry();
+      var jstsfeat = parser.read(geofeat);
```
pixelfr: This line generates an error in Chrome
```
TypeError: Cannot read property 'Point' of undefined
    at Yn.read (deps.js:1439)
    at zC.c (build.js:983)
    at m.$digest (deps.js:1585)
    at m.$apply (deps.js:1588)
    at FileReader.d.onload (build.js:686)
```

c2corg_ui/static/js/gpxupload.js
```diff
@@ -66,6 +66,20 @@ app.GpxUploadController.prototype.importGpx_ = function(gpx) {
...
+      jstsFeatures.push(jstsfeat);
+    }
+    var jstsFeatureColl = gf.createGeometryCollection(jstsFeatures);
```
tsauerwein: For routes and outings, either a `LineString` or `MultiLineString` is expected. So, the API will complain about a geometry collection.

https://github.com/c2corg/v6_api/blob/c4152e4ffd3f2ff801cebc60dcec4eb310db6e35/c2corg_api/models/route.py#L328-L329

pixelfr: The geometry collection goes through a union, so it should be good no?

tsauerwein:
Thanks for your work on this, @pixelfr!

I would probably try to merge the different geometries manually into a `MultiLineString`. Something like:
```javascript
if (features.length > 1) {
  var layout = features[0].getGeometry().getLayout();
  var track = new ol.geom.MultiLineString([], layout);

  for (var i = 0; i < features.length; i++) {
    var geom = features[i];
    if (geom instanceof ol.geom.LineString) {
      track.appendLineString(geom);
    } else if(geom instanceof ol.geom.MultiLineString) {
      var multiLine = /** type{ol.geom.MultiLineString} */ (geom);
      multiLine.getLineStrings().forEeach(function(line) {
        track.appendLineString(geom);
      });
    }
  }
}
```
(untested)

c2corg_ui/static/js/gpxupload.js
tsauerwein: I am not sure from the docs: https://bjornharrtell.github.io/jsts/doc/api/jsts.geom.GeometryCollection.html#union

stef74:
@pixelfr what is the status of this PR ?

pixelfr:
An unexpected event happened and I couldn't look at it for a while. I can try to have another look. I think I saw another discussion regarding the gps track on the c2c forum. Is there still a use for what I was trying to do?

PR #1858, opened 2017-12-08
Section collapse

brunobesson:
Rework section collapsing so that it works again with the new design.


stef74:
@brunobesson could you fix conflict

Issue #1859, opened 2017-12-09
Forum in guidebook documents


brunobesson:
* case with a topic created but no comment posted (only system one) is not handled
* links are not configuration-based (always points to `https://forum.camptocamp.org/`)

See e.g. https://www.camptocamp.org/outings/947661/fr/cret-du-boeuf

brunobesson:
(issue extracted from #1808)

desnoes:
ok in demo

PR #1860, opened 2017-12-09
handle no comments in document topic

brunobesson:
* use config-defined link to forums
* handle case when there is a topic for the document, but no post yet

relates to #1859

### DIFF
--- a/c2corg_ui/static/js/viewdetails.js
+++ b/c2corg_ui/static/js/viewdetails.js
@@ -136,9 +136,9 @@ app.ViewDetailsController = function($scope, $compile, $uibModal, appApi,
 
   /**
    * @type {string}
-   * @private
+   * @export
    */
-  this.discourseUrl_ = discourseUrl;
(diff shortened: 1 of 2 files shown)


config/bruno
```diff
@@ -1,7 +1,12 @@
 include config/dev
+<<<<<<< 69ea292da60cf6a4790cfa85e3478bb599d2c839
```
asaunier: Are you sure you want to merge conflicts in your config file? :P

package-lock.json
```diff
@@ -0,0 +1,6116 @@
+{
```
asaunier: You don't need to commit this file if it is generated.

config/bruno
brunobesson: F** another rebasing issue. I should have been more careful, I apologize.

brunobesson:
@asaunier back to normal state... thanks for your reviews!

PR #1869, opened 2017-12-10
add img and gps icon

stef74:

### DIFF
--- a/c2corg_ui/static/partials/cards/outings.html
+++ b/c2corg_ui/static/partials/cards/outings.html
@@ -132,7 +132,13 @@
             </div>
           </a>
         </div>
-      </div>    
+      </div>
+      <div class="parent-bubble" ng-if="doc['img_count']">
+        <span class="glyphicon glyphicon-picture">  {{::doc['img_count']}}</span>
+      </div>
+      <div class="parent-bubble" ng-if="doc['geometry']['has_geom_detail']">
+        <span class="glyphicon glyphicon-flag"></span>
+      </div>
       <div class="grow">
       </div>
       <div class="parent-bubble bubble-date">
--- a/c2corg_ui/static/partials/cards/outings.html
+++ b/c2corg_ui/static/partials/cards/outings.html
@@ -77,7 +77,7 @@
             </div>
           </div>
         </div>
-        <div class="group-infop elevation_max"">
+        <div class="group-infop elevation_max">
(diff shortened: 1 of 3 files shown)


stef74:
some improvement needed

stef74:
capture d ecran 2017-12-10 a 13 16 38
capture d ecran 2017-12-10 a 13 16 11

stef74:
Is probaly enough for the 1st version ...

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -132,7 +132,13 @@
             </div>
           </a>
         </div>
-      </div>    
+      </div>
+      <div class="parent-bubble" ng-if="doc['img_count']">
+          <span class="glyphicon glyphicon-picture" uib-tooltip="({{::doc['img_count']}})" tooltip-append-to-body="true"></span>
```
brunobesson: Wouldn't a text like `3 pictures for this outing` be better than `(3)`?

c2corg_ui/static/partials/cards/outings.html
```diff
@@ -132,7 +132,13 @@
             </div>
           </a>
         </div>
-      </div>    
+      </div>
```
brunobesson: this is not related to your PR, but there is a typo on line 80 (double quotes). Could you also correct this?

brunobesson: Also, on line 128, the tag is not closed, the greater char is missing.

c2corg_ui/static/partials/cards/outings.html
brunobesson: indentation should be 2 spaces, not 4

stef74: 3 picture(s) is probaly enough !

c2corg_ui/static/partials/cards/outings.html
stef74: fixed

c2corg_ui/static/partials/cards/outings.html
stef74: fixed

brunobesson: Yeah, sure. If we could handle plurals by the same occasion :pray:

PR #1870, opened 2017-12-10
Rework annoucement design

brunobesson:
* no right and left margin, use full space (there is a problem with the preferences and personal feed configuration buttons, but this is already the case anyway)
* only first paragraph is show by default, other are hidden and shown only when expanded.
  This implies the annoucement consists only of paragraphs! But the advantage is that we are not limited to a single line anymore (which was moreover cut so this is problematic e.g. on mobile)
* cleanup and simplify HTML and CSS
* use only one button to expand and reduce
* remove useless scope variable in JS
* rename visible atrribute in JS with expanded, which is more explicit

snapshot6
snapshot7
snapshot8
snapshot9
snapshot10
snapshot11

### DIFF
--- a/c2corg_ui/static/js/announcement.js
+++ b/c2corg_ui/static/js/announcement.js
@@ -18,7 +18,6 @@ app.announcementDirective = function() {
 app.module.directive('appAnnouncement', app.announcementDirective);
 
 /**
- * @param {!angular.Scope} $scope Scope.
  * @param {app.Api} appApi Api service.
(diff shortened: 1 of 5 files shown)


brunobesson:
@fbunoz I reworked the button positioning so that:

* it is always positioned on the right
* text is only shown when size is big enough (i.e. not on tablet nor phone)

Are you OK with such output?

snapshot14
snapshot15
snapshot16
snapshot17

PR #1875, opened 2017-12-11
allow es6

brunobesson:
* closure transpiles from es6 strict to es5 strict (no need to be constrained to es3 anymore)
* update eslint and jshint configuration for es6 (and also packages)
* small test in forumfeed.js

Some alternate questions:

* could/should we think about updating closure-util too?
* what's the role of pyramid_closure in the process? I don't get it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 *.egg-info
 *.po~
 Dockerfile_deployment_environment
+package-lock.json
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,10 @@
   "parserOptions": {
     "ecmaVersion": 6
   },
-  "extends": "openlayers",
+  "extends": [
+    "eslint:recommended",
+    "openlayers"
+  ],
   "globals": {
     "app": false,
     "goog": false,
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,7 @@
 {
+  "parserOptions": {
+    "ecmaVersion": 6
+  },
   "extends": "openlayers",
   "globals": {
     "app": false,
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,5 +2,6 @@
   // suppresses warnings about using [] notation (person['name']) W069
   "sub": true,
   // suppresses errors that occur when adding @typedef
-  "-W030": true
+  "-W030": true,
+  "esversion": 6
 }
--- a/build.json
+++ b/build.json
@@ -65,7 +65,8 @@
     "compilation_level": "ADVANCED",
     "warning_level": "VERBOSE",
     "generate_exports": true,
-    "language_in": "ECMASCRIPT5_STRICT",
+    "language_in": "ECMASCRIPT6_STRICT",
+    "language_out": "ECMASCRIPT5_STRICT",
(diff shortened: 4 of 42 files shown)


brunobesson:
Huhu... I see that it is now failing, because the default eslint configuration is more strict. I think we should use event more constrained rules :1st_place_medal: 

I'll correct them to make it pass!

.gitignore
```diff
@@ -12,3 +12,4 @@
 *.egg-info
 *.po~
 Dockerfile_deployment_environment
+package-lock.json
```
asaunier: Where is this file provided?

asaunier:
> I see that it is now failing, because the default eslint configuration is more strict. I think we should use even more constrained rules

Have you fixed the problem? It seems I have such a problem when trying to build the JS with this branch?!

.gitignore
brunobesson: `pakage-lock.json` file is generated by recent (5.4.0?) versions of npm when doing an install. It is meant to build a tree of dependencies, or something such.

brunobesson:
>Have you fixed the problem?

Yes I did, that's why the checks have passed now.
Have you updated your installation? jshint and eslint have been upgraded.

asaunier:
> could/should we think about updating closure-util too?

That would be nice indeed. By the way there are several PR from snyk suggesting to upgrade closure-utils but then the JS files fail to build. Perhaps your changes will help upgrading to a recent version of closure-utils.

> what's the role of pyramid_closure in the process? I don't get it

I don't remember. That's a tool to help build a pyramid project with Closure: https://github.com/camptocamp/pyramid_closure
But I don't know if we still use it.

> Have you updated your installation? jshint and eslint have been upgraded.

I still had the problem after running ``make -f config/alex clean upgrade install check``?! But indeed it seems to pass for Travis, don't know what my problem is.

brunobesson:
>don't know what my problem is.

What is the error returned?

brunobesson:
>By the way there are several PR from snyk suggesting to upgrade closure-utils but then the JS files fail to build. Perhaps your changes will help upgrading to a recent version of closure-utils.

Yeah, saw that too. Not sure theses changes can solve this. I'll have a look at it if I get the chance.

asaunier:
> What is the error returned?

```
/home/alex/v6_ui/node_modules/eslint/bin/eslint.js:16
const useStdIn = (process.argv.indexOf("--stdin") > -1),
^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
Makefile:245: recipe for target '.build/eslint.timestamp' failed
make: *** [.build/eslint.timestamp] Error 8
```

brunobesson:
@asaunier could it be that your node version is too old? See e..g https://stackoverflow.com/questions/22603078/syntaxerror-use-of-const-in-strict-mode

Could you try either updating your nodejs version, or using the previous eslint dependency?

PR #1877, opened 2017-12-11
Top navbar offset for anchors

brunobesson:
Ensures that top fixed navbar offset is taken into account for anchors.
The solution used is to add an invisible block on top of items with an id, inspired by https://github.com/twbs/bootstrap/issues/1768#issuecomment-46519033
To limit the number of such blocks, we restrict it to h3 to h6 titles in the view-details-section part.

### DIFF
--- a/less/anchors.less
+++ b/less/anchors.less
@@ -1,0 +1,18 @@
+// ensure that anchors take into account the fixed top navbar
+// see https://github.com/twbs/bootstrap/issues/1768#issuecomment-46519033
+
+// restrict this to titles h3 to h6 in view details sections
+
+@anchor-top-margin: 5px;
+
+.view-details-section {
+  h2, h3, h4, h5, h6 {
+    &[id]:before {
+      display: block;
+      content: " ";
+      margin-top: calc(~'-@{page-header-height} - @{anchor-top-margin}');
+      height: @page-header-height + @anchor-top-margin; 
+      visibility: hidden;
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


fbunoz:
>  we restrict it to h3 to h6 titles

It's possible to have h2 using sub-title tag. It correspond to "Titre 1" in this article : https://www.demov6.camptocamp.org/articles/829712/fr/un-bel-article

Moreover, I don't understand why the section title (Description, Remarques, Matériel, ...) is in h3 level instead of h2 or h1.

brunobesson:
>It's possible to have h2 using sub-title tag

But then, is an anchor to a subtitle useful?

>Moreover, I don't understand why the section title (Description, Remarques, Matériel, ...) is in h3 level instead of h2 or h1.

Dunno. Maybe @Courgetteandratatouille knows? Anyway, this should be another issue IMHO.

fbunoz:
> But then, is an anchor to a subtitle useful?

Yes, because you can have several h2 sub-title, like h3 for the most of articles.

On V5, the ## subtitle had a different design between article and other docs. It was always h2, but in an article, the design was like V6 # subtitle : useful for a long article.
Then, on V6 we can have the same design for a subtitle level in whole topoguide, but use the h2 as a bigger subtitle for long articles.

brunobesson:
Didn't understand. What you are describing are not subtitles but sections, doesn't it?

fbunoz:
No.
I just want to say that we can put an h2 subtitle in an article, using # tag (## tag do a h3 sub-title).
Then, we can use the anchor on this h2 subtitle like other levels.
Then, it would be better to extend your fix until h2, instead to limit it to h3.

brunobesson:
OK, I understand. I thought we didn't use a single sharp character. I'll update the PR accordingly.

stef74:
@brunobesson i'ts not clear what I have to test !

brunobesson:
To test, when you have an article with a table of content (look for on of the association articles), click on one and the links and ensure that the section title is visible and not hidden by the top navbar.

PR #1880, opened 2017-12-11
Reduce useless bottom margins

brunobesson:
Was
snapshot12

Now
snapshot13

### DIFF
--- a/c2corg_ui/templates/area/view.html
+++ b/c2corg_ui/templates/area/view.html
@@ -138,10 +138,6 @@ area['doctype'] = 'areas'
   </section>
 </div>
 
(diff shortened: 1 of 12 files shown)


brunobesson:
On the forum, @fbunoz mentioned that this may have been done so that the association tool results for "empty" documents is visible.
Tomorrow, I'll have a look at it.

brunobesson:
I have made some changes to take typahead into account.

What I did:

* limit suggestions to 350px (if overflows, scroll)
* put again some margin in the bottom
* but in the meantime remove the unused scroll to top button

@fbunoz is it ok for you?

snapshot20
snapshot21

Issue #1874, opened 2017-12-11
on diff screen, hide map if location has not changed


cbeauchesne:
because it's a huge waste of place

brunobesson:
It is quite easy to solve this, I'll provide a PR.
My (aside) question is: how is the diff supposed to present a change in the GPX file (if any) of a route? Currently, there is only a point shown (I assume it the the geocenter computed from associated waypoints). Should it show the difference for the GPX track to it?

desnoes:
Ok on demo

PR #1881, opened 2017-12-11
Hide geometry diff if there is... no diff!

brunobesson:
The problem remaining will be the case of an update GPX track: as this is not handled by the diff, no modification is shown
@asaunier do you think we should create an API issue for this?

### DIFF
--- a/c2corg_ui/templates/document/diff.html
+++ b/c2corg_ui/templates/document/diff.html
@@ -55,31 +55,36 @@
     <p class="bg-info" translate>No content changes.</p>
   % else:
     % for field_diff in diffs:
-      <h3 x-translate>${field_diff.field}</h3>
-      <div class="diff-container">
       % if field_diff.field == 'geometry':
-      <%block name="moduleConstantsValues">
-      module.value('mapFeatureCollection', {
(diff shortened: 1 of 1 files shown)


brunobesson:
Relates to #1874

asaunier:
What's the use of the new file ``package-lock.json``?

> The problem remaining will be the case of an update GPX track: as this is not handled by the diff, no modification is shown
@asaunier do you think we should create an API issue for this?

Sorry, I don't understand the question?! If the geometry is changed I think the diff at least shows the different "central point" geom, doesn't it?

brunobesson:
>What's the use of the new file package-lock.json?

Sorry, must have been appended when rebasing (didn't chek after). This file is added with recent npm versions, but we don't want it.

>If the geometry is changed I think the diff at least shows the different "central point" geom, doesn't it?

Yes, sure. My question would be what is shown if I edit the GPX track of a route : the diff does not show it, does it?

asaunier:
I think the diff never show the full geometries, only the ``geom`` attribute (ie. a point geometry representing the line geometry at its barycenter or something like that).

PR #1883, opened 2017-12-11
outing list improvement

stef74:
after : 

capture d ecran 2017-12-11 a 20 18 56

Before : 

capture d ecran 2017-12-11 a 19 05 08

### DIFF
--- a/less/cards.less
+++ b/less/cards.less
@@ -608,7 +608,7 @@
       }
       &.card-outing {
         .cotationbox {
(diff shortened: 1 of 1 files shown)


less/cards.less
```diff
@@ -607,6 +607,9 @@
         }
       }
       &.card-outing {
+        .cotationbox {
+          width:  75px;
```
brunobesson: too many spaces

Issue #1808, opened 2017-11-06
Issue(s) linked to new design


stef74:
To be completed

* [x] Maybe "Point de passage principal" could be closer to the card (or inside the card)?
And add a line between "Choisir comme point principal" and the "main waypoint" card. For instance https://www.camptocamp.org/routes/edit/54201/fr (low prio)
![point de passage principal](https://user-images.githubusercontent.com/19718055/33188501-9f5c0f46-d09a-11e7-9a69-5f6dfb094375.PNG)

* [ ] xreports home page : the map cannot be folded.

**Page d'accueil**
- [ ] On ne peut pas replier la section sur l'application mobile.

Je liste ici les petits soucis cosmétique que je vois. Chrome, écran 1200px.

- [ ] Bouton d'éditions un peu trop "collés" au corps de la page : 
<img src="https://forum.camptocamp.org/uploads/default/original/2X/2/234fdbfa7a0353ffe5fc18f0b43c0630c1133ea3.png" width="79" height="208">

<img src="https://forum.camptocamp.org/uploads/default/original/2X/a/a9a74fd68a0ce04cb3c19475ef884af65632afd6.png" width="272" height="32">

- [ ] h4 headers font size is bigger than h3 headers 

- [ ] on mobile, action description in feed is too long and always ellipsed
  
![screenshot_20171125-175706](https://user-images.githubusercontent.com/2234024/33503414-ead466b8-d6e3-11e7-8474-2ff42b616f94.png)

- [ ] should editiion map be expandable? If yes, buttons positioning should be improved

![screenshot_20171123-174521](https://user-images.githubusercontent.com/2234024/33504960-e7a5c210-d6e9-11e7-93fa-54559a9b5228.png)

 - [ ] empty associated outings section should have a default text
![screenshot_20171123-174849](https://user-images.githubusercontent.com/2234024/33505076-5b4c3d8e-d6ea-11e7-99a4-a8d2347addb1.png)

- [ ] Date is missing in the xreports cards in the home feed (it could be added below the title as for outings):
![capture du 2017-12-05 09-43-05](https://user-images.githubusercontent.com/1192331/33597770-b6d90fca-d9a0-11e7-91b7-9e32ff6e24bb.png)

- [ ] Initial author has been removed from outing cards listed in a route pages

- [x] ~~In  route list, the ratings are truncated in some cases :
[...]

Issue #1549, opened 2017-03-13
Card content: insert GPS icon


desnoes:
When a route or an outing contains a GPS track, an icon should be inserted in the card. (Location TBD)

See also the images counter issue at https://github.com/c2corg/v6_ui/issues/1780

asaunier:
Proposed solution: add a boolean "has_track" attribute set to true when the ``geom_detail`` is set (no need to transfer the whole geometry when listing cards).
TODO:

common:
* add the attribute in the list of attributes https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_outing.py#L42 and https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_route.py#L32

UI:
* upgrade the "common" requirement
* show icon in the outing/route cards if data is 1 => update the card partials + CSS

API:
* upgrade the "common" requirement
* add a boolean "has_track" attribute to the outing and route models (not needed in the archives? in the ``DocumentGeometry`` part or in the root ``Document`` one? perhaps easier and more relevant in the geometry part itself?!)
* set it to true when a detail geometry is added to the outing/route document
* if the detail geometry is removed, set the attribute to false
* create an alembic script to add the attribute and set it to true for existing outings/routes with a detail geometry

**Note: perhaps an actual boolean attribute is not needed in the tables, it could be replaced by a "fake" attribute automatically/on the fly set to true depending if ``geom_detail`` is not null.**

tsauerwein:
> Note: perhaps an actual boolean attribute is not needed in the tables, it could be replaced by a "fake" attribute automatically/on the fly set to true depending if geom_detail is not null.

For example like here:
https://github.com/c2corg/v6_api/blob/9f97ed5b643e6097680e767876cfd1b70c86f3bb/c2corg_api/models/feed.py#L57-L62
https://github.com/c2corg/v6_api/blob/e783fe73fa224adb3358d499a61e987a7ba4fb3c/c2corg_api/views/feed.py#L165-L169

See also: http://docs.sqlalchemy.org/en/latest/orm/mapped_sql_expr.html#using-column-property

asaunier:
> For example like here:

Thanks! Yes, something like ``column_properties`` indeed. I wonder if it would trigger a SELECT request for every item returned by the list queries?

tsauerwein:
> I wonder if it would trigger a SELECT request for every item returned by the list queries?

The `column_property` could be defined as:
```
[...]

PR #1885, opened 2017-12-11
List route optimization

stef74:
- add GPS track flag #1549 
- optimize display (show complete cotation, optimize to show a route on one line) #1808 

before 1200px : 
capture d ecran 2017-12-11 a 22 06 22

After 1200px : 
capture d ecran 2017-12-11 a 21 53 44

After 1700px : 
capture d ecran 2017-12-11 a 21 48 31

### DIFF
--- a/c2corg_ui/static/partials/cards/routes.html
+++ b/c2corg_ui/static/partials/cards/routes.html
@@ -82,6 +82,9 @@
           </a>
         </div>
       </div>
+      <div class="parent-bubble" ng-if="doc['geometry']['has_geom_detail']" uib-tooltip="{{ 'GPS Track'| translate}}" tooltip-append-to-body="true">
+        <span class="glyphicon glyphicon-flag"></span>
+      </div>
       <div class="grow"></div>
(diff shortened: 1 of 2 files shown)


asaunier:
Are you sure a flag icon is really that obvious to indicate a GPS track is available? I thought you wanted to use the GPX icon?

stef74:
Honnestly ... the 1st time probably not, and after hover one time on it ...
yes !

If someone want to spend time on that point no problem ...

2017-12-11 22:34 GMT+01:00 Alex Saunier <notifications@github.com>:

asaunier:
Then why not using an explicit-at-first-sight icon? For me the flag means nothing here.

brunobesson:
@asaunier @stef74 Maybe we can go on with this icon and look for feedback when it's online. We should see quite easily if people understand the meaning of it. (IMHO a more explicit icon could be good too, but let's not loose time on it in the first time if people are satisfied with it)

asaunier:
OK, no problem to go forward with this icon!

Issue #1891, opened 2017-12-12
Remove grey background below outing edition associated routes


brunobesson:
![screenshot_20171123-174437](https://user-images.githubusercontent.com/2234024/33505037-3c7c1618-d6ea-11e7-9037-6fad11eb73f3.png)

(extracted from #1808)

desnoes:
ok on demo

PR #1892, opened 2017-12-12
Edit outing association tool

brunobesson:
* no background, similar to other editions
* handle size of search field on tablet
* also cleanup some stuff, mainly missing empty items to ensure last row items have the same width as others

snapshot22


brunobesson:
Relates to #1891

Issue #1898, opened 2017-12-13
Remove grey background on small forum avatars in forum feed


brunobesson:
![snapshot30](https://user-images.githubusercontent.com/2234024/33928170-35e32c02-dfe5-11e7-9f49-053e4cba0a97.png)

stef74:
Ok on demo

PR #1899, opened 2017-12-13
Remove grey background for avatars

brunobesson:
See #1898 

Was
snapshot30

Now
snapshot31


brunobesson:
This is not an issue related to this commit. See my email. I'll merge despite this, thanks for the review.

PR #1902, opened 2017-12-14
Unused css

stef74:
remove some unsued css on search (list and card).

Area,user, home topo, homo serac continue to used old list-item class.


stef74:
Yes I have checked if this style is not call ... and also tested ;-)

Issue #1905, opened 2017-12-16
List not available on small screen


stef74:
On demo : 
![capture d ecran 2017-12-16 a 21 51 07](https://user-images.githubusercontent.com/3962997/34074245-607e9de8-e2ab-11e7-97d0-8dfc73cb14c1.png)

On Production : 
![capture d ecran 2017-12-16 a 21 51 26](https://user-images.githubusercontent.com/3962997/34074244-6059ccfc-e2ab-11e7-8f9d-d169e46bbe7b.png)

stef74:
https://github.com/c2corg/v6_ui/blob/13324b5438de7e9f2539ceb16f173d814e9e23c2/less/lists.less#L423

After analyse the list button is only show for class

stef74:
Link to this commit https://github.com/c2corg/v6_ui/commit/10db5e65dee24031dae399ccba5359d4b6ba1f35#diff-a7693b847cd07d7aa4c819d11e5becc1

We have an other similar issue on feed for small screen (ex 1200px), the card not take 100% of width available.

![capture d ecran 2017-12-16 a 22 13 21](https://user-images.githubusercontent.com/3962997/34074398-610d8d5c-e2ae-11e7-9a62-ab1ffc468915.png)

if I have a look  : 
`
  .feeds .in-feed-col-1, .feeds .in-feed-col-2, .feeds .in-feed-col-3 {
    float: left;
    **width: 50%;**
    padding-right: 20px;
}
`

In production the value is 100% for small screeen.

https://github.com/c2corg/v6_ui/blob/13324b5438de7e9f2539ceb16f173d814e9e23c2/less/feed.less#L596

stef74:
@brunobesson your opinion ?

PR #1908, opened 2017-12-18
fix value variable

stef74:
some feature  based on screen size not working after. 

The value are not calculated ( interpreted as string ?)

#1905 

Fixe error on home for little-screen
capture d ecran 2017-12-18 a 21 16 34


lbesson:
ça a l'air bien

Issue #1911, opened 2017-12-21
Indication of images associated to an outing is not translated


brunobesson:
@mdesnoes, could you add some details? (I am on my phone, if this could be a valid excuse)

stef74:
[image: Images intégrées 1]

2017-12-21 7:53 GMT+01:00 Bruno Besson <notifications@github.com>:

PR #1912, opened 2017-12-21
fix picture tooltip

stef74:
#1911


stef74:
@brunobesson I try to do it ;-)

PR #1926, opened 2017-12-28
Fix docker node version

brunobesson:
Use version 8 (LTS), debian jessie-packaged one is too old to handle recent developments


brunobesson:
(commits made using the github interface, I'll squash the PR when accepted)

PR #1930, opened 2018-01-01
Disable mailing lists menu

asaunier:
Until the API part is fixed (see https://github.com/c2corg/v6_api/issues/697), who knows when.


asaunier:
Thanks @brunobesson 

> Although maybe a warning in the mailing list page could be beter?

Perhaps, but we don't even know if the mailing lists will be back, so removing the menu item seems simpler to me rather than a warning message with no real info.

PR #1940, opened 2018-01-04
Do not include <figure> inside a <p> tag

cbeauchesne:
The root cause is that `<figure> `HTML element is a block element, but is parsed by inline MD parser. In consequence, it's included in a `<p>` element, which is not a valid HTML. Try for instance to put this html in your browser:

    <p><figure></figure></p>

it will be rendered as : 

    <p></p><figure></figure><p></p>
  
I've also added some unit tests in img_tag.json

### DIFF
--- a/c2corg_ui/format/img.py
+++ b/c2corg_ui/format/img.py
@@ -8,14 +8,15 @@ to advanced HTML img tags.
 '''
 
 from markdown.extensions import Extension
-from markdown.inlinepatterns import Pattern
+from markdown.blockprocessors import BlockProcessor
 from markdown.util import etree
 
-IMG_RE = r'\[img=(\d+)([a-z_ ]*)(/\]|\](.*?)\[/img\])'
+import re
+
+IMG_RE = r'(?:^|\n)\[img=(\d+)([a-z_ ]*)(/\]|\](.*?)\[/img\])'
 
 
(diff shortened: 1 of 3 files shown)


cbeauchesne:
It will be a consequence on MD : [img] tag must be on a separate line. It's because this tag is a block tag, and it won't be correctly parsed inside a code tag : `\`\`` 

Neverless, I've allowed a text after the tag, even if it's not a clear MD code.

Exemple

    [img=123/]
is ok 

    [img=123/][img=123/]
is ok (two figures)

    [img=123/]new paragrpah
is ok, even if it's dirty

    ...End of previous paragraph[img=123/]
is **not** ok, tag won't be parsed    

    ... in code do `[img=123/]`
is not parsed, but it's what we want : a code block

Discussion on forum : https://forum.camptocamp.org/t/migration-bbcode/199173/151?u=charlesb

lbesson:
Is it because you render [img] as `<figure>` and not `` (which is inline)?

cbeauchesne:
Yes, exactly, (even it's not me that have used figure).

I've choosed to keep figure element, because it's the good way to present illustrations, and for our usage, we present them as block elements. Using image would have been dirty, because we had to force CSS to present them as block element...

cbeauchesne:
ok, thanks for the review

Issue #1941, opened 2018-01-04
[p] tag extension


cbeauchesne:
`[p]`tag must be rendered as : 

    <div style="clear:both"></div>

It's a block element, so it must be parsed with a blockProcessor, and the tag is valid only at a begining of a line

PR #1942, opened 2018-01-05
P tag extension

cbeauchesne:
Markdown extension for [p] tag 
See issue #1941


asaunier:
It works in my instance and the code seems OK to me.
Could you please provide a real example page where this tag is used?

cbeauchesne:
https://www.camptocamp.org/routes/184924/fr/punta-tafunata-d-i-paliri-traversee-s-n-variante-de-la-cheminee-du-kangourou

"Voie" title should start behind picture.

PR #1951, opened 2018-01-09
Toc with only 3 levels, and without emphasis

cbeauchesne:
Check page : 

https://www.camptocamp.org/articles/106728/fr/licence-des-contenus

* Title with level 1, 2 and 3 must be in toc (actually, only 2 and 3 are presents in this article)
* Emphasis (actually, old anchor tags that has not been migrated) should not appear in toc. They must end with a '{'

### DIFF
--- a/c2corg_ui/format/toc.py
+++ b/c2corg_ui/format/toc.py
@@ -1,0 +1,82 @@
+"""
+Table of Contents Extension for Python-Markdown
+===============================================
+
+See <https://pythonhosted.org/Markdown/extensions/toc.html>
+for documentation.
+
+Oringinal code Copyright 2008 [Jack Miller](http://codezen.org)
+
+All changes Copyright 2008-2014 The Python Markdown Project
+
+License: [BSD](http://www.opensource.org/licenses/bsd-license.php)
+
+Modified for C2C : remove title emphasis, and only include <4 levels
+"""
+
+from markdown.extensions.toc import (TocExtension, TocTreeprocessor,
+                                     string_type, stashedHTML2text,
+                                     unique, nest_toc_tokens)
+
+
+class C2CTocTreeprocessor(TocTreeprocessor):
+    def run(self, doc):
+
+        def not_emphasis(elt):
+            return elt.attrib.get("class") != "header-emphasis"
+
+        # Get a list of id attributes
+        used_ids = set()
+        for el in doc.iter():
+            if "id" in el.attrib:
+                used_ids.add(el.attrib["id"])
+
+        toc_tokens = []
+        for el in doc.iter():
+            if isinstance(el.tag, string_type) and \
+                    self.header_rgx.match(el.tag):
+                self.set_level(el)
+
+                # modification for camptocamp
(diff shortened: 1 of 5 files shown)


c2corg_ui/format/toc.py
```diff
@@ -0,0 +1,82 @@
...
+            if "id" in el.attrib:
+                used_ids.add(el.attrib["id"])
+
+        toc_tokens = []
+        for el in doc.iter():
+            if isinstance(el.tag, string_type) and \
+                    self.header_rgx.match(el.tag):
+                self.set_level(el)
+
+                # modification for camptocamp
```
brunobesson: Shouldn't we instead simply point to the file you modified on github, linking with the revision that was used?

cbeauchesne:
Yes, I show only level 1,2,3, but I do not handle anchors.

The page I gave (which have a toc) got by chance old style anchors, that are falsy interpreted as emphasis. The point is that emphasis should not appear in toc. So it's a good page for testing :) 

In substance, instead of : 

> * Quelle licence pour quel document ? { licence-of-content}
>   * Type et nature du contenu { content}
>     * Type du contenu { content-type}
>       * documents collaboratifs { common-documents}
>       * documents personnels { personal-documents}
>     * Nature du contenu { content-nature}
>   * Contenu collaboratif du topoguide { common-content}

You must see : 

> * Quelle licence pour quel document ? { 
>   * Type et nature du contenu { 
>   * Contenu collaboratif du topoguide {

cbeauchesne:
And for the subject of old style anchors, taht is totally independant of this PR, yes, I will use rabot to migrtae them into span with ids

c2corg_ui/format/toc.py
cbeauchesne: Actually, I've tried to modify less code of official extension as possible. Unfortunalty key method run() was quite big. 

As I need to modify only two lines, I've commented them and what I've done for future maintenance.

brunobesson:
What I don't understand is there is no toc on camptocamp.org's article. Appart from this, it's ok for me.

c2corg_ui/format/toc.py
brunobesson: I understand that. What I was suggessting is that pointing to the original source file would be more readable. But it's not important.

cbeauchesne:
on prod site, the very first line of *corps de l'article* is `[toc]`.

is it the same on your test site? if yes, there is an issue...

c2corg_ui/format/toc.py
cbeauchesne: there is a header on top of the file with all mention, is it what you suggest?

brunobesson:
I just checked, and it is OK, there is a toc on production. But I can swear there was none yesterday evening 😲 
Nevermind. What you intend to do with this PR is OK.
Let's merge.

c2corg_ui/format/toc.py
brunobesson: Non. Nevermind. 😉

PR #1952, opened 2018-01-10
Alert tag

cbeauchesne:
To test, check that this markdown : 

```
 !! Ceci est une information neutre
 
 alors que :
 
 !!! Ceci est un warning
 !!! a prendre en compte
 
 mais surtout n'oubliez pas que : 
 
 !!!! Ceci est une
 !!!! information **ultra importante**
```

gives something close to https://getbootstrap.com/docs/4.0/components/alerts/

### DIFF
--- a/c2corg_ui/format/alerts.py
+++ b/c2corg_ui/format/alerts.py
@@ -1,0 +1,73 @@
+from markdown.extensions import Extension
+from markdown.blockprocessors import BlockProcessor
+from markdown import util
+import re
+
+
+class AlertProcessor(BlockProcessor):
+    RE = re.compile(r'(^|\n)[ ]{0,3}(!{2,4})(.*)')
+
+    classes = {
+        "!!": "md-alert md-alert-info",
+        "!!!": "md-alert md-alert-warning",
+        "!!!!": "md-alert md-alert-danger",
+    }
+
+    def test(self, parent, block):
+        return bool(self.RE.search(block))
+
+    def run(self, parent, blocks):
+        block = blocks.pop(0)
+        m = self.RE.search(block)
+        level = m.group(2)
+        tester = re.compile("^[ ]{0,3}" + level + "([^!]|$)")
+
+        before = block[:m.start()]  # Lines before blockquote
+        # Pass lines before alert banner
+        self.parser.parseBlocks(parent, [before])
+
+        after = block[m.start():].split('\n')
+        if len(after[0]) == 0:
+            after.pop(0)
+
+        block = []
+
+        # get all lines starting with the same prefix
+        while len(after):
+            is_same_level = bool(tester.search(after[0]))
+            if is_same_level:
+                block.append(after.pop(0))
+            else:
+                break
+
+        # Remove `!!!` from begining of each line.
+        block = '\n'.join([self.clean(line) for line in block])
+
+        quote = util.etree.SubElement(parent, 'div')
+        quote.set("class", self.classes[level])
+        # Recursively parse block with div as parent.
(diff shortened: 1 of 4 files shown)


less/viewdetails.less
```diff
@@ -178,6 +178,31 @@
...
 
+  .md-alert {
+    padding: 15px;
+    margin-bottom: 20px;
+    border: 1px solid transparent;
+    border-radius: 1px;
+  }
+
+  .md-alert-info {
+    border-color: #bce8f1;
```
brunobesson: Are you re-using colors already defined in the CSS, or simply the bootstrap one? In all cases, I suggest we store them in variables.

less/viewdetails.less
```diff
@@ -178,6 +178,31 @@
     }
   }
 
+  .md-alert {
```
brunobesson: rather than adding the CSS to this already big file, I would rather build a new `markdown.less` file, imported in `c2corg_ui.less`

I know the CSS is a mess, but we should start organizing it now.

c2corg_ui/format/alerts.py
```diff
@@ -0,0 +1,74 @@
...
+            return ""
+        elif m:
+            return m.group(3)
+        else:
+            return line
+
+
+class AlertExtension(Extension):
+    def extendMarkdown(self, md, md_globals):  # noqa
+        # md.registerExtension(self)
```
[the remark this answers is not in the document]
cbeauchesne: sure, legacy from old code

less/viewdetails.less
cbeauchesne: Agree, but I never write a line of CSS. Cold we do a separate issue for this?

less/viewdetails.less
cbeauchesne: Simply use the bootstrap one

cbeauchesne:
>Nice job, thanks. I spent some time trying to understand why some block weren't recognized. The solution was to remove the trailing spaces. Maybe the regex could be less strict about that?

Could you give me the code that wasn't recognized ?

> Last question: what is you plan in order to remove the soon-to-be obsolete warning and danger extensions?

Rabot will migrate DB from old to new tag.

less/viewdetails.less
brunobesson: It's only postponing the current situation. Let's do this later.

less/viewdetails.less
brunobesson: Then, here again, let it this way. But in the medium term, we should have better CSS organisation.

brunobesson:
>Edit : I guess its ..!! coucou where dots stands for spaces?

Yes, indeed.

brunobesson:
>Rabot will migrate DB from old to new tag.

OK. Is there a need to build an issue for removing the obsolete tags afterwards, or maybe you know you won't forget this?

less/viewdetails.less
```diff
@@ -178,6 +178,31 @@
...
+    padding: 15px;
+    margin-bottom: 20px;
+    border: 1px solid transparent;
+    border-radius: 1px;
+  }
+
+  .md-alert-info {
```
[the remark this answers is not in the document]
cbeauchesne: Well seen!

cbeauchesne:
Yes, and issue is better. I will add it

cbeauchesne:
> Maybe the regex could be less strict about that?

Done, up to 3 spaces is ok.

brunobesson:
Thanks (weird to see there is no character such as `\s` in python). Please rebase and merge.

Issue #1954, opened 2018-01-12
Map recenter text field not working properly


asaunier:
This tool, available in the maps, for instance in the advanced search pages, is based upon the https://photon.komoot.de/ service.

It was broken a few days ago according to https://forum.camptocamp.org/t/probleme-recherche-sur-la-carte/204205

Now it's back but it seems searching on a city no longer works, one get results looking like addresses?
Perhaps something has changed in the service and it must be reconfigured?

desnoes:
> Perhaps something has changed in the service and it must be reconfigured?

Do you know where is-it configured ? which file ?

desnoes:
@asaunier : i found how to correct the bug. I add filtering based on osm tags ('&osm_tag=place:city&osm_tag=place:town&osm_tag=place:village'). The result is fine, all the street's names are removed.
I prepare a PR.

desnoes:
ok on demo

PR #1975, opened 2018-01-21
Map search query

desnoes:
Fix for [issue 1954](https://github.com/c2corg/v6_ui/issues/1954). The request is filtered to give results that correspond to the following OSM tags: city, town and village. This remove all street's adresses from the search result.

### DIFF
--- a/c2corg_ui/static/js/map/search.js
+++ b/c2corg_ui/static/js/map/search.js
@@ -132,6 +132,7 @@ app.MapSearchController.SEARCH_URL = 'https://photon.komoot.de/api/';
 app.MapSearchController.prototype.createAndInitBloodhound_ = function() {
   let url = app.MapSearchController.SEARCH_URL;
   url += '?q=%QUERY';
+  // url += '&osm_tag=place:city&osm_tag=place:town&osm_tag=place:village';
 
   const bloodhound = new Bloodhound(/** @type {BloodhoundOptions} */({
     limit: 10,
--- a/c2corg_ui/static/js/map/search.js
+++ b/c2corg_ui/static/js/map/search.js
@@ -131,8 +131,7 @@ app.MapSearchController.SEARCH_URL = 'https://photon.komoot.de/api/';
  */
 app.MapSearchController.prototype.createAndInitBloodhound_ = function() {
   let url = app.MapSearchController.SEARCH_URL;
-  url += '?q=%QUERY';
-  // url += '&osm_tag=place:city&osm_tag=place:town&osm_tag=place:village';
+  url += '?q=%QUERY&distance_sort=false';
 
   const bloodhound = new Bloodhound(/** @type {BloodhoundOptions} */({
     limit: 10,


c2corg_ui/static/js/map/search.js
```diff
@@ -132,6 +132,7 @@ app.MapSearchController.SEARCH_URL = 'https://photon.komoot.de/api/';
 app.MapSearchController.prototype.createAndInitBloodhound_ = function() {
   let url = app.MapSearchController.SEARCH_URL;
   url += '?q=%QUERY';
+  url += '&osm_tag=place:city&osm_tag=place:town&osm_tag=place:village';
```
brunobesson: This should rather be written in a single line, with query in the end (although it's only cosmetics)

`url += '?osm_tag=place:city&osm_tag=place:town&osm_tag=place:village&q=%QUERY';`

asaunier:
@desnoes Nice!
I wonder if your change is not a bit too restrictive though. I remember it was possible to search also for location names such as mountains (eg. "Chamechaude") or even regions "Valais". Don't you think it would be nice to keep such features?
Is there a page listing the available values for "osm_tag"?

In addition: do you really need 2 commits? :P 
Again it is better to tidy commits, see https://github.com/c2corg/v6_ui/wiki/Git-Cookbook#managing-commits

desnoes:
@asaunier : it tested with key word 'mont-blanc' and i did not get any result, so i concluded summits were not returned. This was a mistake, i checked with Chamechaude and i get the result. I will try other tags.
The list of tags is given here: https://wiki.openstreetmap.org/wiki/Map_Features. There is an interesting list of tags for mountain activities in section 'Landform related'.

For the commit, this is an error from my side.

desnoes:
I did some investigations. ThHe problem is not solved with a filer defined with osm_tags. I fact the bug comes from the location bias. 
If i comment the code related to search with priority to a geo position, then the search is working as expected. For instance, when i type 'Chambéry', the city is Rhône-Alpes is return in the first result, whereas it is not in the list with the current version. If i type 'chamechaude', the peak is sorted in the first position and streets containing this string are given after.
I don't see the need for a search with location bias in our map search and thus propose to comment these lines of code:

` const center = this.map.getView().getCenter();
  if (center !== undefined) {
     // give priority to nearby results
     const centerWgs84 = ol.proj.toLonLat(center);
     url += '&lon=' + centerWgs84[0] + '&lat=' + centerWgs84[1];
   }`

Do you agree ?

asaunier:
@desnoes

>  add 'distance_sort=false' in the command line to solve the bug and i works very fine !

OK then, please merge.
Could you:
* remove the useless commit
* fix the typo in the other commit message. By the way prefer using explicit but simple commit message: think that they will appear in the commits log https://github.com/c2corg/v6_ui/commits/master => no need of "v2" etc.

desnoes:
@asaunier : thanks for the review. I will remove the useless commit

>Just a question though: komoot/photon#297 makes distance_sort=false a default value in photon. Then do we really need to add this value explicitly? Perhaps we only have to wait photon service is updated with this new default value?

The default value is not yet applied by photon. I don't know when they will update their service. In the meantime, i think it is usefull to correct the bug.

asaunier:
@desnoes Don't forget to delete the branch after merging the PR => "Delete branch" button above.

PR #1981, opened 2018-01-23
Narrow nbsp before ?, ! and ;

cbeauchesne:
Try this markdown : 

    Hello ! How are you ? Check this :

Html produced must be : 

```html
Hello&#8239;! How are you&#8239;? Check this&#8239;:
```

And try to play with the width of your screen, the puctation mark must never leave it precedent's word.


cbeauchesne:
> The distinction between non-breakable space and narrow nbsp is all but useless in the real world

Completly agree. 

But as any geek who respect himself, I love to give the impression to be a diehard in typography. Steve Jobs rules! 😄

PR #1989, opened 2018-01-30
Add support of rate limiting message

asaunier:
As for now, when a contributor's edition is rejected because of the rate limiting tool (see https://github.com/c2corg/v6_api/pull/716), the default message "Failed saving changes" is displayed.

The aim of this PR is to show an explicit message when a user is rate limited.

The PR is still work in progress because it seems the existing ``errorSaveDocument_`` function (in the ``api `` service) does not work as expected: the status and addition data are indeed provided in the API response but are not provided as expected in the ``response`` argument of callback ``errorSaveDocument_``: 

sans titre

We have the same problem when creating a document, and when the user is blocked (403 status instead of 429) or when the API crashes with a 500 error => "Failed saving changes" is always displayed instead of the expected detailed message.

### DIFF
--- a/c2corg_ui/static/js/apiservice.js
+++ b/c2corg_ui/static/js/apiservice.js
@@ -320,13 +320,13 @@ app.Api.prototype.getDocumentByIdAndDoctype = function(id, doctype, lang) {
  * @private
  */
 app.Api.prototype.errorSaveDocument_ = function(response) {
-  // FIXME: API does not return a valid JSON response for 500/403 errors.
-  // See https://github.com/c2corg/v6_api/issues/85
   let msg;
   if (response['data'] instanceof Object && 'errors' in response['data']) {
     msg = response;
   } else if (response['status'] === 403) {
-    msg = 'You have no permission to modify this document';
+    msg = 'You have no rights to edit this document.';
+  } else if (response['status'] === 429) {
+    msg = 'You have made too many changes recently and thus have been blocked';
   } else {
     msg = 'Failed saving the changes';
   }
--- a/c2corg_ui/templates/i18n.html
+++ b/c2corg_ui/templates/i18n.html
@@ -374,6 +374,8 @@
 
 ## errors from js files
 <span translate>You have no rights to edit this document.</span>
+<span translate>You have made too many changes recently and thus have been blocked.</span>
+<span translate>Failed saving the changes</span>
 <span translate>Form is not valid</span>
 <span translate>You must log in to edit this document.</span>
 <span translate>no associated routes</span>


asaunier:
> the existing errorSaveDocument_ function (in the api service) does not work as expected

Actually it works when uploading new images: the ``status`` info is then actually provided in the ``response`` argument of function ``errorSaveDocument_``.

But we have ``status: -1`` (probably means it is missing) when ``errorSaveDocument_`` is called when creating a document, updating a document or creating a comment topic to the document.

This is a bit surprising because the code is almost the same for creating a document or creating an image after uploading it, both using the :
* creating a document: https://github.com/c2corg/v6_ui/blob/ac2518f7525eb6dc485f9d9642d079321b5954be/c2corg_ui/static/js/apiservice.js#L231-L232
* creating an image: https://github.com/c2corg/v6_ui/blob/ac2518f7525eb6dc485f9d9642d079321b5954be/c2corg_ui/static/js/apiservice.js#L662-L665

asaunier:
I have noticed CORS errors with the demo API:
capture du 2018-01-31 20-06-37

I don't know if it can explain the problem above?
The error is also reported when creating images (but now creating images fails the same way...).

asaunier:
I think the errors reported above are related to missing CORS headers when the request failed. See https://github.com/c2corg/v6_api/issues/720

The present PR is then probably OK and can be reviewed as is.

brunobesson:
>I think the errors reported above are related to missing CORS headers when the request failed

Have you tested with an extension to add CORS headers, or are you confident enough this is CORS-related?

asaunier:
> Have you tested with an extension to add CORS headers, or are you confident enough this is CORS-related?

I have not but that's a good idea. The CORS issue is only a suspected reason but I am not sure.
Anyway the changes suggested by the current PR are relevant and consistent with the other cases (eg. 403 error). As a result I think it is safe to merge it. Thanks for reviewing!

asaunier:
> Have you tested with an extension to add CORS headers, or are you confident enough this is CORS-related?

Actually it is not easy to do so because the missing CORS headers should be in the API response (not in the request) that the UI receives. Do you know a way to intercept the API response and add the CORS headers?

Anyway those missing CORS headers seem like the perfect (?) suspects since the JSON response looks consistent and as shown in the screenshots of https://github.com/c2corg/v6_api/issues/720 the missing CORS headers are quite the only difference in the reponse headers (except the "Transfer-Encoding" header).

brunobesson:
>Do you know a way to intercept the API response and add the CORS headers?

There are browser extensions for this. E.g. https://chrome.google.com/webstore/detail/cors-toggle/jioikioepegflmdnbocfhgmpmopmjkim or https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi for chrome. I don't remember wich one I am using. They generally add the related headers to all responses.

arnaud-morvan:
CORS headers should be placed by server (API in our case), CORS headers are added by cornice, probably in a view wrapper, after view execution.
In case of rate limiting, we probably do not enter such wrapper.
In case of 500, the interruption may occurs before they are added to the response.

arnaud-morvan:
https://github.com/Cornices/cornice/blob/master/cornice/service.py#L518

asaunier:
I have installed the "Cors Toggle" extension https://chrome.google.com/webstore/detail/cors-toggle/jioikioepegflmdnbocfhgmpmopmjkim
and tested it on the demo:
* when the extension is OFF => the default "Failed saving changes" error message is shown
* whent the extension is ON => the "Rate limited" message is shown as expected

It seems to confirm that the problem is related to the missing CORS.

By the way I have realized the change of this PR is mostly useless because the "rate limited" message is  retrieved using the standard message system already available at https://github.com/c2corg/v6_ui/blob/6b2acd4d4ddebfa7517bb0a41aa214c619cb2b3d/c2corg_ui/static/js/apiservice.js#L324-L325
=> the "status=429" condition is never tested

I should cancel some part of the changes, perhaps updating the "translate" part to have the message translated.

PR #1992, opened 2018-01-31
Allow no cell, always parse table, labels, multipitchs

cbeauchesne:
to test, try this code : 

    L#
    L#

### DIFF
--- a/c2corg_ui/format/ltag.py
+++ b/c2corg_ui/format/ltag.py
@@ -1,9 +1,10 @@
+# coding: utf-8
+
 """
 LTag Extension for Python-Markdown
 ====================================
 
-See https://regex101.com/r/g2Zjqn/2/ for regex tests
-
+See https://regex101.com/r/bVjbkp/6
 """
 
 from markdown.extensions import Extension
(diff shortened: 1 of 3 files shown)


brunobesson:
Is this PR ready for review ? CI is failing.

cbeauchesne:
No, I've made it a litlle bit too early, and added a WIP on title. I must do a lot of test before submitting it. Sorry.

c2corg_ui/format/ltag.py
```diff
@@ -15,364 +16,315 @@
...
+                             and match.group("offset") is None \
+                             and match.group("label") is None \
+                             and match.group("local_reference") is None
+
+            modifier = match.group("modifier")
+            typ = match.group("type")
+
+            if modifier is None:
+                label = self.get_value(
+                    match.group("type") if impact_numbering else row_type,
```
olaurendeau: `typ if impact_numbering else row_type` ?

c2corg_ui/format/ltag.py
```diff
@@ -15,364 +16,315 @@
...
+                    to_add=1 if impact_numbering else 0)
+
+                return TEMPLATE_LTAG.format(type=typ,
+                                            modifier=str(label))
+
+            if modifier == "=":
+                return "" if is_first_cell else match.group(0)
```
olaurendeau: `match.group(0)` == `typ` ?

c2corg_ui/format/ltag.py
cbeauchesne: fixed

c2corg_ui/format/ltag.py
cbeauchesne: not understood

cbeauchesne:
@olaurandeau thank you. But it's yet far from a stable code, even without numbering feature... wip...

cbeauchesne:
Well, it starts to look like not so bad.

Here is the code to tests : 

    L#
    L#2
    L#3bis
    L#4-5
    L#6ter-7

Note that a label and a simple L# are for the moment mutually exclusives : you cant' use both in the same markdown field.

And also, if an unvalid pattern is detected, table is rendered, and patterns are in a `<code>` node.

@brunobesson @olaurendeau

olaurendeau:
Look good to me @cbeauchesne. Anyway most important is that the tests stay green :)
Huge work and great feature addition, well done !

cbeauchesne:
Thank you @olaurendeau.

Small question : does translation stuff on L span works? because in-text injection of Html dramaticly decrease performance (three time slower). I would like to remove it, until it's nor working.

olaurendeau:
It was a proposition of @asaunier if I remember properly, no issue with for removing it in the meantime :)

asaunier:
> because in-text injection of Html dramaticly decrease performance (three time slower).

What do you mean?
All translations (not only for the L#) are done in JS by angular-gettext, replacing on-the-fly strings marked as translated (with a ``translate`` directive/attribute). It works for L# as well. For instance switch interface language to EN when viewing https://www.camptocamp.org/routes/220643/fr/cornettes-de-bise-poeme-roc

cbeauchesne:
it's not the translation that's slow, but the injection of span by ltag markdown processor.

By the way, it's a double cause issue : the other cause is we do no reset parser state before converting markdown. If we reset it, internals cache used by this injected HTML does not grow, and the performace issue is not noticable.

BUT, to reset parser state, we must be sure that requests are thread safe (see my mail sent yesterday).

cbeauchesne:
Merged, I will handle the performance issue AND the thread safe issue in another PR

PR #2020, opened 2018-02-18
Ajustements on side menu

cbeauchesne:
I did not yet done active hiding because I'm not sure about the best way to do it...

For the rest : 

* fine compute size for ad display toggler
* remove some small links on the bottom : they must not on more than two lines
* remove home link
* reduce margin top and bottom of menu item 2x8 io 2x10px
* if place is missing, hide social links before hiding ads.

### DIFF
--- a/less/sidemenu.less
+++ b/less/sidemenu.less
@@ -52,7 +52,6 @@ body.menu-toggled #content-toggled {
   background-color: white;
   margin: 0;
   white-space: nowrap;
-  font-size: 1em;
 
   .page-menu {
     width: 100%;
@@ -80,14 +79,35 @@ body.menu-toggled #content-toggled {
     }
   }
 
-  @media (max-height: 745px) {
+  /********************************************************
+  fine compute size for ad display toggler
+  if place is missing, hide social links before hiding ads.
+
+  WHAT            SIZE  COMPUTE
+  Logo            80
+  Navigation      185   5*(1+8+20+8)
+  Margin          15
+  Advertising     320
+  Margin          15
+  Links           30
+  Margin          15
+  Social networks 35
+  Margin          15
+  TOTAL
+  without SN      660    35+15
+  All             710
+  ********************************************************/
+
+  @advertising-height: 660px;
+
+  @media (max-height: @advertising-height) {
     .advertisement {
       display: none !important;
     }
   }
 
-  @media (max-height: 770px) {
-    .version {
+  @media (max-height: 710px) {
+    .social-networks {
       display: none !important;
     }
   }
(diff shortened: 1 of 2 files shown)


cbeauchesne:
And for reviewers, it's my very first less lines. It surely be perfectible....

less/sidemenu.less
```diff
@@ -52,7 +52,7 @@ body.menu-toggled #content-toggled {
   background-color: white;
   margin: 0;
   white-space: nowrap;
-  font-size: 1em;
+  font-size: 14px; /* force font size */
```
brunobesson: with the modifications already applied, this should'nt be needed. Keep 1em.

less/sidemenu.less
```diff
@@ -80,14 +80,33 @@ body.menu-toggled #content-toggled {
...
+  Navigation      185   5*(1+8+20+8)
+  Margin          15
+  Advertising     320
+  Margin          15
+  Links           30
+  Margin          15
+  Social networks 35
+  Margin          15
+
+  Total sans SN   660    35+15
```
brunobesson: without

less/sidemenu.less
```diff
@@ -80,14 +80,33 @@ body.menu-toggled #content-toggled {
     }
   }
 
-  @media (max-height: 745px) {
+  /********************************************************
+  fine compute size for ad display toggler
+  if place is missing, hide social links before hiding ads.
+
+  WHAT            SIZE  CALCUL
```
brunobesson: calcul > english plz

less/sidemenu.less
```diff
@@ -80,14 +80,33 @@ body.menu-toggled #content-toggled {
...
+  Links           30
+  Margin          15
+  Social networks 35
+  Margin          15
+
+  Total sans SN   660    35+15
+  Total All       710
+  ********************************************************/
+
+  @media (max-height: 660px) {
```
brunobesson: you could use variables for size constants, it would be more explicit

less/sidemenu.less
cbeauchesne: i've made it on purpose, to be sure that the menu item do exactly 14px. I don't want that this font-size depend of any other CSS parameter. Otherwise, any modification may break the display of advertising.

less/sidemenu.less
cbeauchesne: done

less/sidemenu.less
cbeauchesne: done

less/sidemenu.less
cbeauchesne: done

less/sidemenu.less
brunobesson: Neverless, if we use 1rem, the base font-size won't change, so there is no fear for this. Moreover, if this is a concern, we could handle the leine hieght another way.
I suggest we use 1rem.

PR #2044, opened 2018-03-06
Add biodiv'sports data in map

brunobesson:
Although there might still be some minor changes, the process is the following:

* there is a biodivsport service that is used by the map on routes view to retrieve putative sensitive areas
* those are displayed in map, clicking on it opens a modal dialog
* if there are sensitive areas, a message is added in the remarks section


brunobesson:
Noone? :cry:

PR #2072, opened 2018-04-03
add biodivsports areas to waypoint's map

brunobesson:


asaunier:
Travis tests were broken. It seems the tests pass when restarting them.

PR #2079, opened 2018-04-09
fix position of bottom buttons in fullscreen map on mobile

brunobesson:
Buttons are hidden by bottom user tollbar when map is fullscreen on mobile. This commit move the bottom elements up by 65px in this case.


brunobesson:
References #2080

PR #2083, opened 2018-04-13
add warning text on edit

brunobesson:
References #2057 

Sample rendering:
snapshot37

### DIFF
--- a/c2corg_ui/templates/helpers/edit.html
+++ b/c2corg_ui/templates/helpers/edit.html
@@ -6,7 +6,6 @@
 <%def name="show_editing_associated_waypoints(model, type='waypoints')">\
   <h3 translate ng-show="${model}.associations.${type}.length > 0">Associated waypoints</h3>
   <section  class="associated-section">
-  
     <article class="associated-documents">
       <div class="list-item waypoints ${'to-route' if model == 'route' else ''}"
            ng-repeat="wp in ${model}.associations.${type} | orderBy: 'waypoint_type'"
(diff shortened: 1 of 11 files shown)


c2corg_ui/templates/helpers/edit.html
```diff
@@ -168,7 +167,19 @@ <h3 class="heading show-phone" translate>summary</h3>
...
 </%def>
+
+<%def name="show_edit_framework(model)">\
+  <section class="section edit-framework" ng-if="!${model}.protected">
+    <h2 class="heading show-phone" translate>Framework of your contribution</h2>
+    <div class="content">
+      <div class="icon">
+        <span class="glyphicon glyphicon-warning-sign"></span>
+      </div>
+      <div class="details" translate>${model} edit framework details</div>
```
asaunier: If you use ``translate`` here, it will extract ``${model} edit framework details``. Which is probably not what you want. Then you have to use ``x-translate`` and add the strings explicitly to the ``i18n.html`` template.

But shouldn't you rather add those contents directly to the editing form templates without relying on a transifex "variable" ?

brunobesson: Right, I forgot this for the `x-translate`.

About the variable, I would like to keep a single helper methd for all documents. How would you avoid a variable in the translation string?
Or maybe you see no interest in this factorization?

asaunier: According to your screenshot, the content of the translated variable will probably contain HTML tags such as list or links etc. Is Transifex really the place to write HTML tags? Shouldn't it only suggest explicit strings to translate?

Factorizing using an helper seems a good thing. Perhaps you could use a ``<%block>`` to override the content in each template?

brunobesson:
OK for the `<%block>` part.
One more remark, then: I chose to keep a single entry for each file, containing HTML, because the content is not dry, I fear it could change a lot, and come after the PR has been merged.
This is already done in some other place, AFAIK.

PR #2067, opened 2018-03-29
preset quality for outings, articles and images

desnoes:
Automatic preselection of the quality criterion for outings, articles and images.
cf. https://github.com/c2corg/v6_ui/issues/1929

PR #2109, opened 2018-05-10
(Bug fix) preset quality for outings, articles and images

desnoes:
Fix a bug in [PR #2067 ](https://github.com/c2corg/v6_ui/pull/2067) for types other than outings, articles and images.


desnoes:
I will do the corrections next sunday.

asaunier:
Is this PR ready for reviewing?

brunobesson:
Yes it is. @mdesnoes confirmed that on the forum. I can have a look tomorrow if needed (I'll also tag v1.2.13)

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -626,8 +626,11 @@ app.DocumentEditingController.prototype.confirmSave = function(isValid) {
   }
 
   const data = this.scope[this.modelName];
-  const score = this.presetQuality(data);
-  this.storeQuality(data, score);
+  const doctype = app.utils.getDoctype(data['type']);
```
brunobesson: I would rather use the condition the opposite way so that a new odcument type would not be included by default.

A shame we cannot use `includes` yet!

desnoes:
ok on demo

brunobesson:
I have directly updated the prod... ;-)

Issue #2145, opened 2018-07-25
Improve doc printing


asaunier:
Suggested by @flob38 at https://forum.camptocamp.org/t/mise-en-page-pour-impression-des-topos/222410

```
/*ajouts*/
/*suppression message en entête*/
.announcement
  {display:none !important}
  
/*suppression des icônes*/
.glyphicon, .icon-glacier, .list-green ul li::before
  {display:none !important}

/*remplacement de l'icône "Activité" par la description*/
.route-activity::before
  {content:attr(uib-tooltip);
   font-family:"Helvetica Neue",Helvetica,Arial,sans-serif;
   font-size:10.5px;
   font-weight:700}
[class^="route-activity icon-"]
  {line-height:0 !important}

/*affichage des infos en ligne*/
.ficontent .row
  {display:flex;
   flex-flow:row wrap}
```
To add to https://github.com/c2corg/v6_ui/tree/master/less-print

PR #2148, opened 2018-07-31
Update viewdetails.less

flob38:
[Issue 2415](https://github.com/c2corg/v6_ui/issues/2145)

Changes : 

- Annoucement, bullets and icons are not printed anymore
- Activity is printed in plain text and not icon
- Informations (general, rating, location...) are displayed in row instead of column
- No more bug of white page after the header with Firefox (no bug with Chrome)

### DIFF
--- a/less-print/viewdetails.less
+++ b/less-print/viewdetails.less
@@ -43,7 +43,11 @@
 .view-details-comments,
 .view-details-map,
 .badge,
-.icons-header {
+.icons-header,
+.announcement,
+.glyphicon,
+.icon-glacier,
+.list-green ul li::before{
   display: none !important;
 }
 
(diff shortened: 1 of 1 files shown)


less-print/viewdetails.less
```diff
@@ -166,3 +170,19 @@ h3 {
   font-weight: bold;
   padding: 5px 0 5px 0;
 }
+
+.route-activity::before {
+  content:attr(uib-tooltip);
+  font-family:"Helvetica Neue",Helvetica,Arial,sans-serif;
+  font-size:10.5px;
```
brunobesson: A float value for a pixel size? WHat's the interest?

flob38: I didn't think a lot, it was the value for the other informations.
In fact, it's 0.75em for the other informations. I need to put this one to 0.35em if I want the same rendering (don't know why). But this way is better than px, that's right, I change that.

brunobesson:
Thanks, I'm merging your PR!

PR #2149, opened 2018-07-31
Hide empty outings in feed

cbeauchesne:
This small modification will hide empty outings in home page.

It handle the case where cards are on user page, and in this case, it won't hide anything.

### DIFF
--- a/c2corg_ui/static/partials/feed.html
+++ b/c2corg_ui/static/partials/feed.html
@@ -16,17 +16,17 @@
   <h3 class="title-feed" translate>Activity feed</h3>
   <div class="in-feed-col-1">
     <article ng-repeat="doc in feedCtrl.documentsCol[0] track by doc.id">
-      <app-card app-card-doc="::doc"></app-card>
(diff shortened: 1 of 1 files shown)


cbeauchesne:
Actually, I think about this way to do thing. Regardless the fact that modify API is far from possible, it will be much more complicated : service should not be variable dependant, eg it should not filter results by default. So we must add a argument to this service, in order to allow &quality=empty,empty in url.

In my opinion, I think it's damn complicated, for such a specific usage.

cbeauchesne:
arf je suis allé trop vite... dsl

Si faut rollbacker, dites moi.

asaunier:
Let say a user has created a bunch of empty outings recently. With this approach the home feed will load but all empty docs will be removed, leaving only a few items visible (or maybe none if all are empty) until the user scrolls the feed (if he figures out to do it...).
This doesn't look like a very robust solution 😉 

I am working on a change on the API side https://github.com/c2corg/v6_api/pull/728
Unfortunately I have no dev environment to work and test :/

PR #2155, opened 2018-08-02
Move all doc from readme.md to wiki

cbeauchesne:
I've totally rewamped wiki, and included in it info from readme.

I've reduced it to a small page, and tried to be fancy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,88 +1,20 @@
-UI Application for camptocamp.org v6
+UI application for [camptocamp.org](https://www.camptocamp.org/)
 ====================================
 
 [![Greenkeeper badge](https://badges.greenkeeper.io/c2corg/v6_ui.svg)](https://greenkeeper.io/)
 
-Requirements
-------------
-
- * Python 3.4 (3.5 seems OK as well)
- * Node (0.10.x or higher, there are known issues with 0.10.29)
- * Java (1.7 or higher)
- * gettext (0.18 or higher)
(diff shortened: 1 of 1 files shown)


asaunier:
Just one thing about the wiki: currently a distinction is made between Linux and Mac OS / Windows and Docker usage is not described for Linux:
* https://github.com/c2corg/v6_ui/wiki/Contributing-with-Linux
* https://github.com/c2corg/v6_ui/wiki/Howto-Contribute-on-Windows-or-Mac

=> is it really relevant? Using Docker is as interesting on Debian/Ubuntu than on Windows or MacOS (even though the other approach is possible and has its own pros) => I guess we should describe the docker way for Linux as well => then it is probably not a good idea to have distinct pages?

cbeauchesne:
IMO, the target for this doc is to have exactly the same way to install the dev environement for three OS, for reasons you gave.

But, for the moment : 

* I do not have linux, so I can't test anything
* And I guess (but not sure) that the env described in readme.md for linux does not have the drawback  of relaunch docker-compose up

So, I've keeped the information I have, without presuming what we should do for the future.

And yes, I also think that docker should be the target, because it does not need to install a lot of dependancies. In a perfet world, this help should look like something like that : 

> for all OS, with docker and git installed : 
>
> 1. git clone ...
> 2. docker-compose up
> 3. and let's code, all your modifications on .py, .js, .html, .css file are taken in account, without any action
> 4. Commit ...

Issue #2174, opened 2018-08-27
Link to rating help for "via ferrata" activity is wrong


loicperrin:
When editing a route with activity "via ferrata", the link of the help for the rating should be https://www.camptocamp.org/articles/925306/fr/cotations-via-ferrata-et-via-corda

PR #2175, opened 2018-08-28
#2174

brunobesson:
Fix some links and translations.

I have removed the anchors and language from URLs, because this should be non-french specific.


brunobesson:
See #2174

brunobesson:
Yeah, sorry, did that in a short time

PR #2177, opened 2018-08-29
computation of completeness criterion for route, WP, Xreport, books and areas

desnoes:
Automatic computation of completeness criteria for books, areas, routes, xreports and waypoints. In addition, the possibility to modify manually the value of this ciriterion has been removed from the confirmation window.


c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -670,7 +671,65 @@ app.DocumentEditingController.prototype.openModal = function(selector, sizem) {
 app.DocumentEditingController.prototype.presetQuality = function(doc) {
   // Do nothing special in the standard editing controller.
```
brunobesson: update this comment

c2corg_ui/static/js/editing/routeediting.js
```diff
@@ -105,4 +105,126 @@ app.RouteEditingController.prototype.handleAssociation = function(data, doc,
...
+      score = score + 0.25;
+    }
+    if ('route_history' in doc.locales[0] && doc.locales[0]['route_history']) {
+      score = score + 0.5;
+    }
+    if ('external_resources' in doc.locales[0] && doc.locales[0]['external_resources']) {
+      score = score + 0.25;
+    }
+
+    score_climb = score + score_geometry * 0.75 + score_configuration * 0.25 + score_caracteristics * 0.5 + score_wp * 0.25 + score_ratings * 0.75;
```
brunobesson: split this line for more readability

c2corg_ui/static/js/editing/waypointediting.js
```diff
@@ -0,0 +1,242 @@
...
+
+
+/**
+ * @param {appx.Document} doc Document attributes.
+ * @return {number}
+ * @override
+ */
+app.WaypointEditingController.prototype.presetQuality = function(doc) {
+  let score = 0;
+  const Type = doc['waypoint_type'];
```
brunobesson: no uppercase, please

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -627,9 +627,11 @@ app.DocumentEditingController.prototype.confirmSave = function(isValid) {
 
   const data = this.scope[this.modelName];
   const doctype = app.utils.getDoctype(data['type']);
-  if ((doctype != 'routes') && (doctype != 'waypoints') && (doctype != 'books') && (doctype != 'xreports') && (doctype != 'users') && (doctype != 'areas')) {
+  if (doctype != 'users') {
```
asaunier: Perhaps prefer the positive logic instead:
```
if (doctype == 'users') {
    data['quality'] = 'medium';
} else {
    const score = this.presetQuality(data);
    this.storeQuality(data, score);
}
```

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -670,7 +671,65 @@ app.DocumentEditingController.prototype.openModal = function(selector, sizem) {
 app.DocumentEditingController.prototype.presetQuality = function(doc) {
   // Do nothing special in the standard editing controller.
   // Will be overridden in inheriting controllers.
-  return 1;
+  let score = 0;
```
asaunier: That's not very consistent: it's said above that the fonction does nothing and should be overriden in inheriting controllers but you add some code below.
That would be more consistent to keep the function as it was and to move your code to inheriting controllers (create them if not already existing).

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -670,7 +671,65 @@ app.DocumentEditingController.prototype.openModal = function(selector, sizem) {
...
+        score++;
+      }
+      if ('author' in doc && doc['author']) {
+        score++;
+      }
+      if (doc['associations']['images'].length) {
+        score++;
+      }
+      if ('url' in doc && doc['url']) {
+        score = score + 0.5;
```
asaunier:
```
score += 0.5;
```
Ditto for the other blocks.

c2corg_ui/static/js/editing/routeediting.js
```diff
@@ -105,4 +105,126 @@ app.RouteEditingController.prototype.handleAssociation = function(data, doc,
...
+ */
+app.RouteEditingController.prototype.presetQuality = function(doc) {
+  let score = 0;
+  let score_climb = 0;
+  let score_other_activities = 0;
+  const activities = doc['activities'];
+
+  let score_wp = 0;
+  if (doc['associations']['waypoints'].length > 1) {
+    score_wp = 1.0;
```
asaunier:
```
score_wp = 1;
```

c2corg_ui/static/js/editing/waypointediting.js
asaunier: ``type`` is probably a reserved keyword. Use ``waypoint_type`` instead.

c2corg_ui/static/js/editing/waypointediting.js
```diff
@@ -0,0 +1,242 @@
...
+    }
+    if (result_search_img !== -1) {
+      score_description++;
+    }
+  }
+
+  if (Type == 'climbing_outdoor') {
+    if ('access' in doc.locales[0] && doc.locales[0]['access'] ||
+        'access_period' in doc.locales[0] && doc.locales[0]['access_period']) {
+      score_caracteristics = score_caracteristics + 0.5;
```
asaunier: caracteristics => characteristics?
I think we rather say "figures" in some other places.

c2corg_ui/static/js/editing/waypointediting.js
```diff
@@ -0,0 +1,242 @@
...
+    const result_search_img = doc['locales'][0]['description'].search(pattern_img);
+    if (result_search_title !== -1) {
+      score_description++;
+    }
+    if (result_search_img !== -1) {
+      score_description++;
+    }
+  }
+
+  if (Type == 'climbing_outdoor') {
```
asaunier: Perhaps use ``switch``, or at least ``if ... else if``

desnoes:
related to #2182

desnoes:
@brunobesson:
> I don't understand how you decided whether to extend the default controller or use a switch case in the document editing controller in order to compute the completeness.

I was simply based on the code length. For WP there was 180 lines which would be too much to insert in the generic DocumentEditingController. Tell me if my choice was incorrect.

> Anyway, maybe it could be a good idea to move all this in a separate service?
which would compute a quality score for each case, with a default value when applies.

I don't understand what you mean. Can you explain or provide an example ?

I will upload tonight the revised version of the code addressing comments from Bruno and Alex .

desnoes:
As suggested by @asaunier, i have created book and area controllers. I also add the modification of the confirmation save modal box that remove the modification of quality criterion by hand. I also modified code according to your suggestions.

c2corg_ui/static/js/editing/documentediting.js
```diff
@@ -627,7 +627,9 @@ app.DocumentEditingController.prototype.confirmSave = function(isValid) {
 
   const data = this.scope[this.modelName];
   const doctype = app.utils.getDoctype(data['type']);
-  if ((doctype != 'routes') && (doctype != 'waypoints') && (doctype != 'books') && (doctype != 'xreports') && (doctype != 'users') && (doctype != 'areas')) {
+  if (doctype == 'users') {
```
brunobesson: Why not use the default `presetQuality` method also for users?
This returns a score of 1, which is not medium quality, but anyway, why not return a score of 2?

desnoes: There is no 'quality' field for users.
Normally the default value (1) should never be set since all other type of documents have their own computation function. We can set it to 2 if you want.

c2corg_ui/templates/helpers/edit.html
```diff
@@ -105,11 +105,6 @@ <h3 class="heading show-phone" translate>summary</h3>
         <input type="text" name="message" ng-model="saveCtrl.message" class="form-control" placeholder="{{'Short description of the applied changes' | translate}}" />
       </div>
       % endif
-      ## QUALITY
-      <div class="full comment" ng-class="{'has-success': saveCtrl.quality}">
-        <label translate>quality</label>
-        <select ng-model="saveCtrl.quality" ng-options="type as saveCtrl.langService.translate(type) for type in ${quality_types}" class="form-control"></select>
-      </div>
```
asaunier: Are you sure we really want to skip all control on the final quality choice?
That looks a bit too radical too me. The quality of a document also depends on the content itself: if you have all the fields set but with a silly or poor content, the quality will never be excellent!
I think we should keep the manual selector, at least the time needed to make sure the automatic setting actually works.

desnoes: This decision comes from a vote: cf. https://forum.camptocamp.org/t/vote-sur-le-calcul-du-critere-de-completude/223776/26

asaunier: I do think this is not safe to introduce such a radical change without a significative period of testing, to make sure the auto computation actually works as expected.

In addition if the goal was to have the quality set with no human interaction, your changes should have been done in the API, not in the UI. By the way what happens if someone makes changes through the mobile app or an other API client?

And as I said, it seems to me your computation only use quantitative criterias, and not really qualitative ones. For instance if you have a topo full of "blabla" it is definitely not as good as if you have a shorter but more usable description.

I remove my change request but am definitely reluctant to give my approval.

desnoes:
J'ai remis le sélecteur de qualité. Je vous laisse juge d'intégrer ou non le PR. Pour ma part, j'arrête là les développements et les discussions.

Issue #2179, opened 2018-09-01
Add warning message/control about "revision comment" length


mgerbaux:
When editing a doc, if the field to describe modification is more than 200 characters, the modification cannot be saved and there is the warning message "unexpected error : please consult the server logs".

But is is very hard for contributor to find what is wrong with its contribution !

asaunier:
2 things are needed:
* add a check on the API side that would either return an error if length > 200 or cut the message off to 200 chars
* add a restriction to the HTML field in the UI side (most likely add a ``maxlength="200"`` attribute to the editing form

cbeauchesne:
Seems to be close, @asaunier @brunobesson do not hesitate to reopen if I'm wrong

PR #2180, opened 2018-09-01
Limit changes comment to 200 chars

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/2179

It's a very simple and minimalistic fix (no error message is shown, the field can contain only 200 chars, period) but it does the job.
Perhaps it could be improved using ``ng-maxlength`` https://docs.angularjs.org/api/ng/directive/ngMaxlength but not sure it is a very common case, I guess it's OK like this.


asaunier:
@brunobesson Thanks for reviewing!

